PR #494, opened 2013-07-08
Extended Job in Job.scala to enable running existing MapReduce jobs with...

tglstory:
Extended Job in Job.scala to enable running existing MapReduce jobs with Scalding.

This commit is referencing this group posting:
https://groups.google.com/forum/#!topic/cascading-user/mdgKP6pffLk

(This is my first pull request, please let me know if I'm doing anything wrong. Thanks :))

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -536,3 +536,68 @@ private[scalding] case class FlowStepStrategies[A]() extends Semigroup[FlowStepS
       }
     }
 }
+
+/*                                                                             
+ * Run a pre-existing MapReduce job with Scalding.
+ * Note that it is up to the user to properly configure the job.                                      
+ */
+class MapReduceJob(args: Args) extends Job(args) {
+  val conf: Configuration = mode match {
+    case h: HadoopMode => h.jobConf
+    case _ => new Configuration
+  }
+  protected val job = new JobConf(conf, classOf[MapReduceJob])
+
+  def setupMapReduce: Flow[_] = {
+    val input = args.getOrElse("mapreducejob.input", "")
+    val output = args.getOrElse("mapreducejob.output", "")
+    FileInputFormat.addInputPath(job, new Path(input))
+    FileOutputFormat.setOutputPath(job, new Path(output))
+    processJobSpecifics
+    new MapReduceFlow(job)
+  }
+
+  def processJobSpecifics(): Unit = {
+    /* Set up map output K/V class and output K/V class here...
+       *
+       * job.setMapOutputKeyClass(classOf[Text])
+       * job.setMapOutputValueClass(classOf[LongWritable])
+       *
+       * job.setOutputKeyClass(classOf[Text])
+       * job.setOutputValueClass(classOf[Text])
+       *
+       * Set up mapper/reducer class here...
+       *
+       * job.setMapperClass(classOf[Mapper])
+       * job.setReducerClass(classOf[Reducer])
+       */
+  }
+
+  def buildFlowMapReduce(flow: Flow[_]): Flow[_] = {
+    listeners.foreach { flow.addListener(_) }
+    stepListeners.foreach { flow.addStepListener(_) }
+    skipStrategy.foreach { flow.setFlowSkipStrategy(_) }
+    stepStrategy.foreach { strategy =>
+      val existing = flow.getFlowStepStrategy
+      val composed =
+        if (existing == null)
+          strategy
+        else
+          FlowStepStrategies[Any].plus(
+            existing.asInstanceOf[FlowStepStrategy[Any]],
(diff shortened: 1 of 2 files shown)


tglstory:
@johnynek @jcoveney Sorry for being so late with the update. I tried to write a test but for some reason it wants me to provide an input file- I'm assuming this is because of a different Flow object. Can someone help in getting this test configured properly?

Thanks.

tglstory:
anybody?

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -536,3 +540,68 @@ private[scalding] case class FlowStepStrategies[A]() extends Semigroup[FlowStepS
       }
     }
 }
+
+/*                                                                             
+ * Run a pre-existing MapReduce job with Scalding.
```
johnynek: can we put this in it's own file?

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -536,3 +540,68 @@ private[scalding] case class FlowStepStrategies[A]() extends Semigroup[FlowStepS
...
+ * Note that it is up to the user to properly configure the job.                                      
+ */
+class MapReduceJob(args: Args) extends Job(args) {
+  val conf: Configuration = mode match {
+    case h: HadoopMode => h.jobConf
+    case _ => new Configuration
+  }
+  protected val job = new JobConf(conf, classOf[MapReduceJob])
```
johnynek: can we call this jobConf or something? Also, can we put a type or make it private?

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -536,3 +540,68 @@ private[scalding] case class FlowStepStrategies[A]() extends Semigroup[FlowStepS
...
 }
+
+/*                                                                             
+ * Run a pre-existing MapReduce job with Scalding.
+ * Note that it is up to the user to properly configure the job.                                      
+ */
+class MapReduceJob(args: Args) extends Job(args) {
+  val conf: Configuration = mode match {
+    case h: HadoopMode => h.jobConf
+    case _ => new Configuration
```
johnynek: does this make sense? Are we really going to be able to run this in local mode? If not, perhaps we should sys.error here, or make `conf: Option[Configuration]`

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -536,3 +540,68 @@ private[scalding] case class FlowStepStrategies[A]() extends Semigroup[FlowStepS
...
+            strategy.asInstanceOf[FlowStepStrategy[Any]])
+      flow.setFlowStepStrategy(composed)
+    }
+    flow
+  }
+
+  override def run: Boolean = {
+    val flow = buildFlowMapReduce(setupMapReduce)
+    flow.complete
+    flow.getFlowStats.isSuccessful
```
johnynek: don't see why we need this line.

scalding-core/src/test/scala/com/twitter/scalding/CoreTest.scala
```diff
@@ -1849,3 +1857,68 @@ class DailySuffixTsvTest extends WordSpec with Matchers {
...
+  }
+}
+
+class MapReduceTestJobTest extends WordSpec with Matchers {
+
+  "A MapReduce Job" should {
+    JobTest(new MapReduceTestJob(_))
+      .arg("mapreducejob.input", "fakeIn")
+      .arg("mapreducejob.output", "fakeOut")
+      .source(TextLine("fakeIn"), List("0" -> "single test", "1" -> "single result"))
```
johnynek: totally blown away that this could work. How is the source plumbed through to the map-reduce job that knows nothing of our test mocking or cascading schemes/etc... Does this really work?

johnynek:
Seems reasonable, but I'd like to see more motivation for why you'd want to do this.

Is it to run in a cascade, or with a `Job.next`? Not clear to me why you don't just run a mapreduce job.

The tests are red (I didn't read why). Getting them green is a necessity before merging.

In an ideal world, I'd rather see a way to run a mapreduce job as an `Execution[Unit]` since they compose much better than Jobs do.

tglstory:
The circumstance for this was, way back in 2011~2013 my project required a workflow scheduler, much like Luigi or the likes that are currently available. We had a bunch of mapreduce jobs along with scalding jobs so it made sense at the time to do something like this. I had made this PR and got carried away with other things, and I came back and noticed somebody else wanted this so I made an attempt.
Since nowdays you can easily connect hadoop jobs via various scheduler, I think you're right, there is a lack of motivation for Scalding to carry that burden. Will close this PR.

johnynek:
@tglstory totally appreciate the effort, and I hope you'll make future contributions.

PR #784, opened 2014-02-10
Adds unfold to TypedPipe

MansurAshraf:
see https://github.com/twitter/algebird/issues/247 for details


johnynek:
Maybe we can revisit this on KeyedListLike.

PR #1059, opened 2014-09-19
Move Scalding to 2.10/2.11

jcoveney:
So this is just the beginning, but it's a big chunk of the work. I've been doing it between other things.

The big things I've done are to get scalding-core and scalding-hadoop-test working. Note: the DistributedCache tests in scalding-core had a dependency on mockito.... for now I commented it out. Do we want to depend on mockito just for one test? That said, rewriting is going to be a slight pain. So I'm not sure what we want to do there.

Either way, this is to get the ball rolling. I haven't started the 2.11 work becasue we need a new algebird release...but the scalatest migration stuff is enough for the time being!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # Generate base tests:::
-# for s in 2.10.4 2.9.3; do
+# for s in 2.10.4 2.11.2; do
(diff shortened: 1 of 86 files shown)


johnynek:
This is done now, so closing this.

PR #1110, opened 2014-12-03
use Cascading 2.6.1 and cascading-jdbc 2.6.0

fs111:
Cascading 2.6.x adds amongst various bugfixes a new set of annotations that can be used to send additional user defined information to a DocumentService like driven. Since it is backwards compatible, updating will not cause any issue.

In case you want to review the changes in Cascading 2.6 before merging, see here: https://github.com/Cascading/cascading/blob/2.6/CHANGES.txt


fs111:
@johnynek any update on this?

Issue #989, opened 2014-08-01
Scalding REPL should use Local mode if all sources are MemoryTaps


bholt:
It came up when working on a new Scalding REPL in HDFS tutorial that we _almost_ have a really cool way to "subset" data in memory and work with it locally even when running in HDFS mode.

``` scala
scalding> val tweets = TypedPipe.from(/* some HDFS source */)
scalding> val subset = TypedPipe.from(tweets.map(_.getText).limit(100).toList)
```

Many subsequent operations on `subset` work properly, running right away locally on the IterablePipe. However, to get the results of any operation involving "reduce", it still thinks it needs to run in real hadoop mode:

``` scala
scalding> loweredHashtags.map(_ -> 1).sumByKey.dump
14/08/01 18:48:49 INFO util.HadoopUtil: resolving application jar from found main method on: com.twitter.scalding.ScaldingShell$
14/08/01 18:48:49 INFO planner.HadoopPlanner: using application jar: /home/bholt/scalding-shell-with-sources.jar
14/08/01 18:48:49 INFO flow.Flow: [ScaldingShell] starting
14/08/01 18:48:49 INFO flow.Flow: [ScaldingShell]  source: MemorySourceTap["MemorySourceScheme[['key', 'value']->[ALL]]"]["/22ccf197-ae70-48fe-98a4-44b988df00ff"]
14/08/01 18:48:49 INFO flow.Flow: [ScaldingShell]  sink: Hfs["SequenceFile[[0]]"]["/tmp/scalding-repl/snapshot-feb0186a-8610-4a1b-be13-0dfce768c85c.seq"]
14/08/01 18:48:49 INFO flow.Flow: [ScaldingShell]  parallel execution is enabled: true
14/08/01 18:48:49 INFO flow.Flow: [ScaldingShell]  starting jobs: 1
14/08/01 18:48:49 INFO flow.Flow: [ScaldingShell]  allocating threads: 1
14/08/01 18:48:49 INFO flow.FlowStep: [ScaldingShell] starting step: (1/1) ...a1b-be13-0dfce768c85c.seq
14/08/01 18:49:33 INFO flow.FlowStep: [ScaldingShell] submitted hadoop job: job_201406090027_920567
14/08/01 18:49:33 INFO flow.FlowStep: [ScaldingShell] tracking url: http://hadoop-dw-jt.smf1.twitter.com:50030/jobdetails.jsp?jobid=job_201406090027_920567
[...]

PR #1113, opened 2014-12-04
Implicit execution context / easier switching between modes

declerambaul:
Would something like this makes sense? Is there a way to recover the hadoop config after parsing the args/mode, or would we need to preserve it at startup (like in this pr)?

Sriram pointed me to https://github.com/twitter/scalding/issues/989 which is related but more general (and more effort).


joshualande:
This is great. Thanks so much for adding this nice feature!

PR #1131, opened 2014-12-15
Ianoc/case class tuple converters

ianoc:
Tests won't pass as relies on macro code in an as yet un-released version of bijection


johnynek:
so mysterious... Why is coveralls pinging this 1.5 year old PR!?

PR #1136, opened 2014-12-17
Add an easier to extent delimited source (TypedText)

johnynek:
The intent here is to have macros that allow something like:

``` scala
import ReadPathSelector._

TextSource[((Int, String), Long)](datePath("/logs/%Y/%H/")): TypedSource[((Int, String), Long)]

case class Job(title: String, salary: Long)
case class MyRecord(name: String, age: Int, job: Option[Job])

TextSource[MyRecord]("/some/path/"): TypedSource[MyRecord]
```

To do this, we need to be able to pass the types to cascading without a manifest required by TypedDelimited.

This is also be a design example of how sources could be cleaned up a bit.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -33,7 +33,7 @@ object ScaldingBuild extends Build {
   val hravenVersion = "0.9.13"
   val jacksonVersion = "2.4.2"
   val protobufVersion = "2.4.1"
-  val elephantbirdVersion = "4.4"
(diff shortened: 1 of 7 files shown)


johnynek:
@isnotinvain @ianoc should we merge this?

Do we want another way to write these text sources?

johnynek:
I think 4.5 years is a timeout.

PR #1148, opened 2015-01-01
Add a sketch of OrderedBufferable

johnynek:
This is a pull request against an experimental branch (fast-binay-sort) that will try to plumb through binary comparators from scalding all the way down to Hadoop.

This has been shown, perhaps not surprisingly, to have huge wins.

@ianoc How does this look for a starting point?

My thought is that we will check with reflection in all Groups if we get an Ordering that is a subclass of OrderedBufferable. If so, we will wire up the correct stuff so that cascading sees everything.  Then we can make a method in TwitterJob like:
```scala
// Handle strings, longs, etc....
implicit def primitiveOrd[T:IsPrimitive]: Ordering[T] = Macros.orderingBufferablePrimitive[T]
implicit def caseClassOrd[T:IsCaseClass]: Ordering[T] = Macros.orderedBufferableCaseClass[T]
implicit def thriftStruftOrd[T:IsThriftStruct] = Macros.orderedBufferableScrooge[T]
```

Does this seem like a sane way to plumb things through? It is minimally invasive in that if you don't opt in, you don't get the performance win.

My thought is that if this looks like a huge win, we consider in scalding 0.14 or 0.15 making this required and you opt of it by creating a Kryo-based OrderedBufferable that will just internally deserialize.


scalding-core/src/main/scala/com/twitter/scalding/typed/OrderedBufferable.scala
```diff
@@ -0,0 +1,99 @@
...
+trait OrderedBufferable[T] extends Bufferable[T] with Ordering[T] {
+  /**
+   * This compares two ByteBuffers. After this call, the position in
+   * the ByteBuffers is mutated. If you don't want this behavior
+   * call duplicate before you pass in (which copies only the position, not the data)
+   */
+  def compareBinary(a: ByteBuffer, b: ByteBuffer): OrderedBufferable.Result
+  def hash(t: T): Int
+  /**
+   * This hashes a ByteBuffer. Should be the same as hash(get(buf)).
```
ianoc: not clear what the function get is here

johnynek: Sorry, `hash(Bufferable.get(b: ByteBuffer)._2)` is the correct way to write it.

scalding-core/src/main/scala/com/twitter/scalding/typed/OrderedBufferable.scala
```diff
@@ -0,0 +1,100 @@
...
+     */
+    def unsafeToInt: Int
+  }
+  final case class CompareFailure(ex: Throwable) extends Result { def unsafeToInt = throw ex }
+  case object Greater extends Result { val unsafeToInt = 1 }
+  case object Equal extends Result { val unsafeToInt = 0 }
+  case object Less extends Result { val unsafeToInt = -1 }
+
+  def serializeThenCompare[T](a: T, b: T)(implicit ordb: OrderedBufferable[T]): Int = {
+    val bb1 = Bufferable.reallocatingPut(ByteBuffer.allocate(128)) { ordb.put(_, a) }
```
ianoc: Should the OrderedBufferable I build call reallocatingPut or should I presume whatever is on the outside will handle reallocations on failure? It looks like the Bufferables in bijection all call the reallocatingPut

johnynek: Don't call reallocatingPut. That was a bad idea in my opinion. The
outermost caller should do that if they want it, not a required feature of
all composed items.

On Tuesday, January 6, 2015, ianoc <notifications@github.com> wrote:

> In
> scalding-core/src/main/scala/com/twitter/scalding/typed/OrderedBufferable.scala
> <https://github.com/twitter/scalding/pull/1148#discussion-diff-22528788>:
>
> > +   * Represents the result of a comparison that might fail due
> > +   * to an error deserializing
> > +   */
> > +  sealed trait Result {
> > +    /**
> > +     * Throws if the items cannot be compared
> > +     */
> > +    def unsafeToInt: Int
> > +  }
> > +  final case class CompareFailure(ex: Throwable) extends Result { def unsafeToInt = throw ex }
> > +  case object Greater extends Result { val unsafeToInt = 1 }
> > +  case object Equal extends Result { val unsafeToInt = 0 }
> > +  case object Less extends Result { val unsafeToInt = -1 }
> > +
> > +  def serializeThenCompare[T](a: T, b: T)(implicit ordb: OrderedBufferable[T]): Int = {
> > +    val bb1 = Bufferable.reallocatingPut(ByteBuffer.allocate(128)) { ordb.put(_, a) }
>
> Should the OrderedBufferable I build call reallocatingPut or should I
> presume whatever is on the outside will handle reallocations on failure? It
> looks like the Bufferables in bijection all call the reallocatingPut
>
> —
> <https://github.com/twitter/scalding/pull/1148/files#r22528788>.
>


-- 

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -0,0 +1,144 @@
...
+    chan = Channels.newChannel(os)
+  }
+  def close(): Unit = { chan = null }
+  def serialize(t: T): Unit = {
+    // allocate a new ByteBuffer, save space for size at the header with the putInt(0)
+    val bb1 = Bufferable.reallocatingPut(ByteBuffer.allocate(128).putInt(0)) { buf.put(_, t) }
+    val len = bb1.position - 4 // 4 for the int for size
+    bb1.position(0)
+    bb1.putInt(len)
+    chan.write(bb1)
```
johnynek: rewind before write

johnynek: maybe use DataOutput stream to write size consistently.

johnynek:
Need to add the WrappedSerialization by default here.

PR #1149, opened 2015-01-01
WIP, for comment. Jdbc Macros

ianoc:
Still in real world testing though has unit tests here. Not for merging yet. Want to solidify on the API and migrate some/all of our internal usages of JDBC to it to verify it covers all cases properly. 

to be contributed also is a sibling source+tap combo that uses the method outlined @ http://www.vertica.com/2012/07/05/teaching-the-elephant-new-tricks/ with vertica. Over jdbc its been 80x faster for people(the api, not my code), and also is inside one transaction. This portion is functional but I want to hammer on it a bit in the real world before trying to place in scalding. 

Sample usage:

```
import com.twitter.scalding.macros._
import MacroImplicits._
import com.twitter.scalding.jdbc._
import com.twitter.scalding.jdbc.macros._
import com.twitter.scalding.macros._

val cc = ConnectionConfig(
    ConnectUrl("jdbc:mysql://yourmysqlserver:3306/scaldingtestdb"),
    UserName("your username"),
    Password("yourpassword"),
    Adapter("mysql"))

case class TypedSource1(
  date_id: Int,
  @size(64) user_name: String,
  age: Int,
  @size(22) gender: String = "male")

case class TypedSource1JdbcSource() extends TypedJDBCSource[TypedSource1(TableName("helloWorld"), cc)

TypedPipe.from(List(TypedSource1(1234, "asdf", 13))).save(TypedSource1JdbcSource())
```


scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/JDBCMacro.scala
```diff
@@ -0,0 +1,25 @@
...
+import com.twitter.scalding.jdbc.macros.impl._
+import com.twitter.scalding.jdbc.{ ColumnDefinitionProvider, JDBCTypeInfo }
+
+sealed trait ScaldingJdbcAnnotation
+// This is the size in characters for a string field
+// For integers its really for display purposes
+@scala.annotation.meta.getter
+class size(val size: Int) extends annotation.StaticAnnotation with ScaldingJdbcAnnotation
+
+@scala.annotation.meta.getter
```
ianoc: add comments on each of these, along with a top one above as to what they are for

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/JDBCMacro.scala
```diff
@@ -0,0 +1,25 @@
...
+class text() extends annotation.StaticAnnotation with ScaldingJdbcAnnotation
+
+@scala.annotation.meta.getter
+class varchar() extends annotation.StaticAnnotation with ScaldingJdbcAnnotation
+
+@scala.annotation.meta.getter
+class date() extends annotation.StaticAnnotation with ScaldingJdbcAnnotation
+
+object JDBCMacro {
+  implicit def toColumnDefinitionProvider[T]: ColumnDefinitionProvider[T] = macro JDBCMacroImpl.caseClassJDBCPayloadImpl[T]
```
ianoc: do these need to be implicit?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JDBCTypeInfoImpl.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.util.{ Success, Failure }
+
+import com.twitter.scalding._
+import com.twitter.bijection.macros.impl.IsCaseClassImpl
+import com.twitter.scalding.jdbc.JDBCTypeInfo
+import com.twitter.scalding.jdbc.macros._
+import com.twitter.scalding.macros.impl._
+
+/**
+ * This class contains the core macro implementations. This is in a separate module to allow it to be in
```
ianoc: comment looks out of date

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JDBCTypeInfoImpl.scala
```diff
@@ -0,0 +1,43 @@
+package com.twitter.scalding.jdbc.macros.impl
+
+import scala.language.experimental.macros
+
+import scala.reflect.macros.Context
+import scala.reflect.runtime.universe._
+import scala.util.{ Success, Failure }
```
ianoc: Imports don't look used

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+import com.twitter.scalding.jdbc.{ ColumnDefinition, ColumnDefinitionProvider }
+import com.twitter.scalding.jdbc.macros._
+import com.twitter.scalding.jdbc.macros.impl.handler._
+
+case class FieldName(toStr: String) {
+  override def toString = toStr
+}
+
+/**
+ * This class contains the core macro implementations. This is in a separate module to allow it to be in
```
ianoc: bad comment again

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+import scala.reflect.runtime.universe._
+import scala.util.{ Success, Failure }
+
+import com.twitter.scalding._
+import com.twitter.bijection.macros.impl.IsCaseClassImpl
+import com.twitter.scalding.jdbc.{ ColumnDefinition, ColumnDefinitionProvider }
+import com.twitter.scalding.jdbc.macros._
+import com.twitter.scalding.jdbc.macros.impl.handler._
+
+case class FieldName(toStr: String) {
```
ianoc: Private/package private?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+}
+
+/**
+ * This class contains the core macro implementations. This is in a separate module to allow it to be in
+ * a separate compilation unit, which makes it easier to provide helper methods interfacing with macros.
+ */
+object JDBCMacroImpl {
+  // Takes a type and its companion objects apply method
+  // based on the args it takes gives back out a field name to symbol
+  private[this] def getDefaultArgs[T](c: Context)(T: c.WeakTypeTag[T]): Map[String, c.Expr[String]] = {
```
ianoc: Should maybe go elsewhere eventually, nothing JDBC specific in this

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+  def caseClassJDBCPayloadImpl[T](c: Context)(implicit T: c.WeakTypeTag[T]): c.Expr[ColumnDefinitionProvider[T]] = {
+    import c.universe._
+
+    if (!IsCaseClassImpl.isCaseClassType(c)(T.tpe))
+      c.abort(c.enclosingPosition, s"""We cannot enforce ${T.tpe} is a case class, either it is not a case class or this macro call is possibly enclosed in a class.
+        This will mean the macro is operating on a non-resolved type.""")
+
+    val defaultArgs = getDefaultArgs[T](c)(T)
+
+    // Field To JDBCColumn
```
ianoc: comment out of date

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+    def matchField(oTpe: Type,
+      fieldName: FieldName,
```
ianoc: params should be tabbed in more

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+        case tpe if tpe =:= typeOf[String] => StringTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable)
+        case tpe if tpe =:= typeOf[Short] => NumericTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable, "SMALLINT")
+        case tpe if tpe =:= typeOf[Int] => NumericTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable, "INT")
+        case tpe if tpe =:= typeOf[Long] => NumericTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable, "BIGINT")
+        case tpe if tpe =:= typeOf[Double] => NumericTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable, "DOUBLE")
+        case tpe if tpe =:= typeOf[Boolean] => NumericTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable, "TINYINT")
+        case tpe if tpe =:= typeOf[scala.math.BigInt] => NumericTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable, "BIGINT")
+        case tpe if tpe =:= typeOf[java.util.Date] => DateTypeHandler(c)(fieldName, defaultValOpt, annotationInfo, nullable)
+
+        case tpe if tpe.erasure =:= typeOf[Option[Any]] =>
```
ianoc: Option[Option[Int]] looks like it would pass through here?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+    val defaultArgs = getDefaultArgs[T](c)(T)
+
+    // Field To JDBCColumn
+    def matchField(oTpe: Type,
+      fieldName: FieldName,
+      defaultValOpt: Option[c.Expr[String]],
+      annotationInfo: List[(Type, Option[Int])],
+      nullable: Boolean): scala.util.Try[c.Expr[ColumnDefinition]] = {
+
+      oTpe match {
```
ianoc: Nested case class support?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
```diff
@@ -0,0 +1,117 @@
...
+        .declarations
+        .collect { case m: MethodSymbol if m.isCaseAccessor => m }
+        .map { m =>
+          val fieldName = m.name.toTermName.toString
+          val defaultVal = defaultArgs.get(fieldName)
+          val annotationInfo: List[(Type, Option[Int])] = m.annotations
+            .map(t => (t.tpe, t.scalaArgs))
+            .collect {
+              case (tpe, List(Literal(Constant(siz: Int)))) if tpe =:= typeOf[com.twitter.scalding.jdbc.macros.size] => (tpe, Some(siz))
+              case (tpe, _) if tpe <:< typeOf[com.twitter.scalding.jdbc.macros.ScaldingJdbcAnnotation] => (tpe, None)
```
ianoc: If a size annotation is above, but the size is a reference to a number elsewhere I don't believe it will be a literal constant so we will lose that value.

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/handler/ColFormatter.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scala.language.experimental.macros
+
+import scala.reflect.macros.Context
+
+import com.twitter.scalding.jdbc.ColumnDefinition
+import com.twitter.scalding.jdbc.macros._
+import com.twitter.scalding.jdbc.macros.impl.FieldName
+/**
+ * This class contains the core macro implementations. This is in a separate module to allow it to be in
```
ianoc: bad comment again

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/handler/DateTypeHandler.scala
```diff
@@ -0,0 +1,38 @@
...
+ */
+object DateTypeHandler {
+
+  def apply[T](c: Context)(implicit fieldName: FieldName,
+    defaultValue: Option[c.Expr[String]],
+    annotationInfo: List[(c.universe.Type, Option[Int])],
+    nullable: Boolean): scala.util.Try[c.Expr[ColumnDefinition]] = {
+    import c.universe._
+
+    val opts = scala.util.Try[Boolean](Monoid.sum(annotationInfo.collect {
```
ianoc: comment on what this does? is it necessary in the case of datetype?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/handler/StringTypeHandler.scala
```diff
@@ -0,0 +1,48 @@
...
+      case (tpe, Some(siz)) if tpe =:= typeOf[com.twitter.scalding.jdbc.macros.size] => (Some(siz), false, false)
+      case (tpe, _) if tpe =:= typeOf[com.twitter.scalding.jdbc.macros.varchar] => (None, true, false)
+      case (tpe, _) if tpe =:= typeOf[com.twitter.scalding.jdbc.macros.text] => (None, false, true)
+      case tpe => sys.error(s"Hit annotation ${tpe} which is not supported on field ${fieldName.toStr} of type String")
+    }))
+
+    opts.flatMap {
+      case (size, forceVarChar, forceText) =>
+        (size, forceVarChar, forceText) match {
+          case (_, true, true) => Failure(new Exception(s"String field $fieldName, has mutually exclusive annotations @text and @varchar"))
```
ianoc: Sealed case objects rather than all these booleans?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
johnynek: +1

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
johnynek: yes it does. Probably good to specify this class we can deal with. Option is a special case Either. You can think of it as Either[Unit, T]. That does not make it easier, I don't think, but it I just wanted to mention. :)

ianoc: It would, but i think similar to the TSV problem with a simple flattening to types we can't really support the nested Option[Option[Int]] right? since we won't know at which level the None is

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/handler/StringTypeHandler.scala
johnynek: +1 just because the matrix of cases is so large.

johnynek:
This needs at least 2 shipits. This is will be the third consecutive version of scalding-jdbc with breaking changes, which is not ideal. I don't want the following one to be the 4th.

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/JdbcMacroImpl.scala
johnynek: Yes, I just mean it would be nice to understand this class.

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/handler/DateTypeHandler.scala
johnynek: the sum of Booleans is XOR, right (since it is what Field[Boolean] needs). Is that what you want here? Seems like you want `.forall(identity)`?

ianoc:
Fair enough, i'm happy with refactoring the thing so its only internal too. Not sure anyone is using the current OSS one. @avibryant you using the OSS one? any thoughts on this?

scalding-jdbc-macros/src/main/scala/com/twitter/scalding/jdbc/macros/impl/handler/DateTypeHandler.scala
ianoc: True, seems like there was likely a bug there the tests never hit. Mostly as part of cleaning up the string case i've refactored all of them so the sum code doesn't exist anymore. (Here the monoid i would have been looking for was OR. But its not clean on what would occur in the presence of a repeated annotation. New code has explicit path for that)

PR #1150, opened 2015-01-06
First pass at adding the TBase support along with the macro

ianoc:
So scala has a bug using its reflection API on java static fields:
https://issues.scala-lang.org/browse/SI-6459

So that was out, so I have a java reflection method.. which of course I attempted to call in my macro. And tried for a while. But eventually I realised it made no sense, if the code of the thrift structure compiles in the same pass as your macro is firing you can't use java reflection on it since its not been compiled/loaded into the classpath yet. Its just the type information available. (Though since its a java class file in reality i guess it should be able to load the full thing since it would need to be fully resolved.)


scalding-commons-macros/src/test/scala/com/twitter/scalding/commons/macros/MacrosUnitTests.scala
```diff
@@ -0,0 +1,94 @@
...
+    @transient lazy val prototype: Nothing = null.asInstanceOf[Nothing]
+    override val classz = classOf[Nothing]
+  }
+
+  def rt[T <: TBase[_, _]](t: T)(implicit orderedBuffer: TBaseOrderedBufferable[T]) {
+    val buf = serialize[T](t)
+    assert(orderedBuffer.get(buf).map(_._2) === Success(t))
+  }
+
+  def serialize[T <: TBase[_, _]](t: T)(implicit orderedBuffer: TBaseOrderedBufferable[T]): ByteBuffer = {
```
johnynek: this does not need to take TBaseOrdered... just OrderedBufferable is enough. (same above).

scalding-commons/src/main/scala/com/twitter/scalding/commons/thrift/TBaseOrderedBufferable.scala
```diff
@@ -0,0 +1,83 @@
...
+abstract class TBaseOrderedBufferable[T <: TBase[_, _]] extends OrderedBufferable[T] {
+  val minFieldId: Short
+
+  @transient private lazy val factory = new TBinaryProtocol.Factory //  new TCompactProtocol.Factory
+
+  @transient protected def prototype: T
+
+  def hash(t: T) = t.hashCode
+
+  def compare(a: T, b: T) = if (a == b) 0 else -1
```
johnynek: can we add a BIG comment here explaining this. I'd love to see a TODO to make it follow the contract of compare == compareBinary

johnynek: actually, TBase extends comparable, why not do:

```scala
a.compareTo(b)
```

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -120,15 +123,25 @@ class BufferableDeserializer[T](buf: Bufferable[T]) extends Deserializer[T] {
...
   private val confKey = "com.twitter.scalding.serialization.WrappedSerialization"
 
-  def setBufferables(conf: Configuration, bufs: Iterable[ClassBufferable[_]]): Unit = {
-    conf.set(confKey,
-      bufs.map { case (cls, buf) => s"${cls.getName}:${serialize(buf)}" }.mkString(","))
+  def rawSetBufferable(bufs: Iterable[ClassBufferable[_]], fn: (String, String) => Unit) = {
+    fn(confKey, bufs.map { case (cls, buf) => s"${cls.getName}:${serialize(buf)}" }.mkString(","))
   }
+  def setBufferables(conf: Configuration, bufs: Iterable[ClassBufferable[_]]): Unit =
+    rawSetBufferable(bufs, { case (k, v) => conf.set(k, v) })
```
johnynek: case is not needed here, I don't think.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -135,7 +135,14 @@ sealed trait ReduceStep[K, V1] extends KeyedPipe[K] {
       val reducedPipe = mapped
         .toPipe(Grouped.kvFields)(fd, mode, tup2Setter)
         .groupBy(Grouped.keySorting(keyOrdering))(gb)
-      TypedPipe.from(reducedPipe, Grouped.kvFields)(fd, mode, tuple2Converter[K, V2])
+
+      val classBuff: Option[com.twitter.scalding.serialization.WrappedSerialization.ClassBufferable[_]] = keyOrdering match {
```
johnynek: can we remove this duplicated code and just call one method for both?

scalding-core/src/main/scala/com/twitter/scalding/typed/OrderedBufferable.scala
```diff
@@ -41,6 +41,8 @@ trait OrderedBufferable[T] extends Bufferable[T] with Ordering[T] {
    * Hadoop always gets the hashcode before serializing
    */
   def hash(t: T): Int
+
+  def classz: Class[T]
```
johnynek: this would not be needed with the boxing approach, right?

johnynek:
shipit

PR #1156, opened 2015-01-09
Cascading 2.6 tracing

JamesIry:
When Scalding talked to document services like Driven it was sending line numbers that essentially mark the boundary between Scalding and Cascading. But that's not very useful to end users. A change was made to Cascading 2.6 that allows higher level APIs to indicate where a document service should consider the boundary. This PR bumps Scalding up to use Cascading 2.6 and to participate in the protocol for registering an API boundary.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -237,7 +237,7 @@ object ScaldingBuild extends Build {
   lazy val scaldingDate = module("date")
 
   lazy val cascadingVersion =
-    System.getenv.asScala.getOrElse("SCALDING_CASCADING_VERSION", "2.5.5")
+    System.getenv.asScala.getOrElse("SCALDING_CASCADING_VERSION", "2.6.1")
 
   lazy val cascadingJDBCVersion =
     System.getenv.asScala.getOrElse("SCALDING_CASCADING_JDBC_VERSION", "2.5.4")
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -77,6 +77,7 @@ object Job {
  * these functions can be combined Monadically using algebird.monad.Reader.
  */
 class Job(val args: Args) extends FieldConversions with java.io.Serializable {
+  Tracing.init()
 
   // Set specific Mode
   implicit def mode: Mode = Mode.getMode(args).getOrElse(sys.error("No Mode defined"))
--- a/scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
@@ -1,0 +1,91 @@
+/*
+Copyright 2012 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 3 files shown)


ianoc:
Thanks for the PR. Just an FYI we are going to be pretty slow about taking the major cascading bump as it'll take lots of testing.

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -77,6 +77,7 @@ object Job {
  * these functions can be combined Monadically using algebird.monad.Reader.
  */
 class Job(val args: Args) extends FieldConversions with java.io.Serializable {
+  Tracing.init()
```
ianoc: Doesn't seem like it should be in the default job, something can be opted into by extending a trait instead

project/Build.scala
```diff
@@ -237,7 +237,7 @@ object ScaldingBuild extends Build {
   lazy val scaldingDate = module("date")
 
   lazy val cascadingVersion =
-    System.getenv.asScala.getOrElse("SCALDING_CASCADING_VERSION", "2.5.5")
+    System.getenv.asScala.getOrElse("SCALDING_CASCADING_VERSION", "2.6.1")
 
   lazy val cascadingJDBCVersion =
     System.getenv.asScala.getOrElse("SCALDING_CASCADING_JDBC_VERSION", "2.5.4")
```
ianoc: Does 2.5 JDBC go with 2.6 cascading?

JamesIry:
Understood. Let me know if there's anything I can do to help.

project/Build.scala
JamesIry: It can, but it's not necessary.

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
JamesIry: Are you suggesting that Job extend a trait that initializes Tracing? Or that end users should explicitly request tracing. If the later I disagree. This functionality isn't something people are likely to discover.

scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
```diff
@@ -0,0 +1,91 @@
...
+object Tracing {
+  private val LOG: Logger = LogManager.getLogger(this.getClass)
+
+  /**
+   * Put a barrier at com.twitter.scalding, but exclude things like Tool
+   * that are common entry points for calling user code
+   */
+  private val defaultRegex = """^com\.twitter\.scalding\.(?!Tool|Job|ExecutionContext).*"""
+
+  register()
```
ianoc: Why is this here and not in init?

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
ianoc: The former, they should extend a trait. If they use driven they will know its there. Why have tracing code running for everyone as this would have? 

(Indeed it seems like tracing here should be a trait anyway that gets extended from. also can allow its methods easier overridden)

JamesIry: A document service is a plugin deployed at runtime. The programmer may not have any idea that one is deployed.

The tracing code change is just registering a regex representing the boundary of the internals of Scalding. It runs once at start up and has no impact on users.

scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
JamesIry: It's a very cheap way to create one-time-init semantics.

johnynek:
Okay, this seems safe enough, and good for Driven users.

PR #1159, opened 2015-01-13
Update to the new typeclass: OrderedSerialization and Serialization

johnynek:
This changes the OrderedBufferable typeclass to OrderedSerialization, which extends Serialization. It works on InputStream and OutputStream as that is what matches Hadoop more closely and is a weaker constraint than having ByteBuffer.

This enables some better performance for cases where the records have a static size.


johnynek:
@ianoc take a look.

This looks pretty close to me. I figured it was better to refactor more now than have you refactor the macros more than once.

I think the scalding.serialization package does not (or does not need to) depend on scalding, so we could pull that out as module like scalding-args or scalding-date if we want to use it without picking up scalding.

scalding-commons/src/main/scala/com/twitter/scalding/commons/thrift/ScroogeOrderedSerialization.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import com.twitter.scalding.serialization.{ OrderedSerialization, Serialization }
+import cascading.tuple.hadoop.io.BufferedInputStream
+import cascading.tuple.StreamComparator
+import org.apache.thrift.protocol.TCompactProtocol
+import org.apache.thrift.protocol.TBinaryProtocol
+import com.twitter.scrooge.ThriftStruct
+import java.io.{ ByteArrayOutputStream, InputStream, OutputStream }
+import com.twitter.bijection.Inversion.attempt
+import com.twitter.bijection.Bufferable
```
ianoc: dead import now?

scalding-commons/src/main/scala/com/twitter/scalding/commons/thrift/TBaseOrderedSerialization.scala
```diff
@@ -0,0 +1,53 @@
...
+import com.twitter.bijection.Inversion.attempt
+import com.twitter.bijection.Bufferable
+import org.apache.thrift.transport.TIOStreamTransport
+
+import scala.util.{ Failure, Success, Try }
+import scala.util.control.NonFatal
+
+import com.twitter.scalding.serialization.JavaStreamEnrichments._
+
+abstract class TBaseOrderedBufferable[T <: TBase[_, _]] extends TProtocolOrderedSerialization[T] {
```
ianoc: Still called bufferable here

scalding-commons/src/main/scala/com/twitter/scalding/commons/thrift/TProtocolOrderedSerialization.scala
```diff
@@ -0,0 +1,37 @@
...
+  @transient protected lazy val factory = new TBinaryProtocol.Factory //  new TCompactProtocol.Factory
+
+  def hash(t: T) = t.hashCode
+
+  def compareBinary(lhs: InputStream, rhs: InputStream): OrderedSerialization.Result = {
+    val leftSize = lhs.readSize
+    val rightSize = rhs.readSize
+
+    val seekingLeft = lhs.markOrBuffer(leftSize)
+    val seekingRight = rhs.markOrBuffer(rightSize)
```
ianoc: strategy here is to jump back to a mark then skip the remainder? is this preferable to a mutating variable tracking how far we've gone but just that we can't go backwards?

johnynek: in practice the Hadoop InputStreams appear to always be backed with byte arrays, so they all support mark and reset, and they support fast skip.

The algorithm here is: if the InputStream supports mark/reset (which hadoop and cascading's do), we can use that at no cost. If they don't we read that number of bytes into a buffer and continue on from there. So, some kind of stream from a network that might not support seeking back would be completely read into our outside buffer, which is returned to us (though Java is so complected, my guess is that most NetworkStreams also have some mark buffer to support this).

At the end of this compare, we will have read the size and skipped that much data (by reset + skip or by reading it directly into a buffer).

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,221 @@
...
+
+    def readBoolean: Boolean = (readUnsignedByte != 0)
+
+    /**
+     * Like read, but throws eof on error
+     */
+    def readByte: Byte = readUnsignedByte.toByte
+
+    def readUnsignedByte: Int = {
+      val c1 = s.read
```
ianoc: could you put a type on c1? it looks like its int? So when you really read a byte from the underlying stream it gives you an unsigned byte in an int?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,221 @@
...
+        ((buf(5) & 255) << 16) +
+        ((buf(6) & 255) << 8) +
+        (buf(7) & 255)
+    }
+
+    def readShort: Short = {
+      val c1 = s.read
+      val c2 = s.read
+      // This is the algorithm from DataInputStream
+      if ((c1 | c2) < 0) eof else ((c1 << 8) | c2).toShort
```
ianoc: should this call readUnsignedShort ? looks the same modulo the to short at the end?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * This reads a varInt encoding that only encodes non-negative
+     * numbers. It uses:
+     * 1 byte for values 0 - 255,
+     * 3 bytes for 256 - 65535,
+     * 7 bytes for 65536 - Int.MaxValue
+     */
+    def writeSize(i: Int): Unit = {
+      require(i >= 0, s"sizes must be non-negative: ${i}")
```
ianoc: What is the overhead of requires do you know? will it block jit or just go away?

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -0,0 +1,123 @@
...
+    val toTry = Success(unsafeToInt)
+  }
+
+  def compare[T](a: T, b: T)(implicit ord: OrderedSerialization[T]): Int =
+    ord.compare(a, b)
+
+  def compareBinary[T](a: InputStream, b: InputStream)(implicit ord: OrderedSerialization[T]): Result =
+    ord.compareBinary(a, b)
+
+  def serializeThenCompare[T](a: T, b: T)(implicit ordb: OrderedSerialization[T]): Result =
```
ianoc: What seems to be a big one in testing that i've seen is to serialize N of these, and then compare. It covers a lot of overflow edge cases that start appearing in the real world. Just doing (0 until N) of a and B in this law and requiring them be distinct type thing. Not sure where you want to put it (object / tests).

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
```diff
@@ -0,0 +1,61 @@
...
+import scala.util.{ Failure, Success, Try }
+
+class Serialization2[A, B](val serA: Serialization[A], val serB: Serialization[B]) extends Serialization[(A, B)] {
+  private val MAX_PRIME = Int.MaxValue // turns out MaxValue is a prime, which we want below
+  override def hash(x: (A, B)) = serA.hash(x._1) + MAX_PRIME * serB.hash(x._2)
+  override def equiv(x: (A, B), y: (A, B)): Boolean =
+    serA.equiv(x._1, y._1) && serB.equiv(x._2, y._2)
+
+  override def read(in: InputStream): Try[(A, B)] =
+    (serA.read(in), serB.read(in)) match {
```
ianoc: since order matters here that these are side effecting it might make sense to pull it out before to make that clearer

scalding-core/src/main/scala/com/twitter/scalding/serialization/Writer.scala
```diff
@@ -0,0 +1,128 @@
...
+  implicit val long: Writer[Long] = new Writer[Long] {
+    def write(os: OutputStream, s: Long) = os.writeLong(s)
+  }
+  implicit val float: Writer[Float] = new Writer[Float] {
+    def write(os: OutputStream, s: Float) = os.writeFloat(s)
+  }
+  implicit val double: Writer[Double] = new Writer[Double] {
+    def write(os: OutputStream, s: Double) = os.writeDouble(s)
+  }
+  implicit val string: Writer[String] = new Writer[String] {
```
ianoc: I'm not sure i'm too keen on the distance between the Reader and Writer variants for a type like String. Though have to admit not sure the other options will really improve it once testing is solid.

scalding-commons/src/main/scala/com/twitter/scalding/commons/thrift/TProtocolOrderedSerialization.scala
ianoc: Thanks for the explanation, in light of that seems like this is totally the right idea

ianoc:
Looks great, few small things. The extra testing stuff can come later. Would like the serialization2 lines just made a bit more explicit, but modulo that looks good to merge. Nice work, very polished

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
johnynek: There is an unsignedShort one (that returns an Int, because Java). This one, will interpret the bytes as a signed number. Java does this for the low 16 bits of an Int

```scala
scala> 65000.toShort
res21: Short = -536
```

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
johnynek: Just an if followed by a throw. Nice thing is the message is lazy, so we don't evaluate that string interpolation unless we hit the exceptional case:

https://github.com/scala/scala/blob/2.10.x/src/library/scala/Predef.scala#L231

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
ianoc: I think we can still just can have

```
def readShort = readUnsignedShort.toShort
```
 ?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
ianoc: cool, sounds good

johnynek:
Thanks for the review. I think it is read to merge. When we get the macros in place, we'll be totally set.

One win of pulling this out into a package like scalding-serialization is that we could put scalding-serialization-macros there and actually include the macros in the default Job class right at the start. Might be better to be conservative in the first release though.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Writer.scala
johnynek: not sure I follow. You mean the code to read strings and write strings is far apart (so possibly easy to get out of sync)? If so, yeah, I see that point. Only testing seems like it can help us here.

ianoc:
Yeah i think we can spin it out into its own package later safely enough? Probably best to just get it out first then do that process.

One other thing have you fired this up on Hadoop to check for gremlins there?

johnynek:
I have not search for hadoop gremlins (and possibly the macros are not working correctly). :(

scalding-core/src/main/scala/com/twitter/scalding/serialization/Writer.scala
ianoc: Yeah exactly, normally one would put them beside each other with a read and write. But i guess that just gives you at a glance to read both, no real better enforcement they match. Testing and more testing

Issue #1121, opened 2014-12-06
Move summingbird LookupJoin implementation to scalding


isnotinvain:
The two implementations have diverged but the SB one has more features, we should just move the SB lookup join to scalding and have SB use that.

declerambaul:
Is the SB version of the LookupJoin stable? From a brief look it might be as easy as to copy the code to a package. In the meantime I would like to add a reducer option to the scalding version - I have forgotten to set the global reducer config one too many times.

isnotinvain:
according to @ianoc the one in SB just went through some more changes, so maybe not quite stable yet.

johnynek:
The code was changed to be a bit more clear, but I think we have tests on it, so I am not worried about correctness.

isnotinvain:
by stable I meant 'not changing often' -- I think @ianoc has said that if it needs frequent changes moving it to scalding will be a pain for summingbird development. So if we think the code won't need to be changed often to support summingbird, then we can move it.

johnynek:
Yeah, that's the conclusion we came to before. Putting this with scalding is a pain since it is really used in summingbird. I don't see periodic backports as a disaster.

johnynek:
That said, I don't think summingbird-scalding depends on anything but summingbird-core and scalding. So, maybe we should just encourage people to use that.

isnotinvain:
that will get funky when / if SB depends on an older version of scalding... does that ever happen?

johnynek:
it is never older than the one we have in production at Twitter. But yes,
you could imagine scalding advancing beyond summingbird.

On Wed, Jan 14, 2015 at 2:15 PM, Alex Levenson <notifications@github.com>
wrote:

> that will get funky when / if SB depends on an older version of
> scalding... does that ever happen?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/scalding/issues/1121#issuecomment-70018807>.
>



-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

PR #1160, opened 2015-01-14
add reducer option to LookupJoin

declerambaul:
There is an open issue to consolidate the summingbird/scalding LookupJoin implementations, https://github.com/twitter/scalding/issues/1121, though in the meantime I would like to add a reducer option.


isnotinvain:
LGTM, thanks!

PR #1164, opened 2015-01-17
Bug fix, and modification.

ianoc:
Removes mark support from the PositionInputStream.

Add in a helper method to figure out how many bytes the length encoding of a given int would take


ianoc:
@johnynek this is the pre-PR, i'll put up the other one off this one to make its diff easier to read

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -24,6 +24,17 @@ object JavaStreamEnrichments {
     def toInputStream: ByteArrayInputStream = new ByteArrayInputStream(baos.toByteArray)
   }
 
+  def sizeBytes(i: Int): Int = {
```
johnynek: I'd rather not the outer { }.

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -156,17 +167,11 @@ object JavaStreamEnrichments {
     def writeBoolean(b: Boolean): Unit = if (b) s.write(1: Byte) else s.write(0: Byte)
 
     def writeBytes(b: Array[Byte], off: Int, len: Int): Unit = {
-      @annotation.tailrec
-      def go(o: Int, count: Int): Unit =
-        if (count == 0) ()
-        else {
-          s.write(b(o))
```
johnynek: this was dumb on my part.

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
```diff
@@ -27,20 +27,23 @@ object PositionInputStream {
...
   def position: Long = pos
 
   override def available = wraps.available
+
   override def close() { wraps.close() }
+
   override def mark(limit: Int) {
-    wraps.mark(limit)
-    markPos = pos
+    sys.error("Mark and reset are not supported on this stream")
```
johnynek: why did you remove this? Any reason the code I had would not have worked?

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
```diff
@@ -27,20 +27,23 @@ object PositionInputStream {
...
   }
-  override def markSupported: Boolean = wraps.markSupported
+
+  override val markSupported: Boolean = false
+
   override def read: Int = {
-    val count = wraps.read
-    if (count > 0) pos += 1
-    count
+    val result = wraps.read
```
johnynek: good catch here. I had some moment of confusion and thought, like read on bytes, that it was returning the count. I realized it was not later (because I had to write `pos += 1` not `pos += count`, but I failed fix the `== 0` case.

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
```diff
@@ -49,13 +52,15 @@ class PositionInputStream(val wraps: InputStream) extends InputStream {
...
-    wraps.reset()
-    pos = markPos
+    sys.error("Mark and reset are not supported on this stream")
   }
+
   override def skip(n: Long): Long = {
+    require(n >= 0, "Must seek fowards")
     val count = skip(n)
-    pos += count
+    if (count > 0) pos += count
```
johnynek: I don't think you need the if guard. `n >= 0` so `count >= 0`, right?

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
```diff
@@ -22,7 +22,67 @@ import scala.util.control.NonFatal
 
 import JavaStreamEnrichments._
 
+object StringOrderedSerialization {
+  final def binaryIntCompare(leftSize: Int, seekingLeft: InputStream, rightSize: Int, seekingRight: InputStream): Int = {
```
johnynek: what is the `Int` here? why not just `compareBinary` to match the terminology we normally use?

ianoc: This is moving a function you already had which returned int, though it suits my purposes that it returns Int so it matches all the standard compare functions.

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
ianoc: The thinking here was that using mark and reset in passed around things was unsafe, and the code I had that used it would break in a meta sense. So making the inner input streams not support it allowed me to hint that only the outermost ones should use it. That said I'm not sure in the current vamp its actually a required change anymore... generally the approach I have at the moment looks like::
```
 def compareBinary($inputStreamA: _root_.java.io.InputStream, $inputStreamB: _root_.java.io.InputStream): _root_.com.twitter.scalding.serialization.OrderedSerialization.Result = {
          try {

            val $lenA = ${readLength(inputStreamA)}
            require($lenA >= 0, "Length was " + $lenA + "which is < 0, invalid")
            val $lenB = ${readLength(inputStreamB)}
            require($lenB >= 0, "Length was " + $lenB + "which is < 0, invalid")


            val subSetRes = if(($lenA == $lenB) && $lenA > 24 && $inputStreamA.markSupported && $inputStreamB.markSupported) {
              $inputStreamA.mark($lenA)
              $inputStreamB.mark($lenB)

              var pos = 0
              var isSame = true

              while(pos < $lenA && isSame == true) {
                $inputStreamA.readByte == $inputStreamB.readByte
                pos = pos + 1
              }

              if(isSame) {
                0
              } else {
                // rewind if they don't match for doing the full compare
                $inputStreamA.reset()
                $inputStreamA.reset()
                -1
              }
            } else {
              -1
            }

            val r = if(subSetRes == 0) {
              0
              } else {
              val $bufferedStreamA = _root_.com.twitter.scalding.serialization.PositionInputStream($inputStreamA)
              val initialPositionA = $bufferedStreamA.position
              val $bufferedStreamB = _root_.com.twitter.scalding.serialization.PositionInputStream($inputStreamB)
              val initialPositionB = $bufferedStreamB.position

              val innerR = ${t.compareBinary(bufferedStreamA, bufferedStreamB)}

              $bufferedStreamA.seekToPosition(initialPositionA + $lenA)
              $bufferedStreamB.seekToPosition(initialPositionB + $lenB)
              innerR
            }

            _root_.com.twitter.scalding.serialization.OrderedSerialization.resultFrom(r)
            }
            catch { case _root_.scala.util.control.NonFatal(e) =>
              _root_.com.twitter.scalding.serialization.OrderedSerialization.CompareFailure(e)
            }
          }

```

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
ianoc: Alas no, by the input stream docs 0 is also any negative number too

PR #1168, opened 2015-01-22
Fixes up the boxes, with more boxes too

ianoc:
Also stop kryo being allowed for boxes

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
@@ -18,6 +18,9 @@ package com.twitter.scalding.serialization
 import java.util.concurrent.atomic.AtomicReference
 import java.io.{ InputStream, OutputStream }
 
+import com.esotericsoftware.kryo.{ Serializer => KSerializer, DefaultSerializer, Kryo }
+import com.esotericsoftware.kryo.io.{ Input, Output }
+
 /**
  * This interface is a way of wrapping a value in a marker class
  * whose class identity is used to control which serialization we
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -50,7 +358,103 @@ object Boxed {
...
+    ({ t: Any => new Boxed90(t) }, classOf[Boxed90[Any]]),
+    ({ t: Any => new Boxed91(t) }, classOf[Boxed91[Any]]),
+    ({ t: Any => new Boxed92(t) }, classOf[Boxed92[Any]]),
+    ({ t: Any => new Boxed93(t) }, classOf[Boxed93[Any]]),
+    ({ t: Any => new Boxed94(t) }, classOf[Boxed94[Any]]),
+    ({ t: Any => new Boxed95(t) }, classOf[Boxed95[Any]]),
+    ({ t: Any => new Boxed96(t) }, classOf[Boxed96[Any]]),
+    ({ t: Any => new Boxed97(t) }, classOf[Boxed97[Any]]),
+    ({ t: Any => new Boxed98(t) }, classOf[Boxed98[Any]]),
+    ({ t: Any => new Boxed99(t) }, classOf[Boxed99[Any]]))
```
johnynek: can you change the type below on next:

`def next[K]: (K => Boxed[K], Class[_ <: Boxed[K]])`

That is actually more correct since Class is invariant.

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -39,11 +40,14 @@ class WrappedSerialization[T] extends HSerialization[T] with Configurable {
     serializations = WrappedSerialization.getBinary(config)
   }
 
-  def accept(c: Class[_]): Boolean =
-    serializations.map(_.exists { case (cls, _) => cls == c }).getOrElse(false)
+  def accept(c: Class[_]): Boolean = {
+    val willAccept = serializations.exists { case (cls, _) => cls == c }
+    serializations.map(_._1).foreach(println)
```
johnynek: let's not println if it is working.

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -99,16 +103,22 @@ object WrappedSerialization {
...
+      .asScala
+      .map { it =>
+        (it.getKey, it.getValue)
       }
-    }
+      .filter(_._1.startsWith(confKey))
+      .map {
+        case (_, clsbuf) =>
+          clsbuf.split(":") match {
+            case Array(className, bufferable) =>
```
johnynek: change bufferable to serialization or something.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -104,8 +104,27 @@ object Grouped {
       val keyF = new Fields("key")
       keyF.setComparator("key", new CascadingBinaryComparator(boxordSer))
       val pipe = op(ts, keyF)
+
+      @annotation.tailrec
+      def go(p: Pipe, visited: Set[Pipe], toVisit: List[Pipe]): Set[Pipe] = {
+        val notSeen: Set[Pipe] = p.getPrevious.filter(i => !visited.contains(i)).toSet
+        val nextVisited: Set[Pipe] = visited + p
+        val nextToVisit: List[Pipe] = (toVisit ++ notSeen).distinct
```
johnynek: this will hurt complexity if we get too large. Might be better to keep a `Set[Pipe]` too:

```scala
import scala.collection.immutable.Queue

case class ToVisit[T](queue: Queue[T], inQueue: Set[T]) {
  def maybeAdd(t: T): ToVisit[T] = if (inQueue(t)) this else {
   ToVisit(queue :+ t, inQueue + t)
  }
  def next: Option[(T, ToVisit[T])] =
    if (inQueue.isEmpty) None
    else Some(queue.head, ToVisit(queue.tail, toVisit - queue.head))
}
```

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
ianoc: That i have trouble working in WrappedSerialization, since this spreads out over there as this return type is basically passed in. Unless we want to erase it to Class[_] grouped maybe?

johnynek: never mind then. It's all casts anyway.

PR #1174, opened 2015-01-27
 Use latest algebird, bijection, chill, elephantbird, and scala 2.11.5

isnotinvain:
(creating this also so I can see the tests run)

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -15,7 +15,7 @@ limitations under the License.
 */
 package com.twitter.scalding.typed
 
-import com.twitter.algebird.{ CMS, MurmurHash128 }
+import com.twitter.algebird.{ CMS, CMSHasherImplicits, MurmurHash128 }
 
 /**
  * This class is generally only created by users
@@ -28,14 +28,15 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
   seed: Int)(implicit serialization: K => Array[Byte],
     ordering: Ordering[K])
   extends HasReducers {
+  import CMSHasherImplicits._
 
   val reducers = Some(numReducers)
 
   private lazy val murmurHash = MurmurHash128(seed)
   def hash(key: K): Long = murmurHash(serialization(key))._1
 
-  private lazy implicit val cms = CMS.monoid(eps, delta, seed)
-  lazy val sketch: TypedPipe[CMS] =
+  private lazy implicit val cms = CMS.monoid[Long](eps, delta, seed)
+  lazy val sketch: TypedPipe[CMS[Long]] =
     pipe
       .map{ kv => cms.create(hash(kv._1)) }
       .groupAll
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -15,7 +15,32 @@ limitations under the License.
 */
 package com.twitter.scalding.typed
 
-import com.twitter.algebird.{ CMS, CMSHasherImplicits, MurmurHash128 }
+import com.twitter.algebird.{ CMS, CMSHasher }
+
+object Sketched {
+
+  // TODO: there are more efficient orderings we could use here if this turns
+  // out to be a bottleneck.
+  implicit val byteArrayOrdering = Ordering.by((_: Array[Byte]).toIterable)
+
+  /**
+   * This is based on the CMSHasherBigInt found in algebird (see docs for in depth explanation):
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -28,14 +28,15 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
...
+  import CMSHasherImplicits._
 
   val reducers = Some(numReducers)
 
   private lazy val murmurHash = MurmurHash128(seed)
   def hash(key: K): Long = murmurHash(serialization(key))._1
 
-  private lazy implicit val cms = CMS.monoid(eps, delta, seed)
-  lazy val sketch: TypedPipe[CMS] =
+  private lazy implicit val cms = CMS.monoid[Long](eps, delta, seed)
```
johnynek: actually it is better to fix this to be a `CMS[Array[Byte]]` and make implicit hasher following this approach:

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1113

The way we had it before is because we didn't have this ability.

isnotinvain: OK thanks will do

johnynek:
Just did a minor fix I forgot about now that we are on the latest algebird.

johnynek:
looks like the build is failing due to something with elephantbird. Did they change their packaging? Looks like a bunch of types are not found.

isnotinvain:
I had that issue locally, and it went away after a `git clean -fdx` so I thought it was a local environment issue. I'll take a look

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
isnotinvain: Does this have any implications for persisted data, or is this OK to change?

isnotinvain:
This is strange, EB didn't change it's packaging. Those missing classes are from elephantbird-core, and the test failures are in scalding-commons, which depends on elephantbird-cascading2, and elephantbird-cascading2 depends on elephantbird-core. I don't see anything turning off transitive dependencies, so that is strange. 

At least locally, adding EB-core explicitly to scalding-commons seems to fix it. We should be listing this explicitly because we use it directly anyway, but I am puzzled as to why this didn't work as is (or what change triggered it).

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
johnynek: Those CMS values are not persisted beyond a single join.  Of course,
someone could do something crazy and save them, but the migration is their
problem at that point.

On Tue, Jan 27, 2015 at 11:06 AM, Alex Levenson <notifications@github.com>
wrote:

> In scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
> <https://github.com/twitter/scalding/pull/1174#discussion_r23642791>:
>
> >
> >    val reducers = Some(numReducers)
> >
> >    private lazy val murmurHash = MurmurHash128(seed)
> >    def hash(key: K): Long = murmurHash(serialization(key))._1
> >
> > -  private lazy implicit val cms = CMS.monoid(eps, delta, seed)
> > -  lazy val sketch: TypedPipe[CMS] =
> > +  private lazy implicit val cms = CMS.monoid[Long](eps, delta, seed)
>
> Does this have any implications for persisted data, or is this OK to
> change?
>
> —
> <https://github.com/twitter/scalding/pull/1174/files#r23642791>.
>



-- 

isnotinvain: K thanks just wanted to double check

isnotinvain:
@johnynek @ianoc I think this is ready assuming the tests pass, let me know what you think, thanks!

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -28,16 +54,16 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
   seed: Int)(implicit serialization: K => Array[Byte],
     ordering: Ordering[K])
   extends HasReducers {
+  import Sketched._
 
-  val reducers = Some(numReducers)
+  def serialize(k: K): Array[Byte] = serialization(k)
```
johnynek: why is this here?

johnynek: ahh, because hash is no longer public.

isnotinvain: actually I removed the hash() method entirely, we actually need to share the serialization function with the estimation code that happens below so we can pass the right key into the cms

PR #1175, opened 2015-01-27
parquet-avro support in Scalding

jonbender:
Would love to have parquet-avro support in Scalding, have been using this locally and it has worked great.  Is this something that you'd consider merging?  Or are we hung up on waiting for Scala 2.10+ in the main Scalding repo?

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -17,8 +17,7 @@ object ScaldingBuild extends Build {
   val sharedSettings = Project.defaultSettings ++ assemblySettings ++ scalariformSettings ++ Seq(
     organization := "com.twitter",
 
-    //TODO: Change to 2.10.* when Twitter moves to Scala 2.10 internally
-    scalaVersion := "2.9.3",
+    scalaVersion := "2.10.4",
 
     crossScalaVersions := Seq("2.9.3", "2.10.4"),
 
@@ -207,6 +206,7 @@ object ScaldingBuild extends Build {
   val bijectionVersion = "0.6.3"
   val chillVersion = "0.4.0"
   val slf4jVersion = "1.6.6"
+  val parquetVersion = "1.5.0"
 
   lazy val scaldingCore = module("core").settings(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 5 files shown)


ianoc:
This doesn't look like its mergable ?  Scalding is on 2.10 and 2.11 these days so that shouldn't be an issue

project/Build.scala
```diff
@@ -258,14 +258,15 @@ object ScaldingBuild extends Build {
 
   lazy val scaldingParquet = module("parquet").settings(
     libraryDependencies ++= Seq(
-      "com.twitter" % "parquet-cascading" % "1.4.0",
+      "com.twitter" % "parquet-cascading" % parquetVersion,
+      "com.twitter" % "parquet-avro" % parquetVersion,
```
ianoc: This is some dependency pollution, all parquet users will end up picking this up. Maybe mark this as provided. Downstream users who want to use avro will then need to bring their own parquet-avro

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/avro/ParquetAvroSource.scala
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+  }
+
+  def sinkConfInit(flowProcess: FlowProcess[JobConf], tap: Tap[JobConf, RecordReader[NullWritable, Container[A]], OutputCollector[NullWritable, A]], conf: JobConf): Unit = {
+    conf.set(ParquetOutputFormat.WRITE_SUPPORT_CLASS, classOf[AvroWriteSupport].getName)
+    conf.setOutputFormat(classOf[DeprecatedParquetOutputFormat[A]])
+    // Constants from AvroWriteSupport.java
+    conf.set("parquet.avro.schema", schemaString)
+    conf.set("parquet.compression", "SNAPPY")
```
ianoc: Why are these being set here/hard coded? What is AvroWriteSupport.java?

jonbender:
Thanks for taking a look, I'm happy to open another PR off my own fork and get it to a mergeable state / move out the hard coded bits.  As this was originally @epishkin 's I figured I would just open the PR as is.

project/Build.scala
isnotinvain: I think it would make more sense to split into a separate sub module per format, so that dependencies don't grow here.

ianoc: We might end up with module creep in dependencies with lots of modules then
for all combinations. Could we maybe go the provided route to keep them
separate ? It's own form of uglyness... But avoids build file growing hugely

isnotinvain: well maybe go the other direction then -- treat parquet as a common thing, and push this up into scalding-avro, and put parquet-thrift in scalding-thrift?

PR #1179, opened 2015-01-29
Covert LzoTextDelimited to Cascading scheme.

reconditesea:
Just to be extremely safe.

### DIFF
--- a/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoTraits.scala
+++ b/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoTraits.scala
@@ -85,12 +85,12 @@ trait LzoText extends LocalTapSource with SingleMappable[String] with TypedSink[
 }
 
 trait LzoTsv extends DelimitedScheme with LocalTapSource {
-  override def hdfsScheme = HadoopSchemeInstance(new LzoTextDelimited(fields, skipHeader, writeHeader, separator, strict, quote, types, safe))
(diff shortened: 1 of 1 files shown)


johnynek:
so, to review, we are casting it to a specific kind of Scheme

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Source.scala#L60

Can you describe how this is increasing safety? I don't understand that.

reconditesea:
@johnynek  By "to be safe", I mean to make sure it will work. I remember seeing this issue before that without calling asInstanceOf[Scheme], the compiler will complain a typeMismatch in local environment. I discussed with @ianoc on this offline. We both saw this issue before but not sure exactly why sometimes this casting is needed and sometimes is not. So I guess it's more conservative to add it here.

ianoc:
Alas we've had to put these in almost everywhere internally, and there are a few others in OSS already.

johnynek:
My concern was twofold:

1) all `HadoopSchemeInstance` is, is a cast. So we don't need two casts .asInstanceOf[A].asInstanceOf[B]. Just put the last one.

2) given that this compiles, I'm confused how this could be a problem. I have seen the need to cast before, but to get it to compile. If we need a cast, why can't the strange situation do: super.hdfsScheme.asInstanceOf[...] Why put the voodoo in the base where it will be cargo culted everywhere?

ianoc:
The strange situation is our common one internally. Though consuming
external ones which don't have the cast (some do, some don't?), seems to
work ok too.

PR #1180, opened 2015-01-29
Make TraceUtil support versions of cascading older than 2.6

isnotinvain:
This was the original intention, but unfortunately it still triggers trying to load a class that doesn't exist pre 2.6


isnotinvain:
The logs look like this:

ianoc:
LGTM merge when green

PR #1183, opened 2015-02-02
update the instance creation of CHTextDelimited in hdfsScheme function i...

jeremieSimon:
...n the  DelimitedScheme trait. Use DEFAULT mode to allow compression of the output. Default will only compress the output if the user is overwritting the config and set the property mapred.output.compress to true


johnynek:
I don't want to break anyone with this change, but this would certainly break us internally, as I imagine others that have come to rely on this. What I'd suggest is that you ether make a new trait and update the PR, or you just keep this trait in your project and use it there.

If you do update the PR: we could consider deprecating the old one.

In general, it is nice to clean up design warts, but the cost to migrate to the non-warty version will sometimes limit how fast we can make the change.

jeremieSimon:
I see. Actually, we already created that trait for our projects, I was just trying to help so that other people won't have this problem. I can definitely create a new trait if you think it's any useful. 

Thanks

PR #1186, opened 2015-02-03
Use java.util.Random instead of scala.util.Random

johnynek:
This can cause an issue in jobs that can't be serialized with Kryo using the Externalizer.


isnotinvain:
+1

PR #1187, opened 2015-02-04
Back out 4 changes to be binary compatible

johnynek:
a few of these are binary surprises (a private method and some closure captured vals), but a few others we should have known.

The way these are used is safe (lazy or otherwise). Not worth it to break binary compatibility again.


isnotinvain:
+1

PR #1189, opened 2015-02-04
add .unit to Execution object

rubanm:


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -423,6 +423,9 @@ object Execution {
    */
   def fromFuture[T](fn: ConcurrentExecutionContext => Future[T]): Execution[T] = FutureConst(fn)
 
+  /** This creates a constant Execution[Unit] */
+  def unit: Execution[Unit] = from(())
```
isnotinvain: could be a val right?

reconditesea: .from is a def, what will be benefit of making this a val?

johnynek:
merge when green IMO

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: If it is a val, then you only have one of them, and don't reallocate them
(that said, Executions are not in the hot path, so no big deal).

On Wed, Feb 4, 2015 at 12:54 PM, Kevin Lin <notifications@github.com> wrote:

> In scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
> <https://github.com/twitter/scalding/pull/1189#discussion_r24128280>:
>
> > @@ -423,6 +423,9 @@ object Execution {
> >     */
> >    def fromFuture[T](fn: ConcurrentExecutionContext => Future[T]): Execution[T] = FutureConst(fn)
> >
> > +  /** This creates a constant Execution[Unit] */
> > +  def unit: Execution[Unit] = from(())
>
> .from is a def, what will be benefit of making this a val?
>
> —
> <https://github.com/twitter/scalding/pull/1189/files#r24128280>.
>



-- 

PR #1190, opened 2015-02-05
Override hashCode for Args

stephbian:
Currently, Args overrides equals but does not override hashCode.
This leads to unexpected behavior when hashing and checking for equality
on objects that depend on Args.


johnynek:
Wow. Sorry about that one.

PR #1193, opened 2015-02-10
Move main simple example to be the typed API, and put the .'s at the sta...

ianoc:
...rt of lines like we usually do

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
@@ -4,8 +4,9 @@ import com.twitter.scalding._
 
 class WordCountJob(args: Args) extends Job(args) {
   TypedPipe.from(TextLine(args("input")))
-    .flatMap { line: String => line.split("\\s+") }
-    .map (_ -> 1L)
+    .flatMap { line  => line.split("\\s+") }
+    .map { word => (word, 1L) }
     .sumByKey
+     // The compiler will enforce the type coming out of the sumByKey is the same as the type we have for our sink
     .write(TypedTsv[(String, Long)](args("output")))
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
@@ -4,9 +4,9 @@ import com.twitter.scalding._
 
 class WordCountJob(args: Args) extends Job(args) {
   TypedPipe.from(TextLine(args("input")))
-    .flatMap { line  => line.split("\\s+") }
+    .flatMap { line => line.split("\\s+") }
     .map { word => (word, 1L) }
     .sumByKey
-     // The compiler will enforce the type coming out of the sumByKey is the same as the type we have for our sink
+    // The compiler will enforce the type coming out of the sumByKey is the same as the type we have for our sink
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
```diff
@@ -3,8 +3,9 @@ package com.twitter.scalding.examples
 import com.twitter.scalding._
 
 class WordCountJob(args: Args) extends Job(args) {
-  TextLine(args("input")).read.
-    flatMap('line -> 'word) { line: String => line.split("\\s+") }.
-    groupBy('word) { _.size }.
-    write(Tsv(args("output")))
+  TypedPipe.from(TextLine(args("input")))
+    .flatMap { line: String => line.split("\\s+") }
```
johnynek: you want to drop the type of line? `.flatMap { line => line.split("\\s+") }`

Also, I think `.map(_ -> 1L)` is a bit too compact for an example. Can we do: `.map { word => (word, 1L) }`?

scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
```diff
@@ -3,8 +3,9 @@ package com.twitter.scalding.examples
...
+    .map (_ -> 1L)
+    .sumByKey
+    .write(TypedTsv[(String, Long)](args("output")))
```
johnynek: you could drop the types here, or comment that we list the types to get the compiler to error if we are writing something other than what we expect.

ianoc: Useful to put a comment on each line saying the type of the typed pipe there?

ianoc: Another variant of this in one of your sets of slides has a custom tokenizer method, should that be included to show udf calling?

johnynek: if you you like, as it is a tutorial, but it could get confusing since .sumByKey will return an UnsortedGrouped.

ianoc: I would have relied on the implicit back to typed pipe and pretended it was back as the pipe. But yes, might be better without...

johnynek:
This looks good to me (given all the other material we have). merge when green.

scalding-core/src/main/scala/com/twitter/scalding/examples/WordCountJob.scala
```diff
@@ -3,8 +3,10 @@ package com.twitter.scalding.examples
 import com.twitter.scalding._
 
 class WordCountJob(args: Args) extends Job(args) {
-  TextLine(args("input")).read.
-    flatMap('line -> 'word) { line: String => line.split("\\s+") }.
-    groupBy('word) { _.size }.
-    write(Tsv(args("output")))
+  TypedPipe.from(TextLine(args("input")))
+    .flatMap { line  => line.split("\\s+") }
```
isnotinvain: could add:
`// the compiler knows line is a string, so we can call split without specifying a type unlike in the fields API`

isnotinvain:
LGTM

johnynek:
This build failure is legit. There is a unit test against this job which is now failing.

PR #1194, opened 2015-02-10
Add an exclusiveUpper method to DateRange

johnynek:
DateRanges are inclusive, and this causes a ton of pain. Mea culpa, but it is too much of an issue to break this now.


ianoc:
LGTM, merge when green

Issue #1192, opened 2015-02-08
A wiki for tutorials and usage tips of Execution


reconditesea:
Execution is still a relatively new feature and there are certain tips of how to use it correctly and efficiently. e.g. when to use .waitFor and when to use .run. Difference between .onComplete and .unit. 

There are comments in the code and scaladoc. But still we see many confusions both internal  and external. It would be really great if we can have a new wiki (or maybe update this one: https://github.com/twitter/scalding/wiki/Calling-Scalding-from-inside-your-application) with more tutorials and tips.

jarutis:
That would be of great help indeed.

johnynek:
So let's not overlook: http://twitter.github.io/scalding/#com.twitter.scalding.Execution

The docs there are pretty good, I think.

One problem we have is that everyone wants all the docs all the ways. Wiki's are not the greatest because you can't see all the wiki's that refer to code when you are editting that code, so it gets out of sync: it is not in CI, etc...

The scaladoc at least is right there with the code. I hope users will look at that and only use the wiki as a getting started example point.

Please feel free to edit the wiki and report back to this page. I'll add a couple of comments now.

johnynek:
I updated the wiki with some comments.

reconditesea:
@johnynek  Thanks, I think those comments are really helpful and answered many questions..
I still had only 1 confusion below:

"2. Avoid calling .waitFor or .run AS LONG AS POSSIBLE. Try to compose your entire job into on large Execution using .zip or .flatMap to combine Executions. waitFor is the same as run except it waits on the future."
Here we said avoid using .waitFor and .run. But in the examples above, we both use .waitFor to eventually kick off the jobs. I guess what is missing here is: we prefer using Execution composition to chain all different Executions. But on the end of the day, how should we start the final large Execution? Is it fair to say that there will be 1 (and should be only 1) calling to .waitFor or .run in each Execution App/Job?

johnynek:
Yes. That is a good point: eventually you have to .run/.waitFor an
Execution, but the goal is, as you say, that EXACTLY one Execution do we
call .waitFor/.run. If you call .run on even two, you should try to instead
combine those two Executions with zip (to run in parallel) or flatMap (to
run them in a sequence).

On Tue, Feb 10, 2015 at 9:14 AM, Kevin Lin <notifications@github.com> wrote:

> @johnynek <https://github.com/johnynek> Thanks, I think those comments
> are really helpful and answered many questions..
[...]

PR #1196, opened 2015-02-12
Add a new ExecutionApp tutorial

reconditesea:
Create a tutorial of using Execution in a stand-alone scala/java application.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,6 +100,5 @@ matrix:
     - scala: 2.10.4
       env: BUILD="test execution tutorials"
       script:
-      - "scripts/build_assembly_no_test.sh scalding-core"
       - "scripts/build_assembly_no_test.sh execution-tutorial"
       - "scripts/test_execution_tutorial.sh"
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,5 +97,11 @@ matrix:
       - "scripts/build_assembly_no_test.sh scalding-core"
       - "scripts/test_typed_tutorials.sh"
 
+    - scala: 2.10.4
+      env: BUILD="test execution tutorials"
+      script:
+      - "scripts/build_assembly_no_test.sh execution-tutorial"
+      - "scripts/test_execution_tutorial.sh"
+
 notifications:
   irc: "chat.freenode.net#scalding"
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,6 +100,7 @@ matrix:
     - scala: 2.10.4
       env: BUILD="test execution tutorials"
       script:
+      - "scripts/build_assembly_no_test.sh scalding-core"
       - "scripts/build_assembly_no_test.sh execution-tutorial"
       - "scripts/test_execution_tutorial.sh"
 
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -438,14 +438,10 @@ object ScaldingBuild extends Build {
     settings = sharedSettings
   ).settings(
     name := "execution-tutorial",
-    previousArtifact := None,
-    crossPaths := false,
-    autoScalaLibrary := false,
     libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
       "org.scala-lang" % "scala-library" % scalaVersion,
       "org.scala-lang" % "scala-reflect" % scalaVersion,
-      "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
-      "org.apache.hadoop" % "hadoop-minicluster" % hadoopVersion,
+      "org.apache.hadoop" % "hadoop-core" % hadoopVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion,
       "cascading" % "cascading-hadoop" % cascadingVersion
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -201,7 +201,8 @@ object ScaldingBuild extends Build {
     scaldingJdbc,
     scaldingHadoopTest,
     scaldingMacros,
-    maple
+    maple,
+    executionTutorial
   )
 
   lazy val formattingPreferences = {
@@ -430,4 +431,25 @@ object ScaldingBuild extends Build {
     )
     }
   )
+
+  lazy val executionTutorial = Project(
+    id = "execution-tutorial",
+    base = file("tutorial/execution-tutorial"),
+    settings = sharedSettings
+  ).settings(
+    name := "execution-tutorial",
+    previousArtifact := None,
+    crossPaths := false,
+    autoScalaLibrary := false,
+    libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
+      "org.scala-lang" % "scala-library" % scalaVersion,
+      "org.scala-lang" % "scala-reflect" % scalaVersion,
+      "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
+      "org.apache.hadoop" % "hadoop-minicluster" % hadoopVersion,
+      "org.slf4j" % "slf4j-api" % slf4jVersion,
+      "org.slf4j" % "slf4j-log4j12" % slf4jVersion,
+      "cascading" % "cascading-hadoop" % cascadingVersion
+    )
+    }
+  ).dependsOn(scaldingCore)
 }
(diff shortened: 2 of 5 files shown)


reconditesea:
To close Issue #1192

project/Build.scala
```diff
@@ -430,4 +431,25 @@ object ScaldingBuild extends Build {
     )
     }
   )
+
+  lazy val executionTutorial = Project(
```
johnynek: do we need a whole separate project here?

tutorial/execution-tutorial/ExecutionTutorial.scala
```diff
@@ -0,0 +1,64 @@
...
+Tutorial of using Execution
+
+This tutorial gives an example of use Execution to do MapReduce word count.
+Instead of writing the results in reducers, it writes the data at submitter node.
+
+To test it, first build the assembly jar from root directory:
+ ./sbt execution-tutorial/assembly
+
+Run:
+  scala -classpath  tutorial/execution-tutorial/target/execution-tutorial-assembly-0.13.1.jar \
```
johnynek: can we add this to the script that runs the tutorials to make sure this stays tested?

tutorial/execution-tutorial/ExecutionTutorial.scala
```diff
@@ -0,0 +1,64 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.tutorial
+
+import java.io._
+import scala.util.{Failure, Success}
+
+import com.twitter.scalding._
+import com.twitter.algebird.monad.Reader
```
johnynek: where is this used?

tutorial/execution-tutorial/ExecutionTutorial.scala
```diff
@@ -0,0 +1,64 @@
...
+  
+  override def job = Execution.getConfig.flatMap { config =>
+    val args = config.getArgs
+    val file = new PrintWriter(new File(args("output")))
+    
+    TypedPipe.from(TextLine(args("input")))
+      .flatMap(_.split("\\s+"))
+      .map((_, 1L))
+      .sumByKey
+      .toIterableExecution
```
johnynek: can we add a comment that we could have gotten a similar effect with:

```.writeExecution(TypedTsv(args("output")))```

File: project/Build.scala (hunk 1 above)
Comment (reconditesea) [inline project/Build.scala]: We don't have a project for tutorial yet. Not sure if that's designed. I created this seprate project for execution only, because in the example I want to show how to run the job as a pure scala/java application. Therefore I need the com.twitter.scalding.tutorial package to be built into the jar.

File: tutorial/execution-tutorial/ExecutionTutorial.scala (hunk 3 above)
Comment (reconditesea) [inline tutorial/execution-tutorial/ExecutionTutorial.scala]: Removed

File: tutorial/execution-tutorial/ExecutionTutorial.scala (hunk 2 above)
Comment (reconditesea) [inline tutorial/execution-tutorial/ExecutionTutorial.scala]: Added

File: tutorial/execution-tutorial/ExecutionTutorial.scala (hunk 4 above)
Comment (reconditesea) [inline tutorial/execution-tutorial/ExecutionTutorial.scala]: Added.

File: .travis.yml [hunk 5]
```diff
@@ -97,5 +97,12 @@ matrix:
       - "scripts/build_assembly_no_test.sh scalding-core"
       - "scripts/test_typed_tutorials.sh"
 
+    - scala: 2.10.4
+      env: BUILD="test execution tutorials"
+      script:
+      - "scripts/build_assembly_no_test.sh scalding-core"
```
johnynek: do we need to do this? I think just running the assembly on the next target will do.

project/Build.scala
```diff
@@ -430,4 +431,25 @@ object ScaldingBuild extends Build {
...
+  ).settings(
+    name := "execution-tutorial",
+    previousArtifact := None,
+    crossPaths := false,
+    autoScalaLibrary := false,
+    libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
+      "org.scala-lang" % "scala-library" % scalaVersion,
+      "org.scala-lang" % "scala-reflect" % scalaVersion,
+      "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
+      "org.apache.hadoop" % "hadoop-minicluster" % hadoopVersion,
```
johnynek: do we really depend on this?

project/Build.scala
```diff
@@ -430,4 +431,25 @@ object ScaldingBuild extends Build {
...
+    settings = sharedSettings
+  ).settings(
+    name := "execution-tutorial",
+    previousArtifact := None,
+    crossPaths := false,
+    autoScalaLibrary := false,
+    libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
+      "org.scala-lang" % "scala-library" % scalaVersion,
+      "org.scala-lang" % "scala-reflect" % scalaVersion,
+      "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
```
johnynek: I don't think this is provided for this case. You will need it to run.

project/Build.scala
```diff
@@ -430,4 +431,25 @@ object ScaldingBuild extends Build {
...
+
+  lazy val executionTutorial = Project(
+    id = "execution-tutorial",
+    base = file("tutorial/execution-tutorial"),
+    settings = sharedSettings
+  ).settings(
+    name := "execution-tutorial",
+    previousArtifact := None,
+    crossPaths := false,
+    autoScalaLibrary := false,
```
johnynek: why is this here? It is unusual. Can you comment on this and the above?

scripts/test_execution_tutorial.sh
```diff
@@ -0,0 +1,20 @@
+set -e # first error should stop execution of this script
+
+# Identify the bin dir in the distribution, and source the common include script
+BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/.. && pwd )"
+source ${BASE_DIR}/scripts/common.sh
+
+# also trap errors, to reenable terminal settings
+trap onExit ERR
+export CLASSPATH=tutorial/execution-tutorial/target/execution-tutorial-assembly-0.13.1.jar
+time $SCALD \
```
johnynek: can we not run with scald.rb? A main goal here is to just run like a normal JVM app. Why can't we run this using `java jar`? That should be the full demo, in my view.

tutorial/execution-tutorial/ExecutionTutorial.scala
```diff
@@ -0,0 +1,65 @@
...
+        case Success(iter) => 
+          val file = new PrintWriter(new File(args("output")))
+          iter.foreach { case (k, v) =>
+              file.write(s"$k\t$v\n")
+          }
+          file.close
+        case Failure(e) => println("Error: " + e.toString)
+        }
+      }
+      .unit
```
isnotinvain: maybe comment why you need .unit?

.travis.yml
reconditesea: Without build scalding-core, travis gave me this failure:
https://travis-ci.org/twitter/scalding/jobs/50681252

project/Build.scala
reconditesea: You're right! This probably also explains why with this I need to build scalding-core in my test.sh

scripts/test_execution_tutorial.sh
reconditesea: I was working with @ianoc on this. Previously what I put here is "scala -classpath ...", but travis complains it couldn't find "scala" command. So we figured out travis uses scald.rb to setup scala version etc. Not sure if java -jar will directly work. I can try.

One side question, @ianoc and I find that the onComplete doesn't always guarantee to run since it waits a Future. The submitter node's jvm can exit quietly before the results are ready. We're wondering should we consider developing an inline side-effect like foreach to Execution?

ianoc:
Doesn't merge green anymore, other than that, lgtm. @reconditesea merge issues?

.travis.yml
```diff
@@ -96,3 +96,10 @@ matrix:
       script:
       - "scripts/build_assembly_no_test.sh scalding-core"
       - "scripts/test_typed_tutorials.sh"
+
+    - scala: 2.10.4
+      env: BUILD="test execution tutorials"
+      script:
+      - "scripts/build_assembly_no_test.sh scalding-core"
```
johnynek: do we still need this with the fixes you put in place in the bash script? I'm trying to minimize the time the tests take here.

reconditesea: Yup, have removed this and merged with most recent develop. It's
merge-green now.

PR #1197, opened 2015-02-12
Config/Cascading updater

ianoc:
Exporting some code from internal that will pull out kryo registered classes to ask Cascading to register tokens for them, and allow the user to easily add more tokens to those registered

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
@@ -1,0 +1,61 @@
+/*
+Copyright 2014 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding
+
+object CascadingTokenUpdater {
+  private final val lowestAllowed = 128 // cascading rules
+
+  // Take a cascading string of tokens and turns it into a map
+  // from token index to class
+  private def parseTokens(tokClass: String): Map[Int, String] =
+    if (tokClass == null || tokClass.isEmpty)
+      Map[Int, String]()
+    else
+      tokClass
+        .split(",")
+        .view
+        .map { _.split("=") }
+        .map { ary => (ary(0).toInt, ary(1)) }
+        .toMap
+
+  // does the inverse of the previous function, given a Map of index to class
+  // return the cascading token format for it
+  private def toksToString(m: Map[Int, String]): String =
+    m.map { case (tok, clazz) => tok.toString + "=" + clazz }.mkString(",")
+
+  // Given the map of already assigned tokens, what is the next available one
+  private def firstAvailableToken(m: Map[Int, String]): Int =
+    if (m.isEmpty) lowestAllowed
+    else scala.math.max(m.keys.toList.max + 1, lowestAllowed)
+
+  // Given the first free token spot
+  // assign each of the class names given to al the subsequent
+  // positions
+  private def assignTokens(first: Int, names: Iterable[String]): Map[Int, String] =
+    names.foldLeft((first, Map[Int, String]())) { (idMap, clz) =>
+      val (id, m) = idMap
+      (id + 1, m + (id -> clz))
+    }._2
+
+  def update(config: Config, clazzes: Set[Class[_]]): Config = {
+    val toks: Map[Int, String] = parseTokens(config.get(Config.CascadingSerializationTokens).map(_.toString).orNull)
+    // We don't want to assign tokens to classes already in the map
+    val newClasses: Iterable[String] = clazzes.map { _.getName } -- toks.values
+
+    config + (Config.CascadingSerializationTokens -> toksToString(toks ++ assignTokens(firstAvailableToken(toks), newClasses)))
+  }
+
+}
--- a/scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
@@ -20,7 +20,7 @@ object CascadingTokenUpdater {
 
   // Take a cascading string of tokens and turns it into a map
   // from token index to class
-  private def parseTokens(tokClass: String): Map[Int, String] =
+  def parseTokens(tokClass: String): Map[Int, String] =
     if (tokClass == null || tokClass.isEmpty)
       Map[Int, String]()
     else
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -142,7 +165,11 @@ trait Config {
       case Left((bootstrap, inst)) => ConfiguredInstantiator.setSerialized(chillConf, bootstrap, inst)
       case Right(refl) => ConfiguredInstantiator.setReflect(chillConf, refl)
     }
-    Config(chillConf.toMap + hadoopKV)
+    val withKryo = Config(chillConf.toMap + hadoopKV)
```
johnynek: aren't you disregarding the chillConf here? Don

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2014 Twitter, Inc.
```
johnynek: 2015

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  // Take a cascading string of tokens and turns it into a map
+  // from token index to class
+  private def parseTokens(tokClass: String): Map[Int, String] =
+    if (tokClass == null || tokClass.isEmpty)
+      Map[Int, String]()
+    else
+      tokClass
+        .split(",")
+        .view
```
johnynek: views are kind of deprecated. I think .iterator would work here and be less unclear.

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
```diff
@@ -0,0 +1,61 @@
...
+        .map { _.split("=") }
+        .map { ary => (ary(0).toInt, ary(1)) }
```
isnotinvain: want any validation on ary not having length greater than 2 here?

johnynek:
need some tests on Config. Seems like this should fail the tests due to it overwriting the kryo setup (according to my comment above), but I'm not sure we are actually testing that. We should test that when we set-up Kryo, the Kryo we create is somehow equivalent to what we intend.

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
```diff
@@ -0,0 +1,61 @@
...
+  private final val lowestAllowed = 128 // cascading rules
+
+  // Take a cascading string of tokens and turns it into a map
+  // from token index to class
+  private def parseTokens(tokClass: String): Map[Int, String] =
+    if (tokClass == null || tokClass.isEmpty)
+      Map[Int, String]()
+    else
+      tokClass
+        .split(",")
```
isnotinvain: does this turn into java.lang.String.split?
String.split isn't very flexible on corner cases (like repeated commas, trailing commas, etc.) Not sure how complicated the input string here will be.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
ianoc: Yep, i just realised this myself after staring at some failing hadoop jobs for too long :/ Good catch

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
ianoc: No idea, i guess we've been running with this for a while ok. You mean it includes blank entries in the event of repeated commas? We could filter them out here too

rubanm: split excludes blank entries, repeated commans iirc. I had to use .split(",", -1) once because I wanted to keep the blank entries.

rubanm: wouldn't work for other whitespace though.. tabs etc.

johnynek: we should move this to `Config.getCascadingTokens: Map[Int, String]` We might want unit test it.

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  // does the inverse of the previous function, given a Map of index to class
+  // return the cascading token format for it
+  private def toksToString(m: Map[Int, String]): String =
+    m.map { case (tok, clazz) => s"$tok=$clazz" }.mkString(",")
+
+  // Given the map of already assigned tokens, what is the next available one
+  private def firstAvailableToken(m: Map[Int, String]): Int =
+    if (m.isEmpty) lowestAllowed
+    else scala.math.max(m.keys.toList.max + 1, lowestAllowed)
```
johnynek: I don't think we need the .toList here.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -110,6 +110,29 @@ trait Config {
...
+  def getKryoRegisteredClasses: Set[Class[_]] = {
+    // Get an instance of the Kryo serializer (which is populated with registrations)
+    getKryo.map { kryo =>
+      val cr = kryo.newKryo.getClassResolver
+
+      @annotation.tailrec
+      def kryoClasses(idx: Int, acc: Set[Class[_]]): Set[Class[_]] =
+        Option(cr.getRegistration(idx)) match {
+          case present @ Some(reg) => kryoClasses(idx + 1, acc ++ present.map { _.getType }
+            .filterNot { _.isPrimitive } // Cascading handles primitives and arrays
```
johnynek: should these be here? Seems weird to force the filter in getKryoRegisteredClasses. I think we should filter the result when we are making tokens if this is a public method.

ianoc: Any preference on the top options? keep public and move filters, or make private and leave as is, maybe changing the comment/function name to match the fact its filtered?

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -172,6 +199,9 @@ trait Config {
   def getUniqueId: Option[UniqueID] =
     get(UniqueID.UNIQUE_JOB_ID).map(UniqueID(_))
 
+  def addCascadingClassSerializationTokens(clazzes: Set[Class[_]]): Config =
```
johnynek: Let's add a comment here to the effect of:

The serialization of your data will be smaller if any classes passed between tasks in your job are listed here. Without this, strings are used to write the types IN EACH RECORD, which compression probably takes care of, but compression acts AFTER the data is serialized into buffers and spilling has been triggered.

scalding-core/src/main/scala/com/twitter/scalding/CascadingTokenUpdater.scala
johnynek: This input string is always well formed and always looks like: "a=b,c=d,e=f" I don't think we need to freak out here and as Ian says this has been running at Twitter for years now.

johnynek:
Sure do wish we could get a shipit on this.

isnotinvain:
LGTM!

PR #1198, opened 2015-02-14
Typed Parquet Tuple

JiJiTang:
- Used as sink to write tuple with primitive fields and also collection types(List, Set, Map)
- Create Parquet schema generation macro


colinmarc:
Hey! I'm just a rando, but I work on very similar stuff, and I have some rando thoughts:

This is a really cool approach, because it allows you to load parquet directly into tuples/case classes. However (if I'm reading this correctly), the problem with this PR is that it doesn't actually materialize the tuples/case classes for you, because it's using ParquetTupleScheme and you already have a cascading tuple of the fields from that. It also doesn't support nested schemas (I think) for the same reason.

You could bypass this and make the code simpler if you put this macro more or less directly into a `ParquetTypedScheme`; rather than reading the records into cascading tuples and then materializing them into case classes/scala tuples, you use the macro to materialize the case classes/scala tuples directly off of the parquet reader. Then you'd end up with an interface like:

```
case class Foo(a: String, b: Option[Double])
ParquetTypedSource[Foo].map { f => f.a, f.b } // type checking against your parquet schema!
```

Anyway, just a suggestion!

JiJiTang:
Hello Colin,
Thank you for this comment and you are totally right.  At the beginning, I wanted also to generate cascading fields directly from case classes using macros(defined in the  [scalding-macros module](https://github.com/JiJiTang/scalding/tree/develop/scalding-macros)). But this will create a direct dependence on the module from scalding-parquet. So I've defined the fields as part of the method parameters and let users  to decide whether or not to use scalding-macros. And perhaps I will make all these parameters implicit(I will try to make a commit tonight).  And you are also right about the reason why it doesn't  support nested schemas and need a pull request in parquet-mr to make ParquetTupleScheme support nested schema. As you can see  [TupleWriteSupport](https://github.com/Parquet/parquet-mr/blob/master/parquet-cascading/src/main/java/parquet/cascading/TupleWriteSupport.java) only supports primitive type today.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/macros/MacroUnitTests.scala
```diff
@@ -0,0 +1,86 @@
...
+        |  required int32 d.a1.x;
+        |  required binary d.a1.y;
+        |  required int32 d.a2.x;
+        |  required binary d.a2.y;
+        |  required binary d.y;
+        |  required int32 e.a1.x;
+        |  required binary e.a1.y;
+        |  required int32 e.a2.x;
+        |  required binary e.a2.y;
+        |  required binary e.y;
```
julienledem: I'm curious, why do you use "." to capture nesting instead of actually nesting groups?

julienledem:
Nice approach

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/macros/MacroUnitTests.scala
JiJiTang: @julienledem Hi Julien, totally agree with you, it will be better to use nesting groups. The reason I make the generated schema flat is because today [TupleWriteSupport](https://github.com/Parquet/parquet-mr/blob/master/parquet-cascading/src/main/java/parquet/cascading/TupleWriteSupport.java) only supports primitive type for right now. 

```
     if (field.isPrimitive()) {
        writePrimitive(record, field.asPrimitiveType());
      } else {
        throw new UnsupportedOperationException("Complex type not implemented");
      }
```

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/MacroImplicits.scala
```diff
@@ -0,0 +1,10 @@
+package com.twitter.scalding.parquet.tuple.macros
+
+import scala.language.experimental.macros
+
+import com.twitter.scalding.parquet.tuple.macros.impl.SchemaProviderImpl
+import _root_.parquet.schema.MessageType
+
+object MacroImplicits {
+  implicit def materializeCaseClassTypeDescriptor[T]: MessageType = macro SchemaProviderImpl.toParquetSchemaImp[T]
```
ianoc: comment

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquetTuple.scala
```diff
@@ -0,0 +1,62 @@
...
+        withFilter match {
+          case Some(filterPredicate) => new ParquetTupleScheme(filterPredicate, sourceFields)
+          case _ => new ParquetTupleScheme(sourceFields)
+        }
+    }
+    HadoopSchemeInstance(scheme.asInstanceOf[Scheme[_, _, _, _, _]])
+  }
+}
+
+class TypedFixedPathParquetTuple[T](paths: Seq[String],
```
ianoc: doing tuple converters/setters like this if you have macros in the fray is dangerous, you could get one macro generated and not the other making them incompatible

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
```diff
@@ -0,0 +1,9 @@
+package com.twitter.scalding.parquet.tuple.macros
+
+import scala.language.experimental.macros
+
+import com.twitter.scalding.parquet.tuple.macros.impl.SchemaProviderImpl
+
+object Macros {
+  def caseClassParquetSchema[T]: _root_.parquet.schema.MessageType = macro SchemaProviderImpl.toParquetSchemaImp[T]
```
ianoc: Do you need macros and macro implicits?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/SchemaProviderImpl.scala
```diff
@@ -0,0 +1,84 @@
...
+    def expandCaseClass(outerTpe: Type, outerName: String, isOption: Boolean): Tree = {
+      val expanded = expandMethod(outerTpe, outerName, isOption)
+      if (expanded.isEmpty) c.abort(c.enclosingPosition, s"Case class $outerTpe has no primitive types we were able to extract")
+      val messageTypeName = s"${outerTpe}".split("\\.").last
+      q"""new _root_.parquet.schema.MessageType($messageTypeName,
+                  _root_.scala.Array.apply[parquet.schema.Type](..$expanded):_*)
+                """
+    }
+
+    c.Expr[parquet.schema.MessageType](expandCaseClass(T.tpe, "", isOption = false))
```
ianoc: You should just import this, you've used it with and without root. So importing it up the top is safer

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/SchemaProviderImpl.scala
```diff
@@ -0,0 +1,84 @@
...
+          c.abort(c.enclosingPosition, s"Nested options do not make sense being mapped onto a tuple fields in cascading.")
+        case tpe if tpe.erasure =:= typeOf[Option[Any]] =>
+          val innerType = tpe.asInstanceOf[TypeRefApi].args.head
+          matchField(innerType, outerName, fieldName, true)
+        case tpe if IsCaseClassImpl.isCaseClassType(c)(tpe) => expandMethod(tpe, s"$parquetFieldName.", isOption = false)
+        case _ => c.abort(c.enclosingPosition, s"Case class $T is not pure primitives or nested case classes")
+      }
+    }
+
+    def expandMethod(outerTpe: Type, outerName: String, isOption: Boolean): List[Tree] = {
```
ianoc: drop the braces for a single statement

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/SchemaProviderImpl.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    def getRepetition(isOption: Boolean): Tree = {
+      if (isOption) REPETITION_OPTIONAL else REPETITION_REQUIRED
+    }
+
+    def matchField(fieldType: Type, outerName: String, fieldName: String, isOption: Boolean): List[Tree] = {
+      val parquetFieldName = s"$outerName$fieldName"
+      fieldType match {
+        case tpe if tpe =:= typeOf[String] =>
+          List(q"""new _root_.parquet.schema.PrimitiveType(${getRepetition(isOption)}, _root_.parquet.schema.PrimitiveType.PrimitiveTypeName.BINARY, $parquetFieldName)""")
```
ianoc: these are mostly repetition, can we not make a helper function here?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/SchemaProviderImpl.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    implicit val builderLiftable = new Liftable[Builder] {
+      def apply(b: Builder): Tree = b.toTree
+    }
+
+    implicit val extractorLiftable = new Liftable[Extractor] {
+      def apply(b: Extractor): Tree = b.toTree
+    }
+
+    lazy val REPETITION_REQUIRED = q"_root_.parquet.schema.Type.Repetition.REQUIRED"
```
ianoc: why is it lazy?

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
```diff
@@ -0,0 +1,130 @@
...
+      val csv = ScalaSource.fromFile(new java.io.File(tempCsv, "part-00000"))
+      csv.getLines().toList shouldEqual Seq("A,1,4.0", "B,2,3.0", "A,2,5.0")
+
+      //clean temporary files generated by the jobs
+      deletePath(tempCsv)
+      deletePath(tempParquet)
+    }
+  }
+
+  def buildJobArgs(argArray: Array[String]): Args = {
```
ianoc: you should use the platform tests framework for these. None of this code seems too needed/ideal to replicate here?

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/macros/MacroUnitTests.scala
```diff
@@ -0,0 +1,86 @@
...
+        |}
+      """.stripMargin)
+      schema shouldEqual expectedSchema
+    }
+
+    "Generate parquet schema for SampleClassD" in {
+      val schema = Macros.caseClassParquetSchema[SampleClassD]
+      val expectedSchema = MessageTypeParser.parseMessageType("""
+        |message SampleClassD {
+        |  required binary a;
```
ianoc: Does this cover all the data types?

JiJiTang: Hi @ianoc, the current version of [TupleWriteSupport](https://github.com/Parquet/parquet-mr/blob/master/parquet-cascading/src/main/java/parquet/cascading/TupleWriteSupport.java) only supports primitive type. And in its implementation Strings is mapped to binary.

ianoc: Great thanks, looks like here you might want to include Long and Short for completeness?

JiJiTang: @ianoc, yes exactly. You are right, this is not clear, I will add a test with a case class that contains all the primitive types to make this clear.

JiJiTang: @ianoc Thank you so much for this detailed review. I will come with commits to fix all the problems you've mentioned.

ianoc: Thank you for this PR, some great work. Nice new macro

colinmarc:
@JiJiTang just to clarify - I actually meant you could _replace_ `ParquetTupleScheme` with a macro, a lot like this one. Rather than just generating a schema from a case class, you could also generate a parquet `RecordMaterializer` implementation that reads values directly into case class instances.

Then you wouldn't be limited to primitive types, and you'd be able to support nested structs. And you wouldn't have to define setters/getters for those case classes.

JiJiTang:
@colinmarc Good point. And thank you very much for this suggestion. And It's trivial to write a macro to generate RecordMaterializer(as you said this would support nested case classes). But I am not sure we do this in Scalding and this means we skip directly the current ParquetTupleScheme implementation in parquet-mr. Plus this means we close the door for users to provide their own customized parquet schema? Perhaps I get you wrong. And let me think about it.

JiJiTang:
Hi @julienledem @ianoc , this weekend I've added code that supports  nested case classes read/write with nested parquet message type and also some macro facilities to generate r/w support relative classes. Tuple setter and converter don't need to be defined any more. And tuple Materializer is generated now(directly by manipulating the case class). And generated schema will look like this:

```
message SampleClassC {
  required group a {
    required int32 x;
    required binary y;
  }
  required group b {
    required group a {
      required int32 x;
      required binary y;
    }
    required binary y;
  }
}

```
Please check the unit tests for details.
Sorry for this commit with too much changes. And thank you guys very much for your time to review the code. And please let me know your thoughts about this.

And hi @colinmarc , I've tried to define macros which allow to generate all the read/write support classes. But I cannot generate WriteSupport or ReadSupport classes. Because macros doesn't allow to expand class at top level, which means all the generated classes will be inner classes and cannot be created using java reflection in ParquetInputFormat. I've tried to simplify as possible the definition of TypedParquet source. Please check the committed unit tests for details and thanks a lot.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
JiJiTang: Hi @ianoc , you are right. There's no need for macro implicits. I keep macros and delete these macro implicits in my latest commits.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/SchemaProviderImpl.scala
JiJiTang: Hi @ianoc , refacto done, please check my latest commits.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/SchemaProviderImpl.scala
JiJiTang: No reason to make this lazy, bad habit of mine, shamed.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
JiJiTang: Refacto done, and have used hadoop platfrom tests framework in the lastest commits.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
```diff
@@ -0,0 +1,110 @@
...
+   * Create readable typed parquet source.
+   * Here is an example:
+   *
+   *  case class SampleClassB(string: String, int: Int, double: Option[Double], a: SampleClassA)
+   *
+   *  class ReadSupport extends ParquetReadSupport[SampleClassB] {
+   *    import com.twitter.scalding.parquet.tuple.macros.Macros._
+   *    override val tupleConverter: ParquetTupleConverter = caseClassParquetTupleConverter[SampleClassB]
+   *    override val rootSchema: String = caseClassParquetSchema[SampleClassB]
+   *  }
```
colinmarc: You can call a macro inside a macro, so you could actually have the `ReadSupport` types generated as well.

colinmarc: Oh, hm, I just saw your comment about this, sorry.

JiJiTang:
Could someone please have a look at this PR?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
```diff
@@ -0,0 +1,50 @@
+package com.twitter.scalding.parquet.tuple.macros
+
+import com.twitter.scalding.parquet.tuple.macros.impl.{ ParquetSchemaProvider, ParquetTupleConverterProvider, WriteSupportProvider }
+import com.twitter.scalding.parquet.tuple.scheme.ParquetTupleConverter
+import parquet.io.api.RecordConsumer
+import parquet.schema.MessageType
+
+import scala.language.experimental.macros
+
+object Macros {
```
julienledem: none of those macros depend on scalding and could be in the parquet project.
You could use them with Spark for example.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/ParquetTupleConverterProvider.scala
```diff
@@ -0,0 +1,89 @@
...
+          List(createConverter(q"new _root_.com.twitter.scalding.parquet.tuple.scheme.FloatConverter($idx, this, $isOption)"))
+        case tpe if tpe =:= typeOf[Double] =>
+          List(createConverter(q"new _root_.com.twitter.scalding.parquet.tuple.scheme.DoubleConverter($idx, this, $isOption)"))
+        case tpe if tpe.erasure =:= typeOf[Option[Any]] =>
+          val innerType = tpe.asInstanceOf[TypeRefApi].args.head
+          matchField(idx, innerType, isOption = true)
+        case tpe if IsCaseClassImpl.isCaseClassType(ctx)(tpe) =>
+          val innerConverterTrees = buildConverterBody(tpe, expandMethod(tpe))
+          List(createConverter(buildGroupConverter(tpe, q"Option(this)", isOption, idx, innerConverterTrees)))
+        case _ => ctx.abort(ctx.enclosingPosition, s"Case class $T is not pure primitives or nested case classes")
```
julienledem: You should mention in the doc that the support for lists is not implemented yet.

project/Build.scala
```diff
@@ -296,9 +296,12 @@ object ScaldingBuild extends Build {
         exclude("com.twitter.elephantbird", "elephant-bird-core"),
       "org.apache.thrift" % "libthrift" % "0.7.0",
       "org.slf4j" % "slf4j-api" % slf4jVersion,
-      "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided"
-    )
-  ).dependsOn(scaldingCore)
+      "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
+      "org.scala-lang" % "scala-reflect" % scalaVersion,
+      "com.twitter" %% "bijection-macros" % bijectionVersion
+    ) ++ (if(isScala210x(scalaVersion)) Seq("org.scalamacros" %% "quasiquotes" % "2.0.1") else Seq())
```
ianoc: can you lift the quasiquotes version up the top of the file here and in the other usages

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/ParquetTupleConverterProvider.scala
```diff
@@ -0,0 +1,89 @@
...
+      fieldType match {
+        case tpe if tpe =:= typeOf[String] =>
+          List(createConverter(q"new _root_.com.twitter.scalding.parquet.tuple.scheme.StringConverter($idx, this, $isOption)"))
+        case tpe if tpe =:= typeOf[Boolean] =>
+          List(createConverter(q"new _root_.com.twitter.scalding.parquet.tuple.scheme.BooleanConverter($idx, this, $isOption)"))
+        case tpe if tpe =:= typeOf[Short] =>
+          List(createConverter(q"new _root_.com.twitter.scalding.parquet.tuple.scheme.ShortConverter($idx, this, $isOption)"))
+        case tpe if tpe =:= typeOf[Int] =>
+          List(createConverter(q"new _root_.com.twitter.scalding.parquet.tuple.scheme.IntConverter($idx, this, $isOption)"))
+        case tpe if tpe =:= typeOf[Long] =>
```
julienledem: and Byte?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/WriteSupportProvider.scala
```diff
@@ -0,0 +1,106 @@
...
+                      $wTree
+                      rc.endField($groupName.getFieldName($idx), $idx)""")
+
+      def writeGroupField(subTree: Tree) =
+        q"""rc.startGroup()
+            rc.startField($groupName.getFieldName($idx), $idx)
+            $subTree
+            rc.endField($groupName.getFieldName($idx), $idx)
+            rc.endGroup()
+         """
```
julienledem: Am I missing something or should the startField/endField calls be around the startGroup/endGroup and not the other way around?

ianoc:
Sorry about the delay looking at this. Looks good to me to me. @johnynek any thoughts before good to go?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/ParquetTupleConverterProvider.scala
```diff
@@ -0,0 +1,89 @@
...
+    def buildGroupConverter(tpe: Type, parentTree: Tree, isOption: Boolean, idx: Int, converterBodyTree: Tree): Tree = {
+      q"""new _root_.com.twitter.scalding.parquet.tuple.scheme.ParquetTupleConverter($parentTree, $isOption, $idx){
+            override def newConverter(i: Int): _root_.parquet.io.api.Converter = {
+              $converterBodyTree
+              throw new RuntimeException("invalid index: " + i)
+            }
+
+            override def createValue(): Any = {
+              if(fieldValues.isEmpty) null
+              else classOf[$tpe].getConstructors()(0).newInstance(fieldValues.toSeq.map(_.asInstanceOf[AnyRef]): _*)
```
julienledem: we should be able to do better than creating through reflection. After all it's a macro.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/ParquetTupleConverter.scala
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Parquet tuple converter used to create user defined tuple value from parquet column values
+ * @param parent parent parquet tuple converter
+ * @param isOption is the field optional
+ * @param outerIndex field index in parent tuple schema
+ */
+abstract class ParquetTupleConverter(val parent: Option[ParquetTupleConverter] = None, val isOption: Boolean = false,
+  val outerIndex: Int = -1) extends GroupConverter {
+  var converters: Map[Int, Converter] = Map()
+  val fieldValues: mutable.ArrayBuffer[Any] = mutable.ArrayBuffer()
```
julienledem: The downside of this is boxing primitives.
temporarily storing the value before creating the case class instance could be delegated to the PrimitiveConverter children. Each can have a field of the right type. There's no need to store it as an Option as we will know to wrap it at the time of creation of the case class instance. ```Option(value)```

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
JiJiTang: hi @julienledem , I will commit a PR in parquet project once it comes to release then I will commit a PR in scalding to delete these macros. Right now could we let these macros stay in Scalding?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/WriteSupportProvider.scala
JiJiTang: fixed in the latest commits

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/ParquetTupleConverter.scala
JiJiTang: Hi @julienledem , thanks a lot for your good advice, this is improved in the latest commit, please check.

JiJiTang:
hi @ianoc @julienledem , thanks you guys so much for the review. Check my latest two commits please and tell me if there's something not ready to go.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/ParquetTupleConverterProvider.scala
JiJiTang: Fixed in the latest commit

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/ParquetTupleConverter.scala
```diff
@@ -0,0 +1,142 @@
+package com.twitter.scalding.parquet.tuple.scheme
+
+import parquet.io.api.{ Binary, Converter, GroupConverter, PrimitiveConverter }
+import scala.util.Try
+
+trait TupleFieldConverter[+T] extends Converter {
```
johnynek: what does this type do? It is unclear to mo. Why use currentValue and hasValue, not Option[T].

johnynek: I guess I'm asking two things: 1) can you add a comment to this type. 2) why not use Option rather than two methods.

JiJiTang: Hi @johnynek thanks a lot for the review. This trait is here to model parquet tuple field reading converter. As a field in parquet tuple can be "REQUIRED" and also "OPTIONAL". Not using Option[T] is to avoid Option boxing overhead for those fields that are required(Advice from @julienledem ). "hasValue" is there to tell if a optinal field has a value or not when reading from parquet. For those optional fields, they will be boxed into Option at the moment parent tuple is created, as in macro implementation we know that if field is a option type or not.

JiJiTang: Hi @johnynek , I've added some comments for the trait. Is that ok for you? Please let me know your thoughts.

JiJiTang:
Hi guys, I have added a commit to make the macros support collection fields type (LIST, SET, MAP), please review my latest commit.

JiJiTang:
Could someone please check the latest commits for managing collection type?

ianoc:
@JiJiTang sorry about the delay getting this merged. Those latest commits look ok to me. I think if there are more things to be done should be done as a follow up PR. This was huge. Its great work and an awesome addition. Many thanks

JiJiTang:
Hi @ianoc, thank you so much for the code review and merging this PR. And also many thanks to @julienledem @colinmarc @johnynek for your reviews and good advices. I will follow this PR for future updates.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/WriteSupportProvider.scala
```diff
@@ -0,0 +1,139 @@
...
+      def writeCollectionField(elementGroupName: TermName, subTree: Tree) =
+        writeGroupField(q"""if(!$fValue.isEmpty) {
+                              val $elementGroupName = $groupName.getType($idx).asGroupType.getType(0).asGroupType
+                              $subTree
+                            }
+                         """)
+
+      fieldType match {
+        case tpe if tpe =:= typeOf[String] =>
+          writePrimitiveField(q"rc.addBinary(Binary.fromString($fValue))")
```
johnynek: this macro is not hygenic. We need the fully qualified path here.

JiJiTang: Hi @johnynek , this is fixed in this commit: https://github.com/JiJiTang/scalding/commit/60e7390ea018f291728aab65414e125108541e79 merged with PR #1303

PR #1205, opened 2015-02-20
Add Execution.withArgs

johnynek:
seems like this will come up enough that it is worth it to have a method.


johnynek:
I guess my compile triggered scalariform on lookupjoin also. I guess that's fine.

isnotinvain:
LGTM

Issue #861, opened 2014-04-30
Remove algebird serializers


johnynek:
Depend on these:

https://github.com/twitter/chill/pull/177

PR #1206, opened 2015-02-24
Remove algebird serializers

adamkozuch:
Issue #861
I added chill-algebird to build.scala and removed AlgebirdSerializers.scala from scalding.


scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
```diff
@@ -61,7 +64,7 @@ class KryoHadoop(config: Config) extends KryoInstantiator {
     newK.register(classOf[DateRange], new DateRangeSerializer())
     newK.register(classOf[Args], new ArgsSerializer)
     // Some of the monoids from Algebird that we use:
-    newK.register(classOf[com.twitter.algebird.AveragedValue], new AveragedValueSerializer)
+    newK.register(classOf[com.twitter.algebird.AveragedValue], new  AveragedValueSerializer)
```
johnynek: Looks like we are committing an extra space. Let's remove that.

johnynek:
Nice. Thanks for doing this.

adamkozuch:
Is it ok now ?

scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
johnynek: I'll commit it, but see in line 64, where before it said: "new AveragedValuSerializer" now it has two spaces: "new  AveragedValueSerializer".

The larger point is that your diff should be as minimal as possible when making a functional change (such as this one). Style changes with zero functional changes can be larger (as review is simpler: it is only making sure all the changes are stylistic).

johnynek:
Close enough. :) Thanks for helping!

scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
adamkozuch: Sorry. I didn't notice space. Thanks for advice. I will remember it.

PR #1207, opened 2015-02-25
use AdaptiveCache in MapsideReduce

avibryant:
This goes with https://github.com/twitter/algebird/pull/419

It switches from `SummingCache` to `AdaptiveCache`, and adds some counters for observability.


scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -27,7 +27,7 @@ package com.twitter.scalding {
 
   import com.esotericsoftware.kryo.Kryo;
 
-  import com.twitter.algebird.{ Semigroup, SummingCache }
+  import com.twitter.algebird.{ Semigroup, AdaptiveCache }
```
reconditesea: Alphabetical order.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -171,24 +172,31 @@ package com.twitter.scalding {
...
+    override def operate(flowProcess: FlowProcess[_], functionCall: FunctionCall[AdaptiveCache[Tuple, V]]) {
       val cache = functionCall.getContext
       val keyValueTE = functionCall.getArguments
       // Have to keep a copy of the key tuple because cascading will modify it
       val key = keyValueTE.selectEntry(keyFields).getTupleCopy
       val value = conv(keyValueTE.selectEntry(valueFields))
-      add(cache.put(Map(key -> value)), functionCall)
+      val (hits, grew, evicted) = cache.putWithStats(Map(key -> value))
+      add(evicted, functionCall)
+      flowProcess.increment(COUNTER_GROUP, "misses", 1 - hits)
```
reconditesea: hits is an an absolute counting, not a percentage.

reconditesea: Never mind. We only put a Map of size 1 here.

reconditesea:
Curiously, is the new change in algebird immediately available in Scalding? No need to bump up versions in build?

avibryant:
@reconditesea well, we really ought to set things to `-SNAPSHOT` versions on the develop branch so that we reliably pick up changes without needing to bump anything...

ianoc:
Well... then our develop branch is tied to algebird releases. Which has other painful aspects to it too, having to release everything at once is pretty painful.

ianoc:
@avibryant develop should be on the latest algebird now, so if you merge/bump this it should be able to work now

reconditesea:
Whoops, I completely forgot this PR and created a new one (https://github.com/twitter/scalding/pull/1295). I will discard mine.

avibryant:
Updated to merge clean. Some smoke tests of this at Twitter scale would probably be useful before we release it. (The conservative thing would probably be to make it a config option that's off by default, for now).

reconditesea:
If we want to really play safe, we can merge my PR first (https://github.com/twitter/scalding/pull/1296), which only uses a SummingWithHitsCache to populate Counters for key counts and hits. So that will make available key hit/miss rates for all scalding jobs that uses MSR w/o any underlying implementation changes.
But I do like the benefits of adaptive cache here and that's why I closed my PR in favor of this.

johnynek:
This looks really promising, but yes, we should run a smoke test of this. @ianoc should we merge and then revert if the e2e tests fail, or what do you think is the way forward?

In practice, this could be really great as it should remove the need to tune the map-side cache. BUT it also causes the cache to grow to fill the available memory, which may be a problem for some use cases.

I do like the idea of a config flag to control this. If we did that, making this an option at first and then testing it, when it looks mature, make it the default seems good.

@avibryant have you run this at Stripe?

avibryant:
I do worry a bit about causing the cache to grow to fill available memory (which is why there's the `growthMargin` fudge factor in `AdaptiveCache`), but my intuition is that almost all of the memory pressure on the map side is going to come from the MSR cache, and that other, less predictable memory pressure is going to be on the reduce side.

I've tested it at Stripe and it does work for us, though I can't say I've extensively stress tested it.

Why don't I work on making this be controlled by a config flag, and then we can merge with more confidence.

ianoc:
Ok that sounds reasonable, if we can put it behind a flag it will just make it easier for us to roll it out internally.

avibryant:
Ok, done a first pass at that, turned out to be a much more invasive change. It builds but I haven't had time to verify it beyond that yet.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -166,30 +156,93 @@ package com.twitter.scalding {
...
+
+    private def getCacheSize(fp: FlowProcess[_]): Int =
+      Option(fp.getStringProperty(SIZE_CONFIG_KEY))
+        .filterNot { _.isEmpty }
+        .map { _.toInt }
+        .getOrElse(DEFAULT_CACHE_SIZE)
+
+    def apply[V: Semigroup](cacheSize: Option[Int], flowProcess: FlowProcess[_]): MapsideCache[V] = {
+      val size = cacheSize.getOrElse{ getCacheSize(flowProcess) }
+      val adaptive = Option(flowProcess.getStringProperty(ADAPTIVE_CACHE_KEY)).isDefined
```
DanielleSucher: What if this is an empty string? (Presumably that's possible, since we check for that with `fp.getStringProperty(SIZE_CONFIG_KEY)` on line 206 above.) Or if someone gets creative and sets it explicitly to false?

I'd think it should be something like `val adaptive = Option(flowProcess.getStringProperty(ADAPTIVE_CACHE_KEY)) == Some("true")` instead.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -166,30 +156,93 @@ package com.twitter.scalding {
...
       val cache = functionCall.getContext
       val keyValueTE = functionCall.getArguments
       // Have to keep a copy of the key tuple because cascading will modify it
       val key = keyValueTE.selectEntry(keyFields).getTupleCopy
       val value = conv(keyValueTE.selectEntry(valueFields))
-      add(cache.put(Map(key -> value)), functionCall)
+      val (stats, evicted) = cache.put(key, value)
+      add(evicted, functionCall)
+
+      //iterator for performance
```
DanielleSucher: Speaking of performance, since we're concerned to the point of wanting iterators instead of closures, using maps at all seems like it'd be slower than using objects here. I haven't done the benchmarking, but might it be worth switching to stats objects instead?

ianoc: In general I think it would be great to get some benchmarks down on this code if we can before merge if we can. Cascading local mode might be good enough to sense check the overheads here

johnynek:
I'm going to merge this. By default it just adds keys we have wanted on mapside cache hitrate, and you can opt in to a an adaptive version that decreases the tunes the cache size.

Ideally we want to see tests for all branches, but since the new branch is not the default, and the other branch is tested by our existing tests, it will be good to see this tested at Stripe.

PR #1208, opened 2015-02-26
OrderedSerialization support for faster key sorting

johnynek:
this does not include the macros. Want to stabilize this first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-args scalding-date"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-args scalding-date"
       script: "scripts/run_test.sh"
 
@@ -19,7 +19,7 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-avro scalding-hraven scalding-commons"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-avro scalding-hraven scalding-commons"
       script: "scripts/run_test.sh"
 
@@ -27,7 +27,7 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-core"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-core"
       script: "scripts/run_test.sh"
 
@@ -35,7 +35,7 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-hadoop-test"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-hadoop-test"
       script: "scripts/run_test.sh"
 
@@ -43,15 +43,31 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-jdbc scalding-json"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-jdbc scalding-json"
+      script: "scripts/run_test.sh"
+
+    - scala: 2.10.4
+      env: BUILD="base" TEST_TARGET="scalding-macros"
+      script: "scripts/run_test.sh"
+
+    - scala: 2.11.5
+      env: BUILD="base" TEST_TARGET="scalding-macros"
+      script: "scripts/run_test.sh"
+
+    - scala: 2.10.4
+      env: BUILD="base" TEST_TARGET="scalding-commons-macros"
+      script: "scripts/run_test.sh"
+
+    - scala: 2.11.5
+      env: BUILD="base" TEST_TARGET="scalding-commons-macros"
       script: "scripts/run_test.sh"
 
     - scala: 2.10.4
       env: BUILD="base" TEST_TARGET="scalding-parquet scalding-parquet-scrooge"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-parquet scalding-parquet-scrooge"
       script: "scripts/run_test.sh"
 
@@ -65,7 +81,7 @@ matrix:
       - "scripts/build_assembly_no_test.sh scalding"
       - "scripts/test_tutorials.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="test tutorials"
       script:
       - "scripts/build_assembly_no_test.sh scalding"
@@ -77,7 +93,7 @@ matrix:
       - "scripts/build_assembly_no_test.sh scalding"
       - "scripts/test_matrix_tutorials.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="test matrix tutorials"
       script:
       - "scripts/build_assembly_no_test.sh scalding"
@@ -91,7 +107,7 @@ matrix:
       - "scripts/build_assembly_no_test.sh scalding-core"
       - "scripts/test_typed_tutorials.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="test typed tutorials"
       script:
       - "scripts/build_assembly_no_test.sh scalding-core"
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,21 +47,22 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-jdbc scalding-json"
       script: "scripts/run_test.sh"
 
-    - scala: 2.10.4
-      env: BUILD="base" TEST_TARGET="scalding-macros"
-      script: "scripts/run_test.sh"
+# not committed yet
+    # - scala: 2.10.4
+    #   env: BUILD="base" TEST_TARGET="scalding-macros"
+    #   script: "scripts/run_test.sh"
 
-    - scala: 2.11.5
-      env: BUILD="base" TEST_TARGET="scalding-macros"
-      script: "scripts/run_test.sh"
+    # - scala: 2.11.5
+    #   env: BUILD="base" TEST_TARGET="scalding-macros"
+    #   script: "scripts/run_test.sh"
 
-    - scala: 2.10.4
-      env: BUILD="base" TEST_TARGET="scalding-commons-macros"
-      script: "scripts/run_test.sh"
+    # - scala: 2.10.4
+    #   env: BUILD="base" TEST_TARGET="scalding-commons-macros"
+    #   script: "scripts/run_test.sh"
 
-    - scala: 2.11.5
-      env: BUILD="base" TEST_TARGET="scalding-commons-macros"
-      script: "scripts/run_test.sh"
+    # - scala: 2.11.5
+    #   env: BUILD="base" TEST_TARGET="scalding-commons-macros"
+    #   script: "scripts/run_test.sh"
 
     - scala: 2.10.4
       env: BUILD="base" TEST_TARGET="scalding-parquet scalding-parquet-scrooge"
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -110,6 +110,18 @@ trait Config {
   def setMapSideAggregationThreshold(count: Int): Config =
     this + (AggregateBy.AGGREGATE_BY_THRESHOLD -> count.toString)
 
+  /**
+   * Set this configuration option to require all grouping/cogrouping
+   * to use OrderedSerialization
+   */
+  def setRequireOrderedSerialization(b: Boolean): Config =
+    this + (ScaldingRequireOrderedSerialization -> (b.toString))
+
+  def getRequireOrderedSerialization: Boolean =
+    get(ScaldingRequireOrderedSerialization)
+      .map(java.lang.Boolean.parseBoolean)
+      .getOrElse(false)
+
   def getCascadingSerializationTokens: Map[Int, String] =
     get(Config.CascadingSerializationTokens)
       .map(CascadingTokenUpdater.parseTokens)
@@ -153,7 +165,8 @@ trait Config {
     // Hadoop and Cascading should come first
     val first: Seq[Class[_ <: HSerialization[_]]] =
       Seq(classOf[org.apache.hadoop.io.serializer.WritableSerialization],
-        classOf[cascading.tuple.hadoop.TupleSerialization])
+        classOf[cascading.tuple.hadoop.TupleSerialization],
+        classOf[serialization.WrappedSerialization[_]])
     // this must come last
     val last: Seq[Class[_ <: HSerialization[_]]] = Seq(classOf[com.twitter.chill.hadoop.KryoSerialization])
     val required = (first ++ last).toSet[AnyRef] // Class is invariant, but we use it as a function
@@ -301,6 +314,7 @@ object Config {
   val ScaldingJobArgs: String = "scalding.job.args"
   val ScaldingVersion: String = "scalding.version"
   val HRavenHistoryUserName: String = "hraven.history.user.name"
+  val ScaldingRequireOrderedSerialization: String = "scalding.require.orderedserialization"
 
   /**
    * Parameter that actually controls the number of reduce tasks.
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -119,7 +119,7 @@ trait Config {
 
   def getRequireOrderedSerialization: Boolean =
     get(ScaldingRequireOrderedSerialization)
-      .map(java.lang.Boolean.parseBoolean)
+      .map(_.toBoolean)
       .getOrElse(false)
 
   def getCascadingSerializationTokens: Map[Int, String] =
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
@@ -1,0 +1,470 @@
+/*
+Copyright 2015 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.serialization
+
+import java.util.concurrent.atomic.AtomicReference
+import java.io.{ InputStream, OutputStream }
+
+import com.esotericsoftware.kryo.{ Serializer => KSerializer, DefaultSerializer, Kryo }
+import com.esotericsoftware.kryo.io.{ Input, Output }
+
+/**
+ * This interface is a way of wrapping a value in a marker class
+ * whose class identity is used to control which serialization we
+ * use. This is an internal implementation detail about how we
+ * interact with cascading and hadoop. Users should never care.
+ */
+trait Boxed[+K] {
+  def get: K
+}
+
+class BoxedDefaultSerialization extends KSerializer[Boxed[_]] {
+  def write(kryo: Kryo, output: Output, t: Boxed[_]) {
+    sys.error(s"Kryo should never be used to serialize a boxed instance: $t")
+  }
+  def read(kryo: Kryo, input: Input, t: Class[Boxed[_]]): Boxed[_] = {
+    sys.error("Kryo should never be used to serialize a boxed instance, class: $t")
+  }
+}
+
+// TODO: Make more of these with a script
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed0[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed1[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed2[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed3[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed4[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed5[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed6[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed7[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed8[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed9[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed10[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed11[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed12[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed13[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed14[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed15[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed16[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed17[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed18[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed19[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed20[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed21[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed22[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed23[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed24[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed25[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed26[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed27[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed28[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed29[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed30[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed31[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed32[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed33[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed34[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed35[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed36[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed37[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed38[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed39[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed40[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed41[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed42[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed43[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed44[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed45[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed46[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed47[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed48[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed49[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed50[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed51[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed52[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed53[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed54[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed55[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed56[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed57[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed58[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed59[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed60[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed61[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed62[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed63[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed64[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed65[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed66[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed67[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed68[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed69[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed70[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed71[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed72[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed73[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed74[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed75[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed76[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed77[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed78[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed79[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed80[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed81[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed82[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed83[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed84[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed85[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed86[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed87[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed88[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed89[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed90[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed91[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed92[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed93[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed94[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed95[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed96[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed97[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed98[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed99[K](override val get: K) extends Boxed[K]
+
+// TODO this could be any general bijection
+case class BoxedOrderedSerialization[K](box: K => Boxed[K],
+  ord: OrderedSerialization[K]) extends OrderedSerialization[Boxed[K]] {
+
+  override def compare(a: Boxed[K], b: Boxed[K]) = ord.compare(a.get, b.get)
+  override def hash(k: Boxed[K]) = ord.hash(k.get)
+  override def compareBinary(a: InputStream, b: InputStream) = ord.compareBinary(a, b)
+  override def read(from: InputStream) = ord.read(from).map(box)
+  override def write(into: OutputStream, bk: Boxed[K]) = ord.write(into, bk.get)
+}
+
+object Boxed {
+  private[this] val allBoxes = List(
+    ({ t: Any => new Boxed0(t) }, classOf[Boxed0[Any]]),
+    ({ t: Any => new Boxed1(t) }, classOf[Boxed1[Any]]),
+    ({ t: Any => new Boxed2(t) }, classOf[Boxed2[Any]]),
+    ({ t: Any => new Boxed3(t) }, classOf[Boxed3[Any]]),
+    ({ t: Any => new Boxed4(t) }, classOf[Boxed4[Any]]),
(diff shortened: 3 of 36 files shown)


johnynek:
@ianoc @isnotinvain please carefully review this code. I do want to merge this.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -110,6 +110,18 @@ trait Config {
...
+  /**
+   * Set this configuration option to require all grouping/cogrouping
+   * to use OrderedSerialization
+   */
+  def setRequireOrderedSerialization(b: Boolean): Config =
+    this + (ScaldingRequireOrderedSerialization -> (b.toString))
+
+  def getRequireOrderedSerialization: Boolean =
+    get(ScaldingRequireOrderedSerialization)
+      .map(java.lang.Boolean.parseBoolean)
```
isnotinvain: random: I think there's also .map(_.toBoolean), which throws for invalid boolean strings, compared to:
```
scala> java.lang.Boolean.parseBoolean("heyyy")
res2: Boolean = false
```

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -0,0 +1,470 @@
...
+ * whose class identity is used to control which serialization we
+ * use. This is an internal implementation detail about how we
+ * interact with cascading and hadoop. Users should never care.
+ */
+trait Boxed[+K] {
+  def get: K
+}
+
+class BoxedDefaultSerialization extends KSerializer[Boxed[_]] {
+  def write(kryo: Kryo, output: Output, t: Boxed[_]) {
```
isnotinvain: override def?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -0,0 +1,470 @@
...
+    ({ t: Any => new Boxed92(t) }, classOf[Boxed92[Any]]),
+    ({ t: Any => new Boxed93(t) }, classOf[Boxed93[Any]]),
+    ({ t: Any => new Boxed94(t) }, classOf[Boxed94[Any]]),
+    ({ t: Any => new Boxed95(t) }, classOf[Boxed95[Any]]),
+    ({ t: Any => new Boxed96(t) }, classOf[Boxed96[Any]]),
+    ({ t: Any => new Boxed97(t) }, classOf[Boxed97[Any]]),
+    ({ t: Any => new Boxed98(t) }, classOf[Boxed98[Any]]),
+    ({ t: Any => new Boxed99(t) }, classOf[Boxed99[Any]]))
+
+  private[this] val boxes: AtomicReference[List[(Any => Boxed[Any], Class[_ <: Boxed[Any]])]] =
```
isnotinvain: If I'm reading this right, it's a concurrent queue / stack?
java.util.concurrent comes with already built things like this, eg: http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html

then next could just be: 
`def next[K](...) = queue.poll()`

isnotinvain: or, if you want blocking, http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -0,0 +1,76 @@
...
+    ob.compareBinary(a, b).unsafeToInt
+}
+
+object CascadingBinaryComparator {
+  /**
+   * This method will walk the flowDef and make sure all the
+   * groupBy/cogroups are using a CascadingBinaryComparator
+   */
+  def checkForOrderedSerialization(fd: FlowDef): Try[Unit] = {
+    // Get the asScala enrichments locally?
```
isnotinvain: is this a question?

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -0,0 +1,76 @@
...
+    import cascading.pipe._
+    import com.twitter.scalding.RichPipe
+
+    // all successes or empty returns success
+    def reduce(it: TraversableOnce[Try[Unit]]): Try[Unit] =
+      it.find(_.isFailure).getOrElse(Success(()))
+
+    def check(s: Splice): Try[Unit] = {
+      val m = s.getKeySelectors.asScala
+      if (m.isEmpty) Failure(new Exception(s"Splice must have KeySelectors: $s"))
```
isnotinvain: RuntimeException? IllegalStateException?

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -0,0 +1,76 @@
...
+      else {
+        reduce(m.map {
+          case (pipename, fields) =>
+            Try {
+              if (fields.getComparators()(0).isInstanceOf[com.twitter.scalding.serialization.CascadingBinaryComparator[_]])
+                ()
+              else sys.error(s"pipe: $s, fields: $fields, comparators: ${fields.getComparators.toList}")
```
isnotinvain: prefix w/ more context? "X is not a CascadingBinaryComparator"

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -0,0 +1,76 @@
...
+      it.find(_.isFailure).getOrElse(Success(()))
+
+    def check(s: Splice): Try[Unit] = {
+      val m = s.getKeySelectors.asScala
+      if (m.isEmpty) Failure(new Exception(s"Splice must have KeySelectors: $s"))
+      else {
+        reduce(m.map {
+          case (pipename, fields) =>
+            Try {
+              if (fields.getComparators()(0).isInstanceOf[com.twitter.scalding.serialization.CascadingBinaryComparator[_]])
```
isnotinvain: why only look at the 0th element?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
johnynek: We are using a built in feature: atomicreference. Do you think we gain by using ConcurrentLinkedQueue? I like using AtomicReferences with immutable data structures, personally. I find it really enjoyable.

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -0,0 +1,76 @@
...
+                ()
+              else sys.error(s"pipe: $s, fields: $fields, comparators: ${fields.getComparators.toList}")
+            }
+        })
+      }
+    }
+
+    val allPipes: Set[Pipe] = fd.getTails.asScala.map(p => RichPipe(p).upstreamPipes).flatten.toSet
+    reduce(allPipes.iterator.map {
+      case gb: GroupBy => check(gb)
```
isnotinvain: GroupBy and CoGroup are the only types that incur hadoop sorting? Would there ever be another type in the future and we could forget to update this here? (probably not right?)

isnotinvain: How does total sort happen in scalding / cascading? Does it go through groupby / cogroup?

johnynek: We only use these two, and as per our discussion: we could imagine only using GroupBy. I do not think cascading has another primitive that does this sorting.

If by total sort, you mean secordary sort on the values, that is handled by GroupBy. I am not 100% sure we plumb through an ordered serialization there. See if you can see where we do that as you review.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Hasher.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  /*
+   * Instances below
+   */
+  implicit val unit: Hasher[Unit] = new Hasher[Unit] {
+    @inline
+    def hash(i: Unit) = 0
+  }
+  implicit val boolean: Hasher[Boolean] = new Hasher[Boolean] {
+    // Here we use the two largest mersenne primes
```
isnotinvain: comment as to why these choices were made would be useful to other readers (like me).
Why largest two mersenne primes? 
Why are floats boxed into Floats() and then converted to Ints and then hashed, but doubles are converted to Longs and then hashed?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+*/
+package com.twitter.scalding.serialization
+
+import java.io.{ ByteArrayInputStream, ByteArrayOutputStream, InputStream, OutputStream, EOFException }
+
+object JavaStreamEnrichments {
+  def eof: Nothing = throw new EOFException()
+
+  implicit class RichByteArrayOutputStream(val baos: ByteArrayOutputStream) extends AnyVal {
+    def toInputStream: ByteArrayInputStream = new ByteArrayInputStream(baos.toByteArray)
```
isnotinvain: baos.toByteArray involves a copy of the internal buffer used by ByteArrayOutputStream -- if the usage pattern is to write to a ByteArrayOutputStream, then fully consume it as an input stream, then we could avoid the copy by creating a ByteArrayInputOutputStream or something like that.

Also, since we care a lot about performance here, be aware that ByteArrayOutputStream works like Vector in that it "grows" by allocating a new buffer twice as big as the current then does a full copy. So choosing a good starting size is important, or for writing large amounts of data, using an alternative strategy like allocating byte array slabs into a an ArrayList (no copying) might make sense.

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * This reads a varInt encoding that only encodes non-negative
+     * numbers. It uses:
+     * 1 byte for values 0 - 255,
+     * 3 bytes for 256 - 65535,
+     * 7 bytes for 65536 - Int.MaxValue
+     */
+    final def readSize: Int = {
```
isnotinvain: readPositiveVarInt?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+        else readInt
+      }
+    }
+
+    final def skipFully(count: Long): Unit = {
+      @annotation.tailrec
+      def go(c: Long): Unit = {
+        val skipped = s.skip(c)
+        if (skipped == c) ()
+        else if (skipped == 0L) throw new Exception(s"count, c, skipped = ${(count, c, skipped)}")
```
isnotinvain: RuntimeException / IOException w/ more context?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+        else go(c - skipped)
+      }
+      if (count != 0L) go(count) else ()
+    }
+  }
+
+  implicit class RichOutputStream(val s: OutputStream) extends AnyVal {
+    def writeBoolean(b: Boolean): Unit = if (b) s.write(1: Byte) else s.write(0: Byte)
+
+    def writeBytes(b: Array[Byte], off: Int, len: Int): Unit = {
```
isnotinvain: what's this for? does the caller already have s and can call this on s directly?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+    def writeBytes(b: Array[Byte]): Unit = writeBytes(b, 0, b.length)
+
+    /**
+     * This reads a varInt encoding that only encodes non-negative
+     * numbers. It uses:
+     * 1 byte for values 0 - 255,
+     * 3 bytes for 256 - 65535,
+     * 7 bytes for 65536 - Int.MaxValue
+     */
+    def writeSize(i: Int): Unit = {
```
isnotinvain: writePositiveVarInt?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+        7
+      }
+    }
+
+  /**
+   * This has a lot of methods from DataInputStream without
+   * having to allocate to get them
+   * This code is similar to those algorithms
+   */
+  implicit class RichInputStream(val s: InputStream) extends AnyVal {
```
isnotinvain: What's the issue with DataInputStream? You have to allocate it as a wrapper too frequently?
If a lot of this comes directly from DataInputStream, can you add property tests that asserts that RichInputStream behaves like DataInputStream? There's a lot of room for subtle mistakes here.

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -0,0 +1,228 @@
...
+     * {code}
+     * val size = 100
+     * val marked = s.markOrBuffer(size)
+     * val y = fn(marked)
+     * marked.reset
+     * marked.skipFully(size)
+     * {/code}
+     */
+    def markOrBuffer(size: Int): InputStream = {
+      val ms = if (s.markSupported) s else {
```
isnotinvain: do we expect to hit non-markable streams, or is this just here for completeness? Would we want to log when this happens?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Laws.scala
```diff
@@ -0,0 +1,31 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.serialization
+
+sealed trait Law[T] {
```
isnotinvain: comments as to why this trait defines satisfied on 3 inputs but Law1 and Law2 ignore some of those inputs?

scalding-core/src/main/scala/com/twitter/scalding/serialization/MurmerHashUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+import java.io.InputStream
+import scala.reflect.ClassTag
+import scala.collection.generic.CanBuildFrom
+import java.lang.Integer.{ rotateLeft => rotl }
+
+// Be careful using this, the product/array or similar will attempt to call system hash codes.
+import scala.util.hashing.MurmurHash3
+
+// Taking a few functions from:
+// https://guava-libraries.googlecode.com/git/guava/src/com/google/common/hash/Murmur3_32HashFunction.java
```
isnotinvain: do we need to copy their License file / header or something here too?

scalding-core/src/main/scala/com/twitter/scalding/serialization/MurmerHashUtils.scala
```diff
@@ -0,0 +1,114 @@
...
+object MurmerHashUtils {
+  final val seed = 0xf7ca7fd2
+  private final val C1: Int = 0xcc9e2d51
+  private final val C2: Int = 0x1b873593
```
isnotinvain: seed is clearly a seed, but what's C1 and C2?

isnotinvain: nvm, this is directly from the guava file

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -0,0 +1,147 @@
...
+    val bbytes = Serialization.toBytes(b)
+    val ain = new ByteArrayInputStream(abytes)
+    val bin = new ByteArrayInputStream(bbytes)
+    ordb.compareBinary(ain, bin)
+  }
+
+  /**
+   * This is slow, but always an option. Avoid this if you can, especially for large items
+   */
+  def readThenCompare[T: OrderedSerialization](as: InputStream, bs: InputStream): Result = try resultFrom {
```
isnotinvain: what does the x in the syntax:
```
try x {

} catch {

}

do?

I would have thought this would need to be:

try {
  resultFrom {
    ... 
  }
} catch {

}

File: scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala [hunk 21]
```diff
@@ -0,0 +1,147 @@
...
+    val b = Serialization.read[T](bs)
+    compare(a.get, b.get)
+  } catch {
+    case NonFatal(e) => CompareFailure(e)
+  }
+
+  /**
+   * The the serialized comparison matches the unserialized comparison
+   */
+  def compareBinaryMatchesCompare[T](implicit ordb: OrderedSerialization[T]): Law2[T] =
```
isnotinvain: any reason these laws are in the source tree not the test tree?

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -0,0 +1,147 @@
...
+    Law2("compare(a, b) == compareBinary(aBin, bBin)",
+      { (a: T, b: T) => resultFrom(ordb.compare(a, b)) == writeThenCompare(a, b) })
```
isnotinvain: you could check that it's also true for:
compare(b, a) == writeCompare(b, a)
compare(null, a) == writeCompare(null, a)
etc.

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * ordering must be transitive. If this is not so, sort-based partitioning
+   * will be broken
+   */
+  def orderingTransitive[T](implicit ordb: OrderedSerialization[T]): Law3[T] =
+    Law3("transitivity",
+      { (a: T, b: T, c: T) =>
+        !(ordb.lteq(a, b) && ordb.lteq(b, c)) || ordb.lteq(a, c)
```
isnotinvain: I had to stare at this for a while to convince myself that this is the same as:
```
if (ordb.lteq(a,b) && ordb.lteq(b,c)) {
  ordb.lteq(a, c)
} else {
  true
}
```
Do you think that is more clear?

Similarly here do we want to check that it's true for not just <=, but also ==, and >=, and >, and <, and then change the orders of a,b,c and check all of those as well?

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  override val markSupported: Boolean = wraps.markSupported
+
+  override def read: Int = {
+    val result = wraps.read
+    // returns -1 on eof, otherwise non-negative number
+    if (result >= 0) pos += 1
+    result
+  }
+  override def read(bytes: Array[Byte]): Int = read(bytes, 0, bytes.length)
```
isnotinvain: can't imagine that it matters, but the abstract class InputStream defaults to delegating this method to read(bytes, off, len) but allows it to be overridden, which means some wacky input stream *might* handle read(bytes) differently than read(bytes, 0, bytes.length) (can't imagine why tho) but here you're not delegating to that method.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Reader.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+  implicit val double: Reader[Double] = new Reader[Double] {
+    def read(is: InputStream) = is.readDouble
+  }
+  implicit val string: Reader[String] = new Reader[String] {
+    def read(is: InputStream) = {
+      val size = is.readSize
+      val bytes = new Array[Byte](size)
+      is.readFully(bytes)
+      new String(bytes, "UTF-8")
```
isnotinvain: fun fact did you know that at once point (not sure if still true)
`new String(bytes, Charsets.UTF_8)` was 40% slower than `new String(bytes, "UTF-8")`

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
```diff
@@ -0,0 +1,149 @@
...
+  def hash[T](t: T)(implicit ser: Serialization[T]): Int =
+    ser.hash(t)
+
+  def read[T](in: InputStream)(implicit ser: Serialization[T]): Try[T] =
+    ser.read(in)
+
+  def write[T](out: OutputStream, t: T)(implicit ser: Serialization[T]): Try[Unit] =
+    ser.write(out, t)
+
+  def toBytes[T: Serialization](t: T): Array[Byte] = {
```
isnotinvain: an instance of Serialization could do this better if it knew the size to allocate ByteArrayOutputStream right?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * forAll(roundTripLaw[T]) in a valid test in scalacheck style
+   */
+  def roundTripLaw[T: Serialization]: Law1[T] =
+    Law1("roundTrip", { (t: T) => equiv(roundTrip(t), t) })
+
+  /**
+   * If two items are equal, they should serialize byte for byte equivalently
+   */
+  def serializationIsEquivalence[T: Serialization]: Law2[T] =
```
isnotinvain: the contract is stronger than this right? Not that we can enforce it, but we should document that these laws should be true across JVMs right? Or I guess that's only true of keys in groups / joins and not necessarily here?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
```diff
@@ -0,0 +1,64 @@
...
+
+package com.twitter.scalding.serialization
+
+import java.io.{ ByteArrayInputStream, ByteArrayOutputStream, InputStream, OutputStream, Serializable }
+
+import scala.util.{ Failure, Success, Try }
+
+class Serialization2[A, B](val serA: Serialization[A], val serB: Serialization[B]) extends Serialization[(A, B)] {
+  private val MAX_PRIME = Int.MaxValue // turns out MaxValue is a prime, which we want below
+  override def hash(x: (A, B)) = serA.hash(x._1) + MAX_PRIME * serB.hash(x._2)
```
isnotinvain: serA and serB are going to do fancy murmur hashing, but then they are combined more simply -- is that OK? Or is this how the murmur hashes are combined in that class from guava as well?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
```diff
@@ -0,0 +1,64 @@
...
+  override def equiv(x: (A, B), y: (A, B)): Boolean =
+    serA.equiv(x._1, y._1) && serB.equiv(x._2, y._2)
+
+  override def read(in: InputStream): Try[(A, B)] = {
+    val a = serA.read(in)
+    val b = serB.read(in)
+    (a, b) match {
+      case (Success(a), Success(b)) => Success((a, b))
+      case (Failure(e), _) => Failure(e)
```
isnotinvain: we silently ignore b's exception here, can it be combined into some kind of composite failure of 2 exceptions?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
```diff
@@ -0,0 +1,64 @@
...
+
+class OrderedSerialization2[A, B](val ordA: OrderedSerialization[A],
+  val ordB: OrderedSerialization[B]) extends Serialization2[A, B](ordA, ordB) with OrderedSerialization[(A, B)] {
+  override def compare(x: (A, B), y: (A, B)) = {
+    val ca = ordA.compare(x._1, y._1)
+    if (ca != 0) ca
+    else ordB.compare(x._2, y._2)
+  }
+  override def compareBinary(a: InputStream, b: InputStream) = {
+    // This mutates the buffers and advances them. Only keep reading if they are different
```
isnotinvain: is this out of date? you have to read both as you mention on the next line right?

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
```diff
@@ -0,0 +1,116 @@
...
+  } catch { case NonFatal(e) => Failure(e) }
+
+  override def write(b: OutputStream, s: String) = try {
+    val bytes = s.getBytes("UTF-8")
+    b.writeSize(bytes.length)
+    b.writeBytes(bytes)
+    Serialization.successUnit
+  } catch { case NonFatal(e) => Failure(e) }
+
+  override def compareBinary(lhs: InputStream, rhs: InputStream) = try {
```
isnotinvain: what does try with no catch do?

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
```diff
@@ -0,0 +1,116 @@
...
+package com.twitter.scalding.serialization
+
+import java.io.{ InputStream, OutputStream }
+import scala.util.{ Failure, Success, Try }
+import scala.util.control.NonFatal
+
+import JavaStreamEnrichments._
+
+object StringOrderedSerialization {
+  final def binaryIntCompare(leftSize: Int, seekingLeft: InputStream, rightSize: Int, seekingRight: InputStream): Int = {
```
isnotinvain: so this implements the same ordering as java.lang.String.compareTo, on bytes assumed to be UTF-8 encoded strings?

I think that'd be worth documenting, and maybe explain the algorithm a bit as well

isnotinvain: This works because UTF-8 interpreted as a series of *unsigned* bytes sorts correctly when sorted on its bytes, right?

What's the advantage of comparing as many longs as we can, then ints, then bytes? Logically comparing just bytes would be correct, right? So there's some benefit to using the largest type that we can?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
johnynek: This is only used in tests, if I'm not mistaken.

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
johnynek: what more context can we get than the stack? This should never happen.

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -0,0 +1,121 @@
...
+ * WrappedSerialization wraps a value in a wrapper class that
+ * has an associated Binary that is used to deserialize
+ * items wrapped in the wrapper
+ */
+class WrappedSerialization[T] extends HSerialization[T] with Configurable {
+
+  import WrappedSerialization.ClassSerialization
+
+  private var conf: Option[Configuration] = None
+  private var serializations: Iterable[ClassSerialization[_]] = Nil
```
isnotinvain: why not  a Set? order matters?

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -0,0 +1,121 @@
...
+  def accept(c: Class[_]): Boolean =
+    serializations.exists { case (cls, _) => cls == c }
+
+  def getSerialization(c: Class[T]): Option[Serialization[T]] =
+    serializations.collectFirst { case (cls, b) if cls == c => b }
+      // This cast should never fail since we matched the class
+      .asInstanceOf[Option[Serialization[T]]]
+
+  def getSerializer(c: Class[T]): Serializer[T] =
+    new BinarySerializer(getSerialization(c).getOrElse(sys.error(s"Class: ${c} not found")))
```
isnotinvain: No serializer for class c found?

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
```diff
@@ -0,0 +1,121 @@
...
+    new BinaryDeserializer(getSerialization(c).getOrElse(sys.error(s"Class: ${c} not found")))
+
+}
+
+class BinarySerializer[T](buf: Serialization[T]) extends Serializer[T] {
+  private var out: OutputStream = _
+  def open(os: OutputStream): Unit = {
+    out = os
+  }
+  def close(): Unit = { out = null }
```
isnotinvain: you don't want to `out.close()` because it is the caller's responsibility?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
johnynek: Have you looked at the code in DataInputStream? It is not a just a wrapper, it has lots of internal buffers. Allocating those in a tight loop will not be free. This, being an AnyVal, will not require any allocations.

johnynek: Also note, this code is tested pretty heavily as we have some heavy duty property tests going on that call this (scoverage can be used to verify this claim).

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
johnynek: It's a generic method. It's here for completeness. This code only depends on InputStream.

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: It does what you wrote. Just with less `{ }`. `try` is an expression, so, `try x` evaluates the expression `x` in the try block. That expression might be a block itself, or in this case a method call, `resultFrom`. Keeping in mind you can use `{ }` and pass a block whose last line is the return value, I used that here.

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: 1. the laws document in code the requirements here. They are like executable comments.
2. any testing system can use them without putting a dependency on a particular system and version (this has burned us in the past with the algebird laws which we wanted to share but caused coupling to a particular testframework and version, which was not what we wanted internal to Twitter).

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: null will probably throw. Generally scala code is not null-safe.

johnynek: checking `a, b` and `b, a` is probably not needed since this code is expected to be used in random generation cases (a la scalacheck). So, by doing that you are just sampling twice as many with a weird bias. If you want more test cases, just turn them up, don't put the same ones written different ways with the values reused.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -77,15 +87,81 @@ object Grouped {
...
-    val f = new Fields(key)
-    f.setComparator(key, ord)
-    f
+  /**
+   * If we are using OrderedComparable, we need to box the key
+   * to prevent other serializers from handling the key
+   */
+  def maybeBox[K, V](ord: Ordering[K])(op: (TupleSetter[(K, V)], Fields) => Pipe): Pipe = ord match {
+    case ordser: OrderedSerialization[K] =>
+      val (boxfn, cls) = Boxed.next[K]
```
isnotinvain: Is having 99 Box classes needed because something outside of our control (kryo?) requires a Class to be the key in some map?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
johnynek: Good call and we actually do have such a method: dynamicSize.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
johnynek: Yes that's true that we expect serialization to be a pure function (which only depends on the input, which does not include the identity of the computer running the code). Would you like a comment that effect?

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
johnynek: seems like, but we're not 100% sure the contract is well defined between Hadoop and what Cascading does. Want to check into that?

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
johnynek: Both Cascading and Kryo dispatch on class.

johnynek:
@avibryant @colinmarc @snoble if y'all have time to take a look, that would be great. The basic approach is the OrderedSerialization typeclass and see how we wire it into cascading (it is a bit painful, but that is life with cascading).

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
isnotinvain: What does immutability buy us here if we're swapping out the variable in an atomic reference?
```
val boxes = ConcurrentLinkedQueue[...](allBoxes)
def next = Option(boxes.poll()).getOrElse(sys.error("Exhausted the boxed classes"))
```
I think that is easy to read + understand, comes with all the guarantees and extensive testing that has gone into ConcurrentLinkedQueue, and is no less mutable than an atomic ref that keeps changing what it points to right?

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
isnotinvain: well, for both of these points, I'm getting at the java Comparator / equality laws that are defined, which includes what comparisons with null are supposed to return, as well as transitive laws etc. I see the transitive parts covered later, but I think java has a contract around comparison with null too, but maybe we don't care (cascading might though?)

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
johnynek: so, in your example, the reader has to know that this API is not going to block when it polls, and that poll when empty indeed returns null. I'm not sure I see that the compare and swap is any more error prone than code you wrote (which we can still get wrong). And at the end of the day, we need to test the behavior anyway, right? It's not like we don't need test coverage if we go with the ConcurrentLinkedQueue.

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: since we use scala Orderings all the time in the typedAPI, I guess cascading does not care. Thankfully, it does not quietly pass nulls you didn't write in your data to your comparator.

This library is following scala conventions, so I think exception on unexpected null is the rule. Use `Option[T]` if a missing value is allowed.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
johnynek: serA and serB might do any kind of hashing. There is no contract on that at all. Would you like a stronger combining of hashes here? I don't know how Guava combines hashes (though I don't think it is gospel). We looked at how scala does and cascading does. Combining with a large prime is fast and generally okay, but doing so with murmur would probably be better hashing (but slower).

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
johnynek: We could, and I've done this kind of thing before. It does not fit really well with java since you can't compose the stack traces, but I guess you can print two error messages and maybe put the second stack trace into the error message.

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
johnynek: looks like a bug.

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
johnynek: This works because comparing the binary of utf-8 as unsigned bytes gives the same sort as comparing the unicode code-points by index, which is what String.compare does. This is only coupled to the encoding in that the serialized data was written with UTF-8, but we are doing the reading and writing. We can add some comments.

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
johnynek: order does not matter, but it is private and Iterable is cheaper. If you get a duplicate, there is a bug somewhere because this code both generates and reads these lists (and as you recall with the atomicreference, each Boxed is only used for one type.

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
isnotinvain: if I defined an ordering: `def compare(a: T, b: T) = -1` it would pass this check.
Maybe we need a check:
```
ab = Integer.signum(compare(a,b))
ab == 0 || ab == -Integer.signum(compare(b,a))
```

ianoc: It would pass the check, but i don't think this check is designed/should necessarily catch that. Its catching a transitive property, which if you have a constant result from compare is actually a valid transitive compare ? 

If we don't have it then i guess we need the law that compare(a, b) + compare(b, a) == 0 (if things are clamped between -1 and 1)

.travis.yml
```diff
@@ -43,15 +43,32 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-jdbc scalding-json"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.4
+    - scala: 2.11.5
       env: BUILD="base" TEST_TARGET="scalding-jdbc scalding-json"
       script: "scripts/run_test.sh"
 
+# not committed yet
```
ianoc: Lets just move this to one of the other PR's? and remove from here -- if its not in those PR's we might forget to uncomment, and if its in them doesn't need to be here

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -0,0 +1,470 @@
...
+class BoxedDefaultSerialization extends KSerializer[Boxed[_]] {
+  def write(kryo: Kryo, output: Output, t: Boxed[_]) {
+    sys.error(s"Kryo should never be used to serialize a boxed instance: $t")
+  }
+  def read(kryo: Kryo, input: Input, t: Class[Boxed[_]]): Boxed[_] = {
+    sys.error("Kryo should never be used to serialize a boxed instance, class: $t")
+  }
+}
+
+// TODO: Make more of these with a script
```
ianoc: can kill TODO

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -0,0 +1,470 @@
...
+class Boxed97[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed98[K](override val get: K) extends Boxed[K]
+
+@DefaultSerializer(classOf[BoxedDefaultSerialization])
+class Boxed99[K](override val get: K) extends Boxed[K]
+
+// TODO this could be any general bijection
+case class BoxedOrderedSerialization[K](box: K => Boxed[K],
```
ianoc: We could do this i guess easily, but maybe it wouldn't be common enough for much of a win? Should probably add an issue around the todo i suppose if we want to do it or just kill the todo?

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
ianoc: Yeah, seems like we could just kill this comment, adding the implicits locally seems like a fine/good thing.

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
ianoc: Does that really buy us anything? its not specific enough to match on really?

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
ianoc: only write is defined on s(where write is overloaded with several arg signatures), these actually really aid code readability.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
ianoc: Each has a built in skip when it goes, so they do stop reading and fire a skip in once they are done. Each OrderedSerialization should fully consume a,b but i guess not necessarily read it all

ianoc:
I made some small comments on mostly really TODO's and comments. LGTM really. Though I was involved in enough of it to be suitably biased ;) 

Given we are doing this smaller PR we should consider if doing a followup this week/next week that splits out the ordered serialization code that doesn't depend on scalding into its own package is a good idea or left for future work. (If it survives past a release together we need to change the classpath of the classes)

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
johnynek: Agree with Ian on this. I'll change it to RuntimeException (is there another kind? :) but note the purpose of this method: to fail with an unrecoverable error if the user has declared that they require OrderedSerialization on every grouping.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
johnynek: good call.

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: Actually I'm missing 2 out of three laws on Orderings, so I'll add those:
http://en.wikipedia.org/wiki/Total_order

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: I'll use this clearer style.

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
johnynek: good catch. This should be a pass through.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Reader.scala
johnynek: thanks, Java.

.travis.yml
johnynek: I think I uncommented in the follow up (I think). I'm just as worried we won't remember to add it (since that happens to). I'll do what you prefer.

johnynek:
@isnotinvain @ianoc addressed most of the comments. Want to take another look?

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
johnynek: added a comment on the class.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
```diff
@@ -0,0 +1,69 @@
...
+    val cA = ordA.compareBinary(a, b)
+    // we have to read the second ones to skip
+    val cB = ordB.compareBinary(a, b)
+    cA match {
+      case OrderedSerialization.Equal => cB
+      case f @ OrderedSerialization.CompareFailure(_) => f
+      case _ => cA // the first is not equal
+    }
+  }
```
johnynek: actually, staticSize and dynamicSize need to be added. I wonder if they should not have default implementations. Its too easy forget about them.

ianoc: Seems reasonable not to have them have any defaults, make someone think about the cost involved if they don't supply them if they write something manually.

johnynek:
actually there were several bugs with combinators as a result of the default None in the size methods. I think making those required is a good call for performance critical code, such as this.

johnynek:
@isnotinvain any more concerns on this base PR?

isnotinvain:
@johnynek Sorry, I'm still reading your comments. I will try to finish tonight.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
isnotinvain: not a big deal, but, we need to test that this feature works, but we don't need to test that ConcurrentLinkedQueue works as described since we get that for free. We need to test that we're using it correctly though. Writing tests for multithreaded coded is hard, so using a battle tested concurrent data structure with extensive docs seems like a good choice to me. Obviously AtomicReference is similarly well documented and tested, but it is a lower level primitive. That was my thinking, but not that important.

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
isnotinvain: >> is there another kind?
Yes! 1/2 of the exceptions are not RuntimeExceptions :)
RuntimeExcept should be named "UncheckedException" -- the difference matters more in java, but it is the base class for all unchecked exceptions. IOException is not a RuntimeException... (even though it happens at run time)

However -- it *does* help us with pattern matching. Catching `Exception` is dangerous, it catches InterruptedException and other dangerous-to-catch exceptions, while there are no dangerous to catch RuntimeExceptions.

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
isnotinvain: No I didn't mean sort on secondary values, I meant if you wanted a total sort on all your data, though I guess in scalding you just do groupAll for that? We don't have a distributed total sort operator (that samples and partitions and merges?)

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
isnotinvain: oh I see, writeBytes is a better name than write, yes

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
johnynek: Right. There is no total sort or even controlling partitions in scalding (beyond keys).

With this new code, you can control hashing, but even then who knows how cascading will make hashcodes to reducers.

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
isnotinvain: OK, didn't know that DataInputStream had lots of internal buffers

scalding-core/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
isnotinvain: makes sense

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
isnotinvain: well it also includes Set / Map ordering as well -- you've got to serialize those deterministically

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
isnotinvain: no, just don't have any intuition on if it matters. This is a common way to combine though so seems fine.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
johnynek: yeah, I'm thinking actually this may be too strict. If you are this strict, I'm not sure why you would not want to go all the way to OrderedSerialization.

Dropping this strictness might be something we would do in the future if we require Serialization for all Values.

scalding-core/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
isnotinvain: yeah, pretty silly that stacks don't compose. I'd say "just log it" but we're not really in the right place to do logging here.

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
isnotinvain: odd that it compiles

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
isnotinvain: It's cheaper memory-wise but we use it for lookups by traversing it right (slower)? We don't expect this to be large or in the hot path right so probably doesn't matter?

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
isnotinvain: OK

scalding-core/src/main/scala/com/twitter/scalding/serialization/WrappedSerialization.scala
johnynek: I *think* this is not in the hot path. But if it is, we should use a Map. I guess using a Map may be the right thing to do just in case. Lots of times we got burned on Hadoop or cascading doing something repeatedly that you might guess only happens a handful of times.

isnotinvain: OK -- I remember I fixed a Pig UDF that used an ArrayList instead of a Set, expecting it to make a big difference and then it didn't because for a certain size of ArrayList, just iterating is about as fast as a map look up (that certain size being very small)

johnynek:
@isnotinvain changed the Iterable to a Map. Makes the code clearer and possibly faster. Anything else that needs addressing in this?

isnotinvain:
@johnynek 

Do we have Laws for Orderings? I understand that we now have laws for total orderings applied to serializable orderings, but I think there's value in checking the other things that we expect to be true, like `a < b => b > a` -- it seems easy enough to check and could catch a buggy ordering.

The ignored exception in Serialization2.scala would be great to log or compose with the other exception, maybe like this:
` new RuntimeException(s"There were 2 exceptions:\n${e1.stackTraceString}\n${e2.stackTraceString}`

I have not had a chance to look up the contract around `out.close()` -- though I suppose it's a contract we find out empirically. If we close it early, that'll fail when someone tries to use it. If we don't close it, it might leak something, or we might not flush everything. Where does the stream come from? Hadoop or cascading? or cascading passed it through from hadoop?

isnotinvain:
Meant to say, +1 LGTM, the above questions aren't blockers.

johnynek:
@isnotinvain two issues:

1) So, the current laws in this code do assume that compare is the only method people override, and then we don't need to test the Ordering trait from scala that it correctly implements lt, lteq, etc... from compare. I think that is safe. What do you think? I mean, we could go on to verify that lt(a, b) => compare(a, b) < 0, lteq(a, b) => compare(a, b) <=0, equiv(a, b) => compare(a, b) == 0, gteq(a, b) => compare(a, b) >= 0, gt(a, b) => compare(a, b) > 0. But that is not so much about a correct implementation of Ordering but that they did not override some default methods with insane data.

2) imagine if serialization were lazy, you would not see the second exception, so I don't think that hiding it is so bad. Also if you only have one exception, you will see that one, so I don't really feel like it is a case of losing data, it is more about ignoring some evaluated error in favor of showing the error that happened first. Also, there is no method (stackTraceString) so we have to use apache commons or write something here. I'd rather skip this since I think it is a corner case that is best solved with a designed dual exception that we can flatten into a List so we can see all failures, etc...

I'd rather merge this now and move on to the macro's branch. Is that okay?

ianoc:
Given alex's + 1 above, merging

isnotinvain:
For 1), I guess I meant that for any Orderings we or our users have written by hand, it'd be nice to assert those properties. No need to assert them on ones that come built into the language. Similar to how we provide MonoidLaws etc.

For 2) I must be going crazy, I thought stackTraceString was an implicit enrichment on exception, but now I can't find it. Must have been local to some other project. I agree, if we surface the first exception that's fine since often we wouldn't even try the second stream.

PR #1209, opened 2015-02-26
Macros for OrderedSerialization

johnynek:
These are the macros on top of #1208.

We should only merge this after #1208 is merged and probably after we have deployed this and tested in detail at Twitter.


johnynek:
@isnotinvain if you can start reviewing this, that would be great.

I need to get some benchmarks written for this code and compare it to serializing with kryo, for instance.

johnynek:
Testing round tripping long lists of ints:
```
macro:
[info] Parameters(size -> 300000): 10.488
[info] Parameters(size -> 600000): 21.314
[info] Parameters(size -> 900000): 31.55
[info] Parameters(size -> 1200000): 42.981
[info] Parameters(size -> 1500000): 54.297

kryo:
[info] Parameters(size -> 300000): 42.926
[info] Parameters(size -> 600000): 88.108
[info] Parameters(size -> 900000): 134.415
[info] Parameters(size -> 1200000): 183.419
[info] Parameters(size -> 1500000): 233.11
```

ianoc:
Sweet looking ! -- how many iterations does that do ? Have you tried kryo
with unsafe ?

johnynek:
@ianoc scalameter looks pretty good: http://scalameter.github.io/home/gettingstarted/0.5/simplemicrobenchmark/index.html

it actually adaptively runs again and again, with some warmup, until the statistics don't change much. Then it discards outliers that have large values due to GC (which it attempts to detect directly). It seems fairly complete.

I have not tried the unsafe code. I am just comparing it to Kryo as we would have used it in scalding currently.

ianoc:
Cool -- will look more tomorrow. The unsafe comparison would be to see if
we could gain much by using unsafe ourselves (or if we should be using it
in our kryo now)

johnynek:
Some new results:

Sorting `Array[Array[Byte]]` using the OrderedSerialization is MUCH faster if the object is somewhat fat (did not test Long or (Long, Long) yet.)
```
[info] ::Benchmark Serialization.sort typeclass: (Int, (Long, String))::
[info] cores: 4
[info] Parameters(size -> 30000): 52.46
[info] Parameters(size -> 60000): 116.677
[info] Parameters(size -> 90000): 189.692
[info] Parameters(size -> 120000): 252.071
[info] Parameters(size -> 150000): 342.291

[info] ::Benchmark Serialization.sort kryo: (Int, (Long, String))::
[info] cores: 4
[info] Parameters(size -> 30000): 289.104
[info] Parameters(size -> 60000): 621.573
[info] Parameters(size -> 90000): 973.609
[info] Parameters(size -> 120000): 1261.862
[info] Parameters(size -> 150000): 1765.191
```

The rest of these benchmarks are on just serialization, no sorting.

Serialization of Strings are a bit slower (even after I stole Kryo's ascii optimization):
```
[info] ::Benchmark Serialization.typeclass: String::
[info] cores: 4
[info] Parameters(size -> 30000): 15.585
[info] Parameters(size -> 60000): 31.548
[info] Parameters(size -> 90000): 47.677
[info] Parameters(size -> 120000): 63.229
[info] Parameters(size -> 150000): 78.422

[info] ::Benchmark Serialization.kryo: String::
[info] cores: 4
[info] Parameters(size -> 30000): 14.448
[info] Parameters(size -> 60000): 25.268
[info] Parameters(size -> 90000): 39.133
[info] Parameters(size -> 120000): 50.799
[info] Parameters(size -> 150000): 63.219
```
But number and tuples of them are significantly faster with the macros:

```
[info] ::Benchmark Serialization.typeclass: (Int, Long, Short)::
[info] cores: 4
[info] Parameters(size -> 30000): 10.355
[info] Parameters(size -> 60000): 8.495
[info] Parameters(size -> 90000): 12.731
[info] Parameters(size -> 120000): 16.678
[info] Parameters(size -> 150000): 21.154

[info] ::Benchmark Serialization.kryo: (Int, Long, Short)::
[info] cores: 4
[info] Parameters(size -> 30000): 21.616
[info] Parameters(size -> 60000): 43.783
[info] Parameters(size -> 90000): 40.49
[info] Parameters(size -> 120000): 55.555
[info] Parameters(size -> 150000): 68.118
```

Will add more sorting benchmarks tomorrow.

scalding-benchmarks/src/test/scala/com/twitter/scalding/Serialization.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  def collection[T, C[_]](size: Gen[Int])(implicit arbT: Arbitrary[T], cbf: CanBuildFrom[Nothing, T, C[T]]): Gen[C[T]] =
+    collection[T, C](size, arbT.arbitrary)(cbf)
+
+  def collection[T, C[_]](size: Gen[Int], item: scGen[T])(implicit cbf: CanBuildFrom[Nothing, T, C[T]]): Gen[C[T]] =
+    size.map { s =>
+      val builder = cbf()
+      builder.sizeHint(s)
+      // Initialize a fixed random number generator
+      val rng = new scala.util.Random("scalding".hashCode)
```
isnotinvain: nice

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/BooleanOrderedBuf.scala
```diff
@@ -0,0 +1,64 @@
...
+        q"$inputStream.writeByte(if($element) (1: Byte) else (0: Byte))"
+
+      override def get(inputStreamA: ctx.TermName): ctx.Tree =
+        q"($inputStreamA.readByte == (1: Byte))"
+
+      def compare(elementA: ctx.TermName, elementB: ctx.TermName): ctx.Tree =
+        q"_root_.java.lang.Boolean.compare($elementA, $elementB)"
+
+      override def length(element: Tree): CompileTimeLengthTypes[c.type] =
+        ConstantLengthCalculation(c)(1)
```
isnotinvain: macro noob question, does `ConstantLengthCalculation` get evaluated once at compile time, or every time `length` is called at runtime? I see the `q"..."` syntax is used to put some scala code into the output of the macro, but that's not the only way to accomplish that right? How can you tell when some something is being added to the AST vs being evaluated during compilation?

johnynek:
Improved unsigned number sorting, which improves binary sorting on strings. Verified that the readInt and readLong were faster than any obvious alternatives. Here are the results comparing sorting a bunch of binary blobs that are serialized data (should be similar to what Hadoop is doing to merge):
```
[info] ::Benchmark Serialization.sort typeclass: Long::
[info] Parameters(size -> 30000): 4.846
[info] Parameters(size -> 60000): 11.216
[info] Parameters(size -> 90000): 17.372
[info] Parameters(size -> 120000): 11.988
[info] Parameters(size -> 150000): 15.0

[info] ::Benchmark Serialization.sort kryo: Long::
[info] Parameters(size -> 30000): 4.442
[info] Parameters(size -> 60000): 12.365
[info] Parameters(size -> 90000): 20.012
[info] Parameters(size -> 120000): 26.417
[info] Parameters(size -> 150000): 33.134

[info] ::Benchmark Serialization.sort typeclass: String::
[info] Parameters(size -> 30000): 6.063
[info] Parameters(size -> 60000): 7.784
[info] Parameters(size -> 90000): 11.808
[info] Parameters(size -> 120000): 15.242
[info] Parameters(size -> 150000): 19.276

[info] ::Benchmark Serialization.sort kryo: String::
[info] Parameters(size -> 30000): 13.608
[info] Parameters(size -> 60000): 24.729
[info] Parameters(size -> 90000): 35.489
[info] Parameters(size -> 120000): 49.907
[info] Parameters(size -> 150000): 60.606

[info] ::Benchmark Serialization.sort typeclass: (Int, (Long, String))::
[info] Parameters(size -> 30000): 6.223
[info] Parameters(size -> 60000): 7.842
[info] Parameters(size -> 90000): 13.341
[info] Parameters(size -> 120000): 15.995
[info] Parameters(size -> 150000): 20.453

[info] ::Benchmark Serialization.sort kryo: (Int, (Long, String))::
[info] Parameters(size -> 30000): 20.544
[info] Parameters(size -> 60000): 41.908
[info] Parameters(size -> 90000): 65.156
[info] Parameters(size -> 120000): 88.036
[info] Parameters(size -> 150000): 106.301
```

johnynek:
It looks like equality checking is generally not useful. Removed it from the defaults. It might be useful for expensive comparisons like `Set[T]` or `Map[K, V]` but for numbers and strings you are doing almost all the work to check if something is equal.

```
[info] ::Benchmark Serialization.sort typeclass: (Int, (Long, String))::
[info] Parameters(size -> 30000): 5.96
[info] Parameters(size -> 60000): 7.271
[info] Parameters(size -> 90000): 11.918
[info] Parameters(size -> 120000): 14.62
[info] Parameters(size -> 150000): 18.443

[info] ::Benchmark Serialization.sort kryo: (Int, (Long, String))::
[info] Parameters(size -> 30000): 20.689
[info] Parameters(size -> 60000): 43.922
[info] Parameters(size -> 90000): 64.2
[info] Parameters(size -> 120000): 87.772
[info] Parameters(size -> 150000): 113.368
```

johnynek:
Thought of two issues:

1) We don't include Char in our primitives.
2) We don't include the java boxed types. I am pretty sure some people are using those now because they are comparable (and Java users reach for them sometimes).

isnotinvain:
Does scalding / scala require primitive boxing everywhere (except for `@specialized`)?
Any time you use a primitive with something generic it gets boxed (in java at least).

johnynek:
Yes, scala boxes similar to how java does, but like java, the type `Int` is distinct from `java.lang.Integer`. So you if you make a `(java.lang.Integer, String)` tuple, our code won't work (I don't believe). We need to explicitly handle those types separately.

isnotinvain:
Right, I guess everything we do in scalding is generic so there's no opportunity for unboxed primitives as keys or anything like that.

johnynek:
Okay. I think we need to test deploy this internally at Twitter. I think it is pretty complete now: primitives (boxed and unboxed), Option, Either, all common scala collections, tuples, case classes. Only restriction I know of is that null is not supported. If your data can have nulls, you need to use Option.

scalding-core/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala
```diff
@@ -213,17 +228,15 @@ object JavaStreamEnrichments {
      */
     def writePosVarInt(i: Int): Unit = {
       require(i >= 0, s"must be non-negative: ${i}")
-      if (i < ((1 << 8) - 1)) s.write(i.toByte)
+      if (i < ((1 << 8) - 1)) s.write(i)
```
isnotinvain: require is defined as:
`require(requirement: Boolean, message: ⇒ Any)`

Does that mean that the string interpolation won't happen unless `i<0`? Does it also mean that every time this method is called, a closure is created to hold the string interpolation?

We found a performance regression in parquet due to something similar, but using a not-lazy method call (plain java method) and found that we were doing expensive string interpolation for every call, not just in the error condition.

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -0,0 +1,299 @@
...
+import scala.reflect.macros.Context
+import scala.language.experimental.macros
+import scala.util.control.NonFatal
+
+object CommonCompareBinary {
+  import com.twitter.scalding.serialization.JavaStreamEnrichments._
+
+  // If the lengths are equal and greater than this number
+  // we will compare on all the containing bytes
+  val minSizeForFulBinaryCompare = 24
```
isnotinvain: just curious, is 24 a guess or something from your benchmarks?

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -0,0 +1,299 @@
...
+    (lenA > minSizeForFulBinaryCompare &&
+      (lenA == lenB) &&
+      inputStreamA.markSupported &&
+      inputStreamB.markSupported) && {
+        inputStreamA.mark(lenA)
+        inputStreamB.mark(lenB)
+
+        var pos: Int = 0
+        while (pos < lenA) {
+          val a = inputStreamA.read
```
isnotinvain: are these streams usually backed by a byte array? If not, reading 1 byte at a time vs. chunks of bytes could matter, though maybe BufferedInputStreams handle that case for us too?

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -0,0 +1,299 @@
...
+       * string, where the cost to see the serialized size is not cheaper than
+       * directly serializing.
+       */
+      val noLenCalc = q"""
+      // Start with pretty big buffers because reallocation will be expensive
+      val $baos = new _root_.java.io.ByteArrayOutputStream(256)
+      ${t.put(baos, element)}
+      val $len = $baos.size
+      $outerbaos.writePosVarInt($len)
+      $baos.writeTo($outerbaos)
```
isnotinvain: this writes to a side buffer, then copies it into the final buffer right?
Parquet handles this case in a few places by using something similar to a BAOS that instead just collects byte arrays instead of copying them into an internal array. Not sure if that would be better or not, depends on if you most write big chunks or if you mostly write small chunks.

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/ByteBufferOrderedBuf.scala
```diff
@@ -0,0 +1,99 @@
...
+
+      val $queryLength = _root_.scala.math.min($lenA, $lenB)
+      var $incr = 0
+      var $state = 0
+
+      while($incr < $queryLength && $state == 0) {
+        $state = _root_.java.lang.Byte.compare($inputStreamA.readByte, $inputStreamB.readByte)
+        $incr = $incr + 1
+      }
+      if($state == 0) {
```
isnotinvain: do you have to worry about fully consuming the record from the stream? Or is that handled somewhere via mark() / reset() / skip() ?

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/CaseClassOrderedBuf.scala
```diff
@@ -0,0 +1,87 @@
...
+import com.twitter.scalding._
+import com.twitter.scalding.macros.impl.ordered_serialization.{ CompileTimeLengthTypes, ProductLike, TreeOrderedBuf }
+import CompileTimeLengthTypes._
+import com.twitter.scalding.serialization.OrderedSerialization
+import com.twitter.bijection.macros.impl.IsCaseClassImpl
+
+object CaseClassOrderedBuf {
+  def dispatch(c: Context)(buildDispatcher: => PartialFunction[c.Type, TreeOrderedBuf[c.type]]): PartialFunction[c.Type, TreeOrderedBuf[c.type]] = {
+    case tpe if tpe.typeSymbol.isClass && tpe.typeSymbol.asClass.isCaseClass && !tpe.typeConstructor.takesTypeArgs =>
+      CaseClassOrderedBuf(c)(buildDispatcher, tpe)
```
isnotinvain: does this mean that case classes with generic params are not supported? Just trying to see if I'm reading this right.

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/TreeOrderedBuf.scala
isnotinvain: we dono't have control over baos, so moot point

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/StringOrderedBuf.scala
```diff
@@ -0,0 +1,122 @@
...
+         val $charLen = $element.length
+         if ($charLen == 0) {
+           $inputStream.writePosVarInt(0)
+         }
+         else if (isShortAscii($charLen, $element)) {
+           $inputStream.writePosVarInt($charLen)
+           val $bytes = new Array[Byte]($charLen)
+           // This deprecated gets ascii bytes out, but is incorrect
+           // for non-ascii data.
+           $element.getBytes(0, $charLen, $bytes, 0)
```
isnotinvain: we could wrap this in a java method w/ `@SuppressWarnings`

isnotinvain: not a blocker though

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/TreeOrderedBuf.scala
isnotinvain: not an issue

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/TreeOrderedBuf.scala
isnotinvain: not an issue

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/ByteBufferOrderedBuf.scala
isnotinvain: not an issue

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/CaseClassOrderedBuf.scala
isnotinvain: not an issue

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/StringOrderedBuf.scala
```diff
@@ -0,0 +1,122 @@
...
+           $inputStream.write($bytes)
+         }
+         else {
+           // Just use utf-8
+           val $bytes = $element.getBytes("UTF-8")
```
isnotinvain: maybe add a comment here along the lines of
`// TODO: Investigate benefit of using getBytes(Charset) on jvms > 1.6`

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/StringOrderedBuf.scala
```diff
@@ -0,0 +1,122 @@
...
+import scala.language.experimental.macros
+import scala.reflect.macros.Context
+
+import com.twitter.scalding._
+import com.twitter.scalding.macros.impl.ordered_serialization.{ CompileTimeLengthTypes, ProductLike, TreeOrderedBuf }
+import CompileTimeLengthTypes._
+import java.nio.ByteBuffer
+import com.twitter.scalding.serialization.OrderedSerialization
+
+object StringOrderedBuf {
```
isnotinvain: Do the tests for this create arbitrary strings by forming arbitrary sequences of chars? Because some UTF characters in UTF-16 representation are actually two consecutive chars, we should make sure we're covering this case.

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
```diff
@@ -0,0 +1,303 @@
...
+    case tpe if tpe.erasure =:= c.universe.typeOf[sci.HashSet[Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+    case tpe if tpe.erasure =:= c.universe.typeOf[sci.ListSet[Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+
+    case tpe if tpe.erasure =:= c.universe.typeOf[Map[Any, Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+    case tpe if tpe.erasure =:= c.universe.typeOf[sc.Map[Any, Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+    case tpe if tpe.erasure =:= c.universe.typeOf[sci.Map[Any, Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+    case tpe if tpe.erasure =:= c.universe.typeOf[sci.HashMap[Any, Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+    case tpe if tpe.erasure =:= c.universe.typeOf[sci.ListMap[Any, Any]].erasure => TraversablesOrderedBuf(c)(buildDispatcher, tpe, DoSort, NotArray)
+  }
+
```
isnotinvain: Why do we have to list all the collection types here instead of taking anything Traversable, with a special case for Map / Set / Array? Like we the case for all the other kinds of Traversables / Iterables?

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
```diff
@@ -0,0 +1,303 @@
...
+case object DoSort extends ShouldSort
+case object NoSort extends ShouldSort
+
+sealed trait MaybeArray
+case object IsArray extends MaybeArray
+case object NotArray extends MaybeArray
+
+object TraversablesOrderedBuf {
+  def dispatch(c: Context)(buildDispatcher: => PartialFunction[c.Type, TreeOrderedBuf[c.type]]): PartialFunction[c.Type, TreeOrderedBuf[c.type]] = {
+    case tpe if tpe.erasure =:= c.universe.typeOf[Iterable[Any]] => TraversablesOrderedBuf(c)(buildDispatcher, tpe, NoSort, NotArray)
```
isnotinvain: don't a lot of these cases overlap? Sets are Iterable aren't they? And the Set case comes later?

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
```diff
@@ -0,0 +1,540 @@
...
+      case (soFar, Law1(name, fn)) => soFar && Prop.forAll(g)(fn).label(name)
+      case (soFar, Law2(name, fn)) => soFar && Prop.forAll(g, g)(fn).label(name)
+      case (soFar, Law3(name, fn)) => soFar && Prop.forAll(g, g, g)(fn).label(name)
+    }
+}
+
+object ByteBufferArb {
+  implicit def arbitraryTestTypes: Arbitrary[ByteBuffer] = Arbitrary {
+    for {
+      aBinary <- Gen.alphaStr.map(s => ByteBuffer.wrap(s.getBytes("UTF-8")))
```
isnotinvain: for my own understanding, this should be the same as `s.getBytes("ASCII")` right?

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
```diff
@@ -0,0 +1,540 @@
...
+    for {
+      aInt <- arb[Int]
+      anOption <- arb[Option[Long]]
+    } yield new MyData(aInt, anOption)
+  }
+}
+
+class MyData(override val _1: Int, override val _2: Option[Long]) extends Product2[Int, Option[Long]] {
+  override def canEqual(that: Any): Boolean = that match {
+    case o: MyData => this._1 == o._1 && this._2 == o._2
```
isnotinvain: shouldn't this just be:
```
case o: MyData => true
case _ => false
```
?

The docs for canEqual aren't that clear, but it sort of sounds like they want to know if something could possibly equal something else?

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
```diff
@@ -0,0 +1,540 @@
...
+          assert(compareMem > 0, s"Compare binary: $compareBinary, and compareMem : $compareMem must have the same sign")
+        }
+    }
+  }
+
+  def checkMany[T: Arbitrary](implicit obuf: OrderedSerialization[T]) = forAll { i: List[(T, T)] =>
+    checkManyExplicit(i)
+  }
+
+  def clamp(i: Int): Int =
```
isnotinvain: also known as `i.signum` !

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
```diff
@@ -0,0 +1,540 @@
...
+    if (obuf.dynamicSize(a).isDefined) {
+      assert(obuf.dynamicSize(a).get == asize, "dynamic size matches the correct value")
+    }
+    if (obuf.staticSize.isDefined) {
+      assert(obuf.dynamicSize(a).get == asize, "dynamic size matches the correct value")
+      assert(obuf.staticSize.get == asize, "dynamic size matches the correct value")
+    }
+    assert(oBufCompare(rta, a) === 0, s"A should be equal to itself after an RT -- ${rt(a)}")
+    assert(oBufCompare(rtb, b) === 0, s"B should be equal to itself after an RT-- ${rt(b)}")
+    assert(oBufCompare(a, b) + oBufCompare(b, a) === 0, "In memory comparasons make sense")
```
isnotinvain: do we enforce the constraint that compare returns (-1, 0, 1) or is it valid to return some other negative or positive values, which might trip this check?

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
```diff
@@ -0,0 +1,540 @@
...
+  test("Test out Double") { check[Double] }
+  test("Test out jl.Double") {
+    implicit val a = arbMap { b: Double => java.lang.Double.valueOf(b) }
+    check[java.lang.Double]
+  }
+
+  test("Test out String") {
+    primitiveOrderedBufferSupplier[String]
+
+    check[String]
```
isnotinvain: Somewhere I missed where arbitrary strings are generated I think -- do we know for sure it covers all the strange types of characters, including UTF characters that take up 2 jvm chars ?

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
```diff
@@ -0,0 +1,540 @@
...
+    check[Iterable[Int]]
+  }
+
+  test("Test out Set[Int]") {
+    primitiveOrderedBufferSupplier[Set[Int]]
+    check[Set[Int]]
+  }
+
+  test("Test out Set[Double]") {
+    primitiveOrderedBufferSupplier[Set[Double]]
```
isnotinvain: if we're generating the sets from the same jvm, chances are equal sets will have equal orderings, do we exercise equal sets with unequal order anywhere?

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
johnynek: ```Set[Any] <:< Iterable[Any]``` not `=:=`.

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/StringOrderedBuf.scala
johnynek: Scalacheck generates strings by taking random characters and putting them in arrays. It could be more robust, but it is pretty good. The downside is it is biased pretty strongly toward non-ascii strings by default. It would be nice to actually have a large corpus of tweets to use as strings since they should cover a variety of languages.

scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
johnynek: You are thinking in terms of subclasses, but these macros will generate an invariant `OrderedSerialization` that guarantees that the output type is the same as the type resolved. We have to generate the right type signature in that `OrderedSerialization`.

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
johnynek: yes. UTF-8 is a superset of ascii, and the alphaStr is only ASCII.

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
johnynek: no, we don't. JDK does not, so we let that go. The OrderedSerialization.Result.unsafeToInt does return -1, 0, 1 though.

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
johnynek: in scalacheck, as I describe above: random chars (16 bit numbers) in an Array, converted to String.

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
johnynek: Good point. We could make an a ListSet here with a random permutation. I'll add that.

johnynek:
removed the requires to hopefully micro optimize a bit. Here are the latest benchmarks (does not seem to have made a real difference.
```
[info] ::Benchmark Serialization.typeclass: (Int, (Long, String))::
[info] Parameters(size -> 30000): 19.051
[info] Parameters(size -> 60000): 38.298
[info] Parameters(size -> 90000): 57.549
[info] Parameters(size -> 120000): 78.19
[info] Parameters(size -> 150000): 97.151

[info] ::Benchmark Serialization.kryo: (Int, (Long, String))::
[info] Parameters(size -> 30000): 36.547
[info] Parameters(size -> 60000): 49.688
[info] Parameters(size -> 90000): 76.413
[info] Parameters(size -> 120000): 101.659
[info] Parameters(size -> 150000): 127.509

[info] ::Benchmark Serialization.typeclass: (Int, Long, Short)::
[info] Parameters(size -> 30000): 9.982
[info] Parameters(size -> 60000): 8.34
[info] Parameters(size -> 90000): 13.113
[info] Parameters(size -> 120000): 16.44
[info] Parameters(size -> 150000): 21.099

[info] ::Benchmark Serialization.kryo: (Int, Long, Short)::
[info] Parameters(size -> 30000): 22.129
[info] Parameters(size -> 60000): 44.171
[info] Parameters(size -> 90000): 42.774
[info] Parameters(size -> 120000): 57.262
[info] Parameters(size -> 150000): 70.913

[info] ::Benchmark Serialization.sort typeclass: String::
[info] Parameters(size -> 30000): 5.543
[info] Parameters(size -> 60000): 6.832
[info] Parameters(size -> 90000): 10.322
[info] Parameters(size -> 120000): 13.581
[info] Parameters(size -> 150000): 18.841

[info] ::Benchmark Serialization.sort kryo: String::
[info] Parameters(size -> 30000): 13.386
[info] Parameters(size -> 60000): 23.989
[info] Parameters(size -> 90000): 35.569
[info] Parameters(size -> 120000): 47.564
[info] Parameters(size -> 150000): 59.821

[info] ::Benchmark Serialization.sort typeclass: (Int, (Long, String))::
[info] Parameters(size -> 30000): 5.46
[info] Parameters(size -> 60000): 7.003
[info] Parameters(size -> 90000): 10.76
[info] Parameters(size -> 120000): 14.112
[info] Parameters(size -> 150000): 18.597

[info] ::Benchmark Serialization.sort kryo: (Int, (Long, String))::
[info] Parameters(size -> 30000): 20.658
[info] Parameters(size -> 60000): 41.841
[info] Parameters(size -> 90000): 63.058
[info] Parameters(size -> 120000): 85.396
[info] Parameters(size -> 150000): 105.619
```

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
johnynek: Actually, I already have a listset in the tests, and it uses the same code path. I think we're safe here. Can we drop this issue?

johnynek:
@isnotinvain I think this is ready to go. Addressed (almost) all your issues. Please take a look.

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
isnotinvain: Yeah, so that was my point:
`assert(oBufCompare(a, b) + oBufCompare(b, a) === 0`
is too strict right?
Should be:
`assert(oBufCompare(a, b).signum + oBufCompare(b, a).signum === 0`

unless these compare methods always wind up going through `unsafeToInt` ?

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
isnotinvain: so that approach, as best I can tell, would generate strings with invalid characters in it. Not all 16 bit numbers are valid jvm characters -- some of them are only valid when immediately followed by certain other characters.

http://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#unicode

If scalacheck gets around this by filtering those out, then we might not be testing a particular type of character. I'm probably overthinking this though.

scalding-macros/src/test/scala/com/twitter/scalding/macros/MacroOrderingProperties.scala
isnotinvain: ok

PR #1210, opened 2015-02-26
DO NOT MERGE: fast-binary-sort macros scrooge thrift

johnynek:
this is on top of #1209 

There are issues with using this. If you have thrifts with fields that are neither required nor optional, you can have nulls in your fields. In that case, our macros cannot currently see that, and you will get a null pointer exception at runtime.

The only solution to this problem is to treat all fields as optional which has some size and performance implications. The way forward is not 100% clear here.


mcanlas:
Stale PR. Should we close?

PR #1212, opened 2015-02-26
DO NOT MERGE: fast-binary-sort adds TBase OrderedSerialization

johnynek:
This is on top of #1210 only merge after that has been merged.

Tasks:
- [ ] This has disabled the scrooge-tbase comparators. We need to consider which route to go. One solution may be to provide both the macro based and the TBase approach for scrooge and let users opt in to one or the other.
- [ ] Due to string comparison not matching binary comparison of `Array[Byte]` due to string compare effectively being unsigned, but `Array[Byte]`/ByteBuffer comparison being signed we need to be careful about the OrderedSerialization contract. We should do our best to find a high performance way to follow the laws.
- [ ] Internal to Twitter there may be a more recent TBase comparator. We need to make sure we have the latest and greatest here.


mcanlas:
Stale PR. Should we close?

johnynek:
yeah, I think no one I know of is really dying to use OrderedSerialization with scalding with thrift (other than scrooge).

Issue #1213, opened 2015-02-26
Add rowL1Normalize to Matrix2 API


oeddyo:
For some reason rowL1Normalize is in Matrix but not Matrix2. Thus I would like to migrate the function to Matrix2

in Matrix.scala
  def rowL1Normalize(implicit ev: =:=[ValT, Double]): Matrix[RowT, ColT, Double] = rowL1Norm

PR #1214, opened 2015-02-26
Exie/add row l1 normalize

oeddyo:


oeddyo:
First time contributing to a github project. Sorry I'm not sure how to link the issue here, but here's the link:

https://github.com/twitter/scalding/issues/1213

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
```diff
@@ -142,6 +142,19 @@ sealed trait Matrix2[R, C, V] extends Serializable {
     MatrixLiteral(result, SizeHint.asDiagonal(this.sizeHint.setRowsToCols)) * matD
   }
 
+  /**
+   * Row L1 normalization (can only be called for Double)
+   * After this operation, the sum(|x|) alone each row will be 1.
+   */
+  def rowL1Normalize(implicit ev: =:=[V, Double], mj: MatrixJoiner2): Matrix2[R, C, Double] = {
```
johnynek: can we do this:
```scala
def rowL1Normalize(implicit num: Numeric[V], mj: MatrixJoiner2): Matrix[R, C, Double] = {...
```
then use `num.toDouble(x).abs` on line 151? This will let us work with Int, Long, Float, etc...

Also if you could make the same change to rowL2Normalize, I'd be pumped!

johnynek:
Thanks so MUCH!!

oeddyo:
Made it numeric[V]

the * operator needs two matrix to be the same type thus in rowL1Normalize I have to convert it to Double type first (since the output would be double anyway) and then use the * operator. 

Please kindly let me know how can I improve

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
```diff
@@ -142,6 +142,19 @@ sealed trait Matrix2[R, C, V] extends Serializable {
     MatrixLiteral(result, SizeHint.asDiagonal(this.sizeHint.setRowsToCols)) * matD
   }
 
+  /**
+   * Row L1 normalization
+   * After this operation, the sum(|x|) alone each row will be 1.
+   */
+  def rowL1Normalize(implicit num: Numeric[V], mj: MatrixJoiner2): Matrix2[R, C, Double] = {
+    val matD = MatrixLiteral(this.asInstanceOf[Matrix2[R, C, V]].toTypedPipe.map{ case (r, c, x) => (r, c, num.toDouble(x).abs) }, this.sizeHint)
```
johnynek: you don't need the asInstanceOf anymore right? Can we remove that?

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
```diff
@@ -130,10 +130,10 @@ sealed trait Matrix2[R, C, V] extends Serializable {
   }
 
   /**
-   * Row L2 normalization (can only be called for Double)
+   * Row L2 normalization
    * After this operation, the sum(|x|^2) along each row will be 1.
    */
-  def rowL2Normalize(implicit ev: =:=[V, Double], mj: MatrixJoiner2): Matrix2[R, C, Double] = {
+  def rowL2Normalize(implicit num: Numeric[V], mj: MatrixJoiner2): Matrix2[R, C, Double] = {
     val matD = this.asInstanceOf[Matrix2[R, C, Double]]
```
johnynek: this cast needs to be removed as you did below.

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
```diff
@@ -221,6 +221,29 @@ class Matrix2Cosine(args: Args) extends Job(args) {
...
+
+  val p1: Pipe = Tsv("mat1", ('x1, 'y1, 'v1)).read
+  val tp1 = p1.toTypedPipe[(Int, Int, Double)](('x1, 'y1, 'v1))
+  val mat1 = MatrixLiteral(tp1, NoClue)
+
+  // Now test for the case when value is Long type
+  val matL1Norm = mat1.rowL1Normalize
+  matL1Norm.write(TypedTsv[(Int, Int, Double)]("normalized"))
+
+  val p2: Pipe = Tsv("mat2", ('x2, 'y2, 'v2)).read // test Long type as value is OK
```
johnynek: let's use `TypedTsv` here and not use Pipe at all (just go right to `TypedPipe.from(TypedTsv[(Int, Int, Long)]("long_normalized")`)

oeddyo:
@johnynek Can we have this added please? I'm using implicit class to inject this part in our project in Twitter, but would love to use this patch in Matrix2 directly :)

If there's anything I could improve, please kindly let me know

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
```diff
@@ -130,18 +130,31 @@ sealed trait Matrix2[R, C, V] extends Serializable {
...
   /**
-   * Row L2 normalization (can only be called for Double)
+   * Row L2 normalization
    * After this operation, the sum(|x|^2) along each row will be 1.
    */
-  def rowL2Normalize(implicit ev: =:=[V, Double], mj: MatrixJoiner2): Matrix2[R, C, Double] = {
-    val matD = this.asInstanceOf[Matrix2[R, C, Double]]
-    lazy val result = MatrixLiteral(matD.toTypedPipe.map { case (r, c, x) => (r, c, x * x) }, this.sizeHint)
+  def rowL2Normalize(implicit num: Numeric[V], mj: MatrixJoiner2): Matrix2[R, C, Double] = {
+    val matD = MatrixLiteral(this.toTypedPipe.map{ case (r, c, x) => (r, c, num.toDouble(x).abs) }, this.sizeHint)
```
johnynek: why is there `.abs` here?  I think it is only needed for L1 normational.

johnynek:
Very sorry for the delay. We're planning on publishing a new verion of scalding soon to test some other code, so this will also be available with that.

In the mean time, you can just copy this code and use the enrichment pattern (implicit class that adds this method).

oeddyo:
Thanks @johnynek !

I just removed the abs :)

Sure thing. I'm using enrichment patterns now. It's wried that my own L1 normalization's implementation is about 20x slower than the one implemented following your L2 normalization. I guess it's because some extensive optimization for matrix multiplication...

I'm gonna carefully read through matrix2 api and see what I could do to improve it... any tickets would be welcomed :)

johnynek:
Just to be clear: the Matrix API is 20x faster than the one you wrote by hand?

That is great to hear. I can only guess why. There are two things that are going on here: 1) optimization of matrix multiplication and deferring some additions to later. 2) the representation used may be different. We use a sparse representation.

Issue #1211, opened 2015-02-26
Clean up warnings about type erasure


johnynek:
@adamkozuch want to take a stab at this?

```
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Cumula[130/1833]
ala:42: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]         .sortBy { case (u: U, _) => u }
[warn]                            ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.sc
ala:44: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]           case (acc, (u: U, v: V)) =>
[warn]                          ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.sc
ala:44: abstract type pattern V is unchecked since it is eliminated by erasure
[warn]           case (acc, (u: U, v: V)) =>
[warn]                                ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.sc
ala:66: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]         .map { case (k, (u: U, v: V)) => (k, partition(u)) -> v }
[warn]                             ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.sc
ala:66: abstract type pattern V is unchecked since it is eliminated by erasure
[warn]         .map { case (k, (u: U, v: V)) => (k, partition(u)) -> v }
[warn]                                   ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.sc
ala:90: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]           case (k, (u: U, v: V)) =>
[warn]                        ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.sc
ala:90: abstract type pattern V is unchecked since it is eliminated by erasure
[warn]           case (k, (u: U, v: V)) =>
```

adamkozuch:
Of course! I will look at this.

PR #1215, opened 2015-02-28
remove warnings in CumulativeSum 

adamkozuch:
Issue #1211

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
@@ -37,11 +37,13 @@ object CumulativeSum {
     val pipe: TypedPipe[(K, (U, V))]) {
     /** Takes a sortable field and a monoid and returns the cumulative sum of that monoid **/
     def cumulativeSum(
-      implicit sg: Semigroup[V], ordU: Ordering[U], ordK: Ordering[K]): SortedGrouped[K, (U, V)] = {
+      implicit sg: Semigroup[V],
+      ordU: Ordering[U],
+      ordK: Ordering[K]): SortedGrouped[K, (U, V)] = {
       pipe.group
-        .sortBy { case (u: U, _) => u }
+        .sortBy { case (u, _) => u }
         .scanLeft(Nil: List[(U, V)]) {
-          case (acc, (u: U, v: V)) =>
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
```diff
@@ -20,7 +20,7 @@ import java.io.OutputStream
 import java.io.Serializable
```
johnynek: let's not add this change.

scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
```diff
@@ -37,7 +37,12 @@ object CumulativeSum {
     val pipe: TypedPipe[(K, (U, V))]) {
     /** Takes a sortable field and a monoid and returns the cumulative sum of that monoid **/
     def cumulativeSum(
-      implicit sg: Semigroup[V], ordU: Ordering[U], ordK: Ordering[K]): SortedGrouped[K, (U, V)] = {
+      implicit sg: Semigroup[V],
+      ordU: Ordering[U],
+      ordK: Ordering[K],
+      tagK: ClassTag[K],
```
johnynek: let's not use the ClassTag. That will complicate the use of the code.

scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
```diff
@@ -60,7 +65,10 @@ object CumulativeSum {
...
       sg: Semigroup[V],
       ordU: Ordering[U],
-      ordK: Ordering[K]): TypedPipe[(K, (U, V))] = {
+      ordK: Ordering[K],
+      tagK: ClassTag[K],
+      tagV: ClassTag[V],
+      tagU: ClassTag[U]): TypedPipe[(K, (U, V))] = {
 
       val sumPerS = pipe
         .map { case (k, (u: U, v: V)) => (k, partition(u)) -> v }
```
johnynek: actually, I think it is as easy as removing the `U` and `V` and not adding the ClassTags.

adamkozuch:
Thanks for advice. Now it passed. Is that all?

johnynek:
Thank you sir! Very helpful.

PR #1219, opened 2015-03-06
provide Args as an implicit val

tdyas:
The Checkpoint extension (com.twitter.scalding.commons.extensions.Checkpoint) wants an implicit Args provided (along with FlowDef and Mode). Args is similar to the already implicitly provided FlowDef and Mode in Job and should be made available implicitly as well.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -116,6 +116,9 @@ class Job(val args: Args) extends FieldConversions with java.io.Serializable {
   implicit def iterableToRichPipe[T](iter: Iterable[T])(implicit set: TupleSetter[T], conv: TupleConverter[T]): RichPipe =
     RichPipe(toPipe(iter)(set, conv))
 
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -116,6 +116,9 @@ class Job(val args: Args) extends FieldConversions with java.io.Serializable {
   implicit def iterableToRichPipe[T](iter: Iterable[T])(implicit set: TupleSetter[T], conv: TupleConverter[T]): RichPipe =
     RichPipe(toPipe(iter)(set, conv))
 
+  // Provide args as an implicit val for extensions such as the Checkpoint extension.
+  implicit protected val _args: Args = args
```
johnynek: 1) let's make this a def (avoid having two fields with the same data).
2) to decrease the chances with collide with another name can we give it a longer name? `_implicitJobArgs` or something?

tdyas:
Updated commit as suggested.

johnynek:
Thank you!

PR #1222, opened 2015-03-09
add module to classpath on hdfs mode

mchataigner:
This fixes the missing libraries when running in hdfs mode.

### DIFF
--- a/scripts/scald.rb
+++ b/scripts/scald.rb
@@ -534,7 +534,9 @@ def hadoop_command
 end
 
 def jar_mode_command
-  "HADOOP_CLASSPATH=#{JARBASE} hadoop jar #{JARBASE} #{hadoop_opts} #{JOB} --hdfs " + JOB_ARGS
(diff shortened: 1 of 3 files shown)


scripts/scald.rb
```diff
@@ -534,7 +534,9 @@ def hadoop_command
 end
 
 def jar_mode_command
-  "HADOOP_CLASSPATH=#{JARBASE} hadoop jar #{JARBASE} #{hadoop_opts} #{JOB} --hdfs " + JOB_ARGS
+  hadoop_libjars = ([MODULEJARPATHS.map{|n| File.basename(n)}].select { |s| s != "" }).flatten.join(",")
```
ianoc: jar mode i think expects things to be a fatjar. Adding modules just to the hadoop classpath, but not also on libjars is likely to cause it to fail on the mappers/reducers?

mchataigner:
You are correct. I fixed it. Now it should work

mchataigner:
I rebased my branch to develop and messed up this pull request.
I closed it and reopened another one.

PR #1224, opened 2015-03-10
add module to classpath on hdfs mode

mchataigner:
This fixes the missing libraries when running in hdfs mode.

### DIFF
--- a/scripts/scald.rb
+++ b/scripts/scald.rb
@@ -534,7 +534,9 @@ def hadoop_command
 end
 
 def jar_mode_command
-  "HADOOP_CLASSPATH=#{JARBASE} hadoop jar #{JARBASE} #{hadoop_opts} #{JOB} --hdfs " + JOB_ARGS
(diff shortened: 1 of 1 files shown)


ianoc:
I think the notion is here is that jar mode is when you have a fat jar, which negates the notion of modules.  looks like the hadoop_command is the one using lib jars. Can you give some context as to why that doesn't work and what this change is about?

mchataigner:
When you want to launch a remote repl on a hdfs cluster with avro module, you use:
./scripts/scald.rb -a --repl --hdfs --host HOST
but then you don't have avro module loaded by the repl.
Maybe there is another way to do it, but I am not very familiar with ruby.

And one other point would be to make possible to load external libraries other than scalding modules.

ianoc:
Right that last part i'm not sure on, maybe the repl launch command is hitting the wrong method. 

-- e.g. if you look at one method above the one you've been editing, the changes your trying to add are already in there, maybe the repl launch command should be calling down to it instead?

PR #1228, opened 2015-03-17
call sourceConfInit when reading from taps in local mode

tdyas:
sourceConfInit should be called even in local mode when accessing a Tap.


johnynek:
Thanks!

Issue #1098, opened 2014-11-17
sequencefile sources won't serialize


avibryant:
The underlying issue is that SequenceFile in Cascading has a broken default constructor.

I ran into this trying to do something like this:

``` scala
someTypedPipe.toIterableExecution.flatMap{iterable =>
    lazy val list = iterable.toList
    otherTypedPipe.map{x =>
         doSomethingWith(list)
    }.writeExecution(sink)
}
```

Because `list` is lazy, the .seq actually gets read from inside the `map`, in the individual mapper tasks (except that it doesn't because there's an NPE when it tries to deserialize). If it weren't lazy, it would get read at the submitter, which would work (but then it would need to serialize the data). If I use, say, `TypedTsv`, by doing a `writeThrough`, it also works.

Arguably this is just a bad idea (because reading directly from the source in the mapper tasks could put too much pressure on the name node or something?) and I should be using `mapWithValue` or something else that uses the distributed cache? But this API seems more convenient.

Thoughts?

cwensel:
Just ran across this thread. Just pushed a fix for the default ctor in 2.6.2 wip if it matters at this point.

bwwinthehouse:
Thank you

PR #1230, opened 2015-03-18
Fix toIterableExecution

### LINKED ISSUE #1229: When using toIterableExecution it can result on the iterator being read on every mapper
users have been using toIterableExecution, taking the resulting Iterable and then using it inside a mapper function in a subsequent step. This results in every mapper loading this data from disk (possibly more than once). We should change the implementation here to use a stream or other lazy load that when serialized will materialize the entire thing in memory.

johnynek:
fixes #1229 
related to #1098

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
@@ -54,6 +54,12 @@ class KryoHadoop(config: Config) extends KryoInstantiator {
    */
   override def newKryo: Kryo = {
     val newK = (new ScalaKryoInstantiator).newKryo
+    if (!newK.alreadyRegistered[Stream[Any]]) {
+      // This is being added to chill, but make sure it
+      // is here since we rely on it
+      newK.forTraversableSubclass(Stream.empty[Any])
+    }
+
     // These are scalding objects:
     newK.register(classOf[RichDate], new RichDateSerializer())
     newK.register(classOf[DateRange], new DateRangeSerializer())
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -937,9 +937,17 @@ class TypedPipeInst[T] private[scalding] (@transient inpipe: Pipe,
       // for us. So unwind until you hit the first filter, snapshot,
       // then apply the unwound functions
       case Some((tap, fields, Converter(conv))) =>
-        Execution.from(new Iterable[T] {
-          def iterator = m.openForRead(conf, tap).asScala.map(tup => conv(tup.selectEntry(fields)))
-        })
+        /*
+           * We create a Stream here so we are lazy with construction, but we don't read twice.
+           */
+        val tryStream = scala.util.Try(m.openForRead(conf, tap)
+          .asScala
+          .map(tup => conv(tup.selectEntry(fields)))
+          .toStream)
+        // from is lazy, so the get will throw in evaluation
+        // we want to eagerly create the stream, but not throw
+        // in the factory method.
+        Execution.from(tryStream.get)
       case _ => forceToDiskExecution.flatMap(_.toIterableExecution)
     }
   }


scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
```diff
@@ -54,6 +54,12 @@ class KryoHadoop(config: Config) extends KryoInstantiator {
    */
   override def newKryo: Kryo = {
     val newK = (new ScalaKryoInstantiator).newKryo
+    if (!newK.alreadyRegistered[Stream[Any]]) {
```
ianoc: Should we open some issue to kill this again, lest we forget... Though we'll probably forget about the issue too

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -937,9 +937,17 @@ class TypedPipeInst[T] private[scalding] (@transient inpipe: Pipe,
...
-        Execution.from(new Iterable[T] {
-          def iterator = m.openForRead(conf, tap).asScala.map(tup => conv(tup.selectEntry(fields)))
-        })
+        /*
+           * We create a Stream here so we are lazy with construction, but we don't read twice.
+           */
+        val tryStream = scala.util.Try(m.openForRead(conf, tap)
+          .asScala
+          .map(tup => conv(tup.selectEntry(fields)))
+          .toStream)
```
ianoc: Do we know how this will be serialized? Kryo field serializing this I think would serialize the Head element followed by a function0 which has the Iterator in its closure?

ianoc: I believe a platform test could validate it, though i don't think it can validate easily the excessive reading problem

johnynek: Good point, when this using the KryoHadoop instantiator that I updated, so for Stream it should serialize by iterating all the data and forcing the stream. That might not be ideal, but since there is no contract that Iterable is lazy, this should be fine. If you send this Iterable to 1000 machines, the assumption should be that it is small.

If it is huge, then capturing in a closure is the wrong way to go (use a join or cross or manual closure that you create to force the laziness).

ianoc: Ok, i guess there's some special case in that its a traversable in there. Do we need a test around this? Its probably fine/tiny.

johnynek: This line here:

https://github.com/twitter/scalding/pull/1230/files#diff-950283e698e7d0a9b2bb04551c277715R60

uses the fact that Stream is traversable to use the Traversable serializer in chill. I tested that in the chill pull request:

https://github.com/twitter/chill/pull/222

I agree it would be good to have some test that fails and this fixes, although it is pretty hard to see how this could possibly still have the read from many machines error.

avibryant: Hm, so I can remember being pretty happy when I realized that this was getting read separately on each mapper (though certainly, it's *much* to easy right now to accidentally read it 1000s of times on each mapper). I think it's easy to have data here that's small enough to fit in memory on each mapper but bigger than you really want to serialize into the JobConf. The ideal world would be for it to automatically get put into the distributedcache, but not sure how hard that is to arrange. Failing that, though, my vote would be to ensure that it is only read once per mapper, but not serialized by the submitter.

johnynek: Well, there is this:

http://docs.cascading.org/cascading/2.6/javadoc/cascading-hadoop/cascading/tap/hadoop/DistCacheTap.html

Maybe that's the way to go?

avibryant: Yep, that looks right. We'll want to make sure it's cached in memory on each mapper vs. reading from the tap on every iteration through it. (The argument here being that if it's larger than memory, you should be leaving it as a TypedPipe, not making an Iterable).

avibryant: ... and using DistCacheTap might do some funky things if we try to read it from the submitter? So need to test that usecase as well.

johnynek: This is starting to sound pretty complex actually. I'm not actually sure of a reliable way to see if a function is being called on the submitter or in a task. The only way I can think of right now is to add some static method to update some state variable inside of any of the cascading tasks, and if it is not set, we must be in the submitter.

The testing for all this looks pretty non-trivial as well.

ianoc: How big are the things your reading avi that putting it in the job conf is bad? I guess if your concerned with filling up memory quite large. We are running into the issue of possibly reading it multiple times. But also the previous step can have 100's to 1000's of part files, each being read by a followup set of 1000's to 10,000's of mappers. So the number of read operations even just reading it once per mapper ends up being huge. Our SRE's are not thrilled :)

johnynek: Also note: cascading should use the distributed cache for large JobConfs (if not, that's a bug). On the other hand, I guess it will deserialize it once per map-task rather than once per machine (not sure if once per machine will even be doable with Hadoop since I don't know how it starts tasks).

cwensel: It does use the dist cache if they are largish.

avibryant: @ianoc Yes, if it's 100s of part files, being read over HDFS by 10,000 mappers, that's gonna suck - but if that becomes a single distributed cache file that they're each reading from local disk, not so bad. (And it's not like the reads magically go away just because it's in the jobconf, right? You're just adding serialization overhead).

In practice right now, I'm doing stuff in the 100MB range. So serializing into JobConf will probably work. But more generally, this feels like a primitive you might want to implement hashJoin-like things with; are we comfortable with those always working this way? Given that large JobConfs go out over the distcache anyway, does that make any other use of the distcache redundant?

reconditesea:
I think the deal/contract for toIterableExecution (and the same for toIterable) is that the data is small enough to get materialized in submitter? Otherwise, it should be read in a distributed way on the cluster. And the reason we make it a lazy collection is to reduce serialization cost b/w submitter and cluster. 

If my understanding is correct, I would vote for always materializing in submitter and adding it to JobConf for serialization/transferring. Comparing possible OOM in submitter per app basis, with possible DDOS on the NameNode of the whole cluster. I guess the former is slightly better although both are not good.

avibryant:
Ok, I can cave here. Is stuff in the JobConf still serialized with Java serialization, though, or is it Kryo now?

johnynek:
The closures are always serialized with scalding.Externalizer, which tries
to use Java and falls back to Kryo if that fails.

https://github.com/twitter/chill/blob/develop/chill-scala/src/main/scala/com/twitter/chill/Externalizer.scala#L179

The Externalizer we use is a subclass that uses the same KryoInstantiator
as scalding defaults to.

On Thu, Mar 19, 2015 at 1:53 PM, Avi Bryant <notifications@github.com>
wrote:

> Ok, I can cave here. Is stuff in the JobConf still serialized with Java
> serialization, though, or is it Kryo now?
>
> —
> <https://github.com/twitter/scalding/pull/1230#issuecomment-83757619>.
>



-- 

johnynek:
We've lived years without this fix, and there has been a total rewrite of Execution handling.

PR #1232, opened 2015-03-25
Add distinctCount and distinctValues helper methods to KeyedList.

reconditesea:
Once a while, there will be internal asks about what's a good way to do this. So add 2 helper methods to KeyedList for people to use.

Do we also wanna add a HLL based approximation helper?

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
@@ -308,7 +308,7 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
    * For each key, give the number of unique values. WARNING: May OOM.
    * This assumes the values for each key can fit in memory.
    */
-  def distinctSize[B >: T]: This[K, Long] = toSet[B].mapValues(_.size)
+  def distinctSize: This[K, Long] = toSet[T].mapValues(_.size)
 
   /**
    * For each key, remove duplicate values. WARNING: May OOM.
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
@@ -308,7 +308,7 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
    * For each key, give the number of unique values. WARNING: May OOM.
    * This assumes the values for each key can fit in memory.
    */
-  def distinctCount[B >: T]: This[K, Long] = toSet[B].mapValues(_.size)
+  def distinctSize[B >: T]: This[K, Long] = toSet[B].mapValues(_.size)
 
   /**
    * For each key, remove duplicate values. WARNING: May OOM.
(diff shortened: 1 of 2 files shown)


johnynek:
Nice! distinctCount should be distinctSize to be more consistent with scala
names and it should not have a type parameter on the method.

On Tuesday, March 24, 2015, Kevin Lin <notifications@github.com> wrote:

> Once a while, there will be internal asks about what's a good way to do
> this. So add 2 helper methods to KeyedList for people to use.
>
> Do we also wanna add a HLL based approximation helper?
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/twitter/scalding/pull/1232
> Commit Summary
>
>    - Add distinctCount and distinctValues helper method to KeyedList.
>
> File Changes
>
>    - *M*
>    scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
>    <https://github.com/twitter/scalding/pull/1232/files#diff-0> (13)
>    - *M*
>    scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
>    <https://github.com/twitter/scalding/pull/1232/files#diff-1> (35)
>
> Patch Links:
>
>    - https://github.com/twitter/scalding/pull/1232.patch
>    - https://github.com/twitter/scalding/pull/1232.diff
>
> —
> <https://github.com/twitter/scalding/pull/1232>.
>


-- 

reconditesea:
@johnynek I've updated the name to distinctSize. But I think it needs the type parameter since Set is invariant in scala.

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -303,6 +303,19 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
 
   /** For each key, give the number of values */
   def size: This[K, Long] = mapValues { x => 1L }.sum
+
+  /**
+   * For each key, give the number of unique values. WARNING: May OOM.
+   * This assumes the values for each key can fit in memory.
+   */
+  def distinctSize[B >: T]: This[K, Long] = toSet[B].mapValues(_.size)
```
johnynek:
```scala
def distinctSize: This[K, Long] = toSet[T].mapValues(_.size)
```
should work, right?

reconditesea: Yup, this should work. Just updated the code.

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -303,6 +303,19 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
...
+   * For each key, give the number of unique values. WARNING: May OOM.
+   * This assumes the values for each key can fit in memory.
+   */
+  def distinctSize: This[K, Long] = toSet[T].mapValues(_.size)
+
+  /**
+   * For each key, remove duplicate values. WARNING: May OOM.
+   * This assumes the values for each key can fit in memory.
+   */
+  def distinctValues[B >: T]: This[K, B] = toSet[B].flattenValues
```
johnynek: actually, can't we do the same trick here so that we always return `This[K, T]`? Since keyedlistlike is covariant in the value, this is just as good, and it simplifies the use case for the caller.

reconditesea: Sure thing. Updated.

reconditesea:
Is this good to go now?

johnynek:
Yes. thank you, sir!

PR #1235, opened 2015-03-28
Note the active FlowProcess for Joiners

tdyas:
Certain Scalding features (e.g., counters) require a FlowProcess. Note the FlowProcess passed into Joiner instances.

Made this a pull request to make sure the general approach is agreeable. Please comment!

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/CoGroupBuilder.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/CoGroupBuilder.scala
@@ -44,7 +44,7 @@ class CoGroupBuilder(groupFields: Fields, joinMode: JoinMode) extends GroupBuild
     val pipes = (pipe :: coGroups.map{ _._2 }).map{ RichPipe.assignName(_) }.toArray
     val joinModes = (joinMode :: coGroups.map{ _._3 }).map{ _.booleanValue }.toArray
     val mixedJoiner = new MixedJoin(joinModes)
-    val cg: Pipe = new CoGroup(pipes, fields, null, mixedJoiner)
+    val cg: Pipe = new CoGroup(pipes, fields, null, WrappedJoiner(mixedJoiner))
     overrideReducers(cg)
     evs.foldRight(cg)((op: Pipe => Every, p) => op(p))
   }
--- a/scalding-core/src/main/scala/com/twitter/scalding/JoinAlgorithms.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/JoinAlgorithms.scala
@@ -19,6 +19,7 @@ import cascading.pipe._
 import cascading.pipe.joiner._
 import cascading.tuple._
 
+import java.util.{ Iterator => JIterator }
 import java.util.Random // this one is serializable, scala.util.Random is not
 import scala.collection.JavaConverters._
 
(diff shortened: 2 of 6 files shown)


johnynek:
This approach looks good to me, but we should add some tests, and hopefully see that they fail with the flowprocess addition commented out.

This is great! thanks for the help!

tdyas:
The alternate approach I was considering was to make factory methods to wrap all of the Cascading `Pipe` classes that take a `Joiner` and add the wrapping there. That could be a simpler approach since I would only have to change call sites that actually pass a `Joiner` into Cascading. It would also remove the need to unwrap `WrappedJoiner` to compare for `InnerJoin` and `OuterJoin`.

tdyas:
The commit should be all set to merge. Added some tests. I didn't go with factory methods, and just added the wrapper directly at the applicable call sites that made either a `CoGroup` or a `HashJoin`.

tdyas:
Open question: Should `WrappedJoiner` and its companion be `private[scalding]`?

scalding-core/src/test/scala/com/twitter/scalding/JoinerFlowProcessTest.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  val inA = Tsv("inputA", ('a, 'b))
+  val inB = Tsv("inputA", ('x, 'y))
+
+  val joiner = new CheckFlowProcessJoiner(uniqueID)
+
+  // The .forceToDisk is necessary to have the test work properly.
+  inA.joinWithSmaller('a -> 'x, inB, joiner = joiner).forceToDisk.write(Tsv("output"))
+}
+
```
johnynek: can you also add a test for the Typed-API? We recommend that one now (and it uses a different code path than the field API for joins).

To test, you should be able to add a stat in a `.reduce` following a join in the Typed-API.

tdyas:
I added a Typed-API test but was unable to get it to fail by removing `WrappedJoiner`. I don't know enough about Scalding to understand how the FlowProcess was being recorded, probably due to something peculiar to the test environment. Do you know?

I also made the wrapper `private[scalding]` since it is really an internal API.

johnynek:
@tdyas I assume that since this is being run with the hadoop local mode, that we can't trigger the error as perhaps the work done on the mapper is going to setup the reducer.

We could try using the scalding-hadoop-test module, which actually spins up a mini-cluster. We should be able to repro the issue there.

tdyas:
I split the tests that checked the actual join APIs out into the scalding-hadoop-test module. I still can't get the Typed API test to fail. Additionally, I now can't get the Fields API test to fail. So something peculiar is occurring even for the MiniMRCluster environment used in scalding-hadoop-test. Does it fork off each task attempt or just run them in a thread in the same JVM?

johnynek:
Going to merge this, since it seems no worse than before even if we can't repro the bug with our unit tests.

tdyas:
Thanks.

For clarity, the `WrappedJoinerTest` can be made to fail by removing `WrappedJoiner` behavior.  That test actually takes the Scalding join code out of play by directly creating a Cascading `CoGroup` pipe. The tests in `scalding-hadoop-test` are the ones where I wasn't able to make them fail. My hypothesis is that maybe an `Operation` is in play to make the call into `RuntimeStats` to record the `FlowProcess`. 

Weirdly, I was able to make the Fields-based API test fail when it was run in local mode inside the `scalding-core` tests but not within `scalding-hadoop-test`.

wbertelsen:
Thanks!

Issue #746, opened 2014-01-21
Show an example of shapeless records + TypedAPI


johnynek:
Any takers want to make give us an example to get people to see that records can be as easy as the Fields API?

https://github.com/milessabin/shapeless

PR #1236, opened 2015-03-29
Make it easier to get around Tuple22 restrictions with Shapeless/HLists

matt-martin:
This is just a very basic attempt to integrate Shapeless (specifically HLists) into Scalding.  There's certainly more that could be done, but I _think_ this pull request represents a necessary first step.  I would love some feedback, since I'm basically brand new to Shapeless and did this partly as an exercise to try to learn more about it.  If this looks good/useful, then I suppose the next step would be thinking about how to use [extensible records](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#extensible-records) like Miles did in the [example](https://github.com/milessabin/shapeless/blob/bea51504c6972ff15d341f512487c55f498fc606/examples/src/main/scala/shapeless/examples/scalding.scala) he already put together.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -39,6 +39,7 @@ object ScaldingBuild extends Build {
   val scalaTestVersion = "2.2.2"
   val scalameterVersion = "0.6"
   val scroogeVersion = "3.17.0"
+  val shapelessVersion = "2.1.0"
   val slf4jVersion = "1.6.6"
   val thriftVersion = "0.5.0"
 
@@ -269,6 +270,17 @@ object ScaldingBuild extends Build {
       "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion % "provided"
+    ) ++ (
+      if(isScala210x(scalaVersion.value)) {
+        Seq(
+          "com.chuusai" % "shapeless_2.10.4" % shapelessVersion,
+          compilerPlugin("org.scalamacros" % "paradise_2.10.4" % "2.0.1")
+        )
+      } else {
+        Seq(
+          "com.chuusai" %% "shapeless" % shapelessVersion
+        )
+      }
     )
   ).dependsOn(scaldingArgs, scaldingDate, maple)
 
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -273,8 +273,7 @@ object ScaldingBuild extends Build {
     ) ++ (
       if(isScala210x(scalaVersion.value)) {
         Seq(
-          "com.chuusai" % "shapeless_2.10.4" % shapelessVersion,
-          compilerPlugin("org.scalamacros" % "paradise_2.10.4" % "2.0.1")
+          "com.chuusai" % "shapeless_2.10.4" % shapelessVersion
         )
       } else {
         Seq(
--- a/scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
@@ -103,12 +103,16 @@ object TupleConverter extends GeneratedTupleConverters {
   import shapeless.ops.nat._
 
   implicit def hListConverter[H, T <: HList, N <: Nat]
-    (implicit g: TupleGetter[H], len: Length.Aux[H :: T, N], toIntN : ToInt[N], fl : FromTraversable[H :: T]): TupleConverter[H :: T] =
+    (implicit
+     g: TupleGetter[H],
+     len: Length.Aux[H :: T, N],
+     toIntN : ToInt[N],
+     fl : FromTraversable[H :: T]): TupleConverter[H :: T] =
       new TupleConverter[H :: T] {
         import scala.collection.JavaConverters._
         override def apply(te: TupleEntry): H :: T = {
-          val genTraversable = te.getTupleCopy.asScala
-          val l : Option[H :: T] = fl(genTraversable)
+          val iterable = te.getTupleCopy.asScala
+          val l : Option[H :: T] = fl(iterable)
           l.get
         }
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
@@ -19,6 +19,7 @@ package com.twitter.scalding
 import cascading.tuple.TupleEntry
 import cascading.tuple.{ Tuple => CTuple }
 import shapeless._
+import shapeless.ops.nat._
 
 import scala.collection.{GenTraversable, breakOut}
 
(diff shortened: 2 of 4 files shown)


project/Build.scala
```diff
@@ -269,6 +270,16 @@ object ScaldingBuild extends Build {
       "org.apache.hadoop" % "hadoop-core" % hadoopVersion % "provided",
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-log4j12" % slf4jVersion % "provided"
+    ) ++ (
+      if(isScala210x(scalaVersion.value)) {
+        Seq(
+          "com.chuusai" % "shapeless_2.10.4" % shapelessVersion
```
ianoc: is this actually compatible with 2.10.3 -> 2.10.5?

matt-martin: My understanding (based on the [README](https://github.com/milessabin/shapeless)) is that shapeless 2.1.0 is only compatible with 2.11.x and 2.10.4, but I might be mistaken.  I think the relevant bits are as follows:

>Builds are available for Scala 2.11.x and for Scala 2.10.4
>...
>Note that for Scala 2.10.4 you must provide an explicit Scala version suffix to your shapeless dependency, and it is recommanded to add the macro paradise plugin to your build, for some macros provided by shapeless to work smoothly

For shapeless 2.0.0, they have the following comments:

>Builds are available for Scala 2.11.x and for Scala 2.10.4.
>...
>For Scala 2.10.x you must specify a Scala version of at least 2.10.2, and add either `cross CrossVersion.full` or provide an explicit Scala version suffix to your shapeless dependency
>...
>Note that Scala 2.10.x releases are compatible with each other starting from 2.10.2, so a mismatch in minor versions above would be fine.

scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
```diff
@@ -51,6 +53,37 @@ trait LowPriorityTupleConverters extends java.io.Serializable {
...
+  def apply(l : TupleEntry) : Option[Out]
+}
+
+object FromTupleEntry {
+  def apply[Out <: HList](implicit from: FromTupleEntry[Out, Nat._0]) = from
+
+  implicit def hnilFromTupleEntry[T, I <: Nat](implicit toIntI : ToInt[I]) =
+    new FromTupleEntry[HNil, I] {
+      def apply(te : TupleEntry) =
+        if(te.size() == toIntI()) Some(HNil) else None
```
matt-martin: This forces the TupleEntry to have exactly the same size as the HList.  I'm not sure if it needs to be this strict or if it would be ok for the TupleEntry to have more entries in it than the HList.

scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
```diff
@@ -67,6 +100,23 @@ object TupleConverter extends GeneratedTupleConverters {
...
+  implicit def hListConverter[H, T <: HList, N <: Nat]
+    (implicit
+     g: TupleGetter[H],
+     len: Length.Aux[H :: T, N],
+     toIntN : ToInt[N],
+     fl : FromTupleEntry[H :: T, Nat._0]): TupleConverter[H :: T] =
+      new TupleConverter[H :: T] {
+        override def apply(te: TupleEntry): H :: T = {
+          val l : Option[H :: T] = fl(te)
+          l.get
```
matt-martin: Probably not the best to just "get" the Option here. Open to other suggestions.

matt-martin:
See also: https://github.com/twitter/scalding/issues/746

scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
johnynek: shouldn't we run `toInt` once before creating the instance to guarantee that is only run once? We definitely don't work to run that on each tuple.

scalding-core/src/main/scala/com/twitter/scalding/TupleSetter.scala
```diff
@@ -65,6 +65,21 @@ object TupleSetter extends GeneratedTupleSetters {
...
+
+  implicit def hListSetter[L <: HList, N <: Nat, T <: Any]
+    (implicit len: Length.Aux[L, N], ti: ToInt[N], tl: ToList[L, T]) = new TupleSetter[L] {
+
+        override def apply(arg: L): CTuple = {
+          val list = arg.toList[T].map(_.asInstanceOf[Object])
+          new CTuple(list:_*)
+        }
+
+        override def arity: Int = ti()
```
johnynek: should be a val, to be sure we only call once.

johnynek:
This is great.

There are two thoughts I have:

1) I think we should add a scalding-shapeless subproject since dependencies are a real concern of people. Then you could import com.twitter.scalding.shapeless_support._ or something to get the needed implicits and types you have here.

2) We should show an example with a TypedPipe and do something non-trivial with it.

All said, I am starting to think we don't need this as much as we used to now that we have macros to do this (and they should have no performance issues). You can make a case class for your record and get a TypeDescriptor that has all the stuff you need to flatten into cascading.

project/Build.scala
travisbrown: Shapeless 2.1.0 has full cross-versions for 2.10 ([e.g for 2.10.5](https://repo1.maven.org/maven2/com/chuusai/shapeless_2.10.5/2.1.0/)), so it'd be better to use this:

    "com.chuusai" %% "shapeless" % "2.1.0" cross CrossVersion.full

So that we get the 2.10.5 version when Scalding updates to 2.10.5.

(I don't remember whether you need `%%` for full cross-versioning—I think either will work, and I find it a little clearer with `%%`.)

scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
```diff
@@ -67,6 +100,23 @@ object TupleConverter extends GeneratedTupleConverters {
   def arity[T](implicit tc: TupleConverter[T]): Int = tc.arity
   def of[T](implicit tc: TupleConverter[T]): TupleConverter[T] = tc
 
+  import shapeless.ops.hlist._
+
+  implicit def hListConverter[H, T <: HList, N <: Nat]
+    (implicit
+     g: TupleGetter[H],
+     len: Length.Aux[H :: T, N],
```
travisbrown: I'm still getting a sense of the design here, but any reason you don't constrain the `FromTupleEntry` instances so that the `I` (which could be a type member) matches the length of the `HList`? That should allow you to omit the `N` type parameter here altogether.

scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
```diff
@@ -51,6 +53,37 @@ trait LowPriorityTupleConverters extends java.io.Serializable {
...
+ * Based on `FromTraversable` from shapeless
+ */
+trait FromTupleEntry[Out <: HList, I <: Nat] {
+  def apply(l : TupleEntry) : Option[Out]
+}
+
+object FromTupleEntry {
+  def apply[Out <: HList](implicit from: FromTupleEntry[Out, Nat._0]) = from
+
+  implicit def hnilFromTupleEntry[T, I <: Nat](implicit toIntI : ToInt[I]) =
```
travisbrown: `T` is unused here.

travisbrown:
See [my branch here](https://github.com/travisbrown/scalding/commit/7503fe9bb1604b40663fb5df62d0fd4a4264b4f2) for a quick demonstration of an alternative approach.

The key idea is that you don't need to track the length of the `HList` separately at the type level, so there are no `Nat` type parameters anywhere. You also don't need to check the size of the `TupleEntry` at each element, and my implementation avoids this by using a helper method that assumes the length has already been checked.

More controversially, you don't necessarily even need a separate `FromTupleEntry` since you can get rid of the `Nat` type parameter, and my `FromTupleEntry` extends `TupleConverter` directly. This means that instead of carrying around a `Nat

All of the new tests pass unchanged.

travisbrown:
I should add that this would be a pretty fantastic start even if you _weren't_ brand new to Shapeless, @matt-martin.

matt-martin:
Thanks all for the comments.  Only skimmed them briefly today, but will take a closer look at them tomorrow

milessabin:
I'm very happy to support things like this ... ping me if there's anything I can do to help.

travisbrown:
One other thing: it might be worth trying to do this via `ProductTypeClass`. I haven't tried it, and getting it to play nicely with everything else might be tricky, but if it worked it could give you converters for case classes for free.

ianoc:
One thing on a code perspective, given shapelesses tighter requirements around scala versioning and adding another dep I think it would make sense to move this into a scalding-shapeless module. It'll make it easier to accept/merge.

milessabin:
FWIW, from shapeless-2.0.0 I'm going to drop the specific sub-version requirements for Scala 2.10.x.

scalding-core/src/main/scala/com/twitter/scalding/TupleConverter.scala
johnynek: So, TupleConverter and TupleSetter are the places where types are added and removed before handing to cascading. It can't be type-safe at that point because cascading is not type-safe. There is really nothing you can do but call .get. That said, in a well typed scalding flow, that .get can never throw, so ultimately it would be safe (even though using the TupleConverter on its own, would not be safe).

PR #1241, opened 2015-04-01
Optimize string and (hopefully) number comparisons a bit

johnynek:
This does a couple of things:

1) don't use byte-by-byte binary comparators on primitives. It appears that is not a win in Hadoop, and my guess is that all the branching is not worth it.

2) compare strings Int-by-Int rather than Long-by-Long. The idea here is that unsigned Int compare is faster than Unsigned Long compare (just upcast and compare).

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
@@ -71,7 +71,7 @@ class PositionInputStream(val wraps: InputStream) extends InputStream {
 
   override def skip(n: Long): Long = {
     if (n < 0) illegal("Must seek fowards")
-    val count = skip(n)
+    val count = wraps.skip(n)
     // Make this branch true as much as possible to improve branch prediction
     if (count >= 0) pos += count
     count
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
@@ -42,20 +42,22 @@ class PositionInputStream(val wraps: InputStream) extends InputStream {
   override val markSupported: Boolean = wraps.markSupported
 
   override def read: Int = {
+    // returns -1 on eof or 0 to 255 store 1 byte.
     val result = wraps.read
-    // returns -1 on eof, otherwise non-negative number
     if (result >= 0) pos += 1
     result
   }
   override def read(bytes: Array[Byte]): Int = {
     val count = wraps.read(bytes)
-    if (count > 0) pos += count
+    // Make this branch true as much as possible to improve branch prediction
+    if (count >= 0) pos += count
     count
   }
 
   override def read(bytes: Array[Byte], off: Int, len: Int): Int = {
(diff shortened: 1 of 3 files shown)


scalding-macros/src/main/scala/com/twitter/scalding/macros/impl/ordered_serialization/providers/PrimitiveOrderedBuf.scala
```diff
@@ -76,7 +76,8 @@ object PrimitiveOrderedBuf {
     val bbPutter = newTermName("write" + shortName)
 
     def genBinaryCompare(inputStreamA: TermName, inputStreamB: TermName): Tree =
-      if (Set("Float", "Double", "Character").contains(javaTypeStr)) {
+      //if (Set("Float", "Double", "Character").contains(javaTypeStr)) {
+      if (true) {
```
ianoc: if(true) ?

johnynek: :) I should have changed that. Sorry. That was a quick hack to force that branch (rather than cutting all the code out). Let me see if the unit tests pass here, then I'll fix this.

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
```diff
@@ -70,7 +72,8 @@ class PositionInputStream(val wraps: InputStream) extends InputStream {
   override def skip(n: Long): Long = {
     if (n < 0) illegal("Must seek fowards")
     val count = skip(n)
```
jnievelt: Not so related, but this should be wraps.skip(n)?

johnynek: shit. Yes.

scalding-core/src/main/scala/com/twitter/scalding/serialization/StringOrderedSerialization.scala
```diff
@@ -60,22 +49,23 @@ object StringOrderedSerialization {
     /**
      * Now we start by comparing blocks of longs, then maybe 1 int, then 0 - 3 bytes
```
jnievelt: Should update this

scalding-core/src/main/scala/com/twitter/scalding/serialization/PositionInputStream.scala
```diff
@@ -69,8 +71,9 @@ class PositionInputStream(val wraps: InputStream) extends InputStream {
 
   override def skip(n: Long): Long = {
     if (n < 0) illegal("Must seek fowards")
-    val count = skip(n)
```
ianoc: how did this work before? do we never hit skip?

johnynek: no, we don't use it. We use skipFully, which calls the right one. Just hear to be a complete wrapper.

ianoc: Ok cool

ianoc:
LGTM in terms of code, merge when green modulo comment fix

Issue #1237, opened 2015-03-30
Execution.zip can result in recomputation


johnynek:
Consider the case with #1217 and we do
```scala
val a = in.make(storage)
val a1 = a.flatMap(fn1)
val a2 = a1.flatMap(fn2)
val a3 = a1.flatMap(fn3)
val a4 = a2.zip(a3)
```

here is a classic diamond. I think your case can be relabeled to look like this. a = Execution.getMode. So, why is it evaluating everything twice. I think the answer is a subtle distinction between the evaluation of Execution[T] => Future[T] and the evaluations of cascading flows => T. Really what happens in some cases is `Execution[T] => Flow`, then `Flow => Future[T]`. I think the problem here is that we are sharing an evaluation of `Execution[T] => Future[T]` to on the first step, but then you have a `Future[TypedPipe[U]]`, which is actually still a program that will be run to create some sequence of `U`. So, both sides of the diamond share the same `TypedPipe[U]`, but that does not mean they will share the same on disk (unfortunately).

This case is a bit confusing, and I think it brings up that we need to:

1) be a little more clear about what the contracts and recomputation scenarios are in Execution. Right now, I guess the truth is when you use zip, you could wind up recomputing everything twice (although flatMap should not have that problem, but could reduce parallelism and increase latency).
2) Try to minimize these cases more so that there are fewer cases (hopefully none) of recomputation. I'm not sure that is always possible.

This may be a particular issue with the `writeThrough` method. It would be good to have some tests here. We do have one test:

https://github.com/twitter/scalding/blob/develop/scalding-core/src/test/scala/com/twitter/scalding/typed/ExecutionTest.scala#L184

but it actually dodges the issue. Yes we render 1 typed-pipe, but putting that in two flows with zip is going to cause it to be created twice.

johnynek:
@ianoc suggests something leading to the following:

Should we be able to always ensure the same computation from:
```scala
val first = a.flatMap(f).zip(a.flatMap(g))
val second = a.flatMap { r => f(r).zip(g(r)) }
```
Right now, this is not the case. .zip on a flatMapped Execution is different from a .zip on pure cascading operation (from .writeExecution for instance). In the first case, we hide from cascading the structure, but in the second case, we may implement zip by merging flows in cascading, which should give a better plan.

johnynek:
should be closed by #1253

PR #1242, opened 2015-04-02
Improve Execution semantics

johnynek:
Do not merge this yet.

Next week, I'd like to do the following:
- [ ] add a test that exhibits the re-computation issue we have seen.
- [ ] look at using the DagOptimizer code from summingbird to optimize the entire Execution DAG before running.
- [ ] Think about how counters interact with caching.

The problem is when you need to evaluate `a` and `a.zip(b)` in two different positions in an Execution DAG. If we want to merge as many cascading flows together, we can't evaluate a by itself, we want to run a.zip(b) and then evaluate a as the projection of the left side of that, but that does not seem to be sound because we can't un-merge counters. So, I suppose an optimizer is going to have to see if the counters are ever used (MapCounters) and if not, it may be possible to do something smart.

addresses #1237

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,11 @@ matrix:
   include:
 #BASE TESTS
     - scala: 2.10.4
-      env: BUILD="base" TEST_TARGET="scalding-args scalding-date"
+      env: BUILD="base" TEST_TARGET="scalding-args scalding-date scalding-graph"
       script: "scripts/run_test.sh"
 
     - scala: 2.11.5
-      env: BUILD="base" TEST_TARGET="scalding-args scalding-date"
+      env: BUILD="base" TEST_TARGET="scalding-args scalding-date scalding-graph"
       script: "scripts/run_test.sh"
 
     - scala: 2.10.4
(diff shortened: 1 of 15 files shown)


johnynek:
Added some dag optimization rules, and tests on the number of executions. Unfortunately, it looks like looking inside flatMaps is pretty important for optimization. The thing with a Monad is, you can't statically optimize it too much because what happens in the flatMap is a black-box to the optimizer.

I think there may be a nice way around this, but I have yet to figure it out. One idea could be to add a new Execution type:
```scala
case class FutureWrapped[T](fut: Future[Execution[T]]) extends Execution[T]
```
then maybe when we run, we can evaluate the previous node in the flatMap, get a Future which gives us a `Future[Execution[U]]`. We can lift all children into a zip of any such futures, then optimize the inner execution, and continue evaluating.

Another idea is to use the Dag rewriting to Store a Dag of `Either[Execution[T], Future[T]]`. Then we optimize that graph after each `Future[T]` completes.

johnynek:
Abandoning this. See the comments in #1253

PR #1243, opened 2015-04-02
Swap hash and filename for filename-extension-sensitive code

gerashegalov:
This is required if one wants to read a compressed file whose compression codec is determined by the filename extension. It's the case with many hadoop text-based input formats.


ianoc:
LGTM

PR #1244, opened 2015-04-03
Provide useHdfsLocalMode for an easy switch to mapreduce local mode

gerashegalov:
Sometimes users need to quickly switch to local mapreduce. Instead of typing back and forth for hadoop2

```
scalding> config += "mapreduce.framework.name" -> "local"
// do local work
scalding> config += "mapreduce.framework.name" -> "yarn"
// do cluster work
...
```

and even worse for  hadoop1 where you need to type the JT address

we can have `useHdfsLocalMode` similar to `useHdfsMode`

### DIFF
--- a/scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
+++ b/scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
@@ -26,9 +26,11 @@ import scala.concurrent.{ Future, ExecutionContext => ConcurrentExecutionContext
  * Most of these conversions come from the [[com.twitter.scalding.Job]] class.
  */
 object ReplImplicits extends FieldConversions {
-  val mr1Key = "mapred.job.tracker"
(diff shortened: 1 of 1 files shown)


scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -26,6 +26,9 @@ import scala.concurrent.{ Future, ExecutionContext => ConcurrentExecutionContext
  * Most of these conversions come from the [[com.twitter.scalding.Job]] class.
  */
 object ReplImplicits extends FieldConversions {
+  val mr1Key = "mapred.job.tracker"
```
johnynek: can these fields be private? ReplImplicits is imported by everyone using the REPL.

gerashegalov: Yes, done!

johnynek:
LGTM, merge when green.

PR #1251, opened 2015-04-10
Use non-regex split function

oeddyo:


scalding-core/src/test/scala/com/twitter/scalding/StringUtilityTest.scala
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+  "be able to split when separator doesn't show up" in {
+    val text2 = "a"
+    val res2 = StringUtility.fastSplit(text2, "@")
+    res2 should be {
+      Seq("a")
+    }
+  }
+  "be able to generate the same result with Java's split for random strings and separators" in {
```
johnynek: for this case, let's instead use a property check. It will be much easier:

```scala
property("fastSplit on comma is correct") = forAll { (s: String) =>
  s.split(",") == fastSplit(s, ",")
}
```
see some of the other property check style tests in the repo (those that extend Properties) to see an example.

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -125,7 +125,7 @@ object RuntimeStats extends java.io.Serializable {
   def addFlowProcess(fp: FlowProcess[_]) {
     val uniqueJobIdObj = fp.getProperty(UniqueID.UNIQUE_JOB_ID)
     if (uniqueJobIdObj != null) {
-      uniqueJobIdObj.asInstanceOf[String].split(",").foreach { uniqueId =>
+      StringUtility.fastSplit(uniqueJobIdObj.asInstanceOf[String], ",").foreach { uniqueId =>
```
johnynek: this method will be called on every key in a group now (to fix a bug in counters inside of joins).

so, we really need to optimize this method.

1) let's not use .foreach. We should write a while loop.
2) we should check if flowMappingStore has the key already. If it does, there is no need to allocate a WeakReference.

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
```diff
@@ -0,0 +1,39 @@
...
+  private def isTrailingWith(text: String, separator: String) = {
+    val result = text.foldRight((List[Boolean](), true)) {
+      case (c, (tillNow, isTrailingBefore)) =>
+        val trailing = isTrailingBefore && (c.toString == separator)
+        (tillNow ++ List(trailing), trailing)
+    }
+    result._1.reverse
+  }
+
+  private def fastSplitHelper(text: String, key: String, isTrailingAt: List[Boolean]): Seq[String] = {
```
johnynek: use an IndexedSeq here, not a List. List is O(N) to access.

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
```diff
@@ -0,0 +1,39 @@
+package com.twitter.scalding
+
+object StringUtility {
```
johnynek: do we have some benchmarks to show this is faster? I'm concerned that there are a few things in this code that look slow (List ++, foldRight on the string).

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
```diff
@@ -0,0 +1,39 @@
...
+  // Scala's split function uses regex which is too slow thus we create this faster function
+  private def isTrailingWith(text: String, separator: String) = {
+    val result = text.foldRight((List[Boolean](), true)) {
+      case (c, (tillNow, isTrailingBefore)) =>
+        val trailing = isTrailingBefore && (c.toString == separator)
+        (tillNow ++ List(trailing), trailing)
```
johnynek: rather than putting on the back and calling reverse, why not: `trailing :: tilNow` which is O(1) and then you don't need the reverse?

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
```diff
@@ -0,0 +1,39 @@
...
+    }
+    result._1.reverse
+  }
+
+  private def fastSplitHelper(text: String, key: String, isTrailingAt: List[Boolean]): Seq[String] = {
+    if (text.isEmpty) {
+      Seq()
+    } else {
+      val firstIndex = text.indexOf(key)
+      if (firstIndex == -1) {
```
johnynek: this is probably the common case. I'd like to see us optimize for this case so in that case our only call is text.indexOf(key).

Also, why isn't something like this good enough:

```scala
def fastSplit(s: String, sep: String, from: Int): List[String] = {
  val idx = s.indexOf(sep, from)
  if (idx == -1) List(s)
  else {
    (s.substring(from, idx)) :: fastSplit(s, sep, idx + sep.length)
  }
}
```
Seems like something like that is going to be faster and simpler.

isnotinvain:
Don't we already have fast split in chill?
What's the isTrailingWith for?

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
oeddyo: Hmm.. I was trying to make the fastSplit function to be exactly the same with the regex split function. 

One thing is that, for a string trailing with the separator, regex split drops all of them. 

i.e. "a:b::" would be "a", "b". But this would gives "a", "b", ""

Would this be good enough since our usage might not include trailing separators?

oeddyo:
@isnotinvain Yeah but it only works with ":" as separator

oeddyo:
I guess the problem is: whether we want this function to have exactly the same output as java regex split. Could you @Oscar give me some example input and output? 

The complication I added in the code came from the edge case that java split would trim trailing separators. For instance "a:b:c:::" would be -> "a", "b", "c" instead of "a", "b", "c", "", ""

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
oeddyo: I just implemented a new one after reading your comments. Will post soon:
             mean, std, median running time (nano seconds)
Java's: (7770.3,13256.792142520753,4000.0)
Fast:    (4087.7,39296.86258100002,3000.0)

using a small epoc of 100 strings to separate, with a-z with "," only randomly generated. For instance "a,bvc,sdf,fw,f,,,"

I found no matter how large the epoc is, the std is pretty huge. Not sure why. But the median running time should reflect it's faster than java's implementation for most of the time

oeddyo:
BTW: I just found that the java's implementation is actually funny. 

In the code it says:
        /* fastpath if the regex is a
         (1)one-char String and this character is not one of the
            RegEx's meta characters ".$|()[{^?*+\\", or
         (2)two-char String and the first char is the backslash and
            the second is not the ascii digit or ascii letter.
         */
 
Thus if the regex is something like ",", then it's essentially doing the same thing with the fastSplit function. The gain comes from when the separator is "|*$+" etc. That being said, we are using "," all the places as separator thus I expect we can't really gain too much. What do you guys think?

here's the java's split function code
  public String[] split(String regex, int limit) {
        /* fastpath if the regex is a
         (1)one-char String and this character is not one of the
            RegEx's meta characters ".$|()[{^?*+\\", or
         (2)two-char String and the first char is the backslash and
            the second is not the ascii digit or ascii letter.
         */
        char ch = 0;
        if (((regex.value.length == 1 &&
             ".$|()[{^?*+\\".indexOf(ch = regex.charAt(0)) == -1) ||
             (regex.length() == 2 &&
              regex.charAt(0) == '\\' &&
              (((ch = regex.charAt(1))-'0')|('9'-ch)) < 0 &&
              ((ch-'a')|('z'-ch)) < 0 &&
              ((ch-'A')|('Z'-ch)) < 0)) &&
            (ch < Character.MIN_HIGH_SURROGATE ||
             ch > Character.MAX_LOW_SURROGATE))
        {
            int off = 0;
            int next = 0;
            boolean limited = limit > 0;
            ArrayList<String> list = new ArrayList<>();
            while ((next = indexOf(ch, off)) != -1) {
                if (!limited || list.size() < limit - 1) {
                    list.add(substring(off, next));
                    off = next + 1;
                } else {    // last one
                    //assert (list.size() == limit - 1);
                    list.add(substring(off, value.length));
                    off = value.length;
                    break;
                }
            }
            // If no match was found, return this
            if (off == 0)
                return new String[]{this};

            // Add remaining segment
            if (!limited || list.size() < limit)
                list.add(substring(off, value.length));

            // Construct result
            int resultSize = list.size();
            if (limit == 0)
                while (resultSize > 0 && list.get(resultSize - 1).length() == 0)
                    resultSize--;
            String[] result = new String[resultSize];
            return list.subList(0, resultSize).toArray(result);
        }
        return Pattern.compile(regex).split(this, limit);
    }

oeddyo:
More benchmark:

Using "," and "|" as separators respectively, to split 100,000 randomly generated "," separated strings each 50 characters. Run the experiment 100 times to make reduce randomness

image

The gain is more obvious when "|" is used.

oeddyo:
Will fix it later tmr.

johnynek:
Nice speed up!

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
```diff
@@ -0,0 +1,25 @@
+package com.twitter.scalding
+
+import java.util
+
+import scala.collection.mutable.ArrayBuffer
```
johnynek: not using this anymore right? Or java.util?

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -125,9 +125,15 @@ object RuntimeStats extends java.io.Serializable {
...
-      uniqueJobIdObj.asInstanceOf[String].split(",").foreach { uniqueId =>
+      // for speed concern, use a while loop instead of foreach here
+      val splitted = StringUtility.fastSplit(uniqueJobIdObj.asInstanceOf[String], ",")
+      val size = splitted.length
+      var i = 0
+      while (i < size) {
+        val uniqueId = splitted(i)
         logger.debug("Adding flow process id: " + uniqueId)
-        flowMappingStore.put(uniqueId, new WeakReference(fp))
+        flowMappingStore.getOrElse(uniqueId, new WeakReference(fp))
```
johnynek: can you make this:

```scala
if(!flowMappingStore.contains(uniqueId)) {
  logger.debug("Adding flow process id: " + uniqueId)
  flowMappingStore += (uniqueId -> new WeakReference(fp))
}
```
The lazy parameter does an allocation that we could skip for this method if we are calling it a lot.

johnynek:
Also, can you add something like this:

```scala
126   /*
127    * Here is variable which prove that we have already processed a FlowProcess
128    */                                             
129   private[this] var lastFlowProcess: WeakReference[FlowProcess[_]] = null
130   def addFlowProcess(fp: FlowProcess[_]) {
131     if (lastFlowProcess != null && lastFlowProcess.underlying.get eq fp) {
132       // we have already processed this fp for sure
133     }   
```
to addFlowProcess. Basically, if we just called this method on the same referenced FlowProcess, don't do anything. This should be a way to avoid doing any work in the usual case in joins, which unfortunately have to call this method on each key.

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
johnynek: My fault: I was wrong. .contains won't work here because the WeakReference may be out of scope (and is in the tests), I think. We just need to put each time.

To save the work, maybe keep a
```scala
private[this] var prevFP: FlowProcess = null
```
and `if (prevFP == fp)` don't do any work. This will be the common case (especially in the WrappedJoiner), so the branch predictor should see that is almost always true, and it should be almost free for users that don't use stats.

isnotinvain: why change `.put` to `.getOrElse` ? that seems very different, was that intentional?

johnynek: I think .getOrElseUpdate was intended.

scalding-core/src/test/scala/com/twitter/scalding/StringUtilityTest.scala
```diff
@@ -0,0 +1,95 @@
+package com.twitter.scalding
+
+import org.scalatest.{ Matchers, WordSpec }
+
+class StringUtilityTest extends WordSpec with Matchers {
```
isnotinvain: how about a test / property test that says "fast split should behave the same as String.split?

oeddyo: @isnotinvain Now that I dropped the part to match java's split, I'm dropping this test for comparing fastSplit and java's regex split.

isnotinvain: wait, why?
I think we should assert that this matches String.split, or `String.split(s, -1)` if you want that behavior. But we have a reference that we know works.

johnynek: we don't want to support regexs for splits. So it should match as long is the separator is not a regex.

isnotinvain: yes, that's what I meant.

assert that it matches `String.split(Arbitrary[String], OneOf(",", "foo", ":"), -1)`

the -1 makes split behave the way you'd expect, eg, just splits on the delimiter, includes empty strings when there are double delimiters and so on.

oeddyo: @isnotinvain see your point now. Will fix it. Thanks :+1:

oeddyo:
Can we merge this? And also is it possible to add me as a contributor? :D

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -122,12 +122,20 @@ object RuntimeStats extends java.io.Serializable {
...
+      val uniqueJobIdObj = fp.getProperty(UniqueID.UNIQUE_JOB_ID)
+      if (uniqueJobIdObj != null) {
+        // for speed concern, use a while loop instead of foreach here
+        var splitted = StringUtility.fastSplit(uniqueJobIdObj.asInstanceOf[String], ",")
+        while (!splitted.isEmpty) {
+          val uniqueId = splitted.head
+          splitted = splitted.tail
+          logger.debug("Adding flow process id: " + uniqueId)
+          flowMappingStore.put(uniqueId, new WeakReference(fp))
+          prevFP = fp
```
johnynek: actually, this line should be moved outside of the if on line 129. We want to record that we processed this, even when there is no UniqueID.

scalding-core/src/test/scala/com/twitter/scalding/StringUtilityTest.scala
```diff
@@ -0,0 +1,67 @@
...
+    s <- Gen.pick(5, List.fill(100)(List("k", "l", "m", "x", "//.")).flatten)
+
+  } yield s
+
+  // test for one separator and two
+  val randomSeparator = for {
+    s <- Gen.oneOf("@@", "@", "x", "//.")
+  } yield s
+
+  property("blah") {
```
johnynek: let's give this a better name. What about "fastSplit(s, sep) should match s.split(sep, -1) for non-regex sep"

johnynek:
@oeddyo Thanks! almost there, just a couple of small changes, then we can merge. Can't wait!

Also, please add your name in the README.md in the root directory so your name shows as a contributor there (and of course will show up in the stats page of github).

PS: our style for merging is when someone other than the author likes the result, and the tests are green, we merge.

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -122,13 +122,21 @@ object RuntimeStats extends java.io.Serializable {
...
   }
 
+  private[this] var prevFP: FlowProcess[_] = null
   def addFlowProcess(fp: FlowProcess[_]) {
-    val uniqueJobIdObj = fp.getProperty(UniqueID.UNIQUE_JOB_ID)
-    if (uniqueJobIdObj != null) {
-      uniqueJobIdObj.asInstanceOf[String].split(",").foreach { uniqueId =>
-        logger.debug("Adding flow process id: " + uniqueId)
-        flowMappingStore.put(uniqueId, new WeakReference(fp))
+    if (prevFP != fp) {
```
johnynek: actually, we want:

```scala
if (!(prevFP eq fp)) {
```
because we just want to use reference equality on this check (to make sure we never trigger some expensive structural equals method that could exist or be added to FlowProcess). Sorry, just noticed this.

oeddyo:
@johnynek I have a thing called fiber internet :D

johnynek:
merge when green!

Thanks Eddie!

scalding-core/src/main/scala/com/twitter/scalding/StringUtility.scala
johnynek: the std. dev is probably huge due to occasionally hitting GC. Properly benchmarking is a bit of an art. There are some packages to make it easier, but this is fine for now.

oeddyo:
As an author of a PR I guess I shouldn't merge myself even it's green? or is the "merge when green" imperative sentence? :D

johnynek:
You can merge if someone else say "merge when green". We really would like a button on github that does that. I assume @caniszczyk got you setup in the Twitter org, so you should be able to click merge (following the rules: one shipit/LGTM and no -1/do-not-ship).

oeddyo:
@johnynek Ah thanks. Scalding is my first serious open source project actually. Learning a lot from you guys! :)

Issue #1248, opened 2015-04-10
Sketched always requites the #reducers as the parameter


reconditesea:
In
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala#L61,

We defined reducers = Some(numberReducers), but never used it. Therefore numberReducers becomes a required parameter. Either we need to add an overrideReducers method, which checks if numberReducers is set and if not use default # of reducers in the job setting, or we should at least remove the two defs for reducers = Some(numberReducers) since they are not used and confusing.

PR #1252, opened 2015-04-13
Add MustHasReducers trait.

reconditesea:
For #1248.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -58,7 +58,7 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
 
   def serialize(k: K): Array[Byte] = serialization(k)
 
-  override def reducers = Some(numReducers)
+  def reducers = Some(numReducers)
 
   private lazy implicit val cms = CMS.monoid[Array[Byte]](eps, delta, seed)
   lazy val sketch: TypedPipe[CMS[Array[Byte]]] =
@@ -93,7 +93,7 @@ case class SketchJoined[K: Ordering, V, V2, R](left: Sketched[K, V],
   numReducers: Int)(joiner: (K, V, Iterable[V2]) => Iterator[R])
   extends MustHasReducers {
 
-  override def reducers = Some(numReducers)
+  def reducers = Some(numReducers)
 
   //the most of any one reducer we want to try to take up with a single key
   private val maxReducerFraction = 0.1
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/WithReducers.scala
```diff
@@ -24,6 +24,18 @@ trait HasReducers {
 }
 
 /**
+ * used for types that must know how many reducers they need
+ * e.g. Sketched
+ */
+trait MustHasReducers {
+  def reducers: Option[Int]
```
johnynek: Let's not do the runtime exception. Instead can we do this:

```scala
trait MustHaveReducers extends HasReducers {
  def reducers: Some[Int]
}
```
This should show that reducers is a Some[T] which is a subclass of Option[T]. That should allow us at compile time to verify the property you are checking at runtime below.

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -91,9 +91,9 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
 case class SketchJoined[K: Ordering, V, V2, R](left: Sketched[K, V],
   right: TypedPipe[(K, V2)],
   numReducers: Int)(joiner: (K, V, Iterable[V2]) => Iterator[R])
-  extends HasReducers {
+  extends MustHasReducers {
```
johnynek: this is both, right? HasReducers and MustHasReducers? Or MustHasReducers should extend HasReducers.

johnynek: I actually prefer the latter (MustHas extends Has).

reconditesea: I meant to make MustHas extends Has, but forgot to actually do it in the code.
Will update this.

scalding-core/src/main/scala/com/twitter/scalding/typed/WithReducers.scala
```diff
@@ -24,6 +24,14 @@ trait HasReducers {
 }
 
 /**
+ * used for types that must know how many reducers they need
+ * e.g. Sketched
+ */
+trait MustHasReducers {
```
azymnis: Sorry to be picky. Can we call this `MustHaveReducers` instead?

reconditesea: Of course!

scalding-core/src/main/scala/com/twitter/scalding/typed/WithReducers.scala
```diff
@@ -24,6 +24,14 @@ trait HasReducers {
 }
 
 /**
+ * used for types that must know how many reducers they need
+ * e.g. Sketched
+ */
+trait MustHaveReducers[+This <: MustHaveReducers[This]] extends HasReducers {
```
johnynek: why do we need the F-bounded polymorphism here? The trait never references `This`. I think this complicates things a bit. Can it be removed?

Issue #1055, opened 2014-09-18
Execution does not cache evaluations


johnynek:
This will result in complex dags recomputing results.

FlatMap chains won't be a problem, but zips may. This should be fixed by keeping a private cache through one entire .run call.

Issue #390, opened 2013-04-11
TempSource: dumping data into a location to persist between Jobs


johnynek:
In page rank, we run a job, write out a value, and before we start the next Job/Flow, we read the value and see if we should stop.

Currently, our implementation is ugly and requires the user to control that.  What would be better is something like:

``` scala
  val error = Temp() // can only write ONCE in the set of flows.
 // some place later in the same "family" of jobs:
  val result = error.read
```

So we need to keep a map of all the open temp files, which jobs write into them (we want write-once semantics) and then clean them up just before the Tool exits.

johnynek:
See the code in TypedPipe.forceToDiskExecution: https://github.com/twitter/scalding/blob/0.13.1/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala#L470

This should be refactored to make a temporary source. Ideally, we could delete these all when the entire execution is complete, something we don't do now.

PR #1253, opened 2015-04-13
Make sure Executions are executed at most once

johnynek:
There was previously an issue with have caching interacted with .zip. This cleans that up. It also guarantees that OnComplete has run by the time the final future is completed.

The approach started in #1242 has some issues. It turns out if you want to guarantee that no Execution is run twice (some cascading flows are going to fail if you try to run them twice, sadly), then the existence of flatMap means you really can't combine any Executions that wrap cascading flows because you can never be sure that a later flatMap is not hiding a reference to some Execution. That by itself would be okay, but when counters come in to play things break. If you read the counters after a single flow, they are different than reading them after that flow has been merged with another flow. So if a flow appears with a .zip (which seems like you should be able to combine) and without, and the counters are read, you have an inconsistency. Unfortunately, flatMap being a black box means we can never rule this out.

I'm not sure we even want such advanced optimizations. If some part of a flow is to be reused, use .forceToDiskExecution, and fan out from there. The reality is that without a macro or compiler plugin, optimizing Monads is very difficult (if we only have .zip, it is possible, interestingly, but that defeats the purpose of Execution). This does imply that static optimization of TypedPipe should be effective and not too difficult to implement reusing some of summingbird's dag optimization code.

related to #1055

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -45,7 +45,7 @@ import cascading.flow.{ FlowDef, Flow }
  * zip to flatMap if you want to run two Executions in parallel.
  */
 sealed trait Execution[+T] extends java.io.Serializable {
-  import Execution.{ EvalCache, FlatMapped, MapCounters, Mapped, OnComplete, RecoverWith, Zipped }
+  import Execution.{ EvalCache, FlatMapped, GetCounters, ResetCounters, Mapped, OnComplete, RecoverWith, Zipped }
 
   /**
    * Scala uses the filter method in for syntax for pattern matches that can fail.
@@ -83,7 +83,7 @@ sealed trait Execution[+T] extends java.io.Serializable {
    * You may want .getAndResetCounters.
    */
   def getCounters: Execution[(T, ExecutionCounters)] =
-    MapCounters[T, (T, ExecutionCounters)](this, { case tc @ (t, c) => (tc, c) })
+    GetCounters(this)
 
   /**
    * Reads the counters and resets them to zero. Probably what
@@ -122,7 +122,7 @@ sealed trait Execution[+T] extends java.io.Serializable {
    * you want to reset before a zip or a call to flatMap
    */
   def resetCounters: Execution[T] =
-    MapCounters[T, T](this, { case (t, c) => (t, ExecutionCounters.empty) })
+    ResetCounters(this)
 
   /**
    * This causes the Execution to occur. The result is not cached, so each call
@@ -249,8 +249,7 @@ object Execution {
         val promise = Promise[JobStats]()
         messageQueue.put(RunFlowDef(conf, mode, fd, promise))
         promise.future
-      }
-      catch {
+      } catch {
         case NonFatal(e) =>
           Future.failed(e)
       }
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -157,10 +164,9 @@ sealed trait Execution[+T] extends java.io.Serializable {
    * composition. Every time someone calls this, be very suspect. It is
    * always code smell. Very seldom should you need to wait on a future.
    */
-  def waitFor(conf: Config, mode: Mode): Try[T] = {
+  def waitFor(conf: Config, mode: Mode): Try[T] =
     Try(Await.result(run(conf, mode)(ConcurrentExecutionContext.global),
```
ianoc: Not an issue, but shame Await API doesn't cover returning a try instead of throwing and having to catch it. liftToTry on twitter futures is nice for that

johnynek: agree.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -457,29 +457,26 @@ trait TypedPipe[+T] extends Serializable {
...
+    .getConfigMode
+    .flatMap {
+      case (conf, mode) =>
+        mode match {
+          case _: CascadingLocal => // Local or Test mode
+            val dest = new MemorySink[T]
+            writeExecution(dest).map { _ => TypedPipe.from(dest.readResults) }
+          case _: HadoopMode =>
+            // come up with unique temporary filename, use the config here
+            // TODO: refactor into TemporarySequenceFile class
```
ianoc: should we open an issue for this?

ianoc:
LGTM, cleaner looking than the old too, so nice win there for readability. Merge when green

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: Looks like #390 is related.

johnynek:
@ianoc thank you for the review!

johnynek:
looks like finding the lo compression jars is failing the build. Restarted.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -196,132 +199,159 @@ object Execution {
...
+        def go(): Unit = messageQueue.take match {
+          case Stop =>
+            ()
+          case RunFlowDef(conf, mode, fd, promise) =>
+            promise.completeWith(
+              try {
+                ExecutionContext.newContext(conf)(fd, mode).run
+              } catch {
+                // Try our best to complete the future
+                case e: Throwable => Future.failed(e)
```
isnotinvain: you probably want NonFatal here, this will catch InterruptedException, ThreadDeathError and so on. Often this can make a progam un-cancellable via control-c, because you are squashing interrupted exceptions from a non-daemon thread.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -196,132 +199,159 @@ object Execution {
...
+     * from multiple threads. This thread does all the Flow starting.
+     */
+    private val thread = new Thread(new Runnable {
+      def run() {
+        @annotation.tailrec
+        def go(): Unit = messageQueue.take match {
+          case Stop =>
+            ()
+          case RunFlowDef(conf, mode, fd, promise) =>
+            promise.completeWith(
```
isnotinvain: if completeWith itself throws, this thread will die, which is why it's common to wrap all of run() in a try catch

johnynek: if completeWith throws, there is a serious logic error. That should never happen unless there is a fatal error. I didn't think I should catch that. What do you think? Should we try catch and exit if there is an exception?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: Okay. I got nervous about using NonFatal because I know it catches less than Twitter's version of that does, and if we don't complete that future, we are going to deadlock elsewhere, but in the unlikely event someone does control-c when we are at this stage, we want to exit. I'll change this.

PR #1254, opened 2015-04-13
Fix Config.getUniqueIDs

johnynek:
getUniqueIDs was broken. We were never using it though.


ianoc:
LGTM, merge when green

PR #1255, opened 2015-04-15
Make sure the EvalCache thread isDaemon

johnynek:
@isnotinvain pointed out we should probably never catch Throwable as it may have some really bad side effects.

Note, that after discussion, we kept the catch Throwable, but made a few other changes including making the thread a daemon thread.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -232,8 +232,7 @@ object Execution {
               try {
                 ExecutionContext.newContext(conf)(fd, mode).run
               } catch {
-                // Try our best to complete the future
-                case e: Throwable => Future.failed(e)
+                case NonFatal(e) => Future.failed(e)
               })
             // Loop
             go()
@@ -418,8 +417,7 @@ object Execution {
   /**
    * This creates a definitely failed Execution.
    */
-  def failed(t: Throwable): Execution[Nothing] =
-    fromFuture(_ => Future.failed(t))
+  def failed(t: Throwable): Execution[Nothing] = fromTry(Failure(t))
 
   /**
    * This makes a constant execution that runs no job.
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -225,15 +225,23 @@ object Execution {
       def run() {
         @annotation.tailrec
         def go(): Unit = messageQueue.take match {
-          case Stop =>
-            ()
+          case Stop => ()
           case RunFlowDef(conf, mode, fd, promise) =>
-            promise.completeWith(
-              try {
-                ExecutionContext.newContext(conf)(fd, mode).run
-              } catch {
-                case NonFatal(e) => Future.failed(e)
-              })
+            try {
+              promise.completeWith(
+                ExecutionContext.newContext(conf)(fd, mode).run)
+            } catch {
+              case t: Throwable =>
+                // something bad happened, but this thread is a daemon
+                // that should only stop if all others have stopped or
+                // we have received the stop message.
+                // Stopping this thread prematurely can deadlock
+                // futures from the promise we have.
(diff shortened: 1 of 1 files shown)


Gabriel439:
Can you clarify what the problems would be?

johnynek:
Well, if a thread is killed, it may throw a ThreadInterrupted exception. Catching that, would stop the interruption (seems Java really needs uncatchable exceptions for such cases, but that's a snap judgement). So, if you used a signal (say from control-c) to stop the program, it might be caught by such a catch-all, and not actually stop the program. NonFatal is supposed to encapsulate the correct logic.

isnotinvain:
+1 merge when green

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -232,8 +232,7 @@ object Execution {
               try {
                 ExecutionContext.newContext(conf)(fd, mode).run
               } catch {
-                // Try our best to complete the future
-                case e: Throwable => Future.failed(e)
+                case NonFatal(e) => Future.failed(e)
```
ianoc: I think we should be catching this and possibly re-throw after we satisfiy the future. Your inclination i think i right in that will manifest poorly otherwise with hung programs with no visible reason why.

isnotinvain:
@johnynek @Gabriel439 

So, the long story about catching Throwable is:
1) Errors extend Throwable, and you aren't supposed to catch them, eg VirtualMachineError / OutOfMemoryError -- you can't really recover from these

2) OK, so why not just catch Exception instead of Throwable? Well, InterruptedException is a special Exception, Java Concurrency in Practice has a lot to say about this, but it boils down to:
When a thread is interrupted (not killed -- interruption is a core aspect of thread communication for better or worse) it throws InterruptedException, and it flips a bit on the thread object. Most of the time, you want this exception to propagate, because this exception is a *request* for you to stop doing what you were doing in this thread. You can also query Thread.currentThread.interrupted() to find out if you are interrupted as well. If you do catch this exception, the interrupted bit gets flipped back, it's assumed that you are handling the interruption. So you need to either re-throw it and call Thread.currentThread.interrupt() to reset the bit, or you need to handle the interruption, which usually means, break out of whatever loop you were in or generally stop doing work.

3) OK so how about RuntimeException then? Well, many common exceptions, like IOException don't extend this :(

That's all java stuff, but don't forget about scala's NonLocalReturn exception and so on as well. So that's what `NonFatal` helps with.

Gabriel439:
What @isnotinvain said would make a really good comment to insert above that modified line

ianoc:
-1 no merge as is imo. We know catching only NonFatal will lead to hard to diagnose blocked programs. It happened to internal apps at twitter when thread pools for futures did this.

Something that is fatal to a thread must propagate that exception to the future.

isnotinvain:
@ianoc well yes, you need to handle InterruptedException if you're the one who needs to handle it. In the context of say a thread pool, often you want the thread pool to handle it. So in this case we might need to catch it, but not by accident but because we know what we're doing with it.

isnotinvain:
FuturePool only special cases non local return it looks like.
https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/util/FuturePool.scala#L113

Now I'm confused though, who is this Future being returned to? run() has no return value, it has to do a mutation somewhere that I'm missing

isnotinvain:
never mind, it's in the body of promise.completeWith

isnotinvain:
@ianoc InterruptedException is not fatal though I guess is what I was saying. You can choose to make it fatal, but if you do you need to terminate your loop (eg stop calling go())

If you receive an interrupted exception here you're being asked to stop looping

isnotinvain:
Blocked applications is often a sign of catching interrupted exception by accident, eg:
```
while (true) {
  try {
    Thread.sleep(100);
  } catch {
     e => log(e)
  }
}
```

If you control-c the above it will send InterruptedException, which will get squashed, and the loop will continue. If that's in a non-daemon thread it will block the app from shutting down.

isnotinvain:
LGTM

PR #1256, opened 2015-04-16
make InputSizeReducerEstimator work for any CompositeTap

rubanm:
Will probably need this internally for using custom CompositeTaps for multiformat sources.


johnynek:
shipit

PR #1257, opened 2015-04-16
TimePathedSource helper methods

rubanm:
Moving some re-usable logic to the companion object.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
@@ -31,6 +31,31 @@ object TimePathedSource {
       "%1$tm" -> Months(1)(tz), "%1$tY" -> Years(1)(tz))
       .find { unitDur: (String, Duration) => pattern.contains(unitDur._1) }
       .map(_._2)
+
+  def allPathsFor(pattern: String, duration: Option[Duration], dateRange: DateRange, tz: TimeZone): Iterable[String] =
+    duration.map { dur =>
+      // This method is exhaustive, but too expensive for Cascading's JobConf writing.
+      dateRange.each(dur)
+        .map { dr: DateRange =>
+          this.toPath(pattern, dr.start, tz)
+        }
+    }.getOrElse(Nil)
+
+  // picks all read paths in the given daterange
+  def readPathsFor(pattern: String, dateRange: DateRange, tz: TimeZone): Iterable[String] = {
+    val stepSize = TimePathedSource.stepSize(pattern, DateOps.UTC)
+    this.allPathsFor(pattern, stepSize, dateRange, DateOps.UTC)
+  }
+
+  // picks the write path based on daterange end
+  def writePathFor(pattern: String, dateRange: DateRange, tz: TimeZone): String = {
+    // TODO this should be required everywhere but works on read without it
+    // maybe in 0.9.0 be more strict
+    assert(pattern.takeRight(2) == "/*", "Pattern must end with /* " + pattern)
+    val lastSlashPos = pattern.lastIndexOf('/')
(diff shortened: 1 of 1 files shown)


isnotinvain:
+1

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
```diff
@@ -31,6 +31,38 @@ object TimePathedSource {
       "%1$tm" -> Months(1)(tz), "%1$tY" -> Years(1)(tz))
       .find { unitDur: (String, Duration) => pattern.contains(unitDur._1) }
       .map(_._2)
+
+  /**
+   * Gives all paths in the given daterange with windows based on the provided duration.
+   */
+  def allPathsWithDuration(pattern: String, duration: Option[Duration], dateRange: DateRange, tz: TimeZone): Iterable[String] =
```
johnynek: I don't like passing an Option just to call .getOrElse(Nil). I'd rather require the duration here and have callers do that.

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
```diff
@@ -31,6 +31,38 @@ object TimePathedSource {
...
+          this.toPath(pattern, dr.start, tz)
+        }
+    }.getOrElse(Nil)
+
+  /**
+   * Gives all read paths in the given daterange.
+   */
+  def readPathsFor(pattern: String, dateRange: DateRange, tz: TimeZone): Iterable[String] = {
+    val stepSize = TimePathedSource.stepSize(pattern, DateOps.UTC)
+    this.allPathsWithDuration(pattern, stepSize, dateRange, DateOps.UTC)
```
johnynek: `this` is weird on an object.

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
```diff
@@ -31,6 +31,38 @@ object TimePathedSource {
...
+  }
+
+  /**
+   * Gives the write path based on daterange end.
+   */
+  def writePathFor(pattern: String, dateRange: DateRange, tz: TimeZone): String = {
+    // TODO this should be required everywhere but works on read without it
+    // maybe in 0.9.0 be more strict
```
johnynek: we are well past 0.9.0 now. Need to update this comment.

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
```diff
@@ -31,6 +31,38 @@ object TimePathedSource {
...
+
+  /**
+   * Gives all paths in the given daterange with windows based on the provided duration.
+   */
+  def allPathsWithDuration(pattern: String, duration: Option[Duration], dateRange: DateRange, tz: TimeZone): Iterable[String] =
+    duration.map { dur =>
+      // This method is exhaustive, but too expensive for Cascading's JobConf writing.
+      dateRange.each(dur)
+        .map { dr: DateRange =>
+          this.toPath(pattern, dr.start, tz)
```
johnynek: `this` is not needed here, right?

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
rubanm: Any idea why this was required?

rubanm: Seems to me like this is actually needed to globify read paths but has to be removed for the write path?

rubanm:
@johnynek addressed your comments. Let me know if we should clarify anything for the "/*" logic. Thanks.

PR #1258, opened 2015-04-17
Remove a plugin we don't need, adds to build time. Bump to newer SBT tha...

ianoc:
...t has some resolution caching. Keep its cache dir and that of ivy

Ideally we'd keep the maven one too. But travis actually times out downloading the cache then sometimes. 

(The Travis cache implementation is effectively a remotely stored tar.gz from what I can tell, so the repo caching perf isn't wonderful since its unpacking lots of tiny files)

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -40,7 +40,7 @@ object ScaldingBuild extends Build {
   val scalameterVersion = "0.6"
   val scroogeVersion = "3.17.0"
   val slf4jVersion = "1.6.6"
-  val thriftVersion = "0.5.0"
+  val thriftVersion = "0.7.0"
 
   val printDependencyClasspath = taskKey[Unit]("Prints location of the dependencies")
 
@@ -52,6 +52,8 @@ object ScaldingBuild extends Build {
     crossScalaVersions := Seq("2.10.4", "2.11.5"),
 
     ScalariformKeys.preferences := formattingPreferences,
+
+    updateOptions := updateOptions.value.withCachedResolution(true),
 
     javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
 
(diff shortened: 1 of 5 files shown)


project/Build.scala
```diff
@@ -40,7 +40,7 @@ object ScaldingBuild extends Build {
   val scalameterVersion = "0.6"
   val scroogeVersion = "3.17.0"
   val slf4jVersion = "1.6.6"
-  val thriftVersion = "0.5.0"
+  val thriftVersion = "0.7.0"
```
johnynek: why the bump here?

ianoc: I was seeing failed builds due to 404's -- looks like the 0.5.0 version is no longer listed on maven central.. too old i guess

johnynek:
looks like travis flakyness:
```
[error] error while loading <root>, Error accessing /home/travis/.ivy2/cache/org.eclipse.jdt/core/jars/core-3.1.1.jar
```

project/Build.scala
johnynek: why can't we have an append only artifact cache? Sadness.

MansurAshraf:
so this version has option to NOT use Ivy altogether for dependency resolution but use Ather instead. Not sure if it will help with our issues but it maybe worthwhile to try it out. https://github.com/sbt/sbt/releases

```text
Maven resolver plugin

sbt 0.13.8 adds an extension point in the dependency resolution to customize Maven resolvers.
This allows us to write sbt-maven-resolver auto plugin, which internally uses Eclipse Aether
to resolve Maven dependencies instead of Apache Ivy.

To enable this plugin, add the following to project/maven.sbt (or project/plugin.sbt the file name doesn't matter):

addMavenResolverPlugin
This will create a new ~/.ivy2/maven-cache directory, which contains the Aether cache of files.
You may notice some file will be re-downloaded for the new cache layout.
Additionally, sbt will now be able to fully construct
maven-metadata.xml files when publishing to remote repositories or when publishing to the local ~/.m2/repository.
This should help erase many of the deficiencies encountered when using Maven and sbt together.

Notes and known limitations:

sbt-maven-resolver requires sbt 0.13.8 and above.
The current implementation does not support Ivy-style dynamic revisions, such as "2.10.+" or "latest.snapshot". This is a fixable situation, but the version range query and Ivy -> Maven version range translation code has not been migrated.
The current implementation does not support Maven-style range revisions if found on transitive dependencies. #1921
```

PR #1234, opened 2015-03-27
Change RichDate to value class (AnyVal)

oeddyo:
Change RichDate to value class (AnyVal)

https://github.com/twitter/scalding/issues/1120

PR #1261, opened 2015-04-17
Revert any val

MansurAshraf:
Reverts twitter/scalding#1234

AnyVal would be nice but due to existing java code paths, this is not an option.

### DIFF
--- a/scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala
+++ b/scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala
@@ -66,7 +66,7 @@ object RichDate {
 /**
  * A value class wrapper for milliseconds since the epoch
  */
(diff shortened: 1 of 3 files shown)


scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala
```diff
@@ -79,6 +80,14 @@ case class RichDate(val timestamp: Long) extends AnyVal with Ordered[RichDate] {
   override def compare(that: RichDate): Int =
     Ordering[Long].compare(timestamp, that.timestamp)
```
johnynek: where we are at it, java.lang.Long.compare will be faster here (no boxing).

johnynek:
looks good. Sorry @oeddyo this turned out to hit a lot of issues trying to merge internally. :/ Thanks for the work anyway.

PR #1263, opened 2015-04-24
Fix for reducer estimation not working correctly if withReducers is set to 1 reducer

sriramkrishnan:
Currently, the reducer estimation is broken if the number of reducers is set to 1. This PR uses a property in JobConf to disambiguate a withReducers(1) from the default (which is also 1). 

Added a test with a groupAll, which should always run a single reducer. And also verified that the existing reducer estimation tests still continue to work.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -325,6 +325,9 @@ object Config {
   /** Whether estimator should override manually-specified reducers. */
   val ReducerEstimatorOverride = "scalding.reducer.estimator.override"
 
+  /** Whether the number of reducers has been set explicitly using a `withReducers` */
+  val WithReducersSetExplicitly = "scalding.with.reducers.set.explicitly"
+
   val empty: Config = Config(Map.empty)
 
   /*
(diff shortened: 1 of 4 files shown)


johnynek:
Hell yeah! I love a manager that still sends PRs!

johnynek:
merge when green

By the way, this is perfect work: first there is a test that is red, then it is fixed. Boom!

sriramkrishnan:
Thanks @johnynek. But sadly it looks like I broke another test. So will update this with a fix.

sriramkrishnan:
Its a bad test I tell you! Thanks for @ianoc figured out that this actually does force a single reducer:
https://github.com/twitter/scalding/blob/develop/scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala#L41

So looks like I need to fix the test instead.

isnotinvain:
LGTM!

sriramkrishnan:
Added a bonus test to verify that the override actually works. Note that if the override is set, the current behavior is to estimate even if .withReducers() is set, for instance in a groupAll. I didn't change that behavior. 

Should we merge this, and file another ticket for that? CC @johnynek @isnotinvain @ianoc

johnynek:
Yes, let's merge and then get another ticket for the override/groupAll issue.

megaserg:
yay!

PR #1265, opened 2015-04-29
Fix SimpleDateFormat caching by default

johnynek:
SimpleDateFormat is not threadsafe, some tests are failing in a way that might be related here.

### DIFF
--- a/scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
+++ b/scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
@@ -51,10 +51,13 @@ object DateOps extends java.io.Serializable {
   /**
    * Return the guessed format for this datestring
    */
-  def getFormat(s: String): Option[String] = {
-    DATE_FORMAT_VALIDATORS.find{ _._2.findFirstIn(prepare(s)).isDefined }.map(_._1)
-  }
+  def getFormat(s: String): Option[String] =
+    DATE_FORMAT_VALIDATORS.find { _._2.findFirstIn(prepare(s)).isDefined }.map(_._1)
 
+  /**
+   * The DateParser returned here is based on SimpleDateFormat, which is not thread-safe.
+   * Do not share the result across threads.
+   */
   def getDateParser(s: String): Option[DateParser] =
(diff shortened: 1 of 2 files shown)


rubanm:
LGTM

isnotinvain:
We could also consider using joda-time, it's pretty similar except a) all of it's objects are thread safe and b) it doesn't let you make some of the common mistakes (like swapping YYYY and yyyy -- joda time renames the confusing one to xxxx). It's basically what java's built in time library should be.

Gabriel439:
Yeah, I'm really surprised that we don't use joda time.  I've never used it personally but I've heard nothing but good things about it.

johnynek:
The reason is to avoid a diamond dependency. You can easily make a
DateParser that uses joda time, but since you don't use user threads almost
ever in Hadoop, this is not compelling outside of testing.

On Thursday, April 30, 2015, Gabriel Gonzalez <notifications@github.com>
wrote:

> Yeah, I'm really surprised that we don't use joda time. I've never used it
> personally but I've heard nothing but good things about it.
>
> —
> <https://github.com/twitter/scalding/pull/1265#issuecomment-97867507>.
>


-- 

isnotinvain:
Is that still true with Execution in the mix?

On Thursday, April 30, 2015, P. Oscar Boykin <notifications@github.com>
wrote:

> The reason is to avoid a diamond dependency. You can easily make a
> DateParser that uses joda time, but since you don't use user threads almost
> ever in Hadoop, this is not compelling outside of testing.
>
> On Thursday, April 30, 2015, Gabriel Gonzalez <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > Yeah, I'm really surprised that we don't use joda time. I've never used
> it
> > personally but I've heard nothing but good things about it.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/twitter/scalding/pull/1265#issuecomment-97867507>.
> >
>
>
> --
> Oscar Boykin :: @posco :: http://twitter.com/posco
>
> —
> <https://github.com/twitter/scalding/pull/1265#issuecomment-97878484>.
>


-- 
Sent from my phone

Issue #1124, opened 2014-12-07
Export TypedJDBC Code from twitter internal (use macros)


johnynek:
Use macro to make a really easy TypedJDBC source. It should be as easy as giving a case class and the table name (maybe the table name could even derive from the underscore version of the class name).

johnynek:
@rubanm @ianoc this code is seeming pretty good to me, and we have some good internal customers. Is it time to release?

ianoc:
Possibly, @rubanm do we have anything left in our internal missing features we'd like to roll out first?

rubanm:
An option to fall back to parallel writes for large tables perhaps, but no one needs it immediately. Ok to start a merge without it I think.

PR #1267, opened 2015-05-01
Adds jdbc macros from internal

rubanm:
Part 1 of the new `scalding-db` and `scalding-db-macros` subprojects. Addresses https://github.com/twitter/scalding/issues/1124

This adds @ianoc's original JDBC macros along with some refactoring and fixes done during internal use at Twitter.

Part 2 to be followed will be the new JDBC source and sink classes based on these macros.

---

The motivation for this is to have an improved way of defining JDBC Typed Source and Sinks using **two simple steps**:

1) Defining a case class that represents you table schema:

``` scala
case class ExampleDBRecord(
  user_id: Long,
  tweet_id: Long,
  created_at: java.util.Date,
  deleted: Boolean)
```

2) Defining a TypedJDBCSource based on the above case class:

``` scala
case class ExampleDBRecordsTable(implicit dbsInEnv : AvailableDatabases)
    extends TypedJDBCSource[ExampleDBRecord](dbsInEnv) {
  override val tableName = TableName("example_table")
  override val database = Database("example_schema") 
}
```

Under the hood, this case class is automatically mapped to the underlying DB schema using a **macro-generated** `DBTypeDescriptor`. As a user, one does not need to specify SQL mappings or Injections to do this back and forth translation.

---

There are also optimizations for avoiding common pitfalls when talking to databases directly from hadoop nodes (too many open connections, inefficient OFFSET based queries, table lock contention). This includes performing a jdbc -> hdfs snapshot via the submitter for smaller datasets and likewise for writes. These will be in a separate PR along with the new jdbc source classes.


project/plugins.sbt
```diff
@@ -13,4 +13,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.10.2")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.5.1")
 
+addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.6")
```
ianoc: mima is already in this file

ianoc:
@rubanm can you add in the description some examples of using the new jdbc stuff to show motivation for this stuff

ianoc:
Looks great though, thanks for splitting this out

rubanm:
@ianoc updated the description

johnynek:
can we add a README or something that explains how to get going with this? It's great, but reading the code to learn it is pretty hard.

rubanm:
@johnynek makes sense. Added a README. It probably needs a few more edits though.

scalding-db-macros/README.md
```diff
@@ -0,0 +1,77 @@
...
+------------- | -------------
+`Int` | `INTEGER`
+`Long` | `BIGINT`
+`Short` | `SMALLINT`
+`Double` | `DOUBLE`
+`@varchar @size(20) String `| `VARCHAR(20)`
+`@text String` | `TEXT`
+`java.util.Date` | `DATETIME`
+`@date java.util.Date` | `DATE`
+`Boolean` | `BOOL`, `BOOLEAN`, `TINYINT`
```
johnynek: what is up with the three values on the right in this column?

scalding-db-macros/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+* Annotations are used for String types to clearly distinguish between TEXT and VARCHAR column types
+* Scala `Option`s can be used to denote columns that are `NULLABLE` in the DB
+* Nested case classes can be used as a workaround for the 22-size limitation on Scala tuples, case classes
```
johnynek: this seems to stop abruptly: case classes... ? what? I think you are going to explain that case classes are also flattened in left to right order. You might want to show an example of that.

scalding-db-macros/README.md
johnynek: Seems like the SQL type is BOOLEAN, from reading the code below.

scalding-db-macros/README.md
rubanm: Added a nested case classes example.

johnynek:
Are the tests being run? Did we update the travis running script?

We really need a more reliable way to make sure we are testing our sub-modules.

rubanm:
Updated the travis script to include the new db modules.

scalding-db-macros/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+Case class representing your DB schema:
+
+    scalding> case class ExampleDBRecord(
+       |   card_id: Long,
+       |   tweet_id: Long,
+       |   created_at: Option[java.util.Date],
+       |   deleted: Boolean = false)
+    defined class ExampleDBRecord
+
```
johnynek: what's the last step? How to I read or write from a Database? Seems like we are stopping short of clearly explaining the use of the whole package here.

rubanm: Yes, the sources will be added in a follow-up PR. So this is a little incomplete in that sense.

PR #1268, opened 2015-05-01
LzoGenericScheme and Source

rubanm:
This adds generic Scheme/Source classes for data stored as lzo-compressed protobuf blocks. Taken from @ianoc's old branch + added changes to support writes.

Depends on https://github.com/twitter/elephant-bird/pull/439 and https://github.com/twitter/elephant-bird/pull/440


scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/**
+ * Provides BinaryConverter serialized in JobConf.
+ */
+private[source] class ConfigBinaryConverterProvider[M] extends BinaryConverterProvider[M] {
+  import ConfigBinaryConverterProvider._
+  override def getConverter(conf: Configuration): BinaryConverter[M] = {
+    val data = conf.get(ProviderConfKey)
+    require(data != null, s"No data in field $ProviderConfKey")
```
rangadi: "$providerConfKey is not set in configuration"

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,118 @@
...
+    val extern = Externalizer(conv)
+    try {
+      ExternalizerSerializer.inj.invert(ExternalizerSerializer.inj(extern)).get
+    } catch {
+      case e: Exception => throw new RuntimeException("Unable to roundtrip the BinaryConverter in the Externalizer.", e)
+    }
+
+    conf.set(ConfigBinaryConverterProvider.ProviderConfKey, ExternalizerSerializer.inj(extern))
+
+    MultiInputFormat.setClassConf(conv.getClass, conf)
```
rangadi: conv.getClass would be "BinaryConverter", not the type parameter for ScroogeBinaryConverter, right? it does not matter except for stats, but just checking.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,118 @@
...
+    tap: Tap[JobConf, RecordReader[_, _], OutputCollector[_, _]],
+    conf: JobConf): Unit = {
+    val extern = Externalizer(conv)
+    try {
+      ExternalizerSerializer.inj.invert(ExternalizerSerializer.inj(extern)).get
+    } catch {
+      case e: Exception => throw new RuntimeException("Unable to roundtrip the BinaryConverter in the Externalizer.", e)
+    }
+
+    LzoGenericBlockOutputFormat.setClassConf(implicitly[ClassTag[M]].runtimeClass, conf)
```
rangadi: would 'implicitly[ClassTag[M]].runtimeClass' be the thrift class? Can we use that in sourceConfInit too? This will help with stats and logging (where we would see the actual thrift class)

rubanm: Fixed in the read path. Thanks.

rangadi:
+1

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/BinaryConverters.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def fromBytes(messageBuffer: Array[Byte]) = messageBuffer
+  override def toBytes(message: Array[Byte]) = message
+}
+
+object ScroogeBinaryConverter {
+  def apply[T <: ThriftStruct: ClassTag]: BinaryConverter[T] = {
+    val ct = implicitly[ClassTag[T]]
+    new BinaryConverter[T] {
+      val serializer = BinaryThriftStructSerializer[T] {
+        val companionClass = Class.forName(ct.runtimeClass.getName + "$")
```
johnynek: this is actually more complex if you are going to support Unions at the outer most layer:

https://github.com/twitter/chill/blob/develop/chill-scrooge/src/main/scala/com/twitter/chill/scrooge/ScroogeThriftStructSerializer.scala#L45

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoCodecSource.scala
```diff
@@ -30,6 +30,6 @@ object LzoCodecSource {
       val hdfsPaths = paths
       val localPath = { assert(paths.size == 1, "Cannot use multiple input files on local mode"); paths(0) }
       val boxed = Externalizer(passedInjection)
-      override def injection = boxed.get
+      override lazy val injection = boxed.get
```
johnynek: Externalizer is already doing the equivalent of lazy, isn't it? Do we need it twice?

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,118 @@
...
+  val inj: Injection[Externalizer[_], String] = {
+    import com.twitter.bijection.Inversion.attemptWhen
+    import com.twitter.bijection.codec.Base64
+
+    implicit val baseInj = JavaSerializationInjection[Externalizer[_]]
+
+    implicit val unwrap: Injection[GZippedBase64String, String] =
+      new AbstractInjection[GZippedBase64String, String] {
+        override def apply(gzbs: GZippedBase64String) = gzbs.str
+        override def invert(str: String) = attemptWhen(str)(Base64.isBase64)(GZippedBase64String(_))
```
johnynek: this is not super precise. It could be Base64, but not compressed, but it's fine, I think we'll catch it in the decompression Injection.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.hadoop.conf.Configuration
+
+import cascading.tap.Tap
+import cascading.flow.FlowProcess
+
+/**
+ * Serializes BinaryConverters to JobConf.
+ */
+private[source] object ExternalizerSerializer {
+  val inj: Injection[Externalizer[_], String] = {
```
johnynek: why not make this:

`def inj[T]: Injection[Externalizer[T], String] = { ...`

so we don't have to deal with ugly `_` issues. Should work fine, right?

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Provides BinaryConverter serialized in JobConf.
+ */
+private[source] class ConfigBinaryConverterProvider[M] extends BinaryConverterProvider[M] {
+  import ConfigBinaryConverterProvider._
+  override def getConverter(conf: Configuration): BinaryConverter[M] = {
+    val data = conf.get(ProviderConfKey)
+    require(data != null, s"$ProviderConfKey is not set in configuration")
+
+    val extern: Externalizer[_] = ExternalizerSerializer.inj.invert(data).get
```
johnynek: we could avoid the cast here if we put the type on inj above.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/BinaryConverters.scala
rubanm: could have re-used that code if the codec methods were public. What do you think? Or maybe add support for BinaryThriftStructSerializer there.

Not sure if we're planning a chill release. I can copy over the code for union here for now.

rubanm:
Merged develop. This is now green.

@johnynek addressed your comments. Could you take another look, thanks.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericSource.scala
```diff
@@ -0,0 +1,42 @@
...
+  override def setter[U <: T] = TupleSetter.asSubSetter[T, U](TupleSetter.singleSetter[T])
+  override def hdfsScheme = HadoopSchemeInstance((new LzoGenericScheme[T](conv)).asInstanceOf[Scheme[_, _, _, _, _]])
+}
+
+object LzoGenericSource {
+  def apply[T: ClassTag](passedConv: BinaryConverter[T], paths: String*) =
+    new LzoGenericSource[T] {
+      override val conv: BinaryConverter[T] = passedConv
+      override val hdfsPaths = paths
+      override val localPaths = { assert(paths.size == 1, "Cannot use multiple input files on local mode"); paths }
```
johnynek: I don't think this is needed anymore. This was fixed. These is no need for the assert, is there?

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Provides BinaryConverter serialized in JobConf.
+ */
+private[source] class ConfigBinaryConverterProvider[M] extends BinaryConverterProvider[M] {
+  import ConfigBinaryConverterProvider._
+  override def getConverter(conf: Configuration): BinaryConverter[M] = {
+    val data = conf.get(ProviderConfKey)
+    require(data != null, s"$ProviderConfKey is not set in configuration")
+
+    val extern = ExternalizerSerializer.inj.invert(data).get
```
johnynek: we often cache things like this if it happens a lot (and on hadoop, such things often happen a lot). Something like:

```scala
private[this] var last: Option[(String, BinaryConverter[M])] = None
last match {
  case Some((k, bc)) if k == data => bc
  case _ => // do the decode and reset last.
}
```

rubanm:
Merged in develop. This is now green again.

ianoc:
@johnynek this one look ok to you now?

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,128 @@
...
+private[source] class ConfigBinaryConverterProvider[M] extends BinaryConverterProvider[M] {
+  import ConfigBinaryConverterProvider._
+
+  private[this] var cached: Option[(String, BinaryConverter[M])] = None
+
+  override def getConverter(conf: Configuration): BinaryConverter[M] = {
+    val data = conf.get(ProviderConfKey)
+    require(data != null, s"$ProviderConfKey is not set in configuration")
+    cached match {
+      case Some((data, conv)) => conv
```
johnynek: I think we need `case Some((d, conv)) if d == data => conv` matching on data, I think, will just shadow the data on line 71.

I think we need this because without it, a different conf passed could give different data.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -0,0 +1,128 @@
...
+        conv
+    }
+  }
+}
+
+/**
+ * Generic scheme for data stored as lzo-compressed protobuf messages.
+ * Serialization is performed using the supplied BinaryConverter.
+ */
+class LzoGenericScheme[M: ClassTag](@transient conv: BinaryConverter[M]) extends LzoBinaryScheme[M, GenericWritable[M]] {
```
johnynek: is the ClassTag serializable? Have we tested that this actually round-trips a java serialization (which should happen when cascading uses this on Hadoop).

I wonder if instead of keeping the ClassTag here if it is safer to just take a Class, which is serializable, and make an apply method that uses ClassTag if needed.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericSource.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import com.twitter.elephantbird.mapreduce.io.BinaryConverter
+import com.twitter.scalding._
+
+import cascading.scheme.Scheme
+
+/**
+ * Generic source with an underlying GenericScheme that uses the supplied BinaryConverter.
+ */
+abstract class LzoGenericSource[T: ClassTag] extends FileSource with SingleMappable[T] with TypedSink[T] with LocalTapSource {
```
johnynek: what if we just add `def class: Class[T]` instead of wiring to ClassTag? What do we gain by using the ClassTag here, since concrete subclasses could chose either way if we go with `Class[T]`.

johnynek:
Can we at least add tests that we can instantiate and roundtrip this with the JavaSerialization Injection?

ianoc:
btw has been tested on hadoop, and ClassTag is serializable, though TypeTag is not. Though i'm fine with proposed changes too, just as an FYI.

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
rubanm: Oops, yes that's what I meant to do. Will fix.

johnynek:
merge when green

Issue #1139, opened 2014-12-18
Clean up the warnings or 2.11


johnynek:
We have a bunch of no-op transient annotations it looks like.

For instance:

```
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:297: no valid targets for annotation on value inputRfn - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient inputRfn: (X, X) => X,
[warn]      ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:298: no valid targets for annotation on value inputMrfn - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient inputMrfn: X => U,
[warn]      ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:396: no valid targets for annotation on value inputMrfn - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient inputMrfn: T => X,
[warn]      ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:397: no valid targets for annotation on value inputRfn - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient inputRfn: (X, X) => X,
[warn]      ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:431: no valid targets for annotation on value init - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient init: I,
[warn]      ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:432: no valid targets for annotation on value inputIterfn - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient inputIterfn: (I, Iterator[T]) => TraversableOnce[X],
[warn]      ^
[warn] /Users/oscar/workspace/scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala:450: no valid targets for annotation on value init - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]     @transient init: I,
[...]

PR #1271, opened 2015-05-02
Change separator from \1 to \u0001 

oeddyo:
To remove warning info for scala 2.11

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -267,8 +267,8 @@ trait Config {
    */
   def addReducerEstimator(clsName: String): Config =
     update(Config.ReducerEstimators) {
-      case None => Some(clsName) -> Unit
-      case Some(lst) => Some(clsName + "," + lst) -> Unit
+      case None => (Some(clsName), ())
+      case Some(lst) => (Some(s"$clsName,$lst"), ())
     }._2
 
   /** Set the entire list of reducer estimators (overriding the existing list) */
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -267,8 +267,8 @@ trait Config {
    */
   def addReducerEstimator(clsName: String): Config =
     update(Config.ReducerEstimators) {
-      case None => Some(clsName) -> ()
-      case Some(lst) => Some(clsName + "," + lst) -> ()
+      case None => Some(clsName) -> Unit
+      case Some(lst) => Some(clsName + "," + lst) -> Unit
     }._2
 
   /** Set the entire list of reducer estimators (overriding the existing list) */
--- a/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
@@ -39,9 +39,9 @@ object CumulativeSum {
     def cumulativeSum(
       implicit sg: Semigroup[V], ordU: Ordering[U], ordK: Ordering[K]): SortedGrouped[K, (U, V)] = {
       pipe.group
-        .sortBy { case (u: U, _) => u }
+        .sortBy { case (u, _) => u }
         .scanLeft(Nil: List[(U, V)]) {
-          case (acc, (u: U, v: V)) =>
+          case (acc, (u, v)) =>
             acc match {
               case List((previousU, previousSum)) => List((u, sg.plus(previousSum, v)))
               case _ => List((u, v))
@@ -63,7 +63,7 @@ object CumulativeSum {
       ordK: Ordering[K]): TypedPipe[(K, (U, V))] = {
 
       val sumPerS = pipe
-        .map { case (k, (u: U, v: V)) => (k, partition(u)) -> v }
+        .map { case (k, (u, v)) => (k, partition(u)) -> v }
         .sumByKey
         .map { case ((k, s), v) => (k, (s, v)) }
         .group
@@ -87,7 +87,7 @@ object CumulativeSum {
 
       val summands = pipe
         .map {
-          case (k, (u: U, v: V)) =>
+          case (k, (u, v)) =>
             (k, partition(u)) -> (Some(u), v)
         } ++ sumPerS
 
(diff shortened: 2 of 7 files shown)


oeddyo:
Seems there're some format issues. I was using 

./sbt ++2.11.5 clean test

How shall I automatically get a nicely formatted code?

scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
```diff
@@ -39,9 +39,9 @@ object CumulativeSum {
     def cumulativeSum(
       implicit sg: Semigroup[V], ordU: Ordering[U], ordK: Ordering[K]): SortedGrouped[K, (U, V)] = {
       pipe.group
-        .sortBy { case (u: U, _) => u }
+        .sortBy { case (u, _) => u }
```
isnotinvain: why drop the types?

johnynek: Those give warnings (I think) by the compiler since scala can't see that the type MUST be U, and U is also erased. Scala type inference does not work great with case matches.

jcoveney: scala can't verify the type, and I imagine it throws an error

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -267,8 +267,8 @@ trait Config {
    */
   def addReducerEstimator(clsName: String): Config =
     update(Config.ReducerEstimators) {
-      case None => Some(clsName) -> ()
-      case Some(lst) => Some(clsName + "," + lst) -> ()
+      case None => Some(clsName) -> Unit
```
johnynek: This does not seem like an error to me. `Unit` is the type or the companion object where as `()` is the value of type `Unit`. This code is different, although it does not matter because we discard the side value at the end. 

Why the change?

johnynek:
Looks good to me other than the `()` to `Unit` change. I want to know more about that.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
oeddyo: @johnynek I think I'm mixing two PR. One is to remove as many warnings as possible(https://github.com/twitter/scalding/issues/1139). For that purpose I'm seeing scala 2.11 gives warning when using () instead of Unit

[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/Config.scala:230: Adaptation of argument list by inserting () has been deprecated: this is unlikely to be what you want.
[warn]         signature: ArrowAssoc.->[B](y: B): (A, B)
[warn]   given arguments: <none>
[warn]  after adaptation: ArrowAssoc.->((): Unit)
[warn]       case Some(lst) => Some(clsName + "," + lst) -> ()


I googled around and found this:
http://stackoverflow.com/questions/25351988/adaptation-of-argument-list-by-inserting-has-been-deprecated

scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
oeddyo: @isnotinvain Dropping the types since there're warnings about type erasure here. 



[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:42: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]         .sortBy { case (u: U, _) => u }
[warn]                            ^
[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:44: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]           case (acc, (u: U, v: V)) =>
[warn]                          ^
[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:44: abstract type pattern V is unchecked since it is eliminated by erasure
[warn]           case (acc, (u: U, v: V)) =>
[warn]                                ^
[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:66: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]         .map { case (k, (u: U, v: V)) => (k, partition(u)) -> v }
[warn]                             ^
[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:66: abstract type pattern V is unchecked since it is eliminated by erasure
[warn]         .map { case (k, (u: U, v: V)) => (k, partition(u)) -> v }
[warn]                                   ^
[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:90: abstract type pattern U is unchecked since it is eliminated by erasure
[warn]           case (k, (u: U, v: V)) =>
[warn]                        ^
[warn] /Users/exie/learn/ScaldingGithub/scalding/scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala:90: abstract type pattern V is unchecked since it is eliminated by erasure
[warn]           case (k, (u: U, v: V)) =>

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
johnynek: Ahh, this is what we want here. So, let's change it to:

```scala
(Some(s"$clsName,$lst"), ())
(Some(clsName), ())
```
and just not use the `->` enrichment.

scalding-core/src/main/scala/com/twitter/scalding/CumulativeSum.scala
isnotinvain: can it tell enough to know the "shape" is `(k, (u, v))`?

This is one thing I'm often confused about with case matching in scala, how to make sure you aren't accidentally doing reflection in a case match in a way that will never even match.

isnotinvain:
LGTM other than the Unit change, it should remain `()`
Any idea why some of the tests that weren't changed go picked up in the diff?

oeddyo:
@isnotinvain Sure. I'll chnage it to (). Not really sure why the test missed this

isnotinvain:
@oeddyo well because in this case, any value at all would work but `Unit` makes a little less sense than `()` -- it's sort of like using `java.lang.Void.class` instead of `null`

oeddyo:
@isnotinvain Thanks for the explanation :)

oeddyo:
@johnynek @isnotinvain Fixed. Shall I merge?

johnynek:
Thanks, @oeddyo!

johnynek:
closes #1215

Issue #1135, opened 2014-12-17
FileSource could allow multiple localPaths


johnynek:
for some terrible reason, in FileSource hdfsPaths is Iterable, but localPath is a string. Both can be used with MultiSourceTap, so there is really no reason for this asymmetry.

Also, several places in the code throw exceptions if there is not one path to pass to localPath, so that is pretty ugly.

oeddyo:
Sorry for the mess. PR: https://github.com/twitter/scalding/pull/1275

PR #1272, opened 2015-05-03
Exie/1135

oeddyo:
address: https://github.com/twitter/scalding/issues/1135

Changed localPath: String to localPaths: Iterable[String] per Oscar's comment in the issue. 

I'm sure I'm doing stupid things here and there. Could someone kindly review my changes please?


oeddyo:
Strange. Why is Travis not building this one?

PR #1275, opened 2015-05-03
Exie/1135 try another branch

oeddyo:
address: #1135

Changed localPath: String to localPaths: Iterable[String] per Oscar's comment in the issue.

I'm sure I'm doing stupid things here and there. Could someone kindly review my changes please?

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -223,7 +223,7 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride {
         val files = localPaths.map{ p => new java.io.File(p) }
         if (strict && !files.forall(_.exists)) {
           throw new InvalidSourceException(
-            "[" + this.toString + s"] Data is missing from: $localPaths.filterNot{ p => new java.io.File(p).exists }")
+            "[" + this.toString + s"] Data is missing from: ${localPaths.filterNot { p => new java.io.File(p).exists }}")
         } else if (!files.exists(_.exists)) {
           throw new InvalidSourceException(
             "[" + this.toString + "] No good paths in: " + hdfsPaths.toString)
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -213,6 +213,13 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride {
             "[" + this.toString + "] No good paths in: " + hdfsPaths.toString)
         }
       }
+
+      case Local(_) => {
+        val file = new java.io.File(localPath)
+        if (!file.exists)
+          throw new InvalidSourceException(
+            "[" + this.toString + "] Nothing at path: " + localPath)
+      }
       case _ => ()
     }
   }
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -214,11 +214,20 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride {
         }
       }
 
-      case Local(_) => {
-        val file = new java.io.File(localPath)
-        if (!file.exists)
+      case Local(strict) => {
+        val files = localPaths.map{
+          p =>
+            new java.io.File(p)
+        }
+
+        if (strict && !files.forall(_.exists)) {
           throw new InvalidSourceException(
-            "[" + this.toString + "] Nothing at path: " + localPath)
+            "[" + this.toString + "] Data is missing from one or more paths in: " +
+              localPaths.toString)
+        } else if (!files.exists(_.exists)) {
+          throw new InvalidSourceException(
+            "[" + this.toString + "] No good paths in: " + hdfsPaths.toString)
+        }
       }
       case _ => ()
     }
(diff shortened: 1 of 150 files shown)


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -72,7 +81,13 @@ trait LocalSourceOverride extends SchemedSource {
    * @param sinkMode The mode for handling output conflicts.
    * @returns A tap.
    */
-  def createLocalTap(sinkMode: SinkMode): Tap[_, _, _] = new FileTap(localScheme, localPath, sinkMode)
+  def createLocalTap(sinkMode: SinkMode): Tap[JobConf, _, _] = {
+    val taps = localPaths.map {
+      p: String =>
+        CastFileTap(new FileTap(localScheme, p, sinkMode))
+    }.toList
```
johnynek: can we do:
```scala
taps match {
  case Nil => throw new InvalidSourceException("LocalPaths is empty")
  case oneTap :: Nil => oneTap
  case many => new ScaldingMultiSourceTap(many)
}
```
This way, we don't use the multisource path, which is new in this case, unless you need the new functionality, protecting users who don't make any change.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -196,11 +214,20 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride {
         }
       }
 
-      case Local(_) => {
-        val file = new java.io.File(localPath)
-        if (!file.exists)
+      case Local(strict) => {
+        val files = localPaths.map{
```
johnynek: let's use: `localPaths.map { p => new java.io.File(p) }` to be more consistent with scalding's style.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -313,12 +340,19 @@ trait SuccessFileSource extends FileSource {
  * Put another way, this runs a Hadoop tap outside of Hadoop in the Cascading local mode
  */
 trait LocalTapSource extends LocalSourceOverride {
-  override def createLocalTap(sinkMode: SinkMode) = new LocalTap(localPath, hdfsScheme, sinkMode).asInstanceOf[Tap[_, _, _]]
+  override def createLocalTap(sinkMode: SinkMode) = {
+    val taps = localPaths.map {
+      p =>
```
johnynek: put this on the previous line.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -313,12 +340,19 @@ trait SuccessFileSource extends FileSource {
...
+        new LocalTap(p, hdfsScheme, sinkMode).asInstanceOf[Tap[JobConf, RecordReader[_, _], OutputCollector[_, _]]]
+    }.toSeq
+    new ScaldingMultiSourceTap(taps)
```
johnynek: let's only use the multisource if there is more than one path.

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
```diff
@@ -131,7 +131,10 @@ abstract class TimePathedSource(val pattern: String,
   //Write to the path defined by the end time:
   override def hdfsWritePath = TimePathedSource.writePathFor(pattern, dateRange, tz)
 
-  override def localPath = pattern
+  override def localPaths = patterns
+    .flatMap{ pattern: String =>
```
johnynek: add space after `p` before `{`

scalding-core/src/main/scala/com/twitter/scalding/source/CodecSource.scala
```diff
@@ -56,9 +56,8 @@ class CodecSource[T] private (val hdfsPaths: Seq[String], val maxFailures: Int =
   lazy val field = new Fields(fieldSym.name)
   val injectionBox = Externalizer(injection andThen BytesWritableCodec.get)
 
-  def localPath = {
-    require(hdfsPaths.size == 1, "Local mode only supports a single path");
-    hdfsPaths(0)
+  def localPaths = {
```
johnynek: `def localPaths = hdfsPaths` is better here. no `{}`

johnynek:
Thanks for this cleanup.

oeddyo:
Something weird is happening in Travis build...

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -64,7 +64,16 @@ abstract class SchemedSource extends Source {
  */
 trait LocalSourceOverride extends SchemedSource {
   /** A path to use for the local tap. */
-  def localPath: String
+  def localPaths: Iterable[String]
+
+  // By default, we write to the last path for local paths
+  def localWritePath = localPaths.last
+
+  object CastFileTap {
```
johnynek: let's not put this as an inner object. Can we make it a `private[scalding]` object in FileSource.scala?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -196,11 +219,16 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride {
...
 
-      case Local(_) => {
-        val file = new java.io.File(localPath)
-        if (!file.exists)
+      case Local(strict) => {
+        val files = localPaths.map{ p => new java.io.File(p) }
+        if (strict && !files.forall(_.exists)) {
           throw new InvalidSourceException(
-            "[" + this.toString + "] Nothing at path: " + localPath)
+            "[" + this.toString + "] Data is missing from one or more paths in: " +
```
johnynek: why not add:

```scala
 Data is missing from: ${localPaths.filterNot { p => (new java.io.File(p)).exists}}
```
so the user can see exactly what is missing?

oeddyo:
Done. @johnynek

johnynek:
merge when green

PR #1276, opened 2015-05-05
Disable reducer estimation for map-only steps

rubanm:
Ran into this internally with a test job. With change https://github.com/twitter/scalding/pull/1263/, the estimator sets num reducers in the job conf for map-only steps, because setExplicitly is now always false for map-only steps.

```scala
if (!setExplicitly || overrideExplicit) {
  numReducers.foreach(conf.setNumReduceTasks)
}
```
The earlier version ``` val setExplicitly = flowNumReducers != stepNumReducers ``` somehow worked because the two had different defaults for map-only steps.

Mappers fail if num reducers are set when the step is actually map-only. I think this fixes that case. Added a test.

It'd be nice if there was a more robust way of looking at the FlowStep to determine if it's map-only. Anybody know of one?

Error on mappers:
```
Caused by: java.lang.NullPointerException
	at org.apache.hadoop.mapred.lib.HashPartitioner.getPartition(HashPartitioner.java:38)
	at org.apache.hadoop.mapred.MapTask$OldOutputCollector.collect(MapTask.java:587)
	at cascading.tap.hadoop.util.MeasuredOutputCollector.collect(MeasuredOutputCollector.java:69)
	at cascading.scheme.hadoop.TextDelimited.sink(TextDelimited.java:1061)
	at cascading.tuple.TupleEntrySchemeCollector.collect(TupleEntrySchemeCollector.java:153)
	... 56 more
```

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
@@ -97,7 +97,8 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
       conf.setInt(EstimatorConfig.estimatedNumReducers, numReducers.getOrElse(-1))
 
       // set number of reducers
-      if (!setExplicitly || overrideExplicit) {
+      // stepNumReducers == 0 implies no reduce phase in this step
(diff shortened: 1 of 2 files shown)


sriramkrishnan:
Thanks for this patch, and for adding the test. Seems reasonable not to want to estimate reducers for map-only jobs.

johnynek:
Good catch

PR #1278, opened 2015-05-05
Hydrate both sides of sampledCounts in skewJoinWithSmaller

### LINKED ISSUE #1277: skewJoinWithSmaller can drop entries
In preparation of the replication step, skewJoinWithSmaller creates sampledCounts using an outer join:

https://github.com/twitter/scalding/blob/f999d89511b9d01850ab031d1e537de5491c94e2/scalding-core/src/main/scala/com/twitter/scalding/JoinAlgorithms.scala#L394

For entries that are sampled on only the left or right, this pipe will have entries that look something like:
> left_key: "foo"
> LEFT_SAMPLED_COUNT: 5
> right_key: null
> RIGHT_SAMPLED_COUNT: null

In skewReplicate, each side joins against their own key:

https://github.com/twitter/scalding/blob/f999d89511b9d01850ab031d1e537de5491c94e2/scalding-core/src/main/scala/com/twitter/scalding/JoinAlgorithms.scala#L455

On the left we'll match for left_key == "foo" which correctly finds the counts of (5, null).  On the right we'll match for right_key == "foo" which won't find anything (yielding (null, null), etc.).  As a result, the left & right pipe will replicate differently, leading to data loss.

Of course, it affects skewJoinWithLarger too.

jnievelt:
Resolves #1277

The test fixes here are sufficient to make two tests fail, both of which are fixed with this change.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/JoinAlgorithms.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/JoinAlgorithms.scala
@@ -374,6 +374,7 @@ trait JoinAlgorithms {
         (otherPipe, fs._2, Fields.NONE)
       else // For now, we are assuming an inner join.
         renameCollidingFields(otherPipe, fs._2, intersection)
+    val mergedJoinKeys = Fields.join(fs._1, rightResolvedJoinFields)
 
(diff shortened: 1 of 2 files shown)


johnynek:
nice work tracking this down. It is so sad this has been causing subtle errors for years

Issue #1068, opened 2014-09-28
Add counters to see stdev of values per key


johnynek:
If on every reduce operation or join we counted the number of values for the key (say `N_v`), and then incremented a counter by `N_v^2`, then since map-reduce gives the number of groups, and the number of input and output records, we could compute the standard deviation of the (values per key). If we compare that to the mean of (values per key) we get some idea of the skew. I don't think we would overflow a long counter with this, in most cases, but it is possible (probably we can't directly compute the third standard moment, which is more directly skew), but I think the stdev might be enough in most cases.

We probably want the incoming skew and the outgoing skew. For standard sums/reduces/folds, the outgoing stdev should be 0 since each key has one value.

oeddyo:
Taking a look.

May I have a bit more hints here?

1) where is the "reduce operation or join" you mentioned located in the code? (I guess it's  scalding/scalding-core/src/main/scala/com/twitter/scalding/ReduceOperations.scala) but not 100% sure

2) why we want to add N_v^2? Is it for the convenience to do sqrt(E[x^2] - E[x]^2) and get std?

3) where (in the code) can I get the number of input and output records? is the number of output records equal to number of keys, and thus we could get the mean?

johnynek:
Hadoop already keeps the number of key groups and the total number of pairs, so this gives you values/key. What we need to get (as you point out) is the `\sum_k v_k^2` where `v_k` is the number of values for key `k`. With that, we can compute the `Var[V]` where `V` is the random variable of the number of values for a randomly selected key.

To get the number of values for a key, we need to look at https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala#L480 TypedBufferOp. Not on 496, we iterate through the iterator, so we could increment a counter at the end with the size squared.

For the case of joins, it is more complex (and probably should use a different stat key). That code is here: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/CoGrouped.scala#L306 - we might want a key for each of the join inputs, or maybe just the output. It's not clear. We can take it in two PRs.

oeddyo:
Thanks @johnynek  for the help. 


Currently I have the "proof of concept" PR. 

Basically what I'm doing is: 

Create a simple test, to do a reduce. The data I'm using is

TypedPipe.from(List((1, 2), (1, 3), (2, 3), (2, 5), (3, 1000)))

[...]

PR #1279, opened 2015-05-07
Adding second moment of values per key for Typed-API reduce operations

oeddyo:
proof of concept for https://github.com/twitter/scalding/issues/1068


oeddyo:
DO NOT MERGE YET.

oeddyo:
@johnynek Hi Oscar. Thanks for walking me through the code today!

I missed one problem to discuss with you, which I note in the code. 

so if I'm doing

      var numValuesPerKey = 0L

      val resIter = reduceFnSer.get(key, values)
      while (resIter.hasNext) {
        val tup = Tuple.size(1)
        val t2 = resIter.next

        numValuesPerKey += 1L

        tup.set(0, t2)
        oc.add(tup)
      }
      val valueCountSum = numValuesPerKey
      println("value count = " + numValuesPerKey)
 

For the test it would print

        value count = 1
        value count = 1
        value count = 1

Which should actually has a "value count = 2" for key 1. (please see test ReduceValueCounterTest for detail)

I have the test in 

branch: exie/1068
test-only com.twitter.scalding.ReduceValueCounterTest

should be easy to replicate. Just uncomment the block and comment the block under it. (in Operation.scala line 509-524)

oeddyo:
The reason should due to: 
      val resIter = reduceFnSer.get(key, caches.toIterator)
      while (resIter.hasNext) {
        val tup = Tuple.size(1)
        val t2 = resIter.next

        tup.set(0, t2)
        oc.add(tup)
      }
 
is trying to iterate the reduced result thus it's iterating through how many keys it has. Thus unfortunately we can't use a var to do a count to see how many values are associated with each key here.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -491,13 +497,23 @@ package com.twitter.scalding {
         .asScala
         .map(_.getObject(0).asInstanceOf[V])
 
+      val caches = values.toList
```
johnynek: so, we can't do this (because it would force everything to memory), but what about something like this:
```scala
class CountingIterator[T](wraps: Iterator[T]) extends Iterator[T] {
  private[this] var nextCalls = 0
  def hasNext = wraps.hasNext
  def next = { nextCalls += 1; wraps.next }
  def seen: Int = nextCalls
}
```
then we could wrap `values` with this, then call `.seen` at the end to see how many values went in?

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -491,13 +497,23 @@ package com.twitter.scalding {
...
       while (resIter.hasNext) {
         val tup = Tuple.size(1)
-        tup.set(0, resIter.next)
+        val t2 = resIter.next
+
+        tup.set(0, t2)
         oc.add(tup)
       }
+      
+      flowProcess.increment(SkewMonitorCounters.KeyCount, SkewMonitorCounters.KeyCount, 1L)
```
johnynek: probably better to use a group like "scalding.debug" since I think we want to group them up so they are easy to see next to each other in the job tracker UI.

scalding-core/src/test/scala/com/twitter/scalding/CoreTest.scala
```diff
@@ -1817,6 +1817,46 @@ class CounterJobTest extends WordSpec with Matchers {
...
+          x should be(3)
+      }
+      .counter(SkewMonitorCounters.ValuesCountSquareSum, SkewMonitorCounters.ValuesCountSquareSum) {
+        x =>
+          // key 1 has two values, thus 2^2 = 4. key 2 and 3 has only one respectively
+          x should be(4 + 1 + 1)
+      }
+      .counter(SkewMonitorCounters.ValuesCountSum, SkewMonitorCounters.ValuesCountSum) {
+        x =>
+          // key 1 has two values, thus 2^2 = 4. key 2 and 3 has only one respectively
```
johnynek: looks like a copied comment, don't want `2^2` right?

oeddyo: Ah you're right. it's a copied commend

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
oeddyo: good point

oeddyo:
@johnynek Does this look good?

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -487,17 +501,25 @@ package com.twitter.scalding {
...
         val tup = Tuple.size(1)
-        tup.set(0, resIter.next)
+        val t2 = resIter.next
+        tup.set(0, t2)
         oc.add(tup)
       }
+
+      val numValuesPerKey = values.seen
+
+      flowProcess.increment(SkewMonitorCounters.KeyCount, SkewMonitorCounters.KeyCount, 1L)
```
johnynek: I don

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -476,7 +476,21 @@ package com.twitter.scalding {
     }
   }
 
-  /** In the typed API every reduce operation is handled by this Buffer */
+  private[scalding] object SkewMonitorCounters {
+    // Strangely, if group name and key name are different, then the counter would be zero
```
johnynek: this is bug by itself. Can you confirm this is still true and not due to another issue? We don't want to make a bunch of groups.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -487,17 +501,25 @@ package com.twitter.scalding {
...
         .asScala
-        .map(_.getObject(0).asInstanceOf[V])
+        .map(_.getObject(0).asInstanceOf[V]))
 
       // Avoiding a lambda here
       val resIter = reduceFnSer.get(key, values)
       while (resIter.hasNext) {
         val tup = Tuple.size(1)
-        tup.set(0, resIter.next)
+        val t2 = resIter.next
```
johnynek: can we remove this change?

scalding-core/src/test/scala/com/twitter/scalding/ReduceOperationsTest.scala
```diff
@@ -163,4 +163,33 @@ class ReduceOperationsTest extends WordSpec with Matchers {
...
+        .write(TypedTsv[(Int, Int, Int)](args("output")))
+    }
+
+    class ExperimentTest extends WordSpec with Matchers {
+      "A PageRank2 job" should {
+        JobTest(new com.twitter.scalding.ReduceValueCountJob(_))
+          .arg("output", "blah")
+          .sink[(Int, Int, Int)](TypedTsv[(Int, Int, Int)]("blah")){
+            tuples =>
+              println("RES = " + tuples)
```
johnynek: let's remove println statements.

scalding-core/src/test/scala/com/twitter/scalding/ReduceOperationsTest.scala
```diff
@@ -163,4 +163,33 @@ class ReduceOperationsTest extends WordSpec with Matchers {
...
       .finish
   }
+
+  "Std for number of values aassocated with each key " should {
+    class SDForReduceOperation(args: Args) extends Job(args) {
+      TypedPipe.from(List((1, (1, 1)), (2, (2, 2)), (1, (3, 3)), (3, (3, 3))))
+        .group
+        .foldLeft((0, 0)){ (a, b) => (a._1 + b._1, a._2 + b._2) }
+        .toTypedPipe
+        .map{
```
johnynek: `.map {` note the space.

scalding-core/src/test/scala/com/twitter/scalding/ReduceOperationsTest.scala
```diff
@@ -163,4 +163,33 @@ class ReduceOperationsTest extends WordSpec with Matchers {
       .runHadoop
       .finish
   }
+
+  "Std for number of values aassocated with each key " should {
+    class SDForReduceOperation(args: Args) extends Job(args) {
+      TypedPipe.from(List((1, (1, 1)), (2, (2, 2)), (1, (3, 3)), (3, (3, 3))))
+        .group
+        .foldLeft((0, 0)){ (a, b) => (a._1 + b._1, a._2 + b._2) }
```
johnynek: you could do `.forceToReducers.sum` and it should get the same effect.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
oeddyo: This seems to be a bug. if you simply put 

flowProcess.increment("TestGroup", "TestKey", 1)

in     def operate(flowProcess: FlowProcess[_], call: BufferCall[Any]) {

and this should print 3 in the test

ReduceValueCounterTest. However it prints

PRINTING KEY AND GROUP! 0

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
oeddyo: Fixed.

scalding-core/src/test/scala/com/twitter/scalding/ReduceOperationsTest.scala
oeddyo: Fixed. Dup-test. Thanks for the catch

scalding-core/src/test/scala/com/twitter/scalding/ReduceOperationsTest.scala
oeddyo: Fixed.

scalding-core/src/test/scala/com/twitter/scalding/ReduceOperationsTest.scala
oeddyo: Good point. FIxed.

oeddyo:
After testing it for a couple more times, I confirm it's a bug. Here's how you could re-produce it:

uncomment line 523 in scalding/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala

Then in sbt do

test-only com.twitter.scalding.ReduceValueCounterTest

It print out a line (corresponding to the code in CoreTest.scala line 1837)
PRINTING KEY AND GROUP! 0

But if you use same group name as key name, then it gives
PRINTING KEY AND GROUP! 3

PR #1280, opened 2015-05-07
faster builds on travis by a bit.

ianoc:
Bumps to 2.10.5 for a the builds

-> Attempt to cache the ivy cache and the local cache data used by sbt.

Avoids the pathologically bad builds. Also i collapsed some of the builds, our parallelism is limited to 5 builds on travis, so having too many was hurting us for overall runtime of the build as we rebuild scalding-date/args/core several times in each one. 

(Also basically put a tar ball in the travis cache and unpack it ourselves while constantly producing output to avoid the previous bad case of timing out the build from travis unpacking it.)


.travis.yml
```diff
@@ -15,48 +22,40 @@ matrix:
...
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.5
-      env: BUILD="base" TEST_TARGET="scalding-core"
-      script: "scripts/run_test.sh"
-
-    - scala: 2.10.4
-      env: BUILD="base" TEST_TARGET="scalding-hadoop-test"
+    - scala: 2.10.5
+      env: BUILD="base" TEST_TARGET="scalding-macros"
```
johnynek: why is this on its own in 2.10, but not 2.11?

.travis.yml
```diff
@@ -2,12 +2,19 @@ language: scala
...
+   - $HOME/.scalding.dep.cache
 script:
   - "echo no op"
+addons:
+  apt:
+    packages:
+    - md5deep
 matrix:
   include:
 #BASE TESTS
```
johnynek: I'm nervous we need a test to be sure we are running all the tests. It would be really easy to leave a target out. Any ideas on how we can be sure we have everything other than just manually checking that all the targets appear somewhere?

ianoc: I think I can hack that in with a bash script that will validate this file refers to each target twice maybe -- other than some blacklists. Its not pretty but I think it should work? Thoughts?

.travis.yml
ianoc: In consistency, will move about

.travis.yml
johnynek: I think that's fine. We can also run that bash script as part of the tests, so the tests use the .travis.yml and the Build.scala as inputs before we run any of the other tests it makes sure we are going to run all the tests.

johnynek:
did this work? Should we merge?

.travis.blacklist
```diff
@@ -0,0 +1,5 @@
+#This describes extra builds our validator will pretend to run in CI but won't
+# Remember we run most builds twice, so if you want it disabled for both 2.10 and 2.11 it needs to be here twice
+scalding-repl
```
johnynek: is this 2.10 or 2.11? How does it know?

johnynek:
How does this work? Can you add some comments?

I'm not very strong with bash, sorry.

.travis.yml
```diff
@@ -39,83 +46,38 @@ matrix:
...
-    - scala: 2.10.4
-      env: BUILD="test tutorials"
-      script:
-      - "scripts/build_assembly_no_test.sh scalding"
-      - "scripts/test_tutorials.sh"
-
-    - scala: 2.11.5
-      env: BUILD="test tutorials"
+    - scala: 2.10.5
+      env: BUILD="test tutorials and matrix tutorials and repl" TEST_TARGET="scalding-repl"
```
johnynek: what does this syntax mean? "test tutorials *and* .. *and*"?

What is consuming the BUILD variable?

.travis.yml
ianoc: Yeah its wired in as its own test target, we sort of had a free simple target we didn't use. so its in there.

.travis.blacklist
ianoc: It doesn't alas, simple hack where it just says we only want one of them..

project/Build.scala
```diff
@@ -27,8 +27,8 @@ object ScaldingBuild extends Build {
   val chillVersion = "0.6.0"
   val dfsDatastoresVersion = "1.3.4"
   val elephantbirdVersion = "4.8"
-  val hadoopLzoVersion = "0.4.16"
-  val hadoopVersion = "1.2.1"
+  val hadoopLzoVersion = "0.4.19"
+  val hadoopVersion = "2.5.0"
```
johnynek: why are these changing at the same time?

scripts/testValidator.sh
```diff
@@ -0,0 +1,44 @@
...
+# Cat the travis build file, ignoring the assembly lines
+# Reformatting any quotes to new lines so we get things nicer split up
+# Then grep this for the scalding- and maple from above
+BUILDS_WE_HAVE=$(cat .travis.yml | grep -v scripts/build_assembly_no_test  | tr '"' '\n' | tr ' ' '\n' | grep -e scalding- -e maple )
+
+# Grab the blacklist, lines starting with #'s
+BLACKLIST_BUILDS=$(cat .travis.blacklist | egrep -v '^\s*#')
+
+
+TEST_ID=$(date '+%s')
```
johnynek: should we use mktemp or tempfile here?

ianoc: Unfortunately those are pretty crappy across OSes, different on osx vs linux vs etc.. :/

johnynek:
So, what's the main approach here? Can you add a doc so someone else can manage this?

Is the caching the main deal? How does the caching work? Is it the dependency graph caching or the downloads of all the jars that is the win?

ianoc:
Added a readme. Its really hard to know what all the wins from everything are, huge variance, depends on lots of external factors. But both the jars and the pom's seem to improve things when cached. But who knows if that'll remain the case over time. (Pom's seem like probably worth caching almost always, or the ivy equiv's, but jars maybe not so if you have a local maven mirror doing the caching for you separately.)

project/Build.scala
ianoc: I rolled in another PR people had green since it effected testing time hugely, so two birds one stone. People had reviewed independently and thought ok.

johnynek:
Looks good. Thanks for doing this Ian. Looks like it could be a big win.

PR #1282, opened 2015-05-08
[Dont Merge/Code Share] Added tests for case class that extends Ordered

MansurAshraf:
for macro testing..


ianoc:
Does this fail?

MansurAshraf:
unfortunately No :-(

PR #1283, opened 2015-05-09
Add `withDescription` for naming MR steps

aposwolsky:
Adding ability to have more meaningful names in MR steps instead of just "(X/Y")

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -40,7 +40,7 @@ trait ExecutionContext {
     if (descriptions.nonEmpty) Some(descriptions.distinct.mkString(", ")) else None
   }
 
-  private def updateStepConfigWithDescriptions(step: BaseFlowStep[JobConf], descriptions: Seq[String]): Unit = {
+  private def updateStepConfigWithDescriptions(step: BaseFlowStep[JobConf]): Unit = {
     val conf = step.getConfig
     getIdentifierOpt(getDesc(step)).foreach(descriptionString => {
       conf.set(Config.StepDescriptions, descriptionString)
@@ -80,8 +80,7 @@ trait ExecutionContext {
           val flowSteps = hadoopFlow.getFlowSteps.asScala
           flowSteps.foreach(step => {
             val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
-            val descriptions = getDesc(baseFlowStep)
-            updateStepConfigWithDescriptions(baseFlowStep, descriptions)
+            updateStepConfigWithDescriptions(baseFlowStep)
           })
         case _ => // descriptions not yet supported in other modes
       }
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -40,12 +40,10 @@ trait ExecutionContext {
     if (descriptions.nonEmpty) Some(descriptions.distinct.mkString(", ")) else None
   }
 
-  private def updateJobName(step: BaseFlowStep[JobConf], descriptions: Seq[String]): Unit = {
+  private def updateStepConfigWithDescriptions(step: BaseFlowStep[JobConf], descriptions: Seq[String]): Unit = {
     val conf = step.getConfig
-    getIdentifierOpt(getDesc(step)).foreach(id => {
-      val newJobName = "%s %s".format(conf.getJobName, id)
-      println("Added descriptions to job name: %s".format(newJobName))
-      conf.setJobName(newJobName)
+    getIdentifierOpt(getDesc(step)).foreach(descriptionString => {
+      conf.set(Config.StepDescriptions, descriptionString)
     })
   }
 
@@ -96,7 +94,7 @@ trait ExecutionContext {
             val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
             val descriptions = getDesc(baseFlowStep)
             updateFlowStepName(baseFlowStep, descriptions)
-            updateJobName(baseFlowStep, descriptions)
+            updateStepConfigWithDescriptions(baseFlowStep, descriptions)
           })
         case _ => // descriptions not yet supported in other modes
       }
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -47,18 +47,6 @@ trait ExecutionContext {
     })
   }
 
-  private def updateFlowStepName(step: BaseFlowStep[JobConf], descriptions: Seq[String]): BaseFlowStep[JobConf] = {
-    getIdentifierOpt(descriptions).foreach(newIdentifier => {
-      val stepXofYData = """\(\d+/\d+\)""".r.findFirstIn(step.getName).getOrElse("")
-      // Reflection is only temporary.  Latest cascading has setName public: https://github.com/cwensel/cascading/commit/487a6e9ef#diff-0feab84bc8832b2a39312dbd208e3e69L175
-      // https://github.com/twitter/scalding/issues/1294
-      val x = classOf[BaseFlowStep[JobConf]].getDeclaredMethod("setName", classOf[String])
-      x.setAccessible(true)
-      x.invoke(step, "%s %s".format(stepXofYData, newIdentifier))
-    })
-    step
-  }
-
   private def getDesc(baseFlowStep: BaseFlowStep[JobConf]): Seq[String] = {
     baseFlowStep.getGraph.vertexSet.asScala.toSeq.flatMap(_ match {
       case pipe: Pipe => RichPipe.getPipeDescriptions(pipe)
@@ -93,7 +81,6 @@ trait ExecutionContext {
           flowSteps.foreach(step => {
             val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
             val descriptions = getDesc(baseFlowStep)
-            updateFlowStepName(baseFlowStep, descriptions)
             updateStepConfigWithDescriptions(baseFlowStep, descriptions)
           })
         case _ => // descriptions not yet supported in other modes
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -17,11 +17,10 @@ package com.twitter.scalding
 
 import cascading.flow.hadoop.HadoopFlow
 import cascading.flow.planner.BaseFlowStep
-import cascading.flow.{ FlowStep, FlowStepStrategy, FlowDef, Flow }
+import cascading.flow.{ FlowDef, Flow }
 import cascading.pipe.Pipe
 import com.twitter.scalding.reducer_estimation.ReducerEstimatorStepStrategy
 import com.twitter.scalding.serialization.CascadingBinaryComparator
-import java.util.{ List => JList }
 import org.apache.hadoop.mapred.JobConf
 import scala.collection.JavaConverters._
 import scala.concurrent.Future
@@ -41,6 +40,15 @@ trait ExecutionContext {
     if (descriptions.nonEmpty) Some(descriptions.distinct.mkString(", ")) else None
   }
 
+  private def updateJobName(step: BaseFlowStep[JobConf], descriptions: Seq[String]): Unit = {
+    val conf = step.getConfig
+    getIdentifierOpt(getDesc(step)).foreach(id => {
+      val newJobName = "%s %s".format(conf.getJobName, id)
+      println("Added descriptions to job name: %s".format(newJobName))
+      conf.setJobName(newJobName)
+    })
+  }
+
   private def updateFlowStepName(step: BaseFlowStep[JobConf], descriptions: Seq[String]): BaseFlowStep[JobConf] = {
     getIdentifierOpt(descriptions).foreach(newIdentifier => {
       val stepXofYData = """\(\d+/\d+\)""".r.findFirstIn(step.getName).getOrElse("")
@@ -58,22 +66,6 @@ trait ExecutionContext {
       case pipe: Pipe => RichPipe.getPipeDescriptions(pipe)
       case _ => List() // no descriptions
     })
-  }
-
-  object WithDescriptionStepStrategy extends FlowStepStrategy[JobConf] {
-    override def apply(
-      flow: Flow[JobConf],
-      preds: JList[FlowStep[JobConf]],
-      step: FlowStep[JobConf]): Unit = {
-
-      val conf = step.getConfig
-      val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
-      getIdentifierOpt(getDesc(baseFlowStep)).foreach(id => {
-        val newJobName = "%s %s".format(conf.getJobName, id)
-        println("Added descriptions to job name: %s".format(newJobName))
-        conf.setJobName(newJobName)
-      })
-    }
   }
 
   final def buildFlow: Try[Flow[_]] =
@@ -102,7 +94,9 @@ trait ExecutionContext {
           val flowSteps = hadoopFlow.getFlowSteps.asScala
           flowSteps.foreach(step => {
             val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
-            updateFlowStepName(baseFlowStep, getDesc(baseFlowStep))
+            val descriptions = getDesc(baseFlowStep)
+            updateFlowStepName(baseFlowStep, descriptions)
+            updateJobName(baseFlowStep, descriptions)
           })
         case _ => // descriptions not yet supported in other modes
       }
@@ -111,11 +105,8 @@ trait ExecutionContext {
       // which instantiates and runs them
       mode match {
         case _: HadoopMode =>
-          if (config.get(Config.ReducerEstimators).isDefined)
-            flow.setFlowStepStrategy(FlowStepStrategies.plus(WithDescriptionStepStrategy, ReducerEstimatorStepStrategy))
-          else
-            flow.setFlowStepStrategy(WithDescriptionStepStrategy)
-
+          config.get(Config.ReducerEstimators)
+            .foreach(_ => flow.setFlowStepStrategy(ReducerEstimatorStepStrategy))
         case _ => ()
       }
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -17,12 +17,13 @@ package com.twitter.scalding
 
 import cascading.flow.hadoop.HadoopFlow
 import cascading.flow.planner.BaseFlowStep
-import cascading.flow.{ FlowDef, Flow }
+import cascading.flow.{ FlowStep, FlowStepStrategy, FlowDef, Flow }
 import cascading.pipe.Pipe
 import com.twitter.scalding.reducer_estimation.ReducerEstimatorStepStrategy
 import com.twitter.scalding.serialization.CascadingBinaryComparator
+import java.util.{ List => JList }
 import org.apache.hadoop.mapred.JobConf
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.util.{ Failure, Success, Try }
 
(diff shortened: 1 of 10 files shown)


johnynek:
This is pretty great. Thanks for doing this. I'll review it carefully this week.

aposwolsky:
will change.  any other issues? @johnynek ?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -31,6 +36,25 @@ trait ExecutionContext {
   def flowDef: FlowDef
   def mode: Mode
 
+  def getIdentifierOpt(descriptions: Seq[String]): Option[String] = {
+    if (descriptions.nonEmpty) {
+      Some(descriptions.distinct.mkString(", "))
```
johnynek: can we sort these to make sure it is stable?

johnynek: let's make this a private method on `object ExecutionContext`.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -31,6 +36,25 @@ trait ExecutionContext {
...
+    } else {
+      None
+    }
+  }
+
+  def updateFlowStepName(step: BaseFlowStep[JobConf], descriptions: Seq[String]): BaseFlowStep[JobConf] = {
+    getIdentifierOpt(descriptions).foreach(newIdentifier => {
+      val stepXofYData = """\(\d+/\d+\)""".r.findFirstIn(step.getName).getOrElse("")
+      // Reflection is only temporary.  Latest cascading has setName public: https://github.com/cwensel/cascading/commit/487a6e9ef#diff-0feab84bc8832b2a39312dbd208e3e69L175
```
johnynek: add an issue to the issue tracker so we don't lose track of this, and link the issue here.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -52,6 +76,25 @@ trait ExecutionContext {
...
 
+      if (flow.isInstanceOf[HadoopFlow]) {
+        val flowSteps = flow.asInstanceOf[HadoopFlow].getFlowSteps.toList
+        flowSteps.foreach(step => {
+          val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
+
+          val descriptions: Seq[String] = {
+            for {
+              vertex <- baseFlowStep.getGraph.vertexSet.toList
+              if (vertex.isInstanceOf[Pipe])
```
johnynek:
```scala
def getDesc(v: Any): Seq[String] = v match {
  case vPipe: Pipe => RichPipe.getPipeDescriptions(vPipe)
  case _ => List() // no descriptions
}
```
then call that rather than the test + cast (do it in one go and no explicit cast).

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -52,6 +76,25 @@ trait ExecutionContext {
       }
       val flow = mode.newFlowConnector(withId).connect(flowDef)
 
+      if (flow.isInstanceOf[HadoopFlow]) {
```
johnynek:
```scala
flow match {
  case hflow: HadoopFlow =>
    // logic here
  case _ =>
    // descriptions not yet supported in other modes
}
```

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -31,6 +36,25 @@ trait ExecutionContext {
   def flowDef: FlowDef
   def mode: Mode
 
+  def getIdentifierOpt(descriptions: Seq[String]): Option[String] = {
+    if (descriptions.nonEmpty) {
+      Some(descriptions.distinct.mkString(", "))
+    } else {
+      None
+    }
+  }
```
johnynek: nit: our style is to only use `{ }` when needed (more than one expression). Many of your `{ }` are redundant and we'd prefer if you not add them.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -52,6 +53,38 @@ object RichPipe extends java.io.Serializable {
     }
     p
   }
+
+  // A pipe can have more than one description when merged together, so we store them delimited witg 254.toChar.
```
johnynek: typo: with

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -52,6 +53,38 @@ object RichPipe extends java.io.Serializable {
...
+  def encodePipeDescriptions(descriptions: Seq[String]): String = {
```
johnynek: make private as much as possible to make it easier to preserve binary compatibility.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -423,11 +440,15 @@ case class ValueSortedReduce[K, V1, V2](
   override def withReducers(red: Int) =
     // copy infers loose types. :(
     ValueSortedReduce[K, V1, V2](
-      keyOrdering, mapped, valueSort, reduceFn, Some(red))
+      keyOrdering, mapped, valueSort, reduceFn, Some(red), descriptions)
+
+  override def withDescription(description: String) =
+    ValueSortedReduce[K, V1, V2](
+      keyOrdering, mapped, valueSort, reduceFn, reducers, descriptions ++ Seq(description))
```
johnynek: `descriptions :+ description` would be better.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -469,6 +492,9 @@ case class IteratorMappedReduce[K, V1, V2](
   override def withReducers(red: Int): IteratorMappedReduce[K, V1, V2] =
     copy(reducers = Some(red))
 
+  override def withDescription(description: String): IteratorMappedReduce[K, V1, V2] =
+    copy(descriptions = descriptions ++ Seq(description))
```
johnynek: `use :+ rather than ++`

scalding-core/src/main/scala/com/twitter/scalding/typed/WithDescription.scala
```diff
@@ -0,0 +1,31 @@
...
+ * Used for objects that may have a description set to be used in .dot and MR step names.
+ */
+trait HasDescription {
+  def descriptions: Seq[String]
+}
+
+/**
+ * Used for objects that may _set_ a description to be used in .dot and MR step names.
+ */
+trait WithDescription[+This <: WithDescription[This]] extends HasDescription {
```
johnynek: why must it have `HasDescription`? without that, we could add this to `TypedPipe[T]` which I think would be useful.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -210,6 +211,9 @@ trait TypedPipe[+T] extends Serializable {
   /** prints the current pipe to stdout */
   def debug: TypedPipe[T] = onRawSingle(_.debug)
 
+  /** adds a description to the pipe */
+  def withDescription(description: String): TypedPipe[T] = new WithDescriptionTypedPipe[T](this, description)
```
johnynek: should add `TypedPipe[T] extends WithDescription[TypedPipe[T]]`, don't you think?

johnynek:
Have you tried this on your cluster at all? It's hard to be sure we are hitting all the cases (hashJoin, joins, groupBys, forceToDisk are all different code paths). Could we beef up the tests at the very least to try each of those cases?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
aposwolsky: agreed to make it private.  I wanted the sorting to be in order and did the distinct just to be safe.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
aposwolsky: sounds good

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
aposwolsky: looks good! will change

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
aposwolsky: I liked this approach because it just wrapped a pipe with a description, and you can see how easy it was.  otherwise every version of TypedPipe is going to have to set the description in their ".toPipe" ... this puts it all in one place and it didn't seem as we lose any functionality...  it is analogous to the way we are doing WithOnComplete on TypedPipe

aposwolsky:
Yes tried it on a bunch of our jobs which used all those cases (except for
hashJoin perhaps).  I'll add more tests!

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: Sounds fine to me. Just wanted to see your thoughts here. I don't really see us writing a method like:

```scala
def describe[T <: WithDescription[T]](t: T, s: String): T = t.withDescription(s)
```
which is the only think I can think of us doing with this.

So, let's leave it as is.

johnynek:
Making a macro to get the current line and file would be useful with this:

`.withDescription(fileAndLine)` could be

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
aposwolsky: I'll also add it in the comment.
https://github.com/twitter/scalding/issues/1294

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
ianoc: The other PR going in now puts a common toPipe wrapper in the TypedPipe trait for setting some state on our pipes before handing them out. When that merges doing the TypedPipe extends... will be more straight forward since all typed pipe instances are just suppling asPipe

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -52,6 +53,38 @@ object RichPipe extends java.io.Serializable {
...
+      })
+      Some(encodedResult)
+        .filterNot(x => x == null || x.isEmpty)
+        .map(decodePipeDescriptions)
+        .getOrElse(Nil)
+    }
+  }
+
+  def setPipeDescriptions(p: Pipe, descriptions: Seq[String]): Pipe = {
+    p.getStepConfigDef().setProperty(
```
ianoc: This will mostly work, but... our OrderedSerialization issues seem to have popped up that these won't always go together :/  -- we tried to use these to configure serializers for each step. Do an op that would do some transform then added the config to that pipe to let it work. But then the two would be moved apart by cascading. Not a blocker to this, but it unfortunately would mean the naming ends up being more best effort than would be ideal I think

aposwolsky: I see your point, and I wonder if we remove "withDescription" from typedPipe and only allow it after you ".group" if that would guarantee better that it is attached to the step where it occurs ... but then it is also less powerful and it is nice to annotate descriptions everywhere

ianoc: Funnily enough after a group by is exactly where we've been seeing it fail :/ I'm not sure its worth altering the plan here. This is a big step up in info/usability. We might just want to include a little note about YMMV i guess in a readme

ianoc:
Looks like this has legit failures against latest develop now, doesn't compile

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1036,6 +1040,19 @@ class WithOnComplete[T](typedPipe: TypedPipe[T], fn: () => Unit) extends TypedPi
     forceToDiskExecution.flatMap(_.toIterableExecution)
 }
 
+class WithDescriptionTypedPipe[T](typedPipe: TypedPipe[T], description: String) extends TypedPipe[T] {
+  override def toPipe[U >: T](fieldNames: Fields)(implicit flowDef: FlowDef, mode: Mode, setter: TupleSetter[U]) = {
```
johnynek: this should be asPipe now.

ianoc:
@aposwolsky we are aiming to freeze for a rls on wednesday. It would be awesome to get this in for that if we can

ianoc:
@aposwolsky we are looking to freeze for our next rls at 5pm pst today, you going to have a chance to look at this before then?

aposwolsky:
I was OOO the past few days, but I rebased and pushed the change from toPipe to asPipe

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -31,6 +36,38 @@ trait ExecutionContext {
...
 
+  private def getIdentifierOpt(descriptions: Seq[String]): Option[String] = {
+    if (descriptions.nonEmpty) Some(descriptions.distinct.mkString(", ")) else None
+  }
+
+  private def updateJobName(step: BaseFlowStep[JobConf], descriptions: Seq[String]): Unit = {
+    val conf = step.getConfig
+    getIdentifierOpt(getDesc(step)).foreach(id => {
+      val newJobName = "%s %s".format(conf.getJobName, id)
+      println("Added descriptions to job name: %s".format(newJobName))
```
johnynek: can we get a log.info or debug here instead of a println?

johnynek: also, this looks like it is changing how all jobs are named. That is somewhat orthogonal to descriptions, right?

We have a lot of tooling that assumes the job names are not changed. Could we instead just add a key to the jobConf?

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -53,6 +54,38 @@ object RichPipe extends java.io.Serializable {
     }
     p
   }
+
+  // A pipe can have more than one description when merged together, so we store them delimited with 254.toChar.
+  private def encodePipeDescriptions(descriptions: Seq[String]): String = {
+    descriptions.map(_.replace(254.toChar, ' ')).filter(_.nonEmpty).mkString(254.toChar.toString)
```
johnynek: can we factor `254.toChar` into a private val? Also, can you comment on the choice here? What impact does this have for UTF-8 descriptions? A guess since means that the 254 code point:  http://www.fileformat.info/info/unicode/char/fe/index.htm can't appear in descriptions. Why not `1.toChar` which is often used as a delimiter on HDFS?

aposwolsky: 1.toChar is just as good, I can switch to that.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
aposwolsky: hmmm, we actually like it to show it in hadoop jobtracker, what about we make it configurable so we will keep it setting name and we can default to something else?

aposwolsky: also, the "job name" always had "(X/Y)" appending to it, all this is doing is appending extra stuff after that, not changing the prefix.

aposwolsky:
Made suggested changes

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
johnynek: We have some analytics that run on our jobs and they match on the job name. For instance, we have a reducer estimator that runs and looks at the names to get the history of the particular step. Others might have other systems running.

I think changing how the jobs are names could be disruptive for people.

Can we find a way to opt-in to this behavior and you can put those options as default in your clusters? Something like a configuration key like `"scalding.description.addtoname" -> "true"` or something? When this is not present, we just add an entry to the jobConf?

aposwolsky: I already pushed a change that would put the descriptions in a different config and not update mapred.name.  Since we want it appended to the name I will add a steplistener on our side that will just add the value to that config to the job name on our side... win win for everyone :)

aposwolsky:
how do we re-run the tests, it seems to have failed on spurious errors:
15/05/28 22:09:48 ERROR mapred.TaskTracker: Caught exception: java.io.IOException: Call to localhost/127.0.0.1:35898 failed on local exception: java.nio.channels.ClosedByInterruptException

ianoc:
@aposwolsky i triggered a rebuild of your branch, not sure you can do it yourself :/

aposwolsky:
What's up with these errors? Is this pull request ready to be merged?

ianoc:
Looks like legit failures, same target fails in 2.10 and 2.11 -- 

[error] Failed tests:
[error] 	com.twitter.scalding.platform.WithDescriptionTest
[error] (scalding-hadoop-test/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 1079 s, completed May 29, 2015 6:19:38 PM

aposwolsky:
Thanks, the log didn't show any interesting details but it showed up when I ran the failing test locally. I had a bug not changing everything from char 254 to char 1, but it also turned out that using char 1 as a delimiter caused issues, so I switched it to 255 and pushed fix.

aposwolsky:
Can you trigger it to run again, the failure is now: "java.lang.OutOfMemoryError: PermGen space"

ianoc:
rebuilding..

aposwolsky:
yay it worked!

aposwolsky:
ping?

johnynek:
looking

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -52,6 +87,18 @@ trait ExecutionContext {
       }
       val flow = mode.newFlowConnector(withId).connect(flowDef)
 
+      flow match {
+        case hadoopFlow: HadoopFlow =>
+          val flowSteps = hadoopFlow.getFlowSteps.asScala
+          flowSteps.foreach(step => {
+            val baseFlowStep: BaseFlowStep[JobConf] = step.asInstanceOf[BaseFlowStep[JobConf]]
+            val descriptions = getDesc(baseFlowStep)
+            updateFlowStepName(baseFlowStep, descriptions)
```
johnynek: looks like you are still updating the job name. Like I said, I don't want to change that without more thought and buy in. Our existing job monitoring will be impacted and probably other systems will as well. I know we should have a test for anything we care about, and we can add an issue to do this here (if indeed we want some contract on the name and not another way to track job identity).

aposwolsky: The mapred.name is already set by the time it gets here, so the job.xml file for hadoop will not have any mention of the descriptions. However, the "stepName" needs to be updated for the dot file generation to have the descriptions

aposwolsky:
@johnynek : ok, I changed it so it only changed the flow stepName in the .dot file generation, otherwise it just sets a different config. This works for our setup since (1) .dot files have descriptions and (2) I added a flowstepstrategy that adds the descriptions to the job name so it looks good on our side.

Does this solution work for you as well? I would like to see this pull request go through because I don't want to have to start maintaining a patch.

aposwolsky:
ping?

aposwolsky:
ping?

johnynek:
Looking now. Sorry.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -31,6 +36,24 @@ trait ExecutionContext {
   def flowDef: FlowDef
   def mode: Mode
 
+  private def getIdentifierOpt(descriptions: Seq[String]): Option[String] = {
+    if (descriptions.nonEmpty) Some(descriptions.distinct.mkString(", ")) else None
+  }
+
+  private def updateStepConfigWithDescriptions(step: BaseFlowStep[JobConf], descriptions: Seq[String]): Unit = {
```
johnynek: descriptions is unused, right?

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -53,6 +54,39 @@ object RichPipe extends java.io.Serializable {
...
+    encoding.split(255.toChar).toSeq
+  }
+
+  def getPipeDescriptions(p: Pipe): Seq[String] = {
+    if (p.getStepConfigDef.isEmpty)
+      Nil
+    else {
+      val encodedResult = p.getStepConfigDef.apply(Config.PipeDescriptions, new Getter {
+        override def update(s: String, s1: String): String = ???
+        override def get(s: String): String = null
```
johnynek: can you comment on what is going on here?

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -53,6 +54,39 @@ object RichPipe extends java.io.Serializable {
...
+      })
+      Some(encodedResult)
```
johnynek: Option(encodedResult)
 .filterNot(_.isEmpty)

might be simpler. But no biggie if you don't want to do this.

johnynek:
One minor issue, then in looks good to merge (remove an unused parameter). Thanks for pushing this through to the end.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
aposwolsky: good catch! thanks will fix

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
aposwolsky: oh good idea, will change.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
aposwolsky: yep, will add note.

aposwolsky:
no problem! made changes

johnynek:
Thanks! :+1: Will merge when it's green.

aposwolsky:
:( can we re-run the CI build?

johnynek:
restarted

aposwolsky:
success!

johnynek:
Thanks for this nice new feature!

aposwolsky:
anytime :)

joshualande:
@aposwolsky, thanks for adding this feature. It is really great!

Issue #1204, opened 2015-02-20
comb the code for deprecated things we can remove.


johnynek:
This will be a judgement call, but let's try to get deprecated things out (especially if no one is using them at Twitter, which will be a good litmus test).

oeddyo:
Need some pointers here :)

Is there any way to tell what is deprecated? 

I can see some "deprecate" warning while doing 
./sbt test

How do we know if someone is using them at Twitter (or anywhere else?)

MansurAshraf:
Remove the deprecated methods and do a sandbox build would be the way to go

oeddyo:
@MansurAshraf I'll take a look

oeddyo:
Closing

PR #1288, opened 2015-05-13
Exie/1204

oeddyo:
Remove deprecate methods:

https://github.com/twitter/scalding/issues/1204

What I did was:

1) Search for "deprecate" in scalding-core
2) If some methods have deprecate mark, I do a code search to see if anyone in Twitter internal is using them
3) If not delete


johnynek:
Thanks. This is helpful. Thank you for searching source.

PR #1289, opened 2015-05-13
Move OrderedSerialization into zero-dep scalding-serialization module

johnynek:
This is pretty general code and nice to have it not so tightly coupled to scalding. Was really easy to do this refactor, so there really is not a coupling.


scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
```diff
@@ -50,6 +50,8 @@ class KryoHadoop(config: Config) extends KryoInstantiator {
     newK.register(classOf[com.twitter.algebird.HyperLogLogMonoid], new HLLMonoidSerializer)
     newK.register(classOf[com.twitter.algebird.Moments], new MomentsSerializer)
     newK.addDefaultSerializer(classOf[com.twitter.algebird.HLL], new HLLSerializer)
+    // Don't serialize Boxed instances using Kryo.
+    newK.addDefaultSerializer(classOf[com.twitter.scalding.serialization.Boxed[_]], new ThrowingSerializer)
```
ianoc: Did you manage to test this?

johnynek: No, I didn't, but it should not be hard to do. I'll add such a test.

ianoc:
LGTM --> test as a follow up pr seems fine with me.

PR #1290, opened 2015-05-13
Reverting elephantbird to 4.6

jnievelt:
In case we want to go with this for 0.14.0


johnynek:
I guess, due to semver, 4.7 is safe to use if you need 4.6.

PR #1291, opened 2015-05-13
[Sln hack share do not merge/full review yet] properly register ordered serialization

MansurAshraf:
@ianoc Let me know if you want me to revert import ordering. Stupid IDE formats that on commit

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/package.scala
+++ b/scalding-core/src/main/scala/com/twitter/package.scala
@@ -73,4 +73,6 @@ package object scalding {
       !filter.accept(p)
     }
   }
+
+  val state = new java.util.concurrent.ConcurrentHashMap[String, String]()
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -15,24 +15,20 @@ limitations under the License.
 */
 package com.twitter.scalding
 
-import com.twitter.algebird.monad.Reader
+import java.io.{ BufferedWriter, FileOutputStream, OutputStreamWriter }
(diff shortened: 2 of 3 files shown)


scalding-core/src/main/scala/com/twitter/package.scala
```diff
@@ -73,4 +73,6 @@ package object scalding {
       !filter.accept(p)
     }
   }
+
+  val state = new java.util.concurrent.ConcurrentHashMap[String, String]()
```
johnynek: 0) this looks really scary and architecturally bad news. I hope we can find some other way.

1) this should be `private[scalding]` at the very least.

2) can we add plenty of comments here as to what the use case is here.

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -214,7 +210,14 @@ class Job(val args: Args) extends FieldConversions with java.io.Serializable {
 
   private def executionContext: Try[ExecutionContext] =
     Config.tryFrom(config).map { conf =>
-      ExecutionContext.newContext(conf)(flowDef, mode)
+      val updatedConfig = if (conf.getRequireOrderedSerialization) {
+        state.asScala.foldLeft(conf) {
```
johnynek: why only add these when we require ordered serialization? We want to get the benefit even when we don't require all of them to be ordered. Some of them can still be a win.

johnynek: ExecutionApp does not hit this code path. We should go deeper and look inside ExecutionContext.run to get to the shared code path.

johnynek:
We can close this now that #1293 is merged right?

ianoc:
Think so

PR #1293, opened 2015-05-15
Adds a test that should break

ianoc:
This should break in the ordered serialization

I've hit exceptions from it like OOM's, missing serializers for BoxedN and so on.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FlowState.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FlowState.scala
@@ -21,12 +21,12 @@ import java.util.WeakHashMap
 /**
  * Immutable state that we attach to the Flow using the FlowStateMap
  */
-case class FlowState(sourceMap: Map[String, Source] = Map.empty, flowBoxes: Set[(String, String)] = Set()) {
+case class FlowState(sourceMap: Map[String, Source] = Map.empty, flowConfigUpdates: Set[(String, String)] = Set()) {
   def addSource(id: String, s: Source): FlowState =
     copy(sourceMap = sourceMap + (id -> s))
 
-  def addBoxed(k: String, v: String): FlowState =
-    copy(flowBoxes = flowBoxes + ((k, v)))
+  def addConfigSetting(k: String, v: String): FlowState =
+    copy(flowConfigUpdates = flowConfigUpdates + ((k, v)))
 
   def getSourceNamed(name: String): Option[Source] =
     sourceMap.get(name)
--- a/scalding-core/src/main/scala/com/twitter/scalding/FlowState.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FlowState.scala
@@ -21,9 +21,12 @@ import java.util.WeakHashMap
 /**
  * Immutable state that we attach to the Flow using the FlowStateMap
  */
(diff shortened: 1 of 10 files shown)


MansurAshraf:
+1

scalding-hadoop-test/src/test/scala/com/twitter/scalding/ordered_serialization/OrderedSerializationTest.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def sample[T: Arbitrary]: T = Arbitrary.arbitrary[T].sample.get
+  val data = sample[List[NestedCaseClass]].take(1000)
+}
+
+case class NestedCaseClass(day: RichDate, key: (String, String))
+
+class OrderedSerializationTest extends FunSuite with HadoopPlatformTest {
+  import OrderedSerializationTest._
+  test("A test job with a fork and join, had previously not had boxed serializations on all branches") {
```
johnynek: this is only testing for no exception, right? Can we add a comment so we don't lose track of that later (or add also a logic test of what the results should be)?

scalding-core/src/main/scala/com/twitter/scalding/FlowState.scala
```diff
@@ -21,9 +21,12 @@ import java.util.WeakHashMap
 /**
  * Immutable state that we attach to the Flow using the FlowStateMap
  */
-case class FlowState(sourceMap: Map[String, Source] = Map.empty) {
+case class FlowState(sourceMap: Map[String, Source] = Map.empty, flowBoxes: Set[(String, String)] = Set()) {
```
johnynek: what about changing flowBoxes to `flowConfigUpdates` or something. It seems like it is not really to do with the serialization and could be used for any global flow config key/value updates.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -660,6 +661,45 @@ class RichPipe(val pipe: Pipe) extends java.io.Serializable with JoinAlgorithms
       .flatten
       .toSet
 
+  /**
+   * This finds all the boxed serializations stored in the flow state map for this
+   * flowdef. We then find all the pipes back in the DAG from this pipe and apply
+   * those serializations.
+   */
+  private[scalding] def applyBoxedSerializations(flowDef: FlowDef): Pipe = {
```
johnynek: can we change this to `applyFlowConfigProperties` or something?

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -93,46 +93,24 @@ object Grouped {
    * If we are using OrderedComparable, we need to box the key
    * to prevent other serializers from handling the key
    */
-  def maybeBox[K, V](ord: Ordering[K])(op: (TupleSetter[(K, V)], Fields) => Pipe): Pipe = ord match {
+  def maybeBox[K, V](ord: Ordering[K], flowDef: FlowDef)(op: (TupleSetter[(K, V)], Fields) => Pipe): Pipe = ord match {
```
johnynek: should this be `private[scalding]`? Seems super low level to me.

johnynek:
Aside from some minor naming comments, this looks great. Glad you got it sorted.

PR #1295, opened 2015-05-16
Add KeyCounts and KeyHits Counters to MapsideReduce. 

### LINKED ISSUE #1069: Add counters to see the cache miss rate in sumByLocalKeys
This would help us diagnose how much it helps us, and when it might be disabled (saving work).

reconditesea:
Only enable it for sumByLocalKey now.

To fix #1069


reconditesea:
I think #1207 is more favorable than this one.

PR #1299, opened 2015-05-21
Add tests around hashcode collisions

MansurAshraf:
potential bug in ProductLikeMacro's hashcode method

### DIFF
--- a/scalding-serialization-macros/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
+++ b/scalding-serialization-macros/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
@@ -221,6 +221,16 @@ class MacroOrderingProperties extends FunSuite with PropertyChecks with ShouldMa
     forAll(minSuccessful(500)) { (a: T, b: T) => checkWithInputs(a, b) }
   }
 
+  def checkCollisions[T: Arbitrary: OrderedSerialization] = {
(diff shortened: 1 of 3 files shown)


scalding-serialization-macros/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
```diff
@@ -238,40 +248,49 @@ class MacroOrderingProperties extends FunSuite with PropertyChecks with ShouldMa
   test("Test out jl.Byte") {
     implicit val a = arbMap { b: Byte => java.lang.Byte.valueOf(b) }
     check[java.lang.Byte]
+    checkCollisions[java.lang.Byte]
```
johnynek: there are only 256 bytes, so taking 100 of them, the chance of collision seems pretty high. Seems like we should use the formula, pass the cardinality of T into checkCollisions, and then have it pick the k (100 in your example).

johnynek: actually, if we are hashing 256 things into 32 bits, it should probably not collide actually.

So, this test is fine (assuming it passes).

johnynek:
wait, this test passed?

johnynek:
dammit, I bet we are not testing scalding-serialization-macros.

MansurAshraf:
arrgh, resubmitting

PR #1300, opened 2015-05-21
Fix performance bug in TypedPipeDiff

isnotinvain:
the specialized array wrappers were never being selected


johnynek:
merge when green

PR #1301, opened 2015-05-22
make serialization modules build on travis

MansurAshraf:


johnynek:
Travis is super slow, so on small targets like this, you might combine it with the scalding-date target or something. It might run faster overall.

I wish Travis were faster.

Issue #1302, opened 2015-05-22
Improve TypedParquetTuple


johnynek:
1) add macros to generate ParquetReadSupport and ParquetWriteSupport
2) We may need to change TypedParquetTupleScheme to accept instances rather than a class if reflecting a macro generated instance fails for some reason. If so, serialize and compress those instances (see Bijection's GzippedBase64String injection and JavaSerialization injection).
3) Make it a one-liner to read and write Parquet using case classes.

JiJiTang:
Hi @johnynek , yes we can improve this. And I've just made a commit about this improvement in my forked repo and will create PR later.

PR #1303, opened 2015-05-25
Improve TypedParquetTuple #1302

JiJiTang:


JiJiTang:
Hi @johnynek here's my commit with macros to generate Parquet read/write support instances. And I'm wondering maybe need to create a module like "scalding-parquet-macro" to put all the macros there and use them from "TypedParquet". Thank you very much for reviewing it and let me know your thoughts.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
```diff
@@ -95,6 +79,45 @@ trait ParquetWriteSupport[T] extends WriteSupport[T] {
   def writeRecord(r: T, rc: RecordConsumer, schema: MessageType): Unit
 }
 
+object ParquetInputOutputFormat {
+  val READ_SUPPORT_INSTANCE = "parquet.read.support.instance"
+  val WRITE_SUPPORT_INSTANCE = "parquet.write.support.instance"
+  val deserialize = (GZippedBase64String(_)) andThen Bijection.bytes2GZippedBase64.inverse andThen KryoInjection.invert
```
johnynek: Let's make one Injection and use that:

something like:
```scala
val injection: Injection[Any, String] = KryoInjection.andThen(Injection.connect[Array[Byte], GZippedBase64String, String])
```

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
```diff
@@ -102,7 +125,7 @@ trait ParquetWriteSupport[T] extends WriteSupport[T] {
  * @param fp filter predicate
  * @tparam T tuple value type
  */
-class TypedParquetTupleScheme[T](val readSupport: Class[_], val writeSupport: Class[_],
+class TypedParquetTupleScheme[T](val readSupport: ParquetReadSupport[T], val writeSupport: ParquetWriteSupport[T],
```
johnynek: these might not be serializable. You might want to use `@transient` here and `com.twitter.chill.Externalizer` which attempts both Java and Kryo serialization. Cascading needs schemes to be serializable, I think.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
```diff
@@ -18,8 +16,10 @@ class TypedParquetTupleTest extends WordSpec with Matchers with HadoopPlatformTe
       import com.twitter.scalding.parquet.tuple.TestValues._
       HadoopPlatformJobTest(new WriteToTypedParquetTupleJob(_), cluster)
         .arg("output", "output1")
-        .sink[SampleClassB](TypedParquet[SampleClassB, BReadSupport](Seq("output1"))) { _.toSet shouldBe values.toSet }
-        .run
+        .sink[SampleClassB](TypedParquet[SampleClassB](Seq("output1"),
+          Macros.caseClassParquetReadSupport[SampleClassB](SampleClassB.schema))) {
```
johnynek: this should be an implicit argument and users should do something like:

```scala
import Macros.caseClassParquetReadSupport
```
in any file they use the implicit. If they want to be explicit, that's fine, implicit parameters can be passed explicitly.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
```diff
@@ -18,8 +16,10 @@ class TypedParquetTupleTest extends WordSpec with Matchers with HadoopPlatformTe
...
+          _.toSet shouldBe values.toSet
```
johnynek: this is not a great pattern which we have, but should avoid. A better one is:

```scala
def toMap[T](i: Iterable[T]): Map[T, Int] = i.groupBy(identity).mapValues(_.size) 

toMap(_) shouldBe toMap(values)
```

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
```diff
@@ -38,17 +36,14 @@ object Macros {
...
 
   /**
-   * Macro used to generate parquet tuple converter for a given case class.
-   *
-   * @tparam T Case class type that contains primitive or collection type fields or nested case class.
-   * @return Generated parquet converter
+   * Macro generated case class read support
    */
-  def caseClassParquetTupleConverter[T]: ParquetTupleConverter[T] = macro ParquetTupleConverterProvider.toParquetTupleConverterImpl[T]
+  def caseClassParquetReadSupport[T](schema: String): ParquetReadSupport[T] = macro ParquetReadSupportProvider.toParquetReadSupportImpl[T]
```
johnynek: why is the schema here? Why not call the toParquetSchemaImpl macro? Seems it just requires extra work for the user.

Can this be implicit?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
```diff
@@ -38,17 +36,14 @@ object Macros {
...
 
   /**
    * Macro used to generate case class write support to parquet.
    * @tparam T User defined case class tuple type.
    * @return Generated case class tuple write support function.
    */
-  def caseClassWriteSupport[T]: (T, RecordConsumer, MessageType) => Unit = macro WriteSupportProvider.toWriteSupportImpl[T]
+  def caseClassParquetWriteSupport[T](schema: String): ParquetWriteSupport[T] = macro WriteSupportProvider.toWriteSupportImpl[T]
```
johnynek: same comments as for readSupport.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
```diff
@@ -66,37 +49,26 @@ object TypedParquetSink {
...
    *
    * @param paths paths of parquet I/O
-   * @param t Read support type tag
    * @tparam T Tuple type
-   * @tparam W Write support type
    * @return a typed parquet source.
    */
-  def apply[T, W <: ParquetWriteSupport[T]](paths: Seq[String])(implicit t: ClassTag[W]) =
-    new TypedFixedPathParquetTuple[T, ParquetReadSupport[T], W](paths, null, t.runtimeClass.asInstanceOf[Class[W]])
+  def apply[T](paths: Seq[String], writeSupport: ParquetWriteSupport[T]) =
```
johnynek: writeSupport should be an implicit argument.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
```diff
@@ -16,48 +14,33 @@ object TypedParquet {
...
    * @tparam T Tuple type
-   * @tparam R Read support type
    * @return a typed parquet source.
    */
-  def apply[T, R <: ParquetReadSupport[T], W <: ParquetWriteSupport[T]](paths: Seq[String])(implicit r: ClassTag[R],
-    w: ClassTag[W]) = {
-    val readSupport = r.runtimeClass.asInstanceOf[Class[R]]
-    val writeSupport = w.runtimeClass.asInstanceOf[Class[W]]
-    new TypedFixedPathParquetTuple[T, R, W](paths, readSupport, writeSupport)
+  def apply[T](paths: Seq[String], readSupport: ParquetReadSupport[T], writeSupport: ParquetWriteSupport[T]) = {
```
johnynek: readSupport should be an implicit argument.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
```diff
@@ -16,48 +14,33 @@ object TypedParquet {
...
-    new TypedFixedPathParquetTuple[T, R, ParquetWriteSupport[T]](paths, t.runtimeClass.asInstanceOf[Class[R]], null)
+  def apply[T](paths: Seq[String], readSupport: ParquetReadSupport[T]) =
+    new TypedFixedPathParquetTuple[T](paths, readSupport, null)
 
   /**
    * Create readable typed parquet source with filter predicate.
    */
-  def apply[T, R <: ParquetReadSupport[T]](paths: Seq[String], fp: Option[FilterPredicate])(implicit t: ClassTag[R]) =
-    new TypedFixedPathParquetTuple[T, R, ParquetWriteSupport[T]](paths, t.runtimeClass.asInstanceOf[Class[R]], null) {
+  def apply[T](paths: Seq[String], readSupport: ParquetReadSupport[T], fp: Option[FilterPredicate]) =
```
johnynek: readSupport should be an implicit arg.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
```diff
@@ -16,48 +14,33 @@ object TypedParquet {
...
    *
    * @param paths paths of parquet I/O
-   * @param t Read support type tag
    * @tparam T Tuple type
-   * @tparam R Read support type
    * @return a typed parquet source.
    */
-  def apply[T, R <: ParquetReadSupport[T]](paths: Seq[String])(implicit t: ClassTag[R]) =
-    new TypedFixedPathParquetTuple[T, R, ParquetWriteSupport[T]](paths, t.runtimeClass.asInstanceOf[Class[R]], null)
+  def apply[T](paths: Seq[String], readSupport: ParquetReadSupport[T]) =
```
johnynek: readSupport should be an implicit arg.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
```diff
@@ -66,37 +49,26 @@ object TypedParquetSink {
...
 /**
  * Typed Parquet tuple source/sink.
  */
-trait TypedParquet[T, R <: ParquetReadSupport[T], W <: ParquetWriteSupport[T]] extends FileSource with Mappable[T]
+trait TypedParquet[T] extends FileSource with Mappable[T]
   with TypedSink[T] with HasFilterPredicate {
 
-  val readSupport: Class[R]
-  val writeSupport: Class[W]
+  val readSupport: ParquetReadSupport[T]
```
johnynek: generally in scala it is better to make `def` abstract. making an abstract `val` can cause some issues (val init order bug).

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/ParquetReadSupportProvider.scala
```diff
@@ -13,7 +13,7 @@ object ParquetTupleConverterProvider {
   private[this] case object SET extends CollectionType
   private[this] case object MAP extends CollectionType
 
-  def toParquetTupleConverterImpl[T](ctx: Context)(implicit T: ctx.WeakTypeTag[T]): ctx.Expr[ParquetTupleConverter[T]] = {
+  def toParquetReadSupportImpl[T](ctx: Context)(schema: ctx.Expr[String])(implicit T: ctx.WeakTypeTag[T]): ctx.Expr[ParquetReadSupport[T]] = {
```
johnynek: When would we use something other than the string created by the macro to generate schemas from case classes?

If we think this is very rare, why not skip the schema here and just call your existing macro?

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
```diff
@@ -95,6 +79,45 @@ trait ParquetWriteSupport[T] extends WriteSupport[T] {
   def writeRecord(r: T, rc: RecordConsumer, schema: MessageType): Unit
 }
 
+object ParquetInputOutputFormat {
+  val READ_SUPPORT_INSTANCE = "parquet.read.support.instance"
```
johnynek: maybe this should be `scalding.parquet.read.support.instance` to be sure we don't collide with something from Parquet.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
```diff
@@ -83,7 +70,7 @@ class WriteToTypedParquetTupleJob(args: Args) extends Job(args) {
 
   val outputPath = args.required("output")
 
-  val sink = TypedParquetSink[SampleClassB, WriteSupport](Seq(outputPath))
+  val sink = TypedParquetSink[SampleClassB](Seq(outputPath), Macros.caseClassParquetWriteSupport[SampleClassB](SampleClassB.schema))
```
johnynek: this should be:

```scala
TypedParquetSink[SampleClassB](outputPath)
```
if it is not that easy, it is too much code.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
```diff
@@ -98,7 +85,7 @@ class ReadWithFilterPredicateJob(args: Args) extends Job(args) {
   val inputPath = args.required("input")
   val outputPath = args.required("output")
 
-  val input = TypedParquet[SampleClassC, CReadSupport](Seq(inputPath), Some(fp))
+  val input = TypedParquet[SampleClassC](Seq(inputPath), Macros.caseClassParquetReadSupport[SampleClassC](SampleClassC.schema), Some(fp))
```
johnynek: Make this work with:

```scala
val input = TypedParquet[SampleClassC](inputPath)
```

johnynek:
Thanks a ton! I really appreciate how much work this is. I think we can make it super easy to read and write parquet: just as easy as `TypedTsv[(Int, Int, String)]("mydata")` is, except with all the benefits of Parquet.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
JiJiTang: fixed

JiJiTang: fixed

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/Macros.scala
JiJiTang: Fixed. At the beginning, I thought if user wanted to create a source/sink for the same I/O path the the schema generation macro will be called twice...so I put it as a parameter so that source/sink could share the same schema instance. But you are right  it requires extra work for the user and I delete it to make it simple to use.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/macros/impl/ParquetReadSupportProvider.scala
JiJiTang: fixed

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
JiJiTang: Agree, fixed.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
JiJiTang: fixed, that's how injection should be used:P

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
JiJiTang: Not sure I understand this well. TypedParquetTupleScheme extends Scheme which implements java serializable and both ParquetReadSupport and WriteSupport have implemented serializable interface.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
JiJiTang: Fixed

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/tuple/TypedParquetTupleTest.scala
JiJiTang: Now it works, please check my latest commit.

JiJiTang:
Hi @johnynek thank you very much for your kind words and your code review and also your good advices. I am so glad to contribute to Scalding. Please check my latest commit.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/TypedParquet.scala
JiJiTang: fixed

johnynek:
Looks great! thanks for making the changes so quickly. Looks very close to merging.

johnynek:
Closes #1302

JiJiTang:
Hi @johnynek thanks a lot. I've just added another commit. Please let me know if any problem before good to go.

johnynek:
Looks good to me!

Will merge when green. Thanks a lot.

Next up: macros to create Filters from scala code:

```scala
Filter[SomeClass] {
  _.y == "foo"
}
// returns:
// FilterApi.eq(binaryColumn("y"), Binary.fromString("foo"))
```

JiJiTang:
Hi @johnynek , thank you very much for merging the PR. For Filter, in parquet-mr/parquet-scala, there's a DSL for defining filter. And good idea, macro generated filters would be even more smoothie.

PR #1304, opened 2015-05-29
Add UnitOrderedSerialization

### LINKED ISSUE #1297: Add OrderedSerialization for Unit and wire it in GroupAll
GroupAll is using scala ordering and breaks when user is mixing RequiredBinaryComparator trait

johnynek:
closes #1297 

I'm not sure we should merge this yet. I'm concerned this is going to increase serialization costs unless we add tokens for the Boxed class wrappers that we use to do the dispatch to find the correct serializers.

Also, We should not merge this until we are sure we worked out all the bugs and perf issues at Twitter, since this will impact anyone using groupAll.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -19,6 +19,7 @@ import java.io.Serializable
 
 import com.twitter.algebird.{ Semigroup, Monoid, Ring, Aggregator }
 
+import com.twitter.scalding.serialization.UnitOrderedSerialization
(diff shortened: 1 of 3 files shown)


ianoc:
Mm I didn't think we were merging this ?

Gabriel439:
I thought we were merging this for an internal release?  Mansur said that he needed this for adding hyperdoop support to Tsar

If you're referring to the travis failure, I fixed the failing tests already.  They were just missing an import

However, I'm okay with reverting this if necessary

Gabriel439:
I talked with Ian offline about this.  I didn't realize that develop was not used for changes being tested in internal releases.  I'll revert this change and open up a new pull request to replace the original.

PR #1307, opened 2015-06-03
Change groupRandomly & groupAll to use OrderedSerialization

MansurAshraf:
@johnynek Not sure if this is the best approach but take a look

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -16,6 +16,9 @@ limitations under the License.
 package com.twitter.scalding.typed
 
 import com.twitter.algebird.{ CMS, CMSHasher }
+import com.twitter.scalding.serialization.OrderedSerialization
+import com.twitter.scalding.serialization.macros.impl
+import scala.language.experimental.macros
 
 object Sketched {
 
@@ -92,7 +95,7 @@ case class SketchJoined[K: Ordering, V, V2, R](left: Sketched[K, V],
   right: TypedPipe[(K, V2)],
   numReducers: Int)(joiner: (K, V, Iterable[V2]) => Iterator[R])
   extends MustHaveReducers {
-
+  private implicit def  ordSer: OrderedSerialization[(Int,K)] = macro impl.OrderedSerializationProviderImpl[(Int,K)]
   def reducers = Some(numReducers)
 
   //the most of any one reducer we want to try to take up with a single key
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -30,7 +30,7 @@ object Sketched {
 
   /**
    * This is based on the CMSHasherBigInt found in algebird (see docs for in depth explanation):
-   * https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1086
+   * https://github.com/twitter/algebird/lob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1086
    *
    * TODO: We need to move this hasher to CMSHasherImplicits in algebird:
    * https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1054
@@ -125,6 +125,16 @@ case class SketchJoined[K: Ordering, V, V2, R](left: Sketched[K, V],
       .withReducers(numReducers)
       .map{ case ((r, k), v) => (k, v) }
   }
+
+  private implicit def intKeyOrd: Ordering[(Int, K)] = {
+    val kord = implicitly[Ordering[K]]
+
+    kord match {
+      case kos: OrderedSerialization[_] => new OrderedSerialization2(ordSer[Int], kos.asInstanceOf[OrderedSerialization[K]])
+      case _ => Ordering.Tuple2[Int, K]
+    }
+  }
+
 }
 
 object SketchJoined {
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -30,7 +30,7 @@ object Sketched {
 
   /**
    * This is based on the CMSHasherBigInt found in algebird (see docs for in depth explanation):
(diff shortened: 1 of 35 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -115,6 +118,7 @@ object TypedPipe extends Serializable {
  */
 trait TypedPipe[+T] extends Serializable {
 
+  private def primitiveOrderedBufferSupplier[T]: OrderedSerialization[T] = macro impl.OrderedSerializationProviderImpl[T]
```
johnynek: what is `primitive` about this? Also, can we stop using the `OrderedBuffer` name which is the old name before `OrderedSerialization`. Why not `orderedSerialialization[T]`

johnynek:
Didn't Gabe revert the merging of the fix for groupAll? Do you want to use the same trick to do groupAll in this PR?

MansurAshraf:
Gabe will merge this and groupAll PR internally once its green. 






On Wed, Jun 3, 2015 at 5:42 PM -0700, "P. Oscar Boykin" <notifications@github.com> wrote:












Didn't Gabe revert the merging of the fix for groupAll? Do you want to use the same trick to do groupAll in this PR?

Gabriel439:
I did revert the groupAll fix, but I'm going to apply it to the internal relase that I cut (and also open up a pull request to replace your old one)

johnynek:
Okay. Well, this and the groupAll fix work differently. So, I don't think that is the ideal way to go. The groupAll fix avoided a dependency on the macros and by hand implemented the UnitOrderedSerialization again. If we are going to add the dependency as done here, why not use the same approach for the groupAll fix (then it becomes a 1 line change on top of this, rather than adding a new file, testing that new code and then using it).

MansurAshraf:
Thats a fair point and if we are cool with dependency on serialization module I can do the same trick with groupAll, otherwise we will have to hand role OrderedSerialization for Int 




On Wed, Jun 3, 2015 at 5:55 PM -0700, "P. Oscar Boykin" <notifications@github.com> wrote:












Okay. Well, this and the groupAll fix work differently. So, I don't think that is the ideal way to go. The groupAll fix avoided a dependency on the macros and by hand implemented the UnitOrderedSerialization again. If we are going to add the dependency as done here, why not use the same approach for the groupAll fix (then it becomes a 1 line change on top of this, rather than adding a new file, testing that new code and then using it).

johnynek:
I'd use the same trick (and as per #1308 we can probably merge that module into scalding-serialization).

MansurAshraf:
@Gabriel439 I have added the fix for groupAll in this PR as well. Based on @johnynek suggestion, you will have to revert the other PR internally as well and just use this PR for both fixes

Gabriel439:
Got it.  I will use your pull request exclusively for both changes.

Gabriel439:
Is it possible that adding the serialization macros dependency caused the permgen failure?

MansurAshraf:
I doubt it but I am not a macro expert

johnynek:
Totally possible, but I guess it should be only used to create two new
classes, which should be small (but maybe we don't really know for sure).

On Thu, Jun 4, 2015 at 10:20 AM, Mansur Ashraf <notifications@github.com>
wrote:

> I doubt it but I am not a macro expert
>
> —
> <https://github.com/twitter/scalding/pull/1307#issuecomment-109036443>.
>



-- 

Gabriel439:
This time the error seems to be that one of the test suites ran too long (2 hours) -.-

I've run these tests for Mansur's branch locally successfully (i.e. everything up to and including commit 0cba5b0)

MansurAshraf:
It seems like these new tests for [TypePipeDiff](https://github.com/twitter/scalding/blob/develop/scalding-core/src/test/scala/com/twitter/scalding/typed/TypedPipeDiffTest.scala) are pushing Travis over the edge.

stats on my MBP

Without TypedPipeDiffTest
```
real	1m51.315s
user	2m17.611s
sys	0m8.015s
```

With TypedPipeDiffTest
```
real	8m13.432s
user	6m52.950s
sys	0m11.701s
```

@isnotinvain is it possible that we can trim down some of these tests or are they all required??

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -115,6 +118,7 @@ object TypedPipe extends Serializable {
  */
 trait TypedPipe[+T] extends Serializable {
 
+  private def orderedSerialialization[T]: OrderedSerialization[T] = macro impl.OrderedSerializationProviderImpl[T]
```
ianoc: Can we not get this from the package itself? not have a macros reference here

ianoc:
LGTM, merge when green

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -125,16 +125,6 @@ case class SketchJoined[K: Ordering, V, V2, R](left: Sketched[K, V],
       .withReducers(numReducers)
       .map{ case ((r, k), v) => (k, v) }
   }
-
-  private implicit def intKeyOrd: Ordering[(Int, K)] = {
```
johnynek: this looks like a merge fail. It is backing out the sketched fix.

johnynek:
Other than the Sketched merge fail, this is fine with me.

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -30,7 +30,7 @@ object Sketched {
 
   /**
    * This is based on the CMSHasherBigInt found in algebird (see docs for in depth explanation):
-   * https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1086
+   * https://github.com/twitter/algebird/lob/develop/algebird-core/src/main/scala/com/twitter/algebird/CountMinSketch.scala#L1086
```
isnotinvain: unintentional?

isnotinvain:
@MansurAshraf maybe we can turn down the number of runs in the property tests for typed pipe diff? can that be done per-test?

johnynek:
@isnotinvain yes, that can be done per test. We might only want 10 tries or something on an expensive test.

:+1:

johnynek:
merge when green

isnotinvain:
That's fine with me, however, 8 minutes is not 2 hours, so these tests are just one of many things slowing the build down right?

MansurAshraf:
8 mins was on my mac which is much faster than the travis machine we get.

Issue #1312, opened 2015-06-05
some old commons sources need to be made TypedSink


johnynek:
https://github.com/twitter/scalding/blob/develop/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/DailySources.scala

See the LongThrift classes there.

PR #1313, opened 2015-06-05
make LongThrift sources TypedSink

rubanm:
Addresses https://github.com/twitter/scalding/issues/1312


scalding-commons/src/main/scala/com/twitter/scalding/commons/source/DailySources.scala
```diff
@@ -60,6 +60,7 @@ abstract class TimePathedLongThriftSequenceFile[V <: TBase[_, _]: Manifest](f: F
   with WritableSequenceFileScheme
   with Serializable
   with Mappable[(Long, V)]
+  with TypedSink[(Long, V)]
```
johnynek: I think we need to override `override def sinkFields = f` and `override setter[...]` with the subsetter of `TupleSetter.of[(Long, V)]`

And also below.

Might be good to run a quick job (even using local HDFS), to see if it works.

rubanm:
Verified on hadoop.

johnynek:
Thanks for doing this! Merge when green.

PR #1317, opened 2015-06-09
scaliform formated source code

MansurAshraf:
Pure scalariform formatted source without any code change.


ianoc:
Is this right ? A simple compile doesn't produce this diff which runs
scalariform ?

johnynek:
I thought we turned off the rule that leaves the trailing `)` on newline.

This looks strange.

scalding-args/src/main/scala/com/twitter/scalding/Args.scala
```diff
@@ -41,16 +41,17 @@ object Args {
...
-        .foldLeft(List("" -> List[String]())) { (acc, arg) =>
-          val noDashes = arg.dropWhile{ _ == '-' }
-          if (arg == noDashes || isNumber(arg))
-            (acc.head._1 -> (arg :: acc.head._2)) :: acc.tail
-          else
-            (noDashes -> List()) :: acc
-        }
-        //Now reverse the values to keep the same order
-        .map { case (key, value) => key -> value.reverse }.toMap)
+      .filter{ a => !a.matches("\\s*") }
```
jnievelt: Is this catching some "multiple lines in a function application should be aligned" thing?  If so, maybe declare val processedArgs = args... ahead of this, so the processing chain isn't inside the Args()

MansurAshraf:
This was against a branch cut from develop with no code change. Can someone else verify that its happening for them as well? I talked to @Gabriel439 and he had the same issue

ianoc:
I ran a compile on a local develop and it had no changes.

Gabriel439:
I got scalariform-related changes when I did `twitter_release` to prepare our internal release

Gabriel439:
So it turns out that the kind of changes that Mansur was talking about were much larger than the ones I saw (just a few small changes).  He identified it as a problem specific to his dev environment

MansurAshraf:
It was https://github.com/ensime/ensime-sbt plugin in my global setting causing the problem

PR #1320, opened 2015-06-14
Added a sealed trait ordered serializer. When it works its great. Not…

ianoc:
… as reliable as we'd like. But hopefully restrictions on it will do the job


scalding-serialization-macros/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
```diff
@@ -572,6 +602,14 @@ class MacroOrderingProperties extends FunSuite with PropertyChecks with ShouldMa
     checkCollisions[TestCC]
   }
 
+  test("Test out Sealed Trait") {
```
johnynek: don't we want one that does:

```scala
  check[SealedTraitTest]
```
the test below looks like it is just using the case class stuff you had before.

johnynek:
This is awesome. Really this should be all we need (sealed traits give us sum types, tuples/case classes give us product types).

scalding-serialization-macros/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
ianoc: sorry cleanup typo. updated

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+
+import scala.language.experimental.macros
+import scala.reflect.macros.Context
+
+import com.twitter.scalding._
+import com.twitter.scalding.serialization.OrderedSerialization
+
+object SealedTraitLike {
+
+  def compareBinary(c: Context)(inputStreamA: c.TermName, inputStreamB: c.TermName)(subData: List[(Int, c.Type, Option[TreeOrderedBuf[c.type]])]): c.Tree = {
```
johnynek: can you comment what is in the subData. I am trying to figure that out, but I'll probably forget by the time I look at it again.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+    import c.universe._
+    def freshT(id: String) = newTermName(c.fresh(id))
+    val valueA = freshT("valueA")
+    val valueB = freshT("valueB")
+    val idxCmp = freshT("idxCmp")
+
+    val compareSameTypes: Tree = subData.foldLeft(Option.empty[Tree]) {
+      case (existing, (idx, tpe, optiTBuf)) =>
+
+        val commonCmp: Tree = optiTBuf.map{ tBuf =>
```
johnynek: space after map.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+              } else {
+                $t
+              }
+            """)
+          case None =>
+            Some(q"""
+                if($valueA == $idx) {
+                  $commonCmp
+                } else {
+                  sys.error("Unable to compare unknown type")
```
johnynek: how could this happen? Can you add a comment?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+    }.get
+
+    q"""
+        val $valueA: Int = $inputStreamA.readByte.toInt
+        val $valueB: Int = $inputStreamB.readByte.toInt
+        val $idxCmp: Int = _root_.java.lang.Integer.compare($valueA, $valueB)
+        if($idxCmp != 0) {
+          $idxCmp
+        } else {
+          $compareSameTypes
```
johnynek: this branch implies valueA == valueB, so in the code above we only check `valueA` against `idx`, right?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
johnynek: I guess the way you are building this, it will be the most nested thing and it can only be reached if somehow none of the ids match, which should be impossible.

If that is right can you put a different message, something like: `unreachable code`.

Also, I wonder if it makes sense to write code to do a binary search rather than a linear one as is done here.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  def hash(c: Context)(element: c.TermName)(subData: List[(Int, c.Type, Option[TreeOrderedBuf[c.type]])]): c.Tree = {
+    import c.universe._
+    def freshT(id: String) = newTermName(c.fresh(id))
+
+    val innerArg = freshT("innerArg")
+    subData.foldLeft(Option.empty[Tree]) {
+      case (optiExisting, (idx, tpe, optiTBuf)) =>
+        val commonPut: Tree = optiTBuf.map { tBuf =>
```
johnynek: instead of commonPut, what about a name like `elementHash` or something?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+          q"""{
+              val $innerArg: $tpe = $element.asInstanceOf[$tpe]
+              ${tBuf.hash(innerArg)}
+            }
+              """
+        }.getOrElse[Tree](q"_root_.scala.Int.MaxValue")
```
johnynek: this is repeated for the same reason on line 100. Could we share that Tree?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+          }
+        }.getOrElse(q"_root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.DynamicLen(1)")
+        val tmpPreLen = freshT("tmpPreLen")
+
+        val lenT = q"""
+        val $tmpPreLen: _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.MaybeLength  = $baseLenT
+
+        ($tmpPreLen match {
+          case _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.ConstLen(l) =>
+            _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.DynamicLen(l + 1)
```
johnynek: why to we always jump out to dynamic here? If you did:
```scala
sealed trait Or
case class Left(left: Int) extends Or
case class Right(right: Int) extends Or
```
You'd have a constant size of `5` bytes. Just not worth it to do that optimization yet for what might be corner case?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+            } else {
+              $t
+            }
+          """)
+          case None =>
+            Some(q"""
+          if($valueA == $idx) {
+            $extract
+          } else {
+            sys.error("Did not understand thrift union idx: " + $valueA)
```
johnynek: should this be `thrift` here?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,323 @@
...
+            } else {
+              $t
+            }
+          """)
+          case None =>
+            Some(q"""
+              if($arg.isInstanceOf[$tpe]) {
+                $idx
+              } else {
+                sys.error("Unable to compare unknown type")
```
johnynek: like above, this is actually unreachable code, right?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/SealedTraitOrderedBuf.scala
```diff
@@ -0,0 +1,73 @@
...
+    import c.universe._
+    def freshT(id: String) = newTermName(c.fresh(s"$id"))
+
+    val knownDirectSubclasses = outerType.typeSymbol.asClass.knownDirectSubclasses
+
+    if (knownDirectSubclasses.isEmpty)
+      c.abort(c.enclosingPosition, s"Unable to access any knownDirectSubclasses for $outerType , a bug in scala 2.10/2.11 makes this unreliable.")
+
+    val subClassesValid = knownDirectSubclasses.forall { sc =>
+      scala.util.Try(sc.asType.asClass.isCaseClass).getOrElse(false)
```
johnynek: what about case objects, are they supported?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/SealedTraitOrderedBuf.scala
```diff
@@ -0,0 +1,73 @@
...
+    if (!subClassesValid)
+      c.abort(c.enclosingPosition, s"We only support the extension of a sealed trait with case classes.")
+
+    val dispatcher = buildDispatcher
+
+    val subClasses: List[Type] = knownDirectSubclasses.map(_.asType.toType).toList
+
+    val subData: List[(Int, Type, Option[TreeOrderedBuf[c.type]])] = subClasses.map { t =>
+      (t, Some(dispatcher(t)))
+    }.zipWithIndex.map{ case ((tpe, tbuf), idx) => (idx, tpe, tbuf) }.toList
```
johnynek: looks like the tbuf here is always Some. Do we have cases where it can be None? If not, we are carrying some extra complexity around.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/SealedTraitOrderedBuf.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    require(subData.size > 0, "Must have some sub types on a union?")
```
johnynek: the question mark makes this confusing. Can you restate? Are you saying that it makes no sense to have subData be empty, yet have a sealed trait? So is this indicative of a logic bug in that it should never happen unless you have an uninhabited type?

Also: `require(subData.nonEmpty, ...`

scalding-serialization-macros/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
```diff
@@ -64,8 +64,38 @@ object TestCC {
...
+    for {
+      cc <- arb[TestCC]
+      bb <- arb[TestCaseClassB]
+      dd <- arb[TestCaseClassD]
+      t <- Gen.oneOf(cc, bb, dd)
+    } yield t
+  }
 }
-case class TestCC(a: Int, b: Long, c: Option[Int], d: Double, e: Option[String], f: Option[List[String]], aBB: ByteBuffer)
+sealed trait SealedTraitTest
```
johnynek: can we deal with `case object?` For instance, I guess this macro generated code could subsume `Option` and `Either` serializers as they are just special cases of these, right?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
ianoc: Its been a while since i wrote this, but the thinking goes if i remember, if a given instance of a sealed trait itself has a constant len then we are still stuck at dynamic len since we are working on the trait level, and we may end up with another instance of the trait before long. Passing through a sealed trait guarantee's you that it'll always be dynamic. Now you could have the case where all the instances are the same length in which case this doesn't handle that very well

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
ianoc: Yeah not sure, maybe it does? but i guess maybe it would be good to see that in benchmarks first?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
ianoc: Yep exactly

ianoc:
@johnynek Have i think it in a better shape now, does the case objects. I'll see if working without Either + Option ordered serializations is good now too

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/SealedTraitLike.scala
```diff
@@ -0,0 +1,318 @@
...
+        val elementHash: Tree = q"""
+              val $innerArg: $tpe = $element.asInstanceOf[$tpe]
+              ${tBuf.hash(innerArg)}
+            """
+
+        optiExisting match {
+          case Some(s) =>
+            Some(q"""
+            if($element.isInstanceOf[$tpe]) {
+              $elementHash ^ _root_.com.twitter.scalding.serialization.Hasher.int.hash($idx)
```
johnynek: `_root_.com.twitter.scalding.serialization.Hasher.int.hash($idx)` is a constant known at compile time. Can we evaluate that using something like:

```scala
val treeHash: c.Tree = c.resetLocalAttrs(q"_root_.com.twitter.scalding.serialization.Hasher.int.hash($idx)")
val intHash: Int = c.eval(c.Expr[Int](treeHash))
// then:
q"""$elementHash ^ $intHash"""
```

ianoc: Evals are pretty much totally broken and unreliable.

ianoc: But your points valid, i'm not sure we need to eval or do anything special here. We can probably just call the method itself

PR #1321, opened 2015-06-14
Add secondary sorting using ordered serialization

ianoc:


scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -53,6 +54,7 @@ object CascadingBinaryComparator {
 
     def check(s: Splice): Try[Unit] = {
       val m = s.getKeySelectors.asScala
+      val sortingSelectors = s.getSortingSelectors.asScala
```
johnynek: can't we just do:

```scala
val m = s.getKeySelectors.asScala ++ s.getSortingSelectors.asScala
```
since the check is exactly the same? then we don't have to duplicate the checking code.

ianoc: We can, actually I had a note here to ask you if you think this should be a separate option entirely? Given we never really expose the key sorting ability to users in a real manner just using our sorting should be fine for them. But secondary sort is where you may/probably want to customize ?

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -181,11 +197,37 @@ sealed trait ReduceStep[K, V1] extends KeyedPipe[K] {
   def mapped: TypedPipe[(K, V1)]
   // make the pipe and group it, only here because it is common
   protected def groupOp[V2](gb: GroupBuilder => GroupBuilder): TypedPipe[(K, V2)] =
+    groupOpWithValueSort[V2](None)(gb)
+
+  protected def groupOpWithValueSort[V2](valueSort: Option[Ordering[_ >: V1]] = None)(gb: GroupBuilder => GroupBuilder): TypedPipe[(K, V2)] =
```
johnynek: we never use it with the default arg, can we not use the default arg at all to make it more clear?

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
johnynek: Yes, good point.

I don't think we should require it here since `.sortBy` will be broken in that case, right?

ianoc: Happy to just drop it entirely, is it worth adding a new option for? I'm not sure it is, secondary sorting is rare enough we can have users manually specify it if in doubt?

johnynek: yes, I agree. Let's drop this and have the user's take responsibility for this one.

johnynek:
merge when green

PR #1322, opened 2015-06-14
Bails out from the length calculation if we don't succeed often, leng…

ianoc:
…th calculations can recurse and become very expensive. This is a short path at just saying if its not working often don't bother. We use long's to avoid sync overhead and since we only need a loose value in the long. Exact doesn't matter

### DIFF
--- a/scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
+++ b/scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
@@ -232,6 +232,9 @@ object TreeOrderedBuf {
 
     t.ctx.Expr[OrderedSerialization[T]](q"""
       new _root_.com.twitter.scalding.serialization.OrderedSerialization[$T] {
+        // Ensure macro hygene for Option/Some/None
+        import _root_.scala.{Option, Some, None}
+
         private[this] var lengthCalculationAttempts: Long = 0L
         private[this] var couldNotLenCalc: Long = 0L
         private[this] var skipLenCalc: Boolean = false
--- a/scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
+++ b/scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
@@ -87,7 +87,7 @@ object TreeOrderedBuf {
 
       fnBodyOpt.map { fnBody =>
         q"""
-        @inline private[this] def payloadLength($element: $T): _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.MaybeLength = {
+        private[this] def payloadLength($element: $T): _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.MaybeLength = {
           $fnBody
         }
         """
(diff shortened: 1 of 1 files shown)


scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -102,19 +102,25 @@ object TreeOrderedBuf {
...
-          case _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.ConstLen(l) => Some(l)
-          case _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.DynamicLen(l) => Some(l)
-        }
-        (if ($tempLen.isDefined) {
-          // Avoid a closure here while we are geeking out
-          val innerLen = $tempLen.get
-          val $lensLen = posVarIntSize(innerLen)
-          Some(innerLen + $lensLen)
-       } else None): Option[Int]
+        writtenRecords += 1L
```
Gabriel439: Why do you increment `writtenRecords` both here and in the `write` method?

ianoc: its really a poorly named variable, will modify. -- Its counting points at which we would measure the length of our payload. So either by calling dynamic size as an accessor or via the write method.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -122,15 +128,21 @@ object TreeOrderedBuf {
         case const: ConstantLengthCalculation[_] => (q"""
           override val staticSize: Option[Int] = Some(${const.toInt})""", q"""
           override def dynamicSize($element: $typeName): Option[Int] = staticSize""")
-        case f: FastLengthCalculation[_] => callDynamic
-        case m: MaybeLengthCalculation[_] => callDynamic
+        case f: FastLengthCalculation[_] =>
+          callDynamic
+        case m: MaybeLengthCalculation[_] =>
+          callDynamic
+        case _ => (q"""
```
johnynek: what case is this? I thought this was a sealed trait and we already had a complete match.

ianoc: I should/will revert this. I was moving towards only allowing callDynamic being used here if the length of the code in it was less than some threshold. In practice that proved too hard to tune though.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -166,19 +178,25 @@ object TreeOrderedBuf {
         case m: MaybeLengthCalculation[_] =>
           val tmpLenRes = freshT("tmpLenRes")
           q"""
-            @inline def noLenCalc = {
-              $noLenCalc
-            }
             @inline def withLenCalc(cnt: Int) = {
```
johnynek: why not manually inline this be doing `${withLenCalc(argTree)` here? Can you add a comment about this choice?

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -102,19 +103,24 @@ object TreeOrderedBuf {
...
+              None
+            case _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.ConstLen(l) => Some(l)
+            case _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.DynamicLen(l) => Some(l)
+          }
+          (if ($tempLen.isDefined) {
+            // Avoid a closure here while we are geeking out
+            val innerLen = $tempLen.get
+            val $lensLen = posVarIntSize(innerLen)
+            Some(innerLen + $lensLen)
+         } else None): Option[Int]
```
johnynek: just realized the macros are not hygienic when it comes to Option/Some/None. Maybe put an:

`import _root_.Predef.{Option, Some, None}` on line 235?

testing for clean imports is such a pain with macros.

scalding-serialization-macros/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
ianoc: Could do it either way, this hasn't really changed in this RB. I'm not really sure making this inline was a wonderful idea to begin with. just easier to inline than put it as its own private function right now. 

Context: In XProf for large objects seeing paths not being JIT'd probably from methods growing too large. The thrift variants bail out and don't fully recurse. for case classes/maps we do allow recursion though

johnynek: Ahhh.. This was a worry we had before (we could be hampering the jit by making methods too large).

Got it. That's probably a TODO for us: look at logs to see if we are hurting the JIT.

johnynek:
looks good to me.

If you want to merge now and add an issue for the Option hygiene stuff, that's fine with me.

ianoc:
I've got the option stuff added now, and just inlined that thing, the little @inline method i can't imagine helps at all.

We probably need to figure out a heuristic as to when it makes sense to bail from inlining everything. The thrift one has the very coarse heuristic that every new thrift struct it meets is a new one, so any Maps/Sets in one struct will inline, but not across thrift struct boundaries.

PR #1323, opened 2015-06-15
increased number of box instances to 250

MansurAshraf:
Increased number of Box instances to 250 as Tsar unit tests easily exhaust 99 instances.


johnynek:
LGTM

PR #1328, opened 2015-06-18
Move to dep on a newer hadoop for mr2 API's for FS. Fixes up the hado…

ianoc:
…op platform test

I coalesed a bunch of the platform tests in to a SharedHadoopTest class, this lets them share the setup/tear down costs of the mini-mr cluster.

Looks like setting up a yarn cluster locally is a bit more expensive, but i suppose this is more representative and we can use methods now that weren't available in MR one should we decide to drop its support.


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -114,10 +114,15 @@ object FileSource {
 
   def glob(glob: String, conf: Configuration, filter: PathFilter = AcceptAllPathFilter): Iterable[FileStatus] = {
     val path = new Path(glob)
-    Option(path.getFileSystem(conf).globStatus(path, filter)).map {
-      _.toIterable // convert java Array to scala Iterable
-    } getOrElse {
-      Iterable.empty
+    val fs = FileSystem.newInstance(path.toUri, conf)
+    try {
+      Option(path.getFileSystem(conf).globStatus(path, filter)).map {
```
rubanm: I think you meant `fs.globStatus` here?

rubanm:
LGTM other than the one comment.
I think this fixes the mysql platform jdbc tests internally as well. Thanks.

Gabriel439:
Looks good to me

Issue #1326, opened 2015-06-17
OrderedSerialization for com.twitter.util.Time does not exist


MansurAshraf:
It is fairly common for Jobs to groupBy on time

johnynek:
that would make a dependency on twitter util, which is painful because they use the "overthewall" (or inside the monorepo) style to develop.

I doubt people outside Twitter are using util much with scalding, but I could be wrong.

Better, in my view, would be to make some combinators to transform OrderedSerializations when there are conversion functions (in this case, there is a mapping to Long).

ianoc:
Yeah very twitter centric, and indeed very centric to the jobs your looking at Mansur. Wouldn't do anything in OSS around it. 

We could add some stuff to build bijected OrderedSerializations maybe?

def bijectedOrderedSerialization[T,U](implicit bij: Bijection[T, U], ordSer: OrderedSerialization[U])

johnynek:
Yeah. I was thinking something like that (but maybe not depending on
bijection and just taking two functions, to keep zero dependencies in
scalding-serialization).

On Wed, Jun 17, 2015 at 10:19 AM, ianoc <notifications@github.com> wrote:

> Yeah very twitter centric, and indeed very centric to the jobs your
> looking at Mansur. Wouldn't do anything in OSS around it.
>
> We could add some stuff to build bijected OrderedSerializations maybe?
>
> def bijectedOrderedSerializationT,U
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/scalding/issues/1326#issuecomment-112936330>.
>



-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

ianoc:
Yeah that seems like a good thing to do we could provide the bijection interface in scalding-core then if we want to have it too

johnynek:
yep. That sounds good.

On Wed, Jun 17, 2015 at 10:24 AM, ianoc <notifications@github.com> wrote:

> Yeah that seems like a good thing to do we could provide the bijection
> interface in scalding-core then if we want to have it too
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/scalding/issues/1326#issuecomment-112939168>.
>



-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

PR #1329, opened 2015-06-18
Add a OrderedSerialization.viaTransform with no dependencies, and a B…

ianoc:
…ijectedOrderedSerialization in scalding core

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
@@ -16,9 +16,13 @@ limitations under the License.
 package com.twitter.scalding
 
 import com.twitter.scalding.serialization.OrderedSerialization
-import com.twitter.bijection.Bijection
+import com.twitter.bijection.ImplicitBijection
 
 object BijectedOrderedSerialization {
-  implicit def fromBijection[T, U](implicit bij: Bijection[T, U], ordSer: OrderedSerialization[U]) =
+  implicit def fromBijection[T, U](implicit bij: ImplicitBijection[T, U], ordSer: OrderedSerialization[U]) =
     OrderedSerialization.viaTransform[T, U](bij.apply(_), bij.invert(_))
+
+  implicit def fromInjection[U, U](implicit bij: Injection[T, U], ordSer: OrderedSerialization[U]) =
+    OrderedSerialization.viaTryTransform[T, U](bij.apply(_), bij.invert(_))
 }
+
--- a/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
@@ -1,0 +1,24 @@
+/*
+Copyright 2014 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+package com.twitter.scalding
+
+import com.twitter.scalding.serialization.OrderedSerialization
+import com.twitter.bijection.Bijection
+
+object BijectedOrderedSerialization {
+  implicit def fromBijection[T, U](implicit bij: Bijection[T, U], ordSer: OrderedSerialization[U]) =
+    OrderedSerialization.viaTransform[T, U](bij.apply(_), bij.invert(_))
+}
--- a/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.scalding
 
 import com.twitter.scalding.serialization.OrderedSerialization
-import com.twitter.bijection.ImplicitBijection
+import com.twitter.bijection.{ImplicitBijection, Injection}
 
 object BijectedOrderedSerialization {
   implicit def fromBijection[T, U](implicit bij: ImplicitBijection[T, U], ordSer: OrderedSerialization[U]) =
--- a/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
@@ -22,7 +22,7 @@ object BijectedOrderedSerialization {
   implicit def fromBijection[T, U](implicit bij: ImplicitBijection[T, U], ordSer: OrderedSerialization[U]) =
     OrderedSerialization.viaTransform[T, U](bij.apply(_), bij.invert(_))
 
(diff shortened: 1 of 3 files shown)


scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -101,6 +101,30 @@ object OrderedSerialization {
     case NonFatal(e) => CompareFailure(e)
   }
 
+  def viaTransform[T, U](
+    packFn: T => U,
+    unpackFn: U => T)(implicit otherOrdSer: OrderedSerialization[U]): OrderedSerialization[T] =
+    new OrderedSerialization[T] {
```
rubanm: Should this be a new type like `TransformedOrderedSerialization`? Might be useful for optimizing serialization chains, or maybe this is not applicable.

rubanm:
LGTM

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
ianoc: I'm not sure how we could use that feature right now? we could always add that in later though if we find a use

scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
```diff
@@ -0,0 +1,24 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+package com.twitter.scalding
+
+import com.twitter.scalding.serialization.OrderedSerialization
+import com.twitter.bijection.Bijection
+
+object BijectedOrderedSerialization {
+  implicit def fromBijection[T, U](implicit bij: Bijection[T, U], ordSer: OrderedSerialization[U]) =
```
johnynek: when taking an implicit Bijection, we need to use the `ImplicitBijection` type because it handles the fact that `Bijection[U, T]` would also work here.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -101,6 +101,30 @@ object OrderedSerialization {
     case NonFatal(e) => CompareFailure(e)
   }
 
+  def viaTransform[T, U](
```
johnynek: what about a variant that takes `U => Try[T]` for the deserialization side? For instance, cases where you serialize to strings (Json for instance), but clearly that can sometimes fail if we are reading random strings.

ianoc: Sure, can do

ianoc: Think it probably makes sense to just make this the base/normal variant. We only hit the Try when we are doing deserialization, wherein we have a try anyway. Not sure there is likely much overhead to another allocation for this path?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -101,6 +101,30 @@ object OrderedSerialization {
...
+    packFn: T => U,
+    unpackFn: U => T)(implicit otherOrdSer: OrderedSerialization[U]): OrderedSerialization[T] =
+    new OrderedSerialization[T] {
+
+      override def hash(t: T) = otherOrdSer.hash(packFn(t))
```
johnynek: I wonder if we should implement a simple cache since we might call write, hash, and dynamicSize in quick succession.

Something like:
```scala
private[this] var cache: (T, U) = null
private def packCache(t: T): U = {
  val readCache = cache
  if (null == readCache || !otherOrdSer.equiv(readCache._1, t)) {
    val u = packFn(t)
    cache = (t, u)
    u
  }
  else {
    readCache._2
  }
}
```

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
ianoc: Actually avoiding allocations and nerding in these paths makes sense, i'll just add it and the Injection one in scalding core

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
ianoc: Thats not thread safe right? And we can't use the equiv from the other serialization since we'll need to compare the pre-transformed version ?

scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
```diff
@@ -0,0 +1,28 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+package com.twitter.scalding
+
+import com.twitter.scalding.serialization.OrderedSerialization
+import com.twitter.bijection.ImplicitBijection
+
+object BijectedOrderedSerialization {
```
MansurAshraf: Looks good! 

Thinking out loud about usage in a scalding job, user will at the very least need to know that she has a type that does not have OrdSer and there is a Bijection from that Type to another Type for which OrdSer does exist and will have to import that bijection. We are slightly moving away from AutoMagically making this work by importing 'RequiredBinaryComparator' but not sure if there is a better way? We should probably improve our error message when we encounter a Type with no OrdSer and nudge user toward this approach.

ianoc: Its only ever going to be best effort in the default case I think, if you have a complex JVM type this isn't a good solution either. Though there is a reasonable notion that for new jobs we should force them to re-write in compatible types for the performance improvements... but hard to put in an error message that should write an injection. They probably should write an ordered serialization I suppose... Its a reasonably advanced thing to have to deal with

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
johnynek: I think it is safe: we do one read (which may be old). Then that one read has linked two values: `(T, U)`. If the `T` matches, you're safe to use the `U` no matter what.

If you are out of date, the worst case is you recompute needlessly, right?

You're right about the `equiv`. Mistake there. That compares on `U` obviously. We could use `==` here and assume that if `a == b` that implies `equiv(a, b)`, seems sane (that's not saying they are the same, just that if equals returns true, then equiv would as well. Can't think of a good counter example to that).

johnynek:
LGTM, merge when green.

scalding-core/src/main/scala/com/twitter/scalding/BijectedOrderedSerialization.scala
```diff
@@ -0,0 +1,28 @@
...
+  implicit def fromBijection[T, U](implicit bij: ImplicitBijection[T, U], ordSer: OrderedSerialization[U]) =
+    OrderedSerialization.viaTransform[T, U](bij.apply(_), bij.invert(_))
+
+  implicit def fromInjection[T, U](implicit bij: Injection[T, U], ordSer: OrderedSerialization[U]) =
```
johnynek: looks like you did not import Injection. Maybe even more coffee? :)

ianoc: In my meager defense i got sidetracked and didn't drink the extra coffee yet ;)

ianoc:
Fixes #1326

PR #1334, opened 2015-06-24
Execution id code

ianoc:


johnynek:
@ianoc I added on to this to simplify how naming was done. By doing so, I removed a place we were using setName, and went ahead and removed some methods we had deprecated for about a year (and I doubt were ever used in the first place).

ianoc:
lgtm, i guess we should test it somehow to ensure it operates as we expect

johnynek:
merge when green.

PR #1335, opened 2015-06-25
Add line numbers at .group and .toPipe boundaries

johnynek:
1. Adds line numbers into the jobConf for the .group and .write methods.
2. Gives the above descriptions in the error message if OrderedSerialization checking fails. This should make it easy to find which group is failing to get an OrderedSerialization.

/cc @ianoc

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/LineNumber.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/LineNumber.scala
@@ -1,0 +1,73 @@
+/*
+Copyright 2015 Twitter, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding
+
+object LineNumber {
+  /**
+   * depth 0 means the StackTraceElement for the caller
+   * of this method (skipping getCurrent and the Thread.currentThread
+   */
+  def getCurrent(depth: Int): StackTraceElement =
+    getCurrent(depth, Thread.currentThread().getStackTrace)
+
+  private def getCurrent(depth: Int, stack: Seq[StackTraceElement]): StackTraceElement =
+    stack(depth + 2)
+
+  def ignorePath(classPrefix: String): Option[StackTraceElement] =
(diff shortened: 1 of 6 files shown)


scalding-core/src/main/scala/com/twitter/scalding/LineNumber.scala
```diff
@@ -0,0 +1,73 @@
...
+  def getCurrent(depth: Int): StackTraceElement =
+    getCurrent(depth, Thread.currentThread().getStackTrace)
+
+  private def getCurrent(depth: Int, stack: Seq[StackTraceElement]): StackTraceElement =
+    stack(depth + 2)
+
+  def ignorePath(classPrefix: String): Option[StackTraceElement] =
+    ignorePath(classPrefix, Thread.currentThread().getStackTrace)
+
+  private def ignorePath(classPrefix: String, stack: Seq[StackTraceElement]): Option[StackTraceElement] =
```
ianoc: prefer private[this] over private unless you need to share with a class

johnynek:
Okay, this looks like like a good compromise to me. You don't get line numbers of every function, but you get a minimally invasive change that gives you line numbers of the groups and toPipes, which include the .writes. This should be enough to deduce where you are in your graph on a job.

Moreover, the OrderedSerialization exception includes the line numbers of the groups that are missing ordered serialization.

I think this is a good step forward for the next release.

ianoc:
lgtm merge when green i think

ianoc:
@johnynek FYI rebased this onto latest develop so it should pickup the build fixes

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -337,7 +346,19 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest
...
+            // should see the .group and the .write show up as line numbers
+            "com.twitter.scalding.platform.TypedPipeWithDescriptionJob.<init>(PlatformTest.scala:215)",
+            "com.twitter.scalding.platform.TypedPipeWithDescriptionJob.<init>(PlatformTest.scala:211)")
+
+          val foundDescs = steps.map(_.getConfig.get(Config.StepDescriptions))
+          descs.foreach { d =>
+            assert(foundDescs.size == 1)
+            assert(foundDescs(0).contains(d))
+          }
+          //steps.map(_.getConfig.get(Config.StepDescriptions)).foreach(s => info(s))
```
ianoc: kill commented code

johnynek: I had to look up how to get a println here, since scalatest hides them. Do you really want to drop this or can we leave it for when someone wants to inspect the actual output of the descriptions?

ianoc: Nah, if you have a reason to have it here that'll help future debugging i'm cool with keeping.

PR #1338, opened 2015-06-26
Ordered Serialization macros for thrift 

MansurAshraf:
continuation of https://github.com/twitter/scalding/pull/1210


scalding-thrift-macros-fixtures/src/main/scala/com/twitter/scalding/thrift/macros/scalathrift/TestEnum.scala
```diff
@@ -0,0 +1,94 @@
+/**
+ * Generated by Scrooge
+ *   version: ?
+ *   rev: ?
+ *   built at: ?
+ */
+package com.twitter.scalding.thrift.macros.scalathrift
```
johnynek: is there any way we could not check in generated code?

MansurAshraf: @ianoc Do you still have the original thrift file? We will have to generate these on the fly if we dont want to check them in.

johnynek: I'd rather just use the scrooge plugin and generate these, personally (also it is a better test, because scrooge will evolve).

scalding-thrift-macros/src/main/scala/com/twitter/scalding/thrift/macros/RequiredBinaryComparators.scala
```diff
@@ -0,0 +1,12 @@
+package com.twitter.scalding.thrift.macros
+
+import com.twitter.scalding.serialization.{ OrderedSerialization, RequiredBinaryComparatorsConfig }
+import com.twitter.scalding.thrift.macros.impl.ScroogeInternalOrderedSerializationImpl
+import scala.language.experimental.{ macros => smacros }
+
+/**
+ * @author Mansur Ashraf.
+ */
+trait RequiredBinaryComparators extends RequiredBinaryComparatorsConfig {
```
johnynek: Can you add a comment to make it clear that this is here for people that use scrooge, but otherwise is identical to the one in scalding-core.

johnynek:
Looks good to me other than the minor confusion I hit when I saw the duplicated RequiredBinaryComparators.

MansurAshraf:
@johnynek let me know if you are okay to merge

PR #1342, opened 2015-06-27
make some repl components extensible

rubanm:
Allows for setting custom implicits and config to load at boot time.


scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
```diff
@@ -77,14 +82,16 @@ class ScaldingILoop
    */
   override def commands: List[LoopCommand] = super.commands ++ scaldingCommands
 
+  protected def shellImports: List[String] = List(
```
johnynek: could this be moved to BaseScaldingShell so there is only one place to configure if it makes sense?

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
```diff
@@ -54,7 +54,12 @@ class ScaldingILoop
    *
    * @return a prompt string to use for this REPL.
    */
-  override def prompt: String = ScaldingShell.prompt()
+  override def prompt: String = shell.prompt()
```
johnynek: could this be in BaseScaldingShell? Seems to make sense there too.

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -77,14 +77,15 @@ object ReplImplicits extends FieldConversions {
   /* Using getter/setter here lets us get the correct defaults set by the mode
      (and read from command-line, etc) while still allowing the user to customize it */
   def config: Config = Config.defaultFrom(mode) ++ customConfig
-  def config_=(c: Config) { customConfig = c }
+
+  protected def shell: BaseScaldingShell = ScaldingShell
```
johnynek: seems strange to have this called BaseReplImplicits. The implicit part is about the imports. What is the core function here? If we are refactoring, I'd like to improve it, seems we make be making it worse here.

johnynek:
If we could have some trait like `ShellDefaults` or something that has all the default constants, that seems like it would solve most of the issues, then maybe we wouldn't need to bolt them onto other code here. Would that work?

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -178,7 +183,8 @@ object ReplImplicits extends FieldConversions {
    * @param source to convert to a Pipe.
    * @return a Pipe that is the result of reading the specified Source.
    */
-  implicit def sourceToPipe(source: Source): Pipe = source.read(flowDef, mode)
+  implicit def sourceToPipe(source: Source)(implicit flowDef: FlowDef, mode: Mode): Pipe =
```
johnynek: I really don't think we need this. I think it was just cargo culted from Dsl by the (wibidata folks that started the REPL).

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingShell.scala
```diff
@@ -60,11 +70,11 @@ object ScaldingShell extends MainGenericRunner {
    */
   override def process(args: Array[String]): Boolean = {
     // Get the mode (hdfs or local), and initialize the configuration
-    val (mode, jobArgs) = parseModeArgs(args)
+    val (cfg, mode) = parseModeArgs(args)
 
     // Process command line arguments into a settings object, and use that to start the REPL.
     // We ignore params we don't care about - hence error function is empty
-    val command = new GenericRunnerCommand(jobArgs.toList, _ => ())
+    val command = new GenericRunnerCommand(List[String](), _ => ())
```
johnynek: seems we are changing the way this works. What's the story?

johnynek:
if the args change makes sense, I'm happy to merge.

rubanm:
I made a few more changes and moved the prompt & imports to `ScaldingILoop`, so it no longer needs a reference to `ScaldingShell`.

The main issue is that `ScaldingShell` needs a reference to `ReplState` to mutate it (setting the mode and custom config based on command line args), and `ReplState` needs `ScaldingShell` to build the jars correctly for executing flows.

I've made `ShellPipe` take in the `ReplState` for execution. This state can be customized with different defaults. Still not completely clean but hopefully doesn't make it worse.

johnynek:
I think this is an improvement.

I'm happy to merge this. @ianoc want to merge if you think this looks good?

ianoc:
LGTM, merged

PR #1344, opened 2015-06-29
add useNonStrictHdfsMode in repl

rubanm:


johnynek:
Looks fine to me, but what was the demand for the non-strict mode (something I think may have been a design flaw in scalding, not sure why you would want to run the same job in a way that ignore missing data and one that does not).

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -47,21 +47,27 @@ object ReplImplicits extends FieldConversions {
...
+
+  def useNonStrictHdfsMode() {
+    useHdfsMode_(false)
     customConfig -= mr1Key
     customConfig -= mr2Key
   }
 
   def useHdfsLocalMode() {
-    useHdfsMode_()
+    useHdfsMode_(true)
```
gerashegalov: We probably need to allow non-strict mode with HdfsLocal as well.

rubanm:
@johnynek Internally someone wanted to run a job against an hourly source with some missing hours. Agree this is not ideal for production. Might be useful in the repl for adhoc stuff though where the user is willing to accept the fact that there's missing data. I don't feel strongly though, so we could hold off till this is a frequent ask if at all.

johnynek:
For instance, what about a method like: `NonValidating(MySource)` which is implemented by:

```scala
object NonValidating {
  def apply[T](s: Source with TypedSource[T]): Source with TypedSource[T] = {
    // make a proxy to s except that def validateTaps = {}
  }
}
```
Seems janky to turn it on or off for the whole job, when you think about it, right?

ianoc:
Thats no validation, i guess maybe ok, but non-strict is different to that though?

johnynek:
Good point, but you might be able to do something similar assuming you have a FileSource.

again, it's just really weird for this to a Mode setting. It's more about the source, right?

PR #1348, opened 2015-07-01
Implement flatMapValues method

danosipov:
Sometimes its useful to have flatMapValues, similar to mapValues, but flatter.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
@@ -17,6 +17,7 @@ package com.twitter.scalding.typed
 
 import java.io.Serializable
 import java.util.PriorityQueue
+import scala.collection.GenTraversableOnce
 import scala.collection.JavaConverters._
 
 import com.twitter.algebird.{ Fold, Semigroup, Ring, Aggregator }
@@ -172,6 +173,13 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
    */
   def mapValues[V](fn: T => V): This[K, V] =
     mapGroup { (_, iter) => iter.map(fn) }
+
+  /**
+   * Similar to mapValues, but works like flatMap, returning a collection of outputs
+   * for each value input.
+   */
+  def flatMapValues[V](fn: T => GenTraversableOnce[V]): This[K, V] =
+    mapGroup { (_, iter) => iter.flatMap(fn) }
 
   /**
    * Use this when you don't care about the key for the group,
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -174,6 +175,13 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
     mapGroup { (_, iter) => iter.map(fn) }
 
   /**
+   * Similar to mapValues, but works like flatMap, returning a collection of outputs
+   * for each value input.
+   */
+  def flatMapValues[V](fn: T => GenTraversableOnce[V]): This[K, V] =
```
johnynek: we generally use `TraversableOnce[V]` here, can we stick to that style?

johnynek:
can you add the same method to `TypedPipe`? we try to add all methods that make sense on both to both. Something like:

```scala
def flatMapValues[K, V1, V2](fn: V1 => TraversableOnce[V2])(implicit ev: T <:< (K, V1)): TypedPipe[(K, V2)] =
  raiseTo[K, V1].flatMap { case (k, v1) => fn(v1).map { v2 => k -> v2 } } 
```

johnynek:
Thanks for adding this. You're right, it comes up, and summingbird added it a long time ago.

danosipov:
@johnynek Added the method to TypedPipe, but there appear to be no tests for it. It type checks though :)

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -300,6 +302,10 @@ trait TypedPipe[+T] extends Serializable {
   def mapValues[K, V, U](f: V => U)(implicit ev: T <:< (K, V)): TypedPipe[(K, U)] =
     raiseTo[(K, V)].map { case (k, v) => (k, f(v)) }
 
+  /** Similar to mapValues, but allows to return a collection of outputs for each input value */
+  def flatMapValues[K, V, U](f: V => TraversableOnce[U])(implicit ev: T <:< (K, V)): TypedPipe[(K, TraversableOnce[U])] =
```
johnynek: I think we want `TypedPipe[(K, U)]` to be the return type. In other words, you have another copy of mapValues here, which you can use if you don't want the flattening, but I think this one should behave exactly as it does on KeyedListLike.

danosipov: But if I use `fn(v1).map { v2 => k -> v2 }` This returns a Traversable[U] (or Traversable[V2] from example above)

johnynek: look at my code. It had a flatMap yours does not have.

danosipov: :dizzy_face: Of course, fixed

johnynek:
Thanks! This is useful.

PR #1355, opened 2015-07-07
Merge scalding-macros into scalding-core

sid-kap:
This change is in preparation for adding TypedText (which depends on
scalding macros) to scalding-core.


project/Build.scala
```diff
@@ -475,16 +477,6 @@ object ScaldingBuild extends Build {
...
 
-  lazy val scaldingMacros = module("macros").settings(
-    libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
-      "org.scala-lang" % "scala-library" % scalaVersion,
-      "org.scala-lang" % "scala-reflect" % scalaVersion,
-      "com.twitter" %% "bijection-macros" % bijectionVersion
-    ) ++ (if(isScala210x(scalaVersion)) Seq("org.scalamacros" %% "quasiquotes" % quasiquotesVersion) else Seq())
-  },
-  addCompilerPlugin("org.scalamacros" % "paradise" % paradiseVersion cross CrossVersion.full)
-  ).dependsOn(scaldingCore, scaldingHadoopTest % "test")
```
sid-kap: scaldingMacros used to depend on scaldingHadoopTest for some reason.
I didn't copy this dependency over to scaldingCore because that would create a circular dependency (scaldingHadoopTest depends on scaldingCore). This didn't cause any problems. So maybe the dependency on scaldingHadoopTest was unnecessary?

ianoc: See if the tests all work here? We might need to move a few tests from the macros if it has platform tests over to the scaldingHadoopTest test suite.

PR #1356, opened 2015-07-07
Migrate typedtext

sid-kap:
Add TypedText and TypedTextTest


sid-kap:
Whoops, I haven't made TypedTsv an alias to TypedText yet.

project/Build.scala
```diff
@@ -303,6 +303,7 @@ object ScaldingBuild extends Build {
       "com.twitter" %% "bijection-macros" % bijectionVersion,
       "com.twitter" %% "chill" % chillVersion,
       "com.twitter" %% "chill-algebird" % chillVersion,
+      "com.twitter.elephantbird" % "elephant-bird-cascading2" % elephantbirdVersion,
```
johnynek: nope. We don't want to make this a dependency of scalding-core. It is a dependency of scalding-commons. This means you should move the LZO related text stuff into scalding-commons, but the other stuff can stay here.

scalding-core/src/main/scala/com/twitter/scalding/source/typedtext/TypedText.scala
```diff
@@ -0,0 +1,169 @@
...
+  private val ONE = "\1"
+  private val COMMA = ","
+
+  /*
+   * To use these, you will generally want to
+   * import com.twitter.scalding_internal.source.typedtext._
+   * to get the implicit TypedDescriptor.
+   * Then use TypedText.lzoTzv[MyCaseClass]("path")
+   */
+  def lzoTsv[T: TypeDescriptor](path: String): TypedTextDelimited[T] = new FixedLzoTypedText[T](path, TAB)
```
johnynek: let's move these to LzoTypedText in scalding-commons. All the non-lzo can stay.

scalding-core/src/test/scala/com/twitter/scalding/source/typedtext/TypedTextTest.scala
```diff
@@ -0,0 +1,24 @@
...
+case class Test1(a: Int, b: Long, c: Option[Double])
+case class Test2(one: Test1, d: String)
+
+class TypedTextTest extends FunSuite {
+  test("Test with a flat tuple") {
+    val source = TypedText.tsv[Test1]("myPath")
+    assert(source.sourceFields.size == 3)
+  }
+  test("Test with a nested tuple") {
+    val source = TypedText.tsv[Test2]("myPath")
```
johnynek: does this work with a tuple `(Int, Int)` or a raw type like `String`? We need that if we are going to replace `TypedTsv`.

sid-kap: Would it be sufficient to add more tests exactly like these two, but with tuples and strings instead?

I was trying to write a test that reads in a TSV and checks that it read in the correct values. The only way I can think of doing that is by converting the `TypedPipe` to a list and checking for equality with the expected value.
However, it looks like I can't do that because the function to convert a pipe to an iterator/list is in `scalding-repl`.
Do you know of another way to test this?

ianoc: It would be good to just write out with those types (Int, Int), and String. Manually verify to yourself once. But then in the test you can read them back into matching case classes to see things round trip as you expect. Or for String cases use the TextLine source and it should be the same then too

sid-kap:
Sorry about the git ugliness above.
I moved the LZO stuff to scalding-commons as you suggested.
I still haven't figured out how to test TypedText.

johnynek:
Looks like #1357 is incorporated now, is that right? Can we close that one out?

sid-kap:
Yep

scalding-core/src/main/scala/com/twitter/package.scala
```diff
@@ -31,6 +32,8 @@ package object scalding {
   type ValuePipe[+T] = com.twitter.scalding.typed.ValuePipe[T]
   type Grouped[K, +V] = com.twitter.scalding.typed.Grouped[K, V]
 
+  implicit def typedDescriptor[T]: TypeDescriptor[T] = macro com.twitter.scalding.macros.impl.TypeDescriptorProviderImpl.caseClassTypeDescriptorImpl[T]
```
johnynek: I'm a bit nervous about this. a lot of people do `import com.twitter.scalding._`. If they also import an implicit for TypedDescriptor (as our internal code has been doing) they will now have a compile error (scala won't find either implicit, I'm pretty sure).

Maybe we want that, but I'm a bit skeptical that it is what we want. I think an explicit import might not be a bad idea, personally.

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -47,15 +48,15 @@ trait TypedSeperatedFile extends Serializable {
 /**
  * Typed tab separated values file
  */
-object TypedTsv extends TypedSeperatedFile {
-  val separator = "\t"
+object TypedTsv {
+  def apply[T: TypeDescriptor](path: String*) = TypedText.tsv[T](path: _*)
```
johnynek: let's add a return type to all public methods.

scalding-core/src/main/scala/com/twitter/package.scala
johnynek: I guess if we don't do this we can't replace TypedTsv, right?

What if we put this method in the `object TypedDescriptor`? I think then it would be found by default if not explicitly overridden by another import.

scalding-core/src/main/scala/com/twitter/scalding/bdd/TBddDsl.scala
```diff
@@ -68,7 +68,7 @@ trait TBddDsl extends FieldConversions with TypedPipeOperationsConversions {
 
       val outputPipe = operation(inputPipes) map { Tuple1(_) }
 
-      outputPipe.write(TypedTsv[Tuple1[OutputType]]("output"))
+      outputPipe.write(FixedPathTypedDelimited[Tuple1[OutputType]]("output", "\t"))
```
johnynek: do we need to make this change? Ideally, we don't need to make any change because we don't want to break a ton of code.

scalding-core/src/main/scala/com/twitter/package.scala
ianoc: Yes this is there for the TypedTsv, it was in the package for typedtext. The companion object might/should work I think...

scalding-core/src/main/scala/com/twitter/scalding/bdd/TBddDsl.scala
ianoc: This change is made just because this code is in scalding core so it can't use the macros

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/TupleConverterImpl.scala
```diff
@@ -37,122 +37,185 @@ object TupleConverterImpl {
...
-          val cachedResult = newTermName(c.fresh(s"opti"))
-          val extractorTypeVal: Tree = if (extractor.tpe =:= innerType)
-            extractor.toTree
-          else
-            q"${innerType.typeSymbol.companionSymbol}.unbox($extractor)"
-
-          val build = Builder(q"""
+        outerTpe match {
+          case tpe if tpe =:= typeOf[String] => getPrimitive(q"""t.getString(${idx})""", typeOf[java.lang.String], None)
+          case tpe if tpe =:= typeOf[Boolean] => getPrimitive(q"""t.getBoolean(${idx})""", typeOf[java.lang.Boolean], Some(q"_root_.java.lang.Boolean.valueOf"))
```
johnynek: we could support `Char`, which `t.getChar` should work. For whatever reason, `Byte` is not supported by cascading.

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformJobTest.scala
```diff
@@ -43,7 +43,7 @@ case class HadoopPlatformJobTest(
 
   def arg(inArg: String, value: String): HadoopPlatformJobTest = arg(inArg, List(value))
 
-  def source[T: Manifest](location: String, data: Seq[T]): HadoopPlatformJobTest = source(TypedTsv[T](location), data)
+  def source[T: Manifest](location: String, data: Seq[T])(implicit t: TypeDescriptor[T]): HadoopPlatformJobTest = source(TypedTsv[T](location), data)
```
johnynek: let's write this as: `T: Manifest: TypeDescriptor` likewise below.

johnynek:
why are we moving code out of examples into tutorial? That means they won't be in the jars and some people may be using this code. Any reason to do that?

If it is a pain to update with this change, that means it is going to be VERY hard to ship this change internal to Twitter. The goal here is a very easy (even trivial) update to get better type-safety and performance.

scalding-core/src/main/scala/com/twitter/scalding/bdd/TBddDsl.scala
johnynek: add, this is not in the tests itself. Same package. Got it.

sid-kap:
PageRank and WeightedPageRank use TypedTSV, which require the implicit typeDescriptor macro.

Since the macro is in scalding-core, we can't keep any code that depends on the macro in scalding-core.

sid-kap:
Sure, we can move the examples to scalding-commons

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
sid-kap: Fixed in latest commit.

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformJobTest.scala
sid-kap: Fixed in latest commit

scalding-core/src/main/scala/com/twitter/package.scala
sid-kap: I moved the macro to the companion object in d4a99e5

sid-kap:
Okay, I think I've included all of the suggested changes.

This would be ready to merge, but several Hadoop tests are failing. I don't know what the errors mean.

scalding-core/src/main/scala/com/twitter/package.scala
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 package com.twitter
 
 import org.apache.hadoop.fs.{ Path, PathFilter }
+import scala.language.experimental.{ macros => sMacros }
```
johnynek: still needed?

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -47,15 +48,15 @@ trait TypedSeperatedFile extends Serializable {
 /**
  * Typed tab separated values file
  */
-object TypedTsv extends TypedSeperatedFile {
```
johnynek: Can we instead just override the methods in TypedSeparatedFile for maximum source code compatibility?

I think we could pretty easily implement the 4 apply methods there, then we are fairly sure this change won't break anyone, which is nice.

scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
```diff
@@ -44,7 +44,7 @@ class TupleAdderTest extends WordSpec with Matchers {
   import Dsl._
   "A TupleAdderJob" should {
     JobTest(new TupleAdderJob(_))
-      .source(TypedTsv[(String, String)]("input", ('a, 'b)), List(("a", "a"), ("b", "b")))
```
johnynek: let's not make this change. This will appear in a code in science and elsewhere, and if it is easy for us to just support it, we should.

scalding-core/src/main/scala/com/twitter/scalding/source/typedtext/TypedText.scala
```diff
@@ -0,0 +1,108 @@
+package com.twitter.scalding.source.typedtext
```
johnynek: maybe this should not be in an extra level of package. Just com.twitter.scalding.source seems to be more consistent with what is already there.

johnynek:
I think we should make `TypedDelimited` and `FixedPathTypedDelimited` deprecated. Someone may have subclassed them in their own code, but we don't have to use them anymore with this change.

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformJobTest.scala
```diff
@@ -43,7 +43,7 @@ case class HadoopPlatformJobTest(
 
   def arg(inArg: String, value: String): HadoopPlatformJobTest = arg(inArg, List(value))
 
-  def source[T: Manifest](location: String, data: Seq[T]): HadoopPlatformJobTest = source(TypedTsv[T](location), data)
+  def source[T: Manifest: TypeDescriptor](location: String, data: Seq[T]): HadoopPlatformJobTest = source(TypedTsv[T](location), data)
```
johnynek: do we need the Manifest still? I don't think so.

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformJobTest.scala
```diff
@@ -52,7 +52,7 @@ case class HadoopPlatformJobTest(
       }
     })
 
-  def sink[T: Manifest](location: String)(toExpect: Seq[T] => Unit): HadoopPlatformJobTest =
+  def sink[T: Manifest: TypeDescriptor](location: String)(toExpect: Seq[T] => Unit): HadoopPlatformJobTest =
```
johnynek: don't think Manifest is needed.

scalding-core/src/main/scala/com/twitter/scalding/TypeDescriptor.scala
```diff
@@ -30,3 +31,6 @@ trait TypeDescriptor[T] extends java.io.Serializable {
   def converter: TupleConverter[T]
   def fields: Fields
 }
+object TypeDescriptor {
+  implicit def typeDescriptor[T]: TypeDescriptor[T] = macro com.twitter.scalding.macros.impl.TypeDescriptorProviderImpl.caseClassTypeDescriptorImpl[T]
```
johnynek: Can we add a pretty beefy comment here. Something like

```scala
/**
 * This type descriptor flattens tuples and case classes left to right, depth first. It supports supports any type T where T is Boolean, String, Short, Int, Long, Float or Double, or an Option of these, or a tuple of a supported type. So, ((Int, Int), Int) is supported, and is flattened into a length 3 cascading Tuple/Fields. ((Int, Int), (Int, Int)) would be a length 4 cascading tuple, similarly with case classes.
* Note, the Fields types are populated at the end of this with the exception that Option[T] is recorded as Object (since recording it as the java type would have different consequences for cascading's null handling.
```

scalding-core/src/main/scala/com/twitter/package.scala
sid-kap: Nope. I'll remove it

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
sid-kap: Would we just discard the Fields parameter and call `new FixedTypedText[t](seperator, paths)`?

scalding-core/src/main/scala/com/twitter/scalding/source/typedtext/TypedText.scala
sid-kap: Sure, I'll do this

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformJobTest.scala
sid-kap: I'll remove it

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformJobTest.scala
sid-kap: I'll remove it

scalding-core/src/main/scala/com/twitter/scalding/TypeDescriptor.scala
sid-kap: Sure

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
johnynek: we could take the fields passed in, require that they have the same size as the fields from the TypedDescriptor, then make a new TypeDescriptor that takes just the names from the input fields (but copies the types out).

sid-kap: So I'm thinking we could add a new macro (not implicit) in TypeDescriptorProviderImpl which is like caseClassTypeDescriptorCommonImpl, but which takes fields as a parameter and uses that value of fields, instead of using the helper macro.

We would then call this macro from the `apply` methods of TypedSeparatedFile which specify a fields parameter.

ianoc: You won't have the type information there. You probably want to take the
implicit as is but maybe copy it with fields changed.

johnynek: I agree with Ian here. We try to only use macros when we really have to. We should be able to solve this problem without a macro, since we already have a macro for the more general case (which we need to get the `Class[T]` for each column of the flattened type).

sid-kap: So something like `val td = implicitly[TypeDescriptor[T]]; val newTd = td.copy(fields = f);` would work?

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -47,15 +48,15 @@ trait TypedSeperatedFile extends Serializable {
...
-  val separator = "\t"
+object TypedTsv {
+  def apply[T: TypeDescriptor](path: String*): TypedTextDelimited[T] = TypedText.tsv[T](path: _*)
 }
 
 /**
  * Typed comma separated values file
  */
-object TypedCsv extends TypedSeperatedFile {
```
sid-kap: top lel, every instance of `TypedSeperatedFile` in our code base spells the word "separated" wrong!

johnynek: :) too bad we don't have smart spell check for code.

Let's leave it as is for compatibility.

On Friday, July 10, 2015, Sidharth Kapur <notifications@github.com> wrote:

> In scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
> <https://github.com/twitter/scalding/pull/1356#discussion_r34380731>:
>
> >  }
> >
> >  /**
> >   * Typed comma separated values file
> >   */
> > -object TypedCsv extends TypedSeperatedFile {
>
> top lel, every instance of TypedSeperatedFile in our code base spells the
> word "separated" wrong!
>
> —
> <https://github.com/twitter/scalding/pull/1356/files#r34380731>.
>


-- 

scalding-core/src/main/scala/com/twitter/scalding/bdd/TBddDsl.scala
sid-kap: In the process of adding Fields support to the apply methods in TypedSeperatedFile, I made both `TypedSeperatedFile.apply` and `FixedPathTypedDelimited.apply` depend on the macro and use `FixedTypedText` instead of `FixedPathTypedDelimited`. I wanted to do this so that we could eventually get rid of the old `FixedPathTypedDelimited` class, since it seems like `FixedTypedText` and `FixedPathTypedDelimited` perform the same function.

Unfortunately, my change broke this code again by making it depend on the macro. I see two work-arounds:
1) Reverse the changes I just made.
2) Change `FixedPathTypedDelimited` here to `new FixedPathTypedDelimited`, which doesn't require an implicit `TypeDescriptor`. The downside of this approach is that we won't be able to kill the old `FixedPathTypedDelimited` class.
3) Move this class to scalding-commons.

I think the decision comes down to whether we care to keep the old `FixedPathTypedDelimited` class.

johnynek:
doesn't seem to compile as is.

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -29,19 +32,26 @@ trait TypedSeperatedFile extends Serializable {
...
 
-  def apply[T: Manifest: TupleConverter: TupleSetter](paths: Seq[String], f: Fields): FixedPathTypedDelimited[T] =
-    new FixedPathTypedDelimited[T](paths, f, skipHeader, writeHeader, separator)
+  def apply[T: TypeDescriptor](paths: Seq[String], f: Fields): TypedTextDelimited[T] = {
+    val td = implicitly[TypeDescriptor[T]]
+    require(f.size == td.fields.size)
+    val newTd = new TypeDescriptor[T] {
+      def setter = td.setter
+      def converter = td.converter
+      def fields = f
```
johnynek: see: https://github.com/Cascading/cascading/blob/3.0/cascading-core/src/main/java/cascading/tuple/Fields.java

I think we want:

```scala
val newFields = new Fields(f.get, td.fields.getTypes)
```
we don't want to lose the types, just keep the names.

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -73,19 +83,26 @@ object TypedOsv extends TypedSeperatedFile {
...
+    new FixedTypedText(source.TypedSep(separator), paths: _*)
   }
 
-  def apply[T: Manifest: TupleConverter: TupleSetter](path: String, f: Fields, separator: String): FixedPathTypedDelimited[T] =
+  def apply[T: TypeDescriptor](path: String, f: Fields, separator: String): TypedTextDelimited[T] =
     apply(Seq(path), f, separator)
 
-  def apply[T: Manifest: TupleConverter: TupleSetter](paths: Seq[String], f: Fields, separator: String): FixedPathTypedDelimited[T] =
-    new FixedPathTypedDelimited[T](paths, f, false, false, separator)
+  def apply[T: TypeDescriptor](paths: Seq[String], f: Fields, separator: String): TypedTextDelimited[T] = {
```
johnynek: wait, we should not duplicate this code with the above. Can't the code above just be:
```scala
apply[T: TypedDescriptor](path: Seq[String], f: Fields): TypedTextDelimited[T] = apply[T](path, f, separator)
```

scalding-core/src/main/scala/com/twitter/scalding/bdd/TBddDsl.scala
johnynek: Actually, I think the easiest fix is to define a simple mock `TypeDescriptor` since this code actually is not using any features of the macro. Then we shouldn't hit the macro path:
```scala
implicit val td: TypeDescriptor[Tuple1[OutputType]] = new TypeDescriptor[Tuple1[OutputType]] {
  def converter = TupleConverter.singleConverter
  def setter = TupleSetter.singleSetter // I think these names are close. ;)
  def fields = new Fields("item")
}
```
if this is in scope, the macro should not be hit and we'll be okay. Since this source is not written to disk, it doesn't matter.

PS: I'm not sure anyone uses this test framework, and I'm a bit unsure we should keep it around, but let's do something quick.

scalding-core/src/main/scala/com/twitter/scalding/source/TypedText.scala
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * This object gives you easy access to text formats (possibly LZO compressed) by
+ * using a case class to describe the field names and types.
+ */
+case class TypedSep(str: String) extends AnyVal
+
+object TypedText {
+
+  private val TAB = TypedSep("\t")
```
johnynek: not sure these should be private.

scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
```diff
@@ -33,7 +33,7 @@ object TUtil {
 
 class TupleAdderJob(args: Args) extends Job(args) {
 
-  TypedTsv[(String, String)]("input", ('a, 'b))
+  TypedTsv[(String, String)]("input")
```
johnynek: let's revert this change to test the version that has Fields as the second parameter.

scalding-core/src/test/scala/com/twitter/scalding/source/TypedTextTest.scala
```diff
@@ -0,0 +1,41 @@
...
+    val source = TypedText.tsv[(Int, Int)]("myPath")
+    assert(source.sourceFields.size == 2)
+  }
+
+  test("Test with a Optional Int") {
+    val source = TypedText.tsv[Option[Int]]("myPath")
+    assert(source.sourceFields.size == 1)
+  }
+
+  test("Test with a Optional Int") {
```
johnynek: incorrect string. Should be `Test with an Int`.

johnynek:
closing in on this. This will be a win since `TypedTsv` will just start being a lot better.

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
sid-kap: This would require having `TypedSeperatedDelimited.apply` call the `FixedPathTypedDelimited.apply` method.

Since we're planning to deprecate FixedPathTypedDelimited and TypedDelimited, this means we'll get deprecation warnings. Is that okay?

scalding-core/src/main/scala/com/twitter/scalding/source/TypedText.scala
sid-kap: Sure, I'll remove the private

johnynek:
there are some real failures here.

Some of them look like there are type mismatches since the strings look the same, it may be that somehow types are winding up as strings that should not be or the tests have them as strings, but they are preserved. 

There are also some type changes (`Some` appearing).

My recommendation is to pick a single test, and run that test locally using `./sbt "scalding-core/test-only *SomeTestName*"` to run the test. Probably we can pretty easily see what is going on here.

One thing may be this: Sources need to have a good equals and hashCode. The TypedText stuff may not have a good equals method. We we should probably compare them based on path and fields, since we can't compare the tupleConverter/tupleSetter.

sid-kap:
Most of the errors look similar to the following error, which is from the PageRankTest:
```
Uncaught exception when running com.twitter.scalding.PageRankTest: 
cascading.flow.planner.PlannerException: could not build flow 
from assembly: [[NullScheme][new PageRankTest() @ 
sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)] 
unable to resolve scheme sink selector: 
[{1}:'element'], with incoming: [{1}:'err']]
```
I think it might be failing when it tries to write a RichPipe to a `TypedTsv[Double]` and then trying to read from the TypedTsv again. It has something to do with the fact that the macros, when given a singleton type, call the field "element", while the field is actually called "err".

I considered overriding RichPipe's write method, but unfortunately RichPipe doesn't have access to the list of fields. I also considered changing the test to use `TypedTsv[Double]("filename", new Fields("err"))` instead of `TypedTsv[Double]("filename")`, but that gives me another error

```
Uncaught exception when running com.twitter.scalding.PageRankTest: 
java.lang.IllegalArgumentException: Failed to create tap for: 
com.twitter.scalding.source.FixedTypedTextList(error), 
with error: given field name was not found: element
```
Does this seem to be on the right track?

johnynek:
It could be the difference in the way cascading handles named vs unnamed fields. We may need to default to using this approach:

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala#L123

which is to say, numbered fields (not named) when we have a primitive value or a tuple, otherwise we might see this breakage with the Fields-API.

I'm not sure how to detect if the case class is a scala tuple other than just having a list of 22 classes and testing specifically. It might be okay to do that, maybe in some big list.

@ianoc any ideas on this?

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
```diff
@@ -80,53 +80,81 @@ object FieldsProviderImpl {
...
-          val innerType = tpe.asInstanceOf[TypeRefApi].args.head
-          matchField(innerType, outerName, fieldName, true)
-        case tpe if IsCaseClassImpl.isCaseClassType(c)(tpe) =>
-          val prefix = outerName.map(pre => s"$pre$fieldName.")
-          expandMethod(tpe, prefix, isOption)
-        case tpe if allowUnknownTypes => simpleRet
-        case _ => c.abort(c.enclosingPosition, s"Case class ${T} is not pure primitives or nested case classes")
+    def maybeHandlePrimitive(outerTpe: Type): Option[List[(Tree, String)]] = {
+
+      def innerLoop(tpe: Type, inOption: Boolean): Option[List[(Tree, String)]] = {
```
johnynek: this is copy-paste, right? the string here is always "element" right? And there is never a list of trees, just a `Option[Tree]` would have been enough, right?

Since we can make `Fields` with `Comparable` which should really be `Either[Int, String]` but `Comparable` is a common superclass of both `java.lang.Integer` and `java.lang.String`, Cascading uses `Comparable` there.

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
```diff
@@ -80,53 +80,81 @@ object FieldsProviderImpl {
...
+      val expanded = expandMethod(T.tpe, prefix, false)
+      if (expanded.isEmpty) c.abort(c.enclosingPosition, s"Case class ${T} has no primitive types we were able to extract")
 
-    val typeTrees = expanded.map(_._1)
+      expanded
+    }
 
+    val typeTrees = expanded.map(_._1)
     val res = if (namingScheme == NamedWithPrefix || namingScheme == NamedNoPrefix) {
       val fieldNames = expanded.map(_._2)
```
johnynek: note this is cast to `Comparable` below, so actually using `Comparable[_]` instead of `String` would have worked, I think.

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
sid-kap: @ianoc

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
johnynek: need a `_root_.java.lang.Comparable` below. Note the missing `_root_`.

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
```diff
@@ -80,53 +80,81 @@ object FieldsProviderImpl {
...
+          case tpe if tpe.erasure =:= typeOf[Option[Any]] =>
+            val innerType = tpe.asInstanceOf[TypeRefApi].args.head
+            innerLoop(innerType, true)
+
+          case tpe => None
+        }
       }
+      innerLoop(outerTpe, false)
     }
+    val expanded: List[(Tree, String)] = maybeHandlePrimitive(T.tpe).getOrElse {
```
johnynek: I think we should have `maybeHandlePrimitive` return `Option[Tree]` then here we can do:

```scala
val expanded: List[(Tree, Comparable[_])]  = maybeHandlePrimitive(T.tpe)
  .map { tree =>
    List((tree, Integer.valueOf(0)))
  }
  .getOrElse { ...
```

sid-kap: Why 0 instead of "element"?

johnynek: see the link to the old TypedTsv code. It used numbers not names and I
think cascading treats numbers differently.

So, if we use 0, we are doing the same thing we were doing before, so
hopefully the tests would pass.

Also, we may need to do the same thing for tuples since there could be
similar failures at runtime for Fields-API users using TypedTsv[(Int, Int)].

On Tuesday, July 14, 2015, Sidharth Kapur <notifications@github.com> wrote:

> In
> scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
> <https://github.com/twitter/scalding/pull/1356#discussion_r34633126>:
>
> >      }
> > +    val expanded: List[(Tree, String)] = maybeHandlePrimitive(T.tpe).getOrElse {
>
> Why 0 instead of "element"?
>
> —
> <https://github.com/twitter/scalding/pull/1356/files#r34633126>.
>


-- 

sid-kap:
Changing "element" to java.lang.Integer.valueOf(0) fixed the PageRankTest, but broke a lot of other stuff.

johnynek:
Let me take a quick stab.

On Wed, Jul 15, 2015 at 2:10 PM, Sidharth Kapur <notifications@github.com>
wrote:

> Changing "element" to java.lang.Integer.valueOf(0) fixed the PageRankTest,
> but broke a lot of other stuff.
>
> —
> <https://github.com/twitter/scalding/pull/1356#issuecomment-121698025>.
>



-- 

sid-kap:
I misspoke. It seems like this change did not introduce new errors.

johnynek:
Fixed the BDD tests. I think what is left is the bug that tuples now get fields `("_1", "_2", "_3")`, for instance, but before it was `new Fields(Array[Comparable[_]](Integer.valueOf(0), Integer.valueOf(1), Integer.valueOf(2))`.

So the macro needs to be updated when the case class is a Tuple.

sid-kap:
I think I know how to do that. The `runtime.ScalaRunTime.isTuple` function should be useful here.

Can you commit your fix to a branch somewhere so that I can merge it in?

johnynek:
@sid-kap can you merge this: https://github.com/sid-kap/scalding/pull/1/files

johnynek:
@sid-kap I don't think `scala.runtime.ScalaRunTime.isTuple` will help. That is a runtime check of a product, not a compile time method that can run in a macro.

I think we might need to do something like checking if the type is `tpe.erasure =:= typeOf[Tuple1[Any]]` or `tpe.erasure =:= typedOf[Tuple2[Any, Any]]` etc... Pain in the ass to have 22 checks, but they run at compile time, so it should not be a runtime cost.

johnynek:
@ianoc any other APIs you know of to handle this?

ianoc:
Nope, not unless you want to use substring on the class name, it should be a fully qualified one... but doing test on the erasure should be more robust

sid-kap:
Changing the implicit in `typedSink` from `TupleConverter` to `TypeDescriptor` brought us down from 6 failing tests to 2.

These 2 tests are still failing: 
- **com.twitter.scalding.TypedPipeJoinTest**: This one passes if I change `.sink` to `.typedSink`. 
- **com.twitter.scalding.mathematics.Matrix2Test**: 
    The failing test code is:

    ```scala
    // Treat (Double, Double, Double) as string because that is what is actually returned
    // when using runHadoop
    val result = Map((1, 1) -> (2.0, 6.0, 10.0), (2, 2) -> (6.0, 4.0, 2.0), (1, 2) -> (8.0, 10.0, 4.0))
        .mapValues(_.toString)
    toSparseMat(ob) shouldBe result
    ```

    and the error is:
    ```
Map((1,1) -> (2.0,6.0,10.0), (2,2) -> (6.0,4.0,2.0), (1,2) -> (8.0,10.0,4.0)) 
was not equal to 
Map((1,1) -> "(2.0,6.0,10.0)", (2,2) -> "(6.0,4.0,2.0)", (1,2) -> "(8.0,10.0,4.0)") 
    ```
    I'm not sure why the test converts the values to strings. The test passes if I get rid of `.mapValues(_.toString)`.

Should I make these changes?

johnynek:
another issue: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/source/DailySources.scala#L49

we need to fix that. It is in use in source.

johnynek:
@sid-kap make the changes you propose to the Matrix test.

BUT: we should not make the typedSink DEFAULT to using TypeDescriptor. That will break tests in source (grep for typedSink, you'll get a lot of hits, and many won't work with that change).

instead, on the few tests that fail, at those calls explicitly pass:
```scala
//where we have a failing test *AND* typedSink:
.typedSink(someSink)(TypeDescriptor.typeDescriptor.converter) { ...
```

sid-kap:
In DailySources, would we just change
```scala
class DailySuffixTypedTsv[T](prefix: String)
  (implicit override val dateRange: DateRange, 
  override val mf: Manifest[T], 
  override val conv: TupleConverter[T],
  override val tset: TupleSetter[T])
  extends DailySuffixSource(prefix, dateRange) with TypedDelimited[T]
```
to
```scala
class DailySuffixTypedTsv[T](prefix: String)
  (implicit val td[TypeDescriptor[T]])
  (implicit override val dateRange: DateRange, 
  override val mf: Manifest[T], 
  override val conv: TupleConverter[T] = td.converter,
  override val tset: TupleSetter[T] = td.setter)
  extends DailySuffixSource(prefix, dateRange) with TypedDelimited[T]
```
?

johnynek:
Yes but we won't need the Manifest, TupleConverter or TupleSetter anymore.

On Wednesday, July 15, 2015, Sidharth Kapur <notifications@github.com>
wrote:

> In DailySources, would we just change
>
> class DailySuffixTypedTsv[T](prefix: String)
>   (implicit override val dateRange: DateRange,
>   override val mf: Manifest[T],
>   override val conv: TupleConverter[T],
>   override val tset: TupleSetter[T])
>   extends DailySuffixSource(prefix, dateRange) with TypedDelimited[T]
>
> to
>
> class DailySuffixTypedTsv[T](prefix: String)
>   (implicit val td[TypeDescriptor[T]])
>   (implicit override val dateRange: DateRange,
>   override val mf: Manifest[T],
>   override val conv: TupleConverter[T] = td.converter,
>   override val tset: TupleSetter[T] = td.setter)
>   extends DailySuffixSource(prefix, dateRange) with TypedDelimited[T]
>
> ?
>
> —
> <https://github.com/twitter/scalding/pull/1356#issuecomment-121793412>.
>


-- 

sid-kap:
Since we are deprecating TypedDelimited, should we rewrite this class using TypedText?

johnynek:
Well, we can make th apply method, use TypedText like we did with the apply
method of TypedTsv. But let's just mark the class as deprecated.

On Wednesday, July 15, 2015, Sidharth Kapur <notifications@github.com>
wrote:

> Since we are deprecating TypedDelimited, should we rewrite this class
> using TypedText?
>
> —
> <https://github.com/twitter/scalding/pull/1356#issuecomment-121794202>.
>


-- 

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -93,6 +106,7 @@ object FixedPathTypedDelimited {
  * If T is a subclass of Product, we assume it is a tuple. If it is not, wrap T in a Tuple1:
  * e.g. TypedTsv[Tuple1[List[Int]]]
  */
+@deprecated("Use TypedText instead")
```
johnynek: should be TypedTextDelimited I think. Also, can we add a date in the since portion (so people can be reminded how long it has been deprecated).

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -124,6 +138,7 @@ trait TypedDelimited[T] extends DelimitedScheme
   final override def sinkFields = fields
 }
 
+@deprecated("Use TypedText instead")
```
johnynek: TypedTextDelimited

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
```diff
@@ -80,59 +80,104 @@ object FieldsProviderImpl {
...
     val res = if (namingScheme == NamedWithPrefix || namingScheme == NamedNoPrefix) {
-      val fieldNames = expanded.map(_._2)
-      q"""
-      new _root_.cascading.tuple.Fields(_root_.scala.Array.apply[java.lang.Comparable[_]](..$fieldNames),
-        _root_.scala.Array.apply[java.lang.reflect.Type](..$typeTrees)) with _root_.com.twitter.bijection.macros.MacroGenerated
-      """
+      val leftFn = { fieldNames: List[String] =>
+        q"""
+        new _root_.cascading.tuple.Fields(_root_.scala.Array.apply[_root_.java.lang.Comparable[_]](..$fieldNames),
+          _root_.scala.Array.apply[java.lang.reflect.Type](..$typeTrees)) with _root_.com.twitter.bijection.macros.MacroGenerated
```
johnynek: I don't think I see the benefit in adding with MacroGenerated on the Fields instance, since this will cause us to create a new class for each macro, whereas we could have just created an instance. I think we should remove this.

scalding-core/src/main/scala/com/twitter/scalding/macros/impl/FieldsProviderImpl.scala
```diff
@@ -80,59 +80,104 @@ object FieldsProviderImpl {
...
+        """
+      }
+      val rightFn = { fieldNames: List[Int] =>
+        q"""
+        val ints: _root_.scala.Array[Comparable[_]] = $fieldNames.map(_root_.java.lang.Integer.valueOf(_)).toArray
+        new _root_.cascading.tuple.Fields(ints,
+          _root_.scala.Array.apply[java.lang.reflect.Type](..$typeTrees)) with _root_.com.twitter.bijection.macros.MacroGenerated
```
johnynek: I think we should move MacroGenerated on Fields.

johnynek:
@sid-kap @ianoc Great work here, guys. I went ahead and merged this.

There are two other issues that can be handled by followup:

1) DailyTypedTsv.apply should call the new code (and the class should be deprecated).
2) I think we should remove MacroGenerated from the Fields we create. If we go down that road of adding MacroGenerated to every instance we create of any non-final class, we are going to wind up creating a lot of extra classes. Not sure I see the benefit. When we are adding it to an instance that must be macro generated, there is less (if any) cost to doing it. I am not sure I see any benefit either, as I don't think we have ever used it.

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
sid-kap: I believe this one should be FixedTypedText? 
since 
FixedTypedText extends FixedPathSource with TypedTextDelimited
and 
FixedPathTypedDelimited extends FixedPathSource with TypedDelimited

johnynek: Sounds good.

On Thursday, July 16, 2015, Sidharth Kapur <notifications@github.com> wrote:

> In scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
> <https://github.com/twitter/scalding/pull/1356#discussion_r34813124>:
>
> > @@ -124,6 +138,7 @@ trait TypedDelimited[T] extends DelimitedScheme
> >    final override def sinkFields = fields
> >  }
> >
> > +@deprecated("Use TypedText instead")
>
> I believe this one should be FixedTypedText?
> since
> FixedTypedText extends FixedPathSource with TypedTextDelimited
> and
> FixedPathTypedDelimited extends FixedPathSource with TypedDelimited
>
> —
> <https://github.com/twitter/scalding/pull/1356/files#r34813124>.
>


-- 

PR #1358, opened 2015-07-08
Runtime reducer estimator

sid-kap:
Refactored reduce estimators, and added new runtime-based reducer estimators

* Created `FlowStepHistory` case class and `HistoryReducerEstimator` so that any reducer estimator that depends on Hadoop history can extend `HistoryReducerEstimator`
* Rewrote existing `RatioBasedEstimator` as descendant of `HistoryReducerEstimator`
* Created `RuntimeReducerEstimator`, which represents any reducer estimator that, in some way, uses previous runtime statistics.
* Created 4 new reducer estimators, which vary in whether they use the median or the mean to estimate runtime, and whether they take into account the input bytes of the current job.


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -25,6 +28,40 @@ object EstimatorConfig {
...
+  def size(f: Hfs, conf: JobConf): Long = {
+    val fs = f.getPath.getFileSystem(conf)
+    fs.globStatus(f.getPath)
+      .map{ s => fs.getContentSummary(s.getPath).getLength }
+      .sum
+  }
+
+  def inputSizes(step: FlowStep[JobConf]): Seq[(String, Long)] = {
+    val conf = step.getConfig
+    unrollTaps(step) flatMap {
```
rubanm: `.flatMap` prefer dots for non-symbolic methods

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -24,48 +24,18 @@ class InputSizeReducerEstimator extends ReducerEstimator {
...
-
   /**
    * Figure out the total size of the input to the current step and set the number
    * of reducers using the "bytesPerReducer" configuration parameter.
    */
   override def estimateReducers(info: FlowStrategyInfo): Option[Int] =
-    inputSizes(info.step) match {
-      case Some(inputSizes) =>
+    Common.inputSizes(info.step) match {
+      case inputSizes if inputSizes.isEmpty =>
```
rubanm: `case Nil` ?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -45,6 +82,23 @@ class ReducerEstimator {
...
+    val maxHistory = EstimatorConfig.getMaxHistory(conf)
+
+    fetchHistory(info, maxHistory).map(estimateReducers(info, _)).recoverWith {
+      case e =>
+        LOG.warn(s"Unable to fetch history in $getClass. Error: $e")
+        Failure(e)
+    }.toOption.flatten
+  }
+
+  def estimateReducers(info: FlowStrategyInfo, history: Seq[FlowStepHistory]): Option[Int]
```
rubanm: can this be `protected`? Looks like it.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -0,0 +1,107 @@
...
+        .map { t => t.finishTime - t.startTime }
+    }
+}
+
+/**
+ * Estimator that uses the input size and a fixed "bytesPerReducer" target.
+ *
+ * Bytes per reducer can be configured with configuration parameter, defaults to 1 GB.
+ */
+trait RuntimeReducerEstimator extends HistoryReducerEstimator {
```
sid-kap: This class hierarchy is too deep.

One solution is to rename this trait to something like `Averager` and making it not extend `HistoryReducerEstimator`. Then `OrdinaryRuntimeReducerEstimator`and `InputScaledRuntimeReducerEstimator` could mixin (or take as a constructor parameter) `Averager`.

rubanm: Yes, keeping the estimation logic out of the hierarchy makes sense to me. How about something like the following?
```scala
trait RuntimeEstimationScheme {
  def estimateTaskTime(times: Seq[Long]): Option[Long]
  def estimateJobTime(times: Seq[Long]): Option[Long]
}
object MedianBasedEstimationScheme extends RuntimeEstimationScheme {
 ...
}
object MeanBasedEstimationScheme extends RuntimeEstimationScheme {
 ...
}
```
And the estimator classes could be:
 ```scala
class HRavenMedianRuntimeBasedEstimator
  extends BasicRuntimeReducerEstimator(MedianBasedEstimationScheme)
  with HRavenHistoryService 
```
etc.

I'd also suggest using `Basic` instead of `Ordinary` for the names.

File: scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala [hunk 5]
```diff
@@ -123,26 +177,52 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
...
-  def reducerBytes: Long
-}
-
-object FlowStepHistory {
-  def apply(m: Long, r: Long) = new FlowStepHistory {
-    override def mapperBytes: Long = m
-    override def reducerBytes: Long = r
-  }
-}
+sealed case class FlowStepHistory(keys: FlowStepKeys,
```
johnynek: do we want sealed or final here? sealed means all subclasses are in this file.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
sid-kap: Yep, that would look better

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
johnynek: +1 to Ruban's suggestion here.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
sid-kap: Whoops, I didn't realize that Scala has a final keyword!
I think final would be more appropriate.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -45,6 +82,23 @@ class ReducerEstimator {
 
 }
 
+trait HistoryReducerEstimator extends ReducerEstimator with HistoryService {
```
sid-kap: We could easily refactor this to use composition instead of inheritance.
Should `HistoryReducerEstimator` be a class that takes `HistoryService` as a parameter?

johnynek: it could still be a trait with a `def history: HistoryService` then in subclasses you `override def history = ...`

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -45,6 +82,23 @@ class ReducerEstimator {
 
 }
```
sid-kap: I think `ReducerEstimator` should be a trait, not a class.

I just caught a bug (in this version of the code) in which every single reducer estimator's `estimateReducers` method returns `None` because I forgot to override the method!

Such a change would prevent this kind of bug.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -0,0 +1,107 @@
...
+
+import scala.collection.JavaConverters._
+import cascading.flow.FlowStep
+import cascading.tap.{ Tap, CompositeTap }
+import cascading.tap.hadoop.Hfs
+import org.apache.hadoop.mapred.JobConf
+import org.slf4j.LoggerFactory
+
+object RuntimeReducerEstimator {
+  val RuntimePerReducer = "scalding.reducer.estimator.runtime.per.reducer"
```
johnynek: runtime per reducer sounds strange. This means if you have `n` reducers you expect `n * time` to finish?

I think a good approach is to set a goal time for how long the reducers run, say 10 minutes. Then we look at the history and get a sequence of pairs: (reducers, time-required). We could take the 80%-ile time to finish instead of the actual maximum (to deal with the case of skew). Finally, we can make least squares fit that gives us the number of reducers we should choose to reach the goal time.

sid-kap: Yep, runtime per reducer isn't a great name. "How long the reducers run" is along the lines of what I meant.

The history gives us a sequence of pairs (reducers, list of times for each reducer). `estimateTaskTime` takes each (reducers, list of times for each reducer) -> (reducers, time-required). Right now we've got an implementation of `estimateTaskTime` that uses the 50th percentile (median), and we could easily make one that uses the 80th percentile.

Using the pairs (num-reducers, time-required) to fit a function that predicts time-required from num-reducers is the interesting part.

Here was my approach (for the `BasicReducerEstimator`): 
I assumed that time-required = k / num-reducers for some constant k. So we calculate time-required * num-reducers for each point, and then use the median/mean of these to estimate k.

So the difference in our approaches is that I'm assuming an inverse relation, and you're assuming a line. I suppose that a linear fit would work if the data points are close enough together? Maybe we should take a better look at HRaven data to decide what kind of curve we want to fit.

johnynek:
this looks good to me, assuming we follow up with an evaluation: if our goal is to get the median time (or x percentile time) to be T, we can check after each job to see how well we did.

It would be good to see us get good control of this, and show the evidence that we have it under control.

@rubanm I'll let you merge this when you get back.

PR #1366, opened 2015-07-15
Fix TypedPipe.limit to be correct, if slighly slower

johnynek:
this comes up too often, let's just do the correct thing.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -283,15 +283,10 @@ trait TypedPipe[+T] extends Serializable {
   def fork: TypedPipe[T] = onRawSingle(identity)
 
   /**
-   * WARNING This is dangerous, and may not be what you think.
-   *
-   * limit the output to AT MOST count items.
-   * useful for debugging, but probably that's about it.
-   * The number may be less than count, and not sampled by any particular method
-   *
-   * This may change in the future to be exact, but that will add 1 MR step
+   * limit the output to at most count items, if at least count items exist.
    */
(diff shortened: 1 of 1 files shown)


johnynek:
@avibryant any objections here? Seems like the right thing.

ianoc:
I'd like if we kept the method around in some way shape or form, I use it in analysis for quick map only jobs. SomeNewSource.limit(100).dump in the repl is a super fast command, this will be much slower

johnynek:
Are you sure it will be slower? You only need one reducer and it will send
at most N * Mappers data to the reducers.

Can you do a quick benchmark in the repl to compare
groupAll.bufferedTake(100).values on a source you care about?

On Wed, Jul 15, 2015 at 5:58 PM, ianoc <notifications@github.com> wrote:

> I'd like if we kept the method around in some way shape or form, I use it
> in analysis for quick map only jobs. SomeNewSource.limit(100).dump in the
> repl is a super fast command, this will be much slower
>
> —
> <https://github.com/twitter/scalding/pull/1366#issuecomment-121761862>.
>



-- 

isnotinvain:
we *could* add a new method that works the way old limit did.

ianoc:
Buffered take will deserialize all of the data in the mapper, the old limit
won't right? It'll terminate after reading the number of records. I'll run
a test in a few mins to confirm

johnynek:
@ianoc you're right that if we have sources that implement deserialization at the scalding layer (as I think LzoCodec does, for instance) limit would not deserialize whereas this implementation would.

We would need an optimizer that could see that limit commutes with map.

ianoc:
Given the old one is only really faster when you delay deserialization as a result think it makes sense to just merge this.

avibryant:
I'm late here but :+1: @johnynek.

PR #1371, opened 2015-07-16
Fix deprecation warnings in TypedDelimited

sid-kap:


scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -138,7 +138,7 @@ trait TypedDelimited[T] extends DelimitedScheme
   final override def sinkFields = fields
 }
 
-@deprecated("Use TypedText instead")
+@deprecated("Use FixedTypedText instead", "16 July 2015")
 class FixedPathTypedDelimited[T](p: Seq[String],
```
sid-kap: I believe this one should be FixedTypedText
since 
FixedTypedText extends FixedPathSource with TypedTextDelimited
and 
FixedPathTypedDelimited extends FixedPathSource with TypedDelimited

johnynek:
looks like this does not merge cleanly.

sid-kap:
It should merge now

PR #1372, opened 2015-07-16
Ianoc/revert changes around making file systems

ianoc:
It turns out doing newInstance here is super expensive. It creates new hadoop configurations/reads lots of XML and does the huge hadoop-hard-to-comprehend workload.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -35,7 +35,7 @@ import cascading.tuple.Fields
 import com.etsy.cascading.tap.local.LocalTap
 
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.{ FileStatus, FileSystem, PathFilter, Path }
+import org.apache.hadoop.fs.{ FileStatus, PathFilter, Path }
 import org.apache.hadoop.mapred.JobConf
 import org.apache.hadoop.mapred.OutputCollector
 import org.apache.hadoop.mapred.RecordReader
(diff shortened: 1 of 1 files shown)


johnynek:
Merge when green.

On Thursday, July 16, 2015, ianoc <notifications@github.com> wrote:

> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/twitter/scalding/pull/1372
> Commit Summary
>
>    - Needed to revert this to 0.15.0 version and before, newInstance
>    builds new hadoop configurations and is really really slow it seems
>    - fix bad merge change
>
> File Changes
>
>    - *M*
>    scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
>    <https://github.com/twitter/scalding/pull/1372/files#diff-0> (15)
>
> Patch Links:
>
>    - https://github.com/twitter/scalding/pull/1372.patch
>    - https://github.com/twitter/scalding/pull/1372.diff
>
> —
> <https://github.com/twitter/scalding/pull/1372>.
>


-- 

isnotinvain:
What was the motivation for creating a new instance?
Is this code supposed to be thread safe? You could use a thread-local FileSystem.
I think FileSystem.get() is cached + synchronized, but I don't think the fs instances it returns are themselves thread safe.

ianoc:
We had intermittent issues around the file systems when executions were starting in multiple threads at once. So each one should remain thread local after we get it....

isnotinvain:
So this probably doesn't fix the (intermittent) issue then? I *think* (not sure) that path.getFileSystem() will return a shared instance when 2 instances ask for it (though it'll synchronize the setup of that instance)

ianoc:
No it doesn't fix that, but the 'fix' here was much worse than the problem... so back to having the problem really

isnotinvain:
right makes sense thanks

PR #1373, opened 2015-07-17
Revert typed tsv behavior

ianoc:
Mostly revert for external compatibility problems. Other than that i've left the tests being on TypedText as I think its what we'd like going forward. But existing debugging usage is proving very hard to migrate to TypedText


sid-kap:
Looks good to me.

isnotinvain:
+1

scalding-core/src/main/scala/com/twitter/scalding/TypedDelimited.scala
```diff
@@ -73,32 +72,20 @@ object TypedOsv extends TypedSeperatedFile {
   val separator = "\u0001"
 }
 
-@deprecated("Use `new FixedTypedText` instead")
```
johnynek: I think we should still deprecate.

PR #1376, opened 2015-07-20
A serialization error we were seeing in repl usage

ianoc:
Test that should fail followed by fix


johnynek:
I don't see why this fixes it. Why isn't calling toPipe their okay?

ianoc:
toPipe there should be fine, it just seemed wrong since we'd be doing the operations twice. though not likely to produce an incorrect result

The fix was the other case, there is a few paths from which we merge flowdef's, and i guess our updater stuff was missing those.

PR #1379, opened 2015-07-21
Add monoid and semigroup for Execution

johnynek:
Just added for completeness. Probably only real application is:

```Monoid.sum(actionList: List[Execution[Unit]]): Execution[Unit]```

which can be nice to combine a bunch of independent execution jobs into one.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.scalding
 
 import com.twitter.algebird.monad.Reader
-import com.twitter.algebird.{ Monoid, Monad }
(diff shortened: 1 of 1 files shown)


Gabriel439:
Yay!  I love this!

+1

jnievelt:
If it'll be used to sum, should we add sumOption which uses Execution.sequence?

johnynek:
I thought about that, but that would force us to materialize an Iterator of these into a Sequence, just to get the sumOption. It's not clear that is always a win. Instead, this is incremental and less code.

If you prefer, I can do it, but given that Execution is generally VERY slow code (it talks to the cluster in the interesting cases), I thought it was better to default to the smaller memory case.

johnynek:
@jnievelt let me know either way.

jnievelt:
@johnynek sure sgtm, can always add it later

PR #1380, opened 2015-07-22
Upgrade parquet to 1.8.1

isnotinvain:
This involves porting over the the org.apache.parquet package (from the parquet package), as well as dealing with some private APIs that got changed / removed between 1.6 and 1.8

This is going to be a binary incompatible change given the package rename (the scalding-parquet* sub modules haven't changed much but they deal in 'org.apache.parquet' objects now) so this will trigger a new release version.


scalding-parquet/src/main/scala/com/twitter/scalding/parquet/HasColumnProjection.scala
```diff
@@ -1,27 +1,67 @@
...
 
-  protected[parquet] final def columnGlobs: Set[ColumnProjectionGlob] = withColumns.map(ColumnProjectionGlob)
+  /**
+   * The format for specifying columns is described here:
+   * https://github.com/apache/parquet-mr/blob/master/parquet_cascading.md#21-projection-pushdown-with-thriftscrooge-records
+   *
+   * Note that the format described there says that multiple globs can be combined with a ; character.
+   * Instead, we use a Set() here and will eventually join the set on the ; character for you.
+   */
+  def withColumnProjections: Set[String] = Set()
```
isnotinvain: If anyone has a better suggestion to differentiate `withColumns` (deprecated) from `withColumnProjections` (new api) let me know. Could be something shorter like `withColumns2` :(

isnotinvain:
Seems like http://typesafe.artifactoryonline.com/ is down so the builds are failing :/

isnotinvain:
merging https://github.com/twitter/scalding/pull/1381 should fix this

johnynek:
are we ready to bump the monorepo up if we merge this?

johnynek:
or will this block any future publishing of develop until monorepo is upgraded?

ianoc:
It will block future publishing, Alex is going to work on our internal upgrades before this goes in so it doesn't cause issues.

isnotinvain:
Yeah I'm going to get an internal RB ready before we merge this.

isnotinvain:
@johnynek @ianoc @tsdeng the internal RB is posted, so this is ready to merge as well

isnotinvain:
@johnynek @ianoc @tsdeng this is ready to go, end to end tests passed and internal RB is ready as well

ianoc:
LGTM, merge when your ready to do the rls

PR #1381, opened 2015-07-23
Upgrade sbt launcher script (sbt-extras)

isnotinvain:
Just grabbed the latest from https://github.com/paulp/sbt-extras, which includes the change from http://typesafe.artifactoryonline.com to http://repo.typesafe.com


isnotinvain:
Thanks!

PR #1382, opened 2015-07-24
scalding-db macro-based sources and sinks

rubanm:
Adding new macro-based jdbc sources and sinks from internal.

More details on the readme:
https://github.com/rubanm/scalding/blob/rubanm/jdbc_part2/scalding-db/README.md


rubanm:
@johnynek @ianoc This probably needs some further cleanup before we can merge. A review would be great, thanks.

scalding-db/README.md
```diff
@@ -1,9 +1,167 @@
...
+Currently, the following databases are supported:
+* MySQL, both reads and writes
+* Vertica, writes only
+
+
+## MySQL usage
+
+### Reads
+
+For reads, a streaming copy from MySQL to HDFS is performed via the submitter prior to kicking off the mapreduce tasks. This HDFS snapshot is used for reads during the lifetime of the Scalding job.
```
ianoc: maybe mention the option here that controls this behavior?

scalding-db/README.md
```diff
@@ -1,9 +1,167 @@
...
+
+For more, see: http://www.vertica.com/2012/07/05/teaching-the-elephant-new-tricks/
+
+We use Vertica's native file format to stage the LOAD output on HDFS: http://my.vertica.com/docs/5.0/HTML/Master/13562.htm
+
+NOTE: Reading from Vertica is not supported.
+
+## Passing database credentials
+
+TODO(rubanm): Add JDBCConfiguration like helper class for loading credentials.
```
ianoc: maybe add an issue instead of the todo?

scalding-db/src/main/java/com/twitter/scalding/db/VerticaNativeOutputFormat.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.hadoop.mapred.TaskAttemptContext;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapred.JobConf;
+import org.apache.hadoop.util.Progressable;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.FSDataOutputStream;
+import org.apache.hadoop.fs.Path;
+
+/**
+ * Base class for Lzo outputformats.
```
ianoc: comment looks wrong, nothing to do with Lzo here right?

scalding-db/src/main/java/com/twitter/scalding/db/VerticaRawWriter.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapred.RecordWriter;
+import org.apache.hadoop.mapred.Reporter;
+
+/**
+ * A writer for LZO-encoded blocks of protobuf or Thrift objects, generally read by
```
ianoc: comment is bad here too

scalding-db/src/main/java/com/twitter/scalding/db/VerticaSinkTap.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.hadoop.mapred.RecordReader;
+
+import cascading.flow.FlowProcess;
+import cascading.scheme.Scheme;
+import cascading.tap.SinkMode;
+import cascading.tap.hadoop.Hfs;
+import cascading.tuple.TupleEntryIterator;
+import cascading.util.Util;
+
+public class VerticaSinkTap extends Hfs {
```
ianoc: could do with a java doc

scalding-db/src/main/java/com/twitter/scalding/db/VerticaSinkTap.java
```diff
@@ -0,0 +1,53 @@
+package com.twitter.scalding.db;
```
ianoc: should have apache header

scalding-db/src/main/scala/com/twitter/scalding/db/HdfsUtil.scala
```diff
@@ -0,0 +1,60 @@
+package com.twitter.scalding.db
```
ianoc: apache header

scalding-db/src/main/scala/com/twitter/scalding/db/JDBCTapBuilder.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import com.twitter.scalding.{ AccessMode, Hdfs, Mode, Source, TestTapFactory }
+import cascading.jdbc.JDBCTap
+import cascading.tap.Tap
+import cascading.tuple.Fields
+import com.twitter.scalding.db.driver.JDBCDriver
+
+/**
+ * Extend this source to let scalding read from or write to a database.
+ * In order for this to work you need to specify the table name, column definitions and DB credentials.
```
ianoc: comment looks wrong?

scalding-db/src/main/scala/com/twitter/scalding/db/JdbcSinkCompletionHandler.scala
```diff
@@ -0,0 +1,22 @@
+package com.twitter.scalding.db
```
ianoc: apache header

scalding-db/src/main/scala/com/twitter/scalding/db/JdbcSinkCompletionHandler.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.IOException;
+import scala.util.{ Failure, Success }
+
+import org.apache.hadoop.mapred.JobConf;
+import org.slf4j.LoggerFactory
+
+case class JdbcSinkCompletionHandler(writer: JdbcWriter) {
```
ianoc: comment

ianoc:
A follow up to this is to rip out the json usage of this instead using the ordered serialization stuff since its inflight data we have the same sort of requirements. Not for tis PR tho..

scalding-db/README.md
```diff
@@ -1,10 +1,177 @@
...
+
+#### Working with nested case classes
+
+Nested case classes can be used as a workaround for the 22-size limitation. They can also be used for logically grouping the table columns. Nested case classes are flattened in left to right order. For example:
+```scala
+case class Person(id: Long, name: String, location: Location)
+case class Location(geo: GeoCode, doorNum: Int, street: String, city: String)
+case class GeoCode(lat: Long, lng: Long)
+```
+is flattened to a table schema with columns `id`, `name`, `lat`, `lng`, `doorNum`, `street`, `city`.
```
johnynek: what happens if you have two Persons? How are the fields assigned? Can you manually rename with annotations?

scalding-db/src/main/scala/com/twitter/scalding/db/JDBCSource.scala
```diff
@@ -0,0 +1,77 @@
...
+ *     MysqlDriver
+ *   )
+ * }
+ *
+ * @author Argyris Zymnis
+ * @author Oscar Boykin
+ * @author Kevin Lin
+ */
+
+abstract class JDBCSource(dbsInEnv: AvailableDatabases) extends Source with JDBCOptions with ColumnDefiner {
```
johnynek: isn't this class duplicative with one in scalding-jdbc?

rubanm: Yes. Should we add a dependency on scalding-jdbc instead, or for long term move everything into one module? https://github.com/twitter/scalding/issues/1385

scalding-db/src/main/scala/com/twitter/scalding/db/JdbcToHdfsCopier.scala
```diff
@@ -0,0 +1,95 @@
...
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{ FileSystem, FSDataOutputStream, Path }
+import org.slf4j.LoggerFactory
+
+object JdbcToHdfsCopier {
+
+  protected val log = LoggerFactory.getLogger(this.getClass)
+
+  def apply[T <: AnyRef: Manifest](connectionConfig: ConnectionConfig,
```
johnynek: can you add a comment why we need the Manifest? It looks like we are converting to JSON. Why is that? Why not take an Injection in writeToHdfs and then set that injection up here or let the user supply?

scalding-db/src/main/scala/com/twitter/scalding/db/MySqlJdbcWriter.scala
```diff
@@ -0,0 +1,152 @@
...
+import org.slf4j.LoggerFactory
+
+import com.twitter.scalding.db.driver.DriverClass
+
+import java.sql._
+import scala.annotation.tailrec
+import scala.io.{ Codec, Source }
+import scala.util.Try
+
+class MySqlJdbcWriter[T](
```
johnynek: can we get some comments here. What is json2CaseClass doing? The role of JSON is unclear to me.

scalding-db/src/main/scala/com/twitter/scalding/db/TryHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+    // attempt a side-effect before returning the original Try if the original Try completes (success or failure)
+    // ignore any failures while doing so
+    def onComplete(action: => Unit): Try[T] = {
+      val effect = (_: Any) => { Try(action); t }
+      t.transform(effect, effect)
+    }
+
+    // attempt a side-effect before returning the orignal Try if the original Try fails
+    // ignore any failures while doing so
+    def onFailure[U >: T](action: => Unit): Try[U] = t.recoverWith { case e => Try(action); Failure(e) }
```
johnynek: I don't think you need the type parameter. I think you can just return `Try[T]` which since it is covariant will already be a `Try[U]` if needed.

scalding-db/src/main/scala/com/twitter/scalding/db/TypedJdbcSource.scala
```diff
@@ -0,0 +1,190 @@
...
+ *     ConnectUrl("jdbc:mysql://mysql01.company.com:3306/production"),
+ *     MysqlDriver
+ *   )
+ * }
+ *
+ * @author Ian O Connell
+ * @author Ruban Monu
+ */
+
+abstract class TypedJDBCSource[T <: AnyRef: DBTypeDescriptor: Manifest](dbsInEnv: AvailableDatabases)
```
johnynek: would be awesome if we didn't need Manifest here (especially given that we have the DBTypeDescriptor). Manifest is deprecated and is going to cause us pain sooner or later.

Also, do we really need the `T <: AnyRef`? Ideally just the TypeDescriptor would do.

johnynek:
This is really great, but I would really like it if we could merge with scalding-jdbc to avoid confusion.

It looks like some classes are either duplicated or changed very minorly here. If it is confusing for me to keep straight, I'm sure it also will be for our users and for the long term maintenance.

johnynek:
Can we make `DBTypeDescriptor` extend `TypeDescriptor`? You might have methods on the base you want to apply with a given DBTypeDescriptor which is a superset.

johnynek:
alternatively, if this is a complete superset of the scalding-jdbc code, we could just deprecate the whole module and not add anything to it, add a README that says you should move to scalding-db, and plan to remove that code in a few versions.

One worry I have with that is this: suppose I have 100 sources using the old code. How can I easily upgrade to the new code? If we can't easily support an upgrade path, (like just make a new import and your code works), then we are screwing the users.

scalding-db/src/main/scala/com/twitter/scalding/db/JDBCSource.scala
johnynek: I'd say move to one module.

johnynek:
somehow we lost this.

I'm closing now since I guess it's not a priority for anyone and we do have some existing macro-based DB support.

PR #1383, opened 2015-07-24
Just move whitespace, add comments, simplify a few methods

johnynek:
This is preparation for a follow up to expand the capabilities of TypeDescriptor (and probably rule out `Option[String]`).


ianoc:
lgtm merge when green

PR #1386, opened 2015-07-25
Don't publish maple when doing 2.11 so we only publish it once -- nee…

ianoc:
…ded for cross publishing to maven repo's

Can now do sbt +publish to a normal maven repo. (Sonatype seems more forgiving about this so it worked before -- otherwise maple gets tried to publish twice)

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -488,6 +488,10 @@ object ScaldingBuild extends Build {
     previousArtifact := None,
     crossPaths := false,
     autoScalaLibrary := false,
+    // Disable cross publishing for this artifact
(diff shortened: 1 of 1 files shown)


project/Build.scala
```diff
@@ -488,6 +488,10 @@ object ScaldingBuild extends Build {
     previousArtifact := None,
     crossPaths := false,
     autoScalaLibrary := false,
+    // Disable cross publishing for this artifact
+    publishArtifact <<= (scalaVersion) { scalaVersion =>
+        if(scalaVersion.startsWith("2.11")) false else true
```
johnynek: when we move to 2.11 + 2.12, this will probably need to be updated. but as will many other things.

PR #1388, opened 2015-07-28
Enable Scalding-REPL for Scala 2.11

sid-kap:
Don't merge, I just opened a pull request so that I can use Travis

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -403,7 +403,6 @@ object ScaldingBuild extends Build {
   lazy val scaldingRepl = module("repl")
     .configs(Unprovided) // include 'unprovided' as config option
     .settings(
-      publishArtifact := true,
       initialCommands in console := """
         import com.twitter.scalding._
         import com.twitter.scalding.ReplImplicits._
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -403,9 +403,9 @@ object ScaldingBuild extends Build {
   lazy val scaldingRepl = module("repl")
     .configs(Unprovided) // include 'unprovided' as config option
     .settings(
-      skip in compile := !isScala210x(scalaVersion.value),
-      skip in test := !isScala210x(scalaVersion.value),
-      publishArtifact := isScala210x(scalaVersion.value),
+      // skip in compile := !isScala210x(scalaVersion.value),
+      // skip in test := !isScala210x(scalaVersion.value),
+      publishArtifact := true,
       initialCommands in console := """
         import com.twitter.scalding._
         import com.twitter.scalding.ReplImplicits._
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -403,8 +403,6 @@ object ScaldingBuild extends Build {
   lazy val scaldingRepl = module("repl")
     .configs(Unprovided) // include 'unprovided' as config option
     .settings(
-      // skip in compile := !isScala210x(scalaVersion.value),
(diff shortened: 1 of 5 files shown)


project/Build.scala
```diff
@@ -403,9 +403,7 @@ object ScaldingBuild extends Build {
   lazy val scaldingRepl = module("repl")
     .configs(Unprovided) // include 'unprovided' as config option
     .settings(
-      skip in compile := !isScala210x(scalaVersion.value),
-      skip in test := !isScala210x(scalaVersion.value),
-      publishArtifact := isScala210x(scalaVersion.value),
+      publishArtifact := true,
```
ianoc: This i believe is the default, can remove this line

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
```diff
@@ -85,14 +89,15 @@ class ScaldingILoop
...
     super.createInterpreter()
-    addThunk {
-      intp.beQuietDuring {
-        addImports(imports: _*)
+    intp.beQuietDuring {
+      addImports(imports: _*)
 
-        // interpret all files named ".scalding_repl" from the current directory up to the root
-        findAllUpPath(".scalding_repl")
-          .reverse // work down from top level file to more specific ones
```
rubanm: We don't reverse the list anymore. Is that no longer needed?

sid-kap: Hmm. I guess reversing the list guarantees that local settings override higher-up settings? So maybe we should add back the reverse.

ianoc: Yeah i think the reverse is intentional

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
```diff
@@ -18,12 +18,16 @@ package com.twitter.scalding
...
 import scala.tools.nsc.interpreter.IR
+import scala.tools.nsc.GenericRunnerSettings
 
 /**
  * A class providing Scalding specific commands for inclusion in the Scalding REPL.
  */
 class ScaldingILoop
   extends ILoopCompat {
+
+  settings = new GenericRunnerSettings({ s => echo(s) })
```
sid-kap: To instantiate the settings, I had to provide an `errorFn :: String -> Unit`.
Does this seem like a suitable error function? I could not copy the error function from the default settings because, by default, ILoop's `settings` is null.

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
sid-kap: Fixed

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
ianoc: I think thats ok, echo should inform the user of any error, so that seems ok.


Can we access the parent settings without overriding them like this?

sid-kap: The parent settings is null

ianoc:
Awesome, thanks Sid!

rubanm:
+1 Nice!

PR #1393, opened 2015-07-30
Handle no history case in RatioBasedEstimator

rubanm:
This was a regression from https://github.com/twitter/scalding/pull/1358/

Manually verified this handles the no history case correctly.


Gabriel439:
:ship:

sid-kap:
Looks good to me!

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RatioBasedEstimator.scala
```diff
@@ -48,6 +48,9 @@ abstract class RatioBasedEstimator extends ReducerEstimator {
     val threshold = RatioBasedEstimator.getInputRatioThreshold(conf)
 
     historyService.fetchHistory(info, maxHistory) match {
+      case Success(Nil) =>
```
isnotinvain: does fetchHistory return a List or a Seq? I'm wondering if Success(Nil) will still match say, Success(Vector()) if the underlying Seq type ever changes?

sid-kap: It returns a Try[Seq[FlowStepHistory]].
Scala Seqs match by contents, so List() == Seq(), but yeah I guess it's possible for that to change.

isnotinvain:
+1

PR #1395, opened 2015-07-31
Inline parquet-scrooge

ianoc:
Required a little bit of manual code in the Build.sbt since the scrooge sbt plugin doesn't seem to generate multiple languages at once easily.

### DIFF
--- a/scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetScroogeOutputFormat.java
+++ b/scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetScroogeOutputFormat.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright 2012 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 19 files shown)


scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetScroogeOutputFormat.java
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright 2012 Twitter, Inc.
```
Gabriel439: Shouldn't we update these copyright notices to 2015?

ianoc: Half the scalding repo has 2012. How about I do a follow up PR after this and update all of them?

Gabriel439: Sounds good

johnynek: actually, I don't think we need to keep adding copyright notices. Seems like make work, and the repo LICENSE should be enough (I think @travisbrown signed off on this statement).

ianoc: Oh, ok cool.

isnotinvain: Yeah, but we *do* probably need to leave the license headers as-is from files that we are copying from apache projects

isnotinvain:
LGTM, lets file a ticket for future cleanup though. This will probably just get deleted once we have a more efficient parquet thrift implementation (inside scalding rather than parquet)

tsdeng:
+ 1

PR #1399, opened 2015-08-03
Fixes the scrooge generator tasks not to generate code in the compile…

ianoc:
… target, we were publishing these


Gabriel439:
Looks good to me

PR #1400, opened 2015-08-03
Ianoc/configure set converter

ianoc:
Allows using the object to set the configuration of which deserializer is being used.

### DIFF
--- a/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
+++ b/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
@@ -87,6 +87,20 @@ object LzoGenericScheme {
 
   def apply[M](conv: BinaryConverter[M], clazz: Class[M]): LzoGenericScheme[M] =
     new LzoGenericScheme(conv, clazz)
+
+  /**
+   * From a Binary Converter passed in configure in the JobConf using of that by ElephantBird
+   */
+  def setConverter[M](conf: JobConf, conv: BinaryConverter[M]): Unit = {
+    val extern = Externalizer(conv)
+    try {
+      ExternalizerSerializer.inj.invert(ExternalizerSerializer.inj(extern)).get
+    } catch {
+      case e: Exception => throw new RuntimeException("Unable to roundtrip the BinaryConverter in the Externalizer.", e)
+    }
+    conf.set(ConfigBinaryConverterProvider.ProviderConfKey, ExternalizerSerializer.inj(extern))
+  }
+
 }
 
 /**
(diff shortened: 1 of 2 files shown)


scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -102,15 +118,7 @@ class LzoGenericScheme[M](@transient conv: BinaryConverter[M], clazz: Class[M])
...
-    val extern = Externalizer(conv)
-    try {
-      ExternalizerSerializer.inj.invert(ExternalizerSerializer.inj(extern)).get
-    } catch {
-      case e: Exception => throw new RuntimeException("Unable to roundtrip the BinaryConverter in the Externalizer.", e)
-    }
-
-    conf.set(ConfigBinaryConverterProvider.ProviderConfKey, ExternalizerSerializer.inj(extern))
-
+    LzoGenericScheme.setConverter(conv, conf)
```
rubanm: So we don't want to set the converter here when it's already been set earlier. I'm guessing this is because we want it to be configurable via the new method, before this scheme code is executed?

rubanm:
LGTM

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
ianoc: Exactly, if something else has gotten there first and wanted to set how the deserialization between the two fixed types is done we don't want to override. I'd view this as the default path per say. 

Use case so far is that I can swap out different thrift implementations for testing by setting this param

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/LzoGenericScheme.scala
```diff
@@ -87,6 +87,22 @@ object LzoGenericScheme {
...
     new LzoGenericScheme(conv, clazz)
+
+  /**
+   * From a Binary Converter passed in configure in the JobConf using of that by ElephantBird
+   */
+  def setConverter[M](conv: BinaryConverter[M], conf: JobConf, overrideConf: Boolean = false): Unit = {
+    if ((conf.get(ConfigBinaryConverterProvider.ProviderConfKey) == null) || overrideConf) {
+      val extern = Externalizer(conv)
+      try {
+        ExternalizerSerializer.inj.invert(ExternalizerSerializer.inj(extern)).get
```
johnynek: we should move this to `ExternalizerSerialization.testRoundTrip(i: Any): Try[Unit]`

PR #1401, opened 2015-08-03
Change hash function in GroupRandomly 

MansurAshraf:
Current hashing is causing collision so switching to identity hash.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -15,7 +15,7 @@ limitations under the License.
 */
 package com.twitter.scalding.typed
 
-import java.io.Serializable
+import java.io.{OutputStream, InputStream, Serializable}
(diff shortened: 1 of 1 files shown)


MansurAshraf:
Working on a unit test....

jnievelt:
This goes better in the TypedPipe object?

MansurAshraf:
I guess I can move it there and make it private. I inlined it because I dont want anyone using it outside this method

Gabriel439:
Looks good to me

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -382,7 +386,17 @@ trait TypedPipe[+T] extends Serializable {
   def groupRandomly(partitions: Int): Grouped[Int, T] = {
```
johnynek: can we use this for `.shard` now?

ianoc:
LGTM, thanks for getting on this quickly

PR #1404, opened 2015-08-04
Adds a function to test if a sink exists at the version we created

ianoc:
In summingbird would like to not try write into a version if it already exists

### DIFF
--- a/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/VersionedKeyValSource.scala
+++ b/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/VersionedKeyValSource.scala
@@ -124,6 +124,25 @@ class VersionedKeyValSource[K, V](val path: String, val sourceVersion: Option[Lo
       }
     }
 
+  def sinkExists(mode: Mode) =
+    sinkVersion match {
+      case Some(version) =>
+        mode match {
+          case Test(buffers) =>
+            buffers(this) map { !_.isEmpty } getOrElse false
+
+          case HadoopTest(conf, buffers) =>
+            buffers(this) map { !_.isEmpty } getOrElse false
+
+          case m: HadoopMode =>
+            val conf = new JobConf(m.jobConf)
+            val store = sink.getStore(conf)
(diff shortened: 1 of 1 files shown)


scalding-commons/src/main/scala/com/twitter/scalding/commons/source/VersionedKeyValSource.scala
```diff
@@ -124,6 +124,25 @@ class VersionedKeyValSource[K, V](val path: String, val sourceVersion: Option[Lo
...
+            buffers(this) map { !_.isEmpty } getOrElse false
+
+          case HadoopTest(conf, buffers) =>
+            buffers(this) map { !_.isEmpty } getOrElse false
+
+          case m: HadoopMode =>
+            val conf = new JobConf(m.jobConf)
+            val store = sink.getStore(conf)
+            store.hasVersion(version)
+          case _ => sys.error(s"Unknown mode $mode")
```
Gabriel439: Maybe use a `Try`/`Maybe` instead of throwing an error.  How likely is it that there will be a mode other than those three?

ianoc: Extremely unlikely, the other methods in this class seem to just cast under the wildcard case

johnynek: Yeah, there is little type safety around Mode, and no way to recover from a bad mode, so I'd leave it as is (Mode is generally configured at runtime).

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/VersionedKeyValSource.scala
```diff
@@ -124,6 +124,25 @@ class VersionedKeyValSource[K, V](val path: String, val sourceVersion: Option[Lo
       }
     }
 
+  def sinkExists(mode: Mode) =
```
johnynek: we should have return types on public methods.

PR #1406, opened 2015-08-05
add .groupWith method to TypedPipe

tdyas:
This method is more convenient than trying to pass the `Ordering[K]` explicitly into the `.group` method because there is no need to also specify the evidence parameter proving that the `TypedPipe[T]` is a `TypedPipe[(K, V)]`.


ianoc:
is it worth it over doing .group(implicitly, myordering) ?

tdyas:
The use of `implicitly` there feels like a workaround and not a natural part of the API.

PR #1409, opened 2015-08-06
add counter verification logic

nevillelyh:
This would allow users to mix in a `CounterVerification` trait and implement their own `verifyCounters` method when the job completes. We use this to fail jobs that produced invalid data, e.g. too many nulls or incorrect distribution.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -498,6 +498,29 @@ class ScriptJob(cmds: Iterable[String]) extends Job(Args("")) {
   }
 }
 
+/**
+ * Allows custom counter verification logic when the job completes.
+ */
+trait CounterVerification extends Job {
+  /**
+   * Verify counter values. The job will fail if this returns false or throws an exception.
+   */
+  def verifyCounters(counters: Map[StatKey, Long]): Try[Unit]
+
+  /**
+   * Override this to false to skip counter verification in tests.
+   */
+  def verifyCountersInTest: Boolean = true
+
+  override def listeners: List[FlowListener] = {
+    if (this.mode.isInstanceOf[TestMode] && !this.verifyCountersInTest) {
+      super.listeners
+    } else {
+      super.listeners :+ new StatsFlowListener(this.verifyCounters)
+    }
+  }
+}
+
 private[scalding] object FlowStepStrategies {
   /**
    * Returns a new FlowStepStrategy that runs both strategies in sequence.
--- a/scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
@@ -1,12 +1,13 @@
 package com.twitter.scalding
 
-import cascading.flow.{ FlowDef, FlowProcess }
+import cascading.flow.{ Flow, FlowListener, FlowDef, FlowProcess }
 import cascading.stats.CascadingStats
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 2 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -498,6 +498,29 @@ class ScriptJob(cmds: Iterable[String]) extends Job(Args("")) {
...
+   * Override this to false to skip counter verification in tests.
+   */
+  def verifyCountersInTest: Boolean = true
+
+  override def listeners: List[FlowListener] = {
+    if (this.verifyCountersInTest) {
+      super.listeners :+ new StatsFlowListener(this.verifyCounters)
+    } else {
+      super.listeners
+    }
```
rubanm: This needs to check for `TestMode` along with the flag here, right? and always add `StatsFlowListener` in other modes.

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -498,6 +498,29 @@ class ScriptJob(cmds: Iterable[String]) extends Job(Args("")) {
...
 }
 
+/**
+ * Allows custom counter verification logic when the job completes.
+ */
+trait CounterVerification extends Job {
+  /**
+   * Verify counter values. The job will fail if this returns false or throws an exception.
+   */
+  def verifyCounters(counters: Map[StatKey, Long]): Boolean
```
johnynek: can we instead return `Try[Unit]` here so we can allow the user a chance to give an exception message?

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -156,3 +159,25 @@ object RuntimeStats extends java.io.Serializable {
...
+/**
+ * FlowListener that checks counter values against a function.
+ */
+class StatsFlowListener(f: Map[StatKey, Long] => Boolean) extends FlowListener {
+
+  override def onCompleted(flow: Flow[_]): Unit = {
+    val stats = flow.getFlowStats
+    val keys = stats.getCounterGroups.asScala.flatMap(g => stats.getCountersFor(g).asScala.map(c => StatKey(c, g)))
+    val values = keys.map(k => (k, stats.getCounterValue(k.group, k.counter))).toMap
+    if (!f(values)) {
```
johnynek: if we used `f: Map[StatKey, Long] => Try[Unit]` we could just do: `f(values).get` here.

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
nevillelyh: Good catch. Forgot to port this from our internal fork.

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
nevillelyh: I was trying to keep this interface simple. The user can include message with `assert(_, msg)` or throw exception directly. Does that make sense?

johnynek: We really try now to keep the type signature guiding how it is used. We don't want users throwing exceptions to the framework because these compose very poorly.

I think:
```
def verifyCounters(m: Map[StatKey, Long]) = Try {
  assert(isGood(m), s"Stats are incorrect, $m")
}
```
If we take the position that Try is too complex to use, we are kind of sunk since much of the API assumes understanding working with types like that.

nevillelyh: You're right. This does look better :)

PR #1411, opened 2015-08-07
Runtime reducer estimator fixes

sid-kap:
Some changes that I made when testing the reducer estimator

- Instead of 4 runtime reducer estimators, now there is only one HRavenRuntimeReducerEstimator with sane defaults (prediction is input-scaled, uses median task time per job as goal). 
- Use Double arithmetic instead of Long.
- Update HRaven version so that I can get Task history. (HRaven versioning is weird, but this seems to be compatible with the version of HRaven in source.)
- Improve logging.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -411,9 +411,20 @@ object ScaldingBuild extends Build {
   lazy val scaldingHRaven = module("hraven").settings(
     libraryDependencies ++= Seq(
       "com.twitter.hraven" % "hraven-core" % hravenVersion
+        // These transitive dependencies cause sbt to give a ResolveException
+        // because they're not available on Maven. We don't need them anyway.
+        // See https://github.com/twitter/cassie/issues/13
         exclude("javax.jms", "jms")
         exclude("com.sun.jdmk", "jmxtools")
-        exclude("com.sun.jmx", "jmxri"),
+        exclude("com.sun.jmx", "jmxri")
+
+        // These transitive dependencies of hRaven cause conflicts when
+        // running scalding-hraven/*assembly and aren't needed
+        // for the part of the hRaven API that we use anyway
+        exclude("com.twitter.common", "application-module-log")
+        exclude("com.twitter.common", "application-module-stats")
+        exclude("com.twitter.common", "args")
+        exclude("com.twitter.common", "application"),
       "org.apache.hbase" % "hbase" % hbaseVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided"
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -410,7 +410,10 @@ object ScaldingBuild extends Build {
 
   lazy val scaldingHRaven = module("hraven").settings(
     libraryDependencies ++= Seq(
-      "com.twitter.hraven" % "hraven-core" % hravenVersion exclude("javax.jms", "jms") exclude("com.sun.jdmk", "jmxtools") exclude("com.sun.jmx", "jmxri"),
+      "com.twitter.hraven" % "hraven-core" % hravenVersion
+        exclude("javax.jms", "jms")
+        exclude("com.sun.jdmk", "jmxtools")
+        exclude("com.sun.jmx", "jmxri"),
       "org.apache.hbase" % "hbase" % hbaseVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided"
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -31,7 +31,7 @@ object ScaldingBuild extends Build {
   val hadoopLzoVersion = "0.4.19"
   val hadoopVersion = "2.5.0"
   val hbaseVersion = "0.94.10"
-  val hravenVersion = "0.9.13"
+  val hravenVersion = "0.9.16"
   val jacksonVersion = "2.4.2"
   val json4SVersion = "3.2.11"
   val paradiseVersion = "2.0.1"
@@ -410,7 +410,7 @@ object ScaldingBuild extends Build {
 
   lazy val scaldingHRaven = module("hraven").settings(
     libraryDependencies ++= Seq(
-      "com.twitter.hraven" % "hraven-core" % hravenVersion,
+      "com.twitter.hraven" % "hraven-core" % hravenVersion exclude("javax.jms", "jms") exclude("com.sun.jdmk", "jmxtools") exclude("com.sun.jmx", "jmxri"),
       "org.apache.hbase" % "hbase" % hbaseVersion,
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided"
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
@@ -29,7 +29,7 @@ object RuntimeReducerEstimator {
       case "mean" => MeanEstimationScheme
       case "median" => MedianEstimationScheme
       case _ =>
-        throw new Exception(s"""Value of $key must be "mean", "median", or not specified.""")
+        throw new Exception(s"""Value of $EstimationScheme must be "mean", "median", or not specified.""")
     }
   }
 
@@ -138,7 +138,7 @@ trait InputScaledRuntimeReducerEstimator extends HistoryReducerEstimator {
     // time-to-byte ratio for a step = time per reducer * number of reducers / number of bytes
     val timeToByteRatios: Seq[Double] =
       jobTimes.zip { history.map(_.hdfsBytesRead) }
-        .collect { case (Some(time), bytes) => time.toDouble / bytes }
+        .collect { case (Some(time), bytes) => time / bytes }
 
     // time-to-byte ratio, averaged over all the steps
     val typicalTimeToByteRatio: Option[Double] = runtimeEstimationScheme.estimateJobTime(timeToByteRatios)
@@ -151,7 +151,9 @@ trait InputScaledRuntimeReducerEstimator extends HistoryReducerEstimator {
       None
     } else {
       // numReducers = time-per-byte * numBytes / desiredRuntime
-      val estimate = typicalTimeToByteRatio.map { t => (t.toDouble * inputBytes / desiredRuntime).ceil.toInt }
+      val estimate = typicalTimeToByteRatio.map { t: Double =>
+        (t * inputBytes / desiredRuntime).ceil.toInt
+      }
 
       LOG.info(s"""
         | - HDFS bytes read: ${history.map(_.hdfsBytesRead)}
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
@@ -105,7 +105,6 @@ trait BasicRuntimeReducerEstimator extends HistoryReducerEstimator {
     val estimate = typicalJobTime.map { t: Double => (t / desiredRuntime).ceil.toInt }
 
     LOG.info(s"""
-      | - Job times: $jobTimes
       | - Typical job time: $typicalJobTime
       | - Desired runtime: $desiredRuntime
       | - Estimate: $estimate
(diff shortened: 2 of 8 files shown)


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -43,69 +72,135 @@ trait RuntimeEstimationScheme {
...
     if (inputBytes == 0) {
+      LOG.warn("Input bytes is zero in current step.")
       None
     } else {
       // numReducers = time-per-byte * numBytes / desiredRuntime
-      typicalTimeToByteRatio.map { t => (t.toDouble * inputBytes / desiredRuntime).ceil.toInt }
+      val estimate = typicalTimeToByteRatio.map { t => (t.toDouble * inputBytes / desiredRuntime).ceil.toInt }
+
+      LOG.info(s"""
+        | - History: $history
```
sid-kap: This is too much debugging. Will get rid of "$history".

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.15.1-SNAPSHOT"
+version in ThisBuild := "0.15.8-SNAPSHOT"
```
sid-kap: I will revert this

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -8,11 +8,40 @@ import org.apache.hadoop.mapred.JobConf
...
+
+  /**
+   * Whether to ignore the input size of the data.
+   * If true, RuntimeReducerEstimator uses a non-input scaled estimator.
+   * If false, RuntimeReducerEstimator uses an input-scaled estimator
+   * first, and uses a non-input-scaled estimator as a fallback.
+   * Default is false.
+   */
+  def getRuntimeIgnoreInputSize(conf: JobConf): Boolean = {
+    val RuntimeIgnoreInputSize = "scalding.reducer.estimator.runtime.ignore.input.size"
```
sid-kap: Will make this less verbose by changing it to "IgnoreInputSize"

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -8,11 +8,40 @@ import org.apache.hadoop.mapred.JobConf
 import org.slf4j.LoggerFactory
 
 object RuntimeReducerEstimator {
-  val RuntimePerReducer = "scalding.reducer.estimator.runtime.per.reducer"
-  val defaultRuntimePerReducer = 10 * 60 * 1000 // 10 mins
```
rubanm: Any reason for removing these? Might be useful for outside code to fetch the defaults.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -43,69 +72,134 @@ trait RuntimeEstimationScheme {
...
 trait BasicRuntimeReducerEstimator extends HistoryReducerEstimator {
   import RuntimeReducerEstimator._
 
+  private val LOG = LoggerFactory.getLogger(this.getClass)
+
   def runtimeEstimationScheme: RuntimeEstimationScheme
 
   def estimateReducers(info: FlowStrategyInfo, history: Seq[FlowStepHistory]): Option[Int] = {
-    val reduceTimes: Seq[Seq[Long]] = getReduceTimes(history)
+    val reduceTimes: Seq[Seq[Double]] = getReduceTimes(history).map(xs => xs.map(_.toDouble))
```
rubanm: Should the `toDouble` conversion be moved to `getReducerTimes`?

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -75,18 +75,29 @@ object HRavenHistoryService extends HistoryService {
...
+    Try(client.fetchFlowsWithConfig(cluster, user, batch, signature, nFetch, RequiredJobConfigs: _*))
+      .flatMap { flows =>
+        Try {
+          // Ugly mutable code to add task info to flows
+          flows.asScala.foreach { flow =>
+            flow.getJobs.asScala.foreach { job =>
+              val tasks = client.fetchTaskDetails(flow.getCluster, job.getJobId)
+              job.addTasks(tasks)
+            }
+          }
```
rubanm: So looks like this was a bug affecting the existing RatioBasedEstimator as well?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
rubanm: Ah scratch that. This looks fine actually.

project/Build.scala
```diff
@@ -410,7 +410,7 @@ object ScaldingBuild extends Build {
 
   lazy val scaldingHRaven = module("hraven").settings(
     libraryDependencies ++= Seq(
-      "com.twitter.hraven" % "hraven-core" % hravenVersion,
+      "com.twitter.hraven" % "hraven-core" % hravenVersion exclude("javax.jms", "jms") exclude("com.sun.jdmk", "jmxtools") exclude("com.sun.jmx", "jmxri"),
```
rubanm: nit - excludes to be in separate lines

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -75,18 +75,29 @@ object HRavenHistoryService extends HistoryService {
...
+
+          val successfulFlows = flows.asScala.filter(_.getHdfsBytesRead > 0).take(max)
+          if (successfulFlows.isEmpty) {
+            LOG.warn("Unable to find any successful flows in the last " + nFetch + " jobs.")
+          }
+          successfulFlows
+        }
+      } recoverWith {
```
rubanm: recoverWith here should still apply to the outer Try right?

ianoc: Also always use .'s between methods. (I see the old didn't have it, but it should have had)

rubanm:
Thanks for the fixes. Would be good to have some basic unit tests for RuntimeReducerEstimator, similar to the ones added for RatioBasedEstimator.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -8,11 +8,40 @@ import org.apache.hadoop.mapred.JobConf
...
+   * Whether to use the median or the mean in the runtime estimation process.
+   * Default is median.
+   */
+  def getRuntimeEstimationScheme(conf: JobConf): RuntimeEstimationScheme = {
+    val key = "scalding.reducer.estimator.runtime.estimation.scheme"
+    val default = "median"
+    conf.get(key, default) match {
+      case "mean" => MeanEstimationScheme
+      case "median" => MedianEstimationScheme
+      case _ => MedianEstimationScheme
```
rubanm: Should we just throw here? I feel it's better for users to be precise if they are providing the scheme than us silently picking median.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
sid-kap: I don't think the ratio-based estimator uses task-level data

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
sid-kap: Sure, I'll add the dot.
recoverWith still applies to the outer Try, I just indented incorrectly so it looks like it changed. I'll fix the indentation.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
sid-kap: Sure, that makes sense

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
sid-kap: Sure, might as well since it's common to both places where getReduceTimes is used

sid-kap:
Should I add unit tests in this PR or make a separate PR?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
sid-kap: This will be useful for writing unit tests, so I think I'll move it out of the method.

rubanm:
@sid-kap ideally this PR I'd say so we can exercise the tests as part of CI.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimatorTest.scala
```diff
@@ -0,0 +1,158 @@
...
+  def fetchHistory(info: FlowStrategyInfo, maxHistory: Int): Try[Seq[FlowStepHistory]] =
+    Success(
+      Seq(
+        makeHistory(inputSize * 2, 0, List(10, 1000, 3000)),
+        makeHistory(inputSize / 2, 0, List(10, 200, 400)),
+        makeHistory(inputSize * 4, 0, List(10, 2400, 3000))))
+}
+
+class Estimator1 extends RuntimeReducerEstimator {
+  override val historyService = HistoryService1
```
rubanm: Probably good to also add test cases for 'empty history' and 'failed to fetch history', given we not share anything with RatioBasedEstimator so don't have coverage for that here.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
```diff
@@ -43,69 +74,136 @@ trait RuntimeEstimationScheme {
...
+    val typicalJobTime: Option[Double] = runtimeEstimationScheme.estimateJobTime(jobTimes.flatten)
 
-    val desiredRuntime = getRuntimePerReducer(info.step.getConfig)
-    typicalJobTime.map { t => (t.toDouble / desiredRuntime).ceil.toInt }
+    val desiredRuntime: Long = getRuntimePerReducer(info.step.getConfig)
+
+    val estimate = typicalJobTime.map { t: Double => (t / desiredRuntime).ceil.toInt }
+
+    LOG.info(s"""
+      | - Job times: $jobTimes
```
rubanm: Just wondering if logging the whole sequence might be too verbose? The other values below are `Option`s so should be fine.

rubanm:
Some minor comments. This lgtm to merge after that. Thanks for adding the tests.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimator.scala
sid-kap: Sure, I can remove this

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimatorTest.scala
sid-kap: Fixed in latest commit

project/Build.scala
```diff
@@ -410,7 +410,21 @@ object ScaldingBuild extends Build {
...
   lazy val scaldingHRaven = module("hraven").settings(
     libraryDependencies ++= Seq(
-      "com.twitter.hraven" % "hraven-core" % hravenVersion,
+      "com.twitter.hraven" % "hraven-core" % hravenVersion
+        // These transitive dependencies cause sbt to give a ResolveException
+        // because they're not available on Maven. We don't need them anyway.
+        // See https://github.com/twitter/cassie/issues/13
+        exclude("javax.jms", "jms")
+        exclude("com.sun.jdmk", "jmxtools")
+        exclude("com.sun.jmx", "jmxri")
```
rubanm: I think most of these excludes would not have been needed if there was a separate `hraven-client` module. There is none sadly.

PR #1412, opened 2015-08-07
Make sure Execution.zip fails fast

johnynek:
This could be an issue on hadoop where some of the executions take a very long time, so failing fast will be useful.

(Also at Twitter, there is an example which deadlocks could happen if people were not careful and zip does not have this property)


ianoc:
LGTM, merge when green

PR #1414, opened 2015-08-11
When using WriteExecution and forceToDisk we can share the same flow …

ianoc:
…def closer in construction

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -571,10 +571,23 @@ object Execution {
 
   /**
    * Creates an Execution to do a write
+   *
+   * This variant allows the user to supply a method using the config and mode to build a new
+   * type U for the resultant execution.
    */
   private[scalding] def write[T, U](pipe: TypedPipe[T], sink: TypedSink[T], generatorFn: (Config, Mode) => U): Execution[U] =
     WriteExecution(ToWrite(pipe, sink), Nil, generatorFn)
 
+  /**
+   * The simplest form, just sink the typed pipe into the sink and get a unit execution back
+   */
+  private[scalding] def write[T](pipe: TypedPipe[T], sink: TypedSink[T]): Execution[Unit] =
+    WriteExecution(ToWrite(pipe, sink), Nil, (Config, Mode) => ())
+
+  /**
+   * Here we allow both the targets to write and the sources to be generated from the config and mode.
+   * This allows us to merge things looking for the config and mode without using flatmap.
+   */
   private[scalding] def write[T, U](fn: (Config, Mode) => (TypedPipe[T], TypedSink[T]), generatorFn: (Config, Mode) => U): Execution[U] =
     WriteExecution(ToWrite(fn), Nil, generatorFn)
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -559,35 +559,36 @@ object Execution {
     // We look up to see if any of our ToWrite elements have already been ran
     // if so we remove them from the cache.
     // Anything not already ran we run as part of a single flow def, using their combined counters for the others
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
-      val cacheLookup: List[(ToWrite, Either[Promise[ExecutionCounters], Future[ExecutionCounters]])] = (head :: tail).map{ tw => (tw, cache.getOrLock(tw)) }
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -467,27 +467,60 @@ object Execution {
    * This is here so we can call without knowing the type T
    * but with proof that pipe matches sink
    */
-  private case class ToWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) {
-    def write(flowDef: FlowDef, mode: Mode): Unit = {
-      // This has the side effect of mutating flowDef
-      pipe.write(sink)(flowDef, mode)
-      ()
+
+  private trait ToWrite[T] {
```
johnynek: don't think we need the `T` here, actually. Your design makes that clearer. Nice.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -539,8 +572,11 @@ object Execution {
   /**
    * Creates an Execution to do a write
    */
-  private[scalding] def write[T](pipe: TypedPipe[T], sink: TypedSink[T]): Execution[Unit] =
-    WriteExecution(ToWrite(pipe, sink), Nil)
+  private[scalding] def write[T, U](pipe: TypedPipe[T], sink: TypedSink[T], generatorFn: (Config, Mode) => U): Execution[U] =
```
johnynek: what if we add the old method back here rather than change in TypedPipe?

ianoc: Sure

johnynek:
Looks great, merge when green!

So nice to get that flatMap out of .forceToDiskExecution and be able to easily optimize zips on write/forceToDisk now. Great work, @ianoc!

johnynek:
This does have one potential downside due to the lack of a whole DAG optimizer. Consider this:

```scala
val write1 = pipe1.writeExecution(sink1)
val write2 = pipe2.writeExecution(sink2)

write1.zip(write2).map(_ => ()).flatMap { _ => write1 }
// previously, write1 was added to the execution cache, so the result is already known and it is not re-evaluated
// now write1.zip(write2) makes a different item that does not point to write1 at all, so evaluating
// that won't make write1 by itself available.
```
basically, if you have writes, or forceToDiskExecutions appearing multiple times in your graph, you could get recomputation.

I guess this is a trade-off on this PR since the alternative is recomputation inside of a flow. But what this really underscores is the need to analyze the whole DAG (TypedPipe + Executions) to minimize work.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -467,27 +498,113 @@ object Execution {
...
+        val (l, r) = unwrapListEither(tail)
+        (l, (a, c) :: r)
+      case Nil => (Nil, Nil)
+    }
+
+    // We look up to see if any of our ToWrite elements have already been ran
+    // if so we remove them from the cache.
+    // Anything not already ran we run as part of a single flow def, using their combined counters for the others
+    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      val cacheLookup: List[(ToWrite, Either[Promise[ExecutionCounters], Future[ExecutionCounters]])] = (head :: tail).map{ tw => (tw, cache.getOrLock(tw)) }
```
johnynek: should we also add `this` to the cache? The only reason I can see is to make sure we call `fn` fewer times (we can still call it twice in the case of `.map` on an execution, but supposedly `.map` is for pure functions, so that should be okay.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -486,27 +486,47 @@ trait TypedPipe[+T] extends Serializable {
...
     }
 
+    val readFn = { (conf: Config, mode: Mode) =>
+      mode match {
+        case _: CascadingLocal => // Local or Test mode
+          TypedPipe.from(inMemoryDest.readResults)
+        case _: HadoopMode =>
+          // come up with unique temporary filename, use the config here
+          // TODO: refactor into TemporarySequenceFile class
+          val tmpDir = conf.get("hadoop.tmp.dir")
```
johnynek: can we factor this into:
```scala
def path(c: Config): String = 
  ...
```
at line 492, and call it in both read/write?

johnynek:
can we fix `.writeThrough` to explicitly use just one `WriteExecution` rather than `write.map` which may or may not be optimized with zip?

johnynek:
big improvement. I bet people will see improved flow plans due to this.

ianoc:
Excellent, thanks for all the reviewing!

PR #1415, opened 2015-08-11
Cache the zipped up write executions

ianoc:


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -291,8 +294,23 @@ object Execution {
      */
     def finished(): Unit = messageQueue.put(Stop)
 
-    def getOrElseInsert[T](ex: Execution[T],
-      res: => Future[(T, ExecutionCounters)])(implicit ec: ConcurrentExecutionContext): Future[(T, ExecutionCounters)] = {
+    def getOrLock(write: ToWrite)(implicit ec: ConcurrentExecutionContext): Either[Promise[ExecutionCounters], Future[ExecutionCounters]] = {
```
johnynek: do we need the `implicit ec`?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -490,20 +511,51 @@ object Execution {
...
+    }
+
+    // We look up to see if any of our ToWrite elements have already been ran
+    // if so we remove them from the cache.
+    // Anything not already ran we run as part of a single flow def, using their combined counters for the others
+    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      val cacheLookup: List[(ToWrite, Either[Promise[ExecutionCounters], Future[ExecutionCounters]])] = (head :: tail).map{ tw => (tw, cache.getOrLock(tw)) }
+      val (weDoOperation, someoneElseDoesOperation) = cacheLookup.partition(_._2.isLeft)
+
+      val localFlowDefCountersFuture: Future[ExecutionCounters] = if (!weDoOperation.isEmpty) {
```
johnynek: `if(weDo.nonEmpty) {`

johnynek: actually what about:
```
weDoOperation match {
  case all@(h :: tail) =>
    // no need to call head and tail

  case Nil => Future.successful(...)
}
```

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -490,20 +511,51 @@ object Execution {
...
+        val trimmed = weDoOperation.map { case (toWrite, eitherP) => toWrite }
+        val futCounters = scheduleToWrites(conf, mode, cache, trimmed.head, trimmed.tail)
+        weDoOperation.foreach {
+          case (toWrite, eitherP) =>
+            eitherP.left.get.completeWith(futCounters)
+        }
+        futCounters
+      } else Future.successful(ExecutionCounters.empty)
+
+      Future.sequence(someoneElseDoesOperation.map(_._2.right.get)).zip(localFlowDefCountersFuture).map {
```
johnynek: we should use the `failFastZip` and probably our own `sequence` that does the same:

```scala
def failFastSequence(t: Iterable[Future[T]]): Future[List[T]] = {
  t.foldLeft(Future.successful(Nil: List[T])) { (f, i) =>
      failFastZip(f, i).map { case (tail, h) => h :: tail }
    }
    .map(_.reverse)
}
```

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -490,20 +511,51 @@ object Execution {
...
+        _ = FlowStateMap.clear(flowDef)
+      } yield (ExecutionCounters.fromJobStats(jobStats))
+    }
+
+    // We look up to see if any of our ToWrite elements have already been ran
+    // if so we remove them from the cache.
+    // Anything not already ran we run as part of a single flow def, using their combined counters for the others
+    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      val cacheLookup: List[(ToWrite, Either[Promise[ExecutionCounters], Future[ExecutionCounters]])] = (head :: tail).map{ tw => (tw, cache.getOrLock(tw)) }
+      val (weDoOperation, someoneElseDoesOperation) = cacheLookup.partition(_._2.isLeft)
```
johnynek: sorry for the geekout, but can we get the types right here:
````val (weDoOperation, someoneElseDoesOperation): List[(ToWrite, Promise[...]), (ToWrite, Future[..])])``` without a cast?

I think something like:
```scala
def go[A, B, C](it: List[(A, Either[B, C])]): (List[(A, B)], List[(A, C)]) = it match {
  case (a, Left(b)) :: tail =>
    val (l, r) = go(tail)
    ((a, b) :: l, r)
  case (a, Right(c)) :: tail =>
    val (l, r) = go(tail)
    (l, (a, c) :: r)
  case Nil => (Nil, Nil)
}
```

ianoc: Which cast are you seeing? or you mean without doing the unwraps lower down for right and left?

ianoc: Updated with the go function, I think thats what you want it used for?

johnynek: just the `.left.get`

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -490,20 +521,55 @@ object Execution {
...
+            futCounters
+          case Nil => Future.successful(ExecutionCounters.empty) // No work to do, provide a fulled set of 0 counters to operate on
+        }
+
+      failFastZip(failFastSequence(someoneElseDoesOperation.map(_._2.right.get)), localFlowDefCountersFuture).map {
+        case (lCounters, fdCounters) =>
+          val summedCounters: ExecutionCounters = Monoid.sum(fdCounters :: lCounters)
+          (fn(conf, mode), summedCounters)
+      }
+    }
```
johnynek: you could `override def map` and get better composition I think.

ianoc: Does it avoid anything to inline Map instead of having the MappedExecution? if its in the presentation function we won't cache it anyway. (Indeed if the map function is expensive we would want to cache it as we do MappedExecutions right?)

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -291,8 +294,23 @@ object Execution {
...
-    def getOrElseInsert[T](ex: Execution[T],
-      res: => Future[(T, ExecutionCounters)])(implicit ec: ConcurrentExecutionContext): Future[(T, ExecutionCounters)] = {
+    def getOrLock(write: ToWrite): Either[Promise[ExecutionCounters], Future[ExecutionCounters]] = {
+      /*
+       * Since we don't want to evaluate res twice, we make a promise
+       * which we will use if it has not already been evaluated
+       */
+      val promise = Promise[ExecutionCounters]()
+      val fut = promise.future
+      toWriteCache.putIfAbsent(write, fut) match {
```
johnynek: do we want to make our two subclasses of ToWrite case classes since they are used as hashkeys now?

```
case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite {
 ...
}

case class PreparedWrite[T](fn: (Config, Mode) => SimpleWrite[T]) extends ToWrite {
 ...
}
```

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: `a.map(fn).zip(b)` would not get this optimization now, but if you inline map it would.

johnynek: also, you still cache the function, in the normal `Execution[T] => Future[T]` cache.

johnynek: right, there is a trade-off of how expensive the map function is. Right now, since you could execute prefixes twice. No big deal either way if you ask me. Maybe keep it simpler now and just stick with what you have.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -490,20 +521,65 @@ object Execution {
...
+      case Nil => (Nil, Nil)
+    }
+
+    // We look up to see if any of our ToWrite elements have already been ran
+    // if so we remove them from the cache.
+    // Anything not already ran we run as part of a single flow def, using their combined counters for the others
+    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      val cacheLookup: List[(ToWrite, Either[Promise[ExecutionCounters], Future[ExecutionCounters]])] = (head :: tail).map{ tw => (tw, cache.getOrLock(tw)) }
+      val (weDoOperation, someoneElseDoesOperation) = unwrapListEither(cacheLookup)
+
```
johnynek: is it worth checking if

```
  val otherResult = failFastSequence(someoneElseDoesOperation.map(_._2))
  otherResult.value match {
    case Some(Failure(err)) => Future.failed(err)
    case _ =>
      // not complete yet, or successful:
      // schedule the write, and do the zip on otherResult.
  }

ianoc: Thats a good idea, can just fail fast the whole process

johnynek:
:+1: nice!

this is much better: compose when possible, still guarantee each write is done at most once.

So, the lesson here: if you use `.zip` with `writeExecution` or `forceToDiskExecution` results, we should build one big flow. If you map the output, or flatMap it, all bets are off, so map after zip for best results.

ianoc:
Merged this into the Other PR for any more comments/review before going to master

PR #1416, opened 2015-08-12
Fix DateOps "match may not be exhaustive" warning

sid-kap:
This fixes the compile warning:
```
[warn] /Users/skapur/workspace/scalding/scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala:53: match may not be exhaustive.
[warn] It would fail on the following input: Some((x: String forSome x not in (DATEHOUR_WITH_DASH, DATETIME_HMSM_WITH_DASH, DATETIME_HMS_WITH_DASH, DATETIME_WITH_DASH, DATE_WITH_DASH)))
[warn]     (DateOps.getFormat(s) match {
[warn]                       ^
[warn] one warning found
```

### DIFF
--- a/scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
+++ b/scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
@@ -33,17 +33,21 @@ object DateOps extends java.io.Serializable {
   val DATETIME_HMS_WITH_DASH = "yyyy-MM-dd HH:mm:ss"
   val DATETIME_HMSM_WITH_DASH = "yyyy-MM-dd HH:mm:ss.SSS"
 
-  sealed abstract class Format(val pattern: String, val validator: Regex)
+  private[scalding] sealed abstract class Format(val pattern: String, val validator: Regex) {
+    def matches(s: String): Boolean = validator.findFirstIn(s).isDefined
+  }
 
-  object Format {
-    private val DATE_RE = """\d{4}-\d{2}-\d{2}"""
-    private val SEP_RE = """(T?|\s*)"""
+  private[scalding] object Format {
+    private val date = """\d{4}-\d{2}-\d{2}"""
+    private val sep = """(T?|\s*)"""
+    private val emptyBegin = """^\s*"""
+    private val emptyEnd = """\s*$"""
 
-    case object DATE_WITH_DASH extends Format("yyyy-MM-dd", new Regex("""^\s*""" + DATE_RE + """\s*$"""))
-    case object DATEHOUR_WITH_DASH extends Format("yyyy-MM-dd HH", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d\s*$"""))
-    case object DATETIME_WITH_DASH extends Format("yyyy-MM-dd HH:mm", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d:\d\d\s*$"""))
-    case object DATETIME_HMS_WITH_DASH extends Format("yyyy-MM-dd HH:mm:ss", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d:\d\d:\d\d\s*$"""))
-    case object DATETIME_HMSM_WITH_DASH extends Format("yyyy-MM-dd HH:mm:ss.SSS", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d:\d\d:\d\d\.\d{1,3}\s*$"""))
+    case object DATE_WITH_DASH extends Format(DateOps.DATE_WITH_DASH, new Regex(emptyBegin + date + emptyEnd))
+    case object DATEHOUR_WITH_DASH extends Format(DateOps.DATEHOUR_WITH_DASH, new Regex(emptyBegin + date + sep + """\d\d""" + emptyEnd))
+    case object DATETIME_WITH_DASH extends Format(DateOps.DATETIME_WITH_DASH, new Regex(emptyBegin + date + sep + """\d\d:\d\d""" + emptyEnd))
+    case object DATETIME_HMS_WITH_DASH extends Format(DateOps.DATETIME_HMS_WITH_DASH, new Regex(emptyBegin + date + sep + """\d\d:\d\d:\d\d""" + emptyEnd))
+    case object DATETIME_HMSM_WITH_DASH extends Format(DateOps.DATETIME_HMSM_WITH_DASH, new Regex(emptyBegin + date + sep + """\d\d:\d\d:\d\d\.\d{1,3}""" + emptyEnd))
   }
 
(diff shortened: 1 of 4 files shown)


scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
```diff
@@ -27,23 +27,19 @@ object DateOps extends java.io.Serializable {
   val PACIFIC = TimeZone.getTimeZone("America/Los_Angeles")
   val UTC = TimeZone.getTimeZone("UTC")
 
-  val DATE_WITH_DASH = "yyyy-MM-dd"
```
johnynek: these were public, can we leave them? Why incur any migration cost?

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
```diff
@@ -27,23 +27,19 @@ object DateOps extends java.io.Serializable {
...
-    DATEHOUR_WITH_DASH -> new Regex("""^\s*""" + DATE_RE +
-      SEP_RE + """\d\d\s*$"""),
-    DATETIME_WITH_DASH -> new Regex("""^\s*""" + DATE_RE +
-      SEP_RE + """\d\d:\d\d\s*$"""),
-    DATETIME_HMS_WITH_DASH -> new Regex("""^\s*""" + DATE_RE +
-      SEP_RE + """\d\d:\d\d:\d\d\s*$"""),
-    DATETIME_HMSM_WITH_DASH -> new Regex("""^\s*""" + DATE_RE +
-      SEP_RE + """\d\d:\d\d:\d\d\.\d{1,3}\s*$"""))
+
+  case object DATE_WITH_DASH extends Format("yyyy-MM-dd", new Regex("""^\s*""" + DATE_RE + """\s*$"""))
```
johnynek: can we name these something else, so they don't conflict with the old strings?

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
```diff
@@ -27,23 +27,19 @@ object DateOps extends java.io.Serializable {
...
+  case object DATEHOUR_WITH_DASH extends Format("yyyy-MM-dd HH", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d\s*$"""))
+  case object DATETIME_WITH_DASH extends Format("yyyy-MM-dd HH:mm", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d:\d\d\s*$"""))
+  case object DATETIME_HMS_WITH_DASH extends Format("yyyy-MM-dd HH:mm:ss", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d:\d\d:\d\d\s*$"""))
+  case object DATETIME_HMSM_WITH_DASH extends Format("yyyy-MM-dd HH:mm:ss.SSS", new Regex("""^\s*""" + DATE_RE + SEP_RE + """\d\d:\d\d:\d\d\.\d{1,3}\s*$"""))
+
+  val formats = List(DATE_WITH_DASH, DATEHOUR_WITH_DASH, DATETIME_WITH_DASH, DATETIME_HMS_WITH_DASH, DATETIME_HMSM_WITH_DASH)
```
johnynek: let's add a type here or make it private.

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
```diff
@@ -51,13 +47,15 @@ object DateOps extends java.io.Serializable {
   /**
    * Return the guessed format for this datestring
    */
-  def getFormat(s: String): Option[String] =
-    DATE_FORMAT_VALIDATORS.find { _._2.findFirstIn(prepare(s)).isDefined }.map(_._1)
+  def getFormat(s: String): Option[Format] =
```
sid-kap: Looks like this is a public method. So maybe it would be a better idea to add a new method of signature `String -> Option[Format]` rather than changing the return type of this method?

sid-kap:
I undid some of the changes so that we don't break public APIs.
Should we make the Format class and Format object `private[scalding]` to minimize the surface area here?

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
```diff
@@ -33,26 +33,38 @@ object DateOps extends java.io.Serializable {
...
-    DATETIME_HMS_WITH_DASH -> new Regex("""^\s*""" + DATE_RE +
-      SEP_RE + """\d\d:\d\d:\d\d\s*$"""),
-    DATETIME_HMSM_WITH_DASH -> new Regex("""^\s*""" + DATE_RE +
-      SEP_RE + """\d\d:\d\d:\d\d\.\d{1,3}\s*$"""))
+  sealed abstract class Format(val pattern: String, val validator: Regex)
+
+  object Format {
+    private val DATE_RE = """\d{4}-\d{2}-\d{2}"""
+    private val SEP_RE = """(T?|\s*)"""
+
```
johnynek: while we're at it, can you add `private val emptyBegin = """^\s*"""` and use it below. Also, can you refer to the `DateOps.DATETIME_...` strings instead of repeating them below?

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
```diff
@@ -33,26 +33,38 @@ object DateOps extends java.io.Serializable {
...
   }
+
+  /**
+   * Return the guessed format for this datestring
+   */
+  def getFormatObject(s: String): Option[Format] = {
+    val formats: List[Format] = List(Format.DATE_WITH_DASH, Format.DATEHOUR_WITH_DASH,
+      Format.DATETIME_WITH_DASH, Format.DATETIME_HMS_WITH_DASH, Format.DATETIME_HMSM_WITH_DASH)
+
+    formats.find { _.validator.findFirstIn(prepare(s)).isDefined }
```
johnynek: can we add this method to `Format` directly:

```scala
def matches(s: String): Boolean = validator.findFirstIn(s).isDefined
```
so this becomes
```scala
formats.find(_.matches(prepare(s)))
```

johnynek:
About making `Format` private, in general, I think it's a good idea unless we really think this adds value. Since there are not many (any?) methods on Format, I'd say go private.

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
sid-kap: Done

scalding-date/src/main/scala/com/twitter/scalding/DateOps.scala
sid-kap: Done. I was able to factor out `emptyEnd = """\s*$"""` as well

johnynek:
merge when green

PR #1419, opened 2015-08-12
Remove unneeded Mode

johnynek:
While looking at some code for someone, I noticed TypedPipeInst does not need or really use the Mode parameter it carries around.

I think Mode is only used by TypedPipe to create Pipes from Iterable, which we do differently in cascading local mode, and on hadoop.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/CoGrouped.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/CoGrouped.scala
@@ -306,7 +306,7 @@ trait CoGrouped[K, +R] extends KeyedListLike[K, R, CoGrouped] with CoGroupable[K
         newPipe.project('key, 'value)
       }
       //Construct the new TypedPipe
-      TypedPipe.from[(K, R)](pipeWithRedAndDescriptions, ('key, 'value))(flowDef, mode, tuple2Converter)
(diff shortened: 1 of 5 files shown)


johnynek:
this is probably of limited value, but I just wanted to document it.

I'm not sure it is worth it to merge it since there is some chance of incompatibility for callers that passed some args explicitly. Also this is a little weaker on safety because the `checkMode` method is removed, which gave a saner error message if the mode (somehow, though I don't think it has ever happened) changed over the scope of the job.

ianoc:
I dunno, I actually like the clean up if we aren't using it. the next scalding release will be binary incompatible already i'm pretty sure right? so if we want to get this clean up in no time like the present -- it might be worth a test drive internally since we have so much scalding code to get a representative idea as to how often this might break users

ianoc:
Also can we add back in the old from method and just call the new one without the parameter marking the old as deprecated? TypedPipeInst is package private so we probably get most of our benefits in cleanups without breaking people?

johnynek:
yeah, doing the internal test drive seems reasonable.

On Wed, Aug 12, 2015 at 5:23 PM, ianoc <notifications@github.com> wrote:

> Also can we add back in the old from method and just call the new one
> without the parameter marking the old as deprecated? TypedPipeInst is
> package private so we probably get most of our benefits in cleanups without
> breaking people?
>
> —
> <https://github.com/twitter/scalding/pull/1419#issuecomment-130489110>.
>



-- 

PR #1423, opened 2015-08-15
Adds much more line number information through the NoStackAndThen class

ianoc:
Found a bunch of cases when we would just be showing toPipe and asPipe in the descriptions. In these cases the scalding line number doesn't actually seem to be helpful at all. It seemed to be as a result of passing through the NoStackAndThen class, so thread some line data through there is the goal here.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -94,7 +94,7 @@ object TypedPipe extends Serializable {
       def mapped = pipe
       def keyOrdering = ord
       def reducers = None
-      val descriptions: Seq[String] = List(LineNumber.tryNonScaldingCaller.toString)
+      val descriptions: Seq[String] = LineNumber.tryNonScaldingCaller.map(_.toString).toList
       def joinFunction = CoGroupable.castingJoinFunction[V]
     }
 
@@ -161,7 +161,7 @@ trait TypedPipe[+T] extends Serializable {
     import Dsl._
     // Ensure we hook into all pipes coming out of the typed API to apply the FlowState's properties on their pipes
     val pipe = asPipe[U](fieldNames).applyFlowConfigProperties(flowDef)
-    RichPipe.setPipeDescriptions(pipe, List(LineNumber.tryNonScaldingCaller.toString))
+    RichPipe.setPipeDescriptionFrom(pipe, LineNumber.tryNonScaldingCaller)
   }
 
   /**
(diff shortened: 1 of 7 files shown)


johnynek:
Is there any test we can add to make sure we don't regress here?

ianoc:
The tests I've managed to do so far I couldn't repo it easily in alas :/ It occurs in some of the execution tests in our test suite. But not too many others. I'll have another look later see if i can do something

johnynek:
If we can't find any test, but you still have evidence this is better, it's fine with me to merge. I'll leave it to you. Obviously it would be better to find some kind of test since we are at a stage where changes easily cause regressions.

ianoc:
@johnynek I have a test that requires the new code to pass here. Though its not quite the problem I set out to solve(and this does solve). It tests probably a stronger contract, if we pass through a NoStackAndThen operation we pick up some line number info, so if we toPipe elsewhere its all kept. Futures are just there to take out any stack traces we can. -- Where I mostly see line number info being lost is in executions futures where after a NoStackAndThen there is no user code on the stacktrace.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -894,23 +894,32 @@ class TypedPipeFactory[T] private (@transient val next: NoStackAndThen[(FlowDef,
...
-  private def unwrap(pipe: TypedPipe[T])(implicit flowDef: FlowDef, mode: Mode): TypedPipe[T] = pipe match {
-    case TypedPipeFactory(n) => unwrap(n(flowDef, mode))
-    case tp => tp
+  private def unwrap(pipe: TypedPipe[T], st: Array[StackTraceElement])(implicit flowDef: FlowDef, mode: Mode): (TypedPipe[T], Array[StackTraceElement]) = pipe match {
+    case TypedPipeFactory(n) =>
+      val fullTrace = n match {
+        case NoStackAndThen.WithStackTrace(_, st) => st
+        case _ => Array[StackTraceElement]()
+      }
+      unwrap(n(flowDef, mode), st ++ fullTrace)
```
johnynek: does the order matter here? Seems to me, each unwrapping might be a different line number. Is that not the case? Do we not want a `List[Array[StackTraceElement]]` and just concat all of them and try to get line numbers out of each?

ianoc: We could get a line number out of each, possibly getting several line numbers. That might be better really, but it could appear more noisey. This was just attempting to do a priority of the first thing you did to the last, but thats pretty arbitrary

johnynek:
sounds good to me for now. We can keep improving. If you're happy with it, merge away!

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
ianoc: Attempts to preserve everything we know now, other than the line numbers being off a little I think in the scala compiler it looks decent

TypedPipe.from(List(1,2,3))
   .map()

reports both line numbers are with the .map

PR #1424, opened 2015-08-17
Allow constructing a PailSource from a PailSpec

amazari:
By allowing a PailSource to be constructed from a PailSpec, this patch gives more flexibility to client code. For example, it is is now possible to set compression and/or sizing policy.
For backward compatibility sake, it is still possible to construct a source from a PailStructure. In that case, a PailSpec is materialized using PailTap.makeSpec facility in the same manner it was previously done.

### DIFF
--- a/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/PailSource.scala
+++ b/scalding-commons/src/main/scala/com/twitter/scalding/commons/source/PailSource.scala
@@ -161,11 +161,12 @@ object PailSource {
   }
 }
 
-class PailSource[T] private (rootPath: String, spec: PailSpec, subPaths: Array[List[String]] = Array.empty)(implicit conv: TupleConverter[T])
-  extends Source with Mappable[T] {
+class PailSource[T] private (rootPath: String, spec: PailSpec, subPaths: Array[List[String]] = Array.empty)(implicit conv: TupleConverter[T], tset: TupleSetter[T])
+  extends Source with Mappable[T] with TypedSink[T] {
   import Dsl._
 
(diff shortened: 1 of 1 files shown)


scalding-commons/src/main/scala/com/twitter/scalding/commons/source/PailSource.scala
```diff
@@ -153,7 +169,6 @@ class PailSource[T] private (rootPath: String, structure: PailStructure[T], subP
   val fieldName = "pailItem"
 
   lazy val getTap = {
-    val spec = PailTap.makeSpec(null, structure)
     val javaSubPath = if ((subPaths == null) || (subPaths.size == 0)) null else subPaths map { _.asJava }
```
ianoc: Since the constructor setting up the null for the list is private, and we only seem to care if the path is empty or not. Can we just default subPaths to an empty list and not need to do null checking?

scalding-commons/src/main/scala/com/twitter/scalding/commons/source/PailSource.scala
```diff
@@ -145,7 +161,7 @@ object PailSource {
   }
 }
 
-class PailSource[T] private (rootPath: String, structure: PailStructure[T], subPaths: Array[List[String]] = null)(implicit conv: TupleConverter[T])
+class PailSource[T] private (rootPath: String, spec: PailSpec, subPaths: Array[List[String]] = Array.empty)(implicit conv: TupleConverter[T])
   extends Source with Mappable[T] {
```
johnynek: can you add with `TypedSink[T]`? Is there any reason why you can't write to this?

PR #1425, opened 2015-08-18
Requires a DateRange's "end" to be after its "start"

benpence:


johnynek:
merge when green

scalding-date/src/main/scala/com/twitter/scalding/DateRange.scala
```diff
@@ -73,6 +73,7 @@ object DateRange extends java.io.Serializable {
  */
 case class DateRange(val start: RichDate, val end: RichDate) {
   import DateOps._
+  require(start <= end)
```
rubanm: An error message here would be nice
```scala
require(start <= end, "end of date range must not occur before the start")
```

rubanm:
+1

benpence:
Should I add an error message? I just remembered you could do that

johnynek:
Yes;

require(start <= end, s"$start is not on or before $end")

On Mon, Aug 17, 2015 at 2:10 PM, Ben Pence <notifications@github.com> wrote:

> Should I add an error message? I just remembered you could do that
>
> —
> <https://github.com/twitter/scalding/pull/1425#issuecomment-132000719>.
>



-- 

PR #1426, opened 2015-08-18
Scalding viz options

ianoc:


scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -89,8 +92,30 @@ trait ExecutionContext {
       // which instantiates and runs them
       mode match {
         case _: HadoopMode =>
-          config.get(Config.ReducerEstimators)
-            .foreach(_ => flow.setFlowStepStrategy(ReducerEstimatorStepStrategy))
+          val reducerEstimatorStrategy: Seq[FlowStepStrategy[JobConf]] = config.get(Config.ReducerEstimators).toList.map(_ => ReducerEstimatorStepStrategy)
+
+
+          val otherStrategies: Seq[FlowStepStrategy[JobConf]] = config.getFlowStepStrategies.map { tTry: Try[(Mode, Config) => FlowStepStrategy[JobConf]] =>
```
johnynek: why take the mode as a parameter if we only pass it a HadoopMode?  Why not do all these in any case and let the user attach an empty FlowListener/etc... (which we can even provide, as there is a monoid on these things).

project/Build.scala
```diff
@@ -25,7 +25,7 @@ object ScaldingBuild extends Build {
   val avroVersion = "1.7.4"
   val bijectionVersion = "0.8.1"
   val cascadingAvroVersion = "2.1.2"
-  val chillVersion = "0.7.0"
+  val chillVersion = "0.7.1-t1439774596000-77b8a73d7643263249f76ae03c0d3f03db330fc3"
```
johnynek: just to be clear, we wouldn't merge this yet, right? This is just to get comments?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
ianoc: Can move it to hadoop mode sure.

You mean have a default strategy so its not an option below?

project/Build.scala
ianoc: Yes exactly, this version isn't on maven central, so this won't pass tests.  -- branch hasn't merged in Chill either, and no rls there.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -476,4 +532,11 @@ object Config {
     is.close()
     md5Hex(bytes)
   }
+
+  private[this] def buildInj[T: ExternalizerInjection: ExternalizerCodec]: Injection[T, String] =
```
johnynek: while we're at it, why not stack in GzippedBase64String:

https://github.com/twitter/bijection/blob/develop/bijection-core/src/main/scala/com/twitter/bijection/BinaryBijections.scala#L33

Might make the jobconfs slightly smaller.

PR #1427, opened 2015-08-18
[DONT_MERGE] Pass scalding descriptions to Driven

MansurAshraf:
This is an attempt to pass scalding description to Driven and it sort of works.

constrains:
1) Driven annotation can only go on a subclass of `Operations`. 
2) I am only able to set descriptions for 'Each' n 'Every' as they are the only subclasses of 'Operator'. I cant figure out how to pass description from a 'Group' or 'Join' Operation
3) Line numbers are unfortunately incorrect (but this is probably a bug in scalding??)

so for a Scalding Job like the one below
screen shot 2015-08-18 at 12 17 51 pm

this is how descriptions show up in Driven
screen shot 2015-08-18 at 12 17 01 pm

Thanks @cwensel for answering my questions, I would appreciate any help to make this work for GroupBy etc.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
@@ -15,6 +15,7 @@ limitations under the License.
 */
 package com.twitter.scalding {
 
+  import cascading.management.annotation.{ Visibility, PropertyDescription, Property }
   import cascading.operation._
   import cascading.tuple._
   import cascading.flow._
@@ -32,6 +33,12 @@ package com.twitter.scalding {
       RuntimeStats.addFlowProcess(flowProcess)
       super.prepare(flowProcess, operationCall)
     }
+
+    var metaData: String = _
+
+    @Property(name = "line.number", visibility = Visibility.PRIVATE)
+    @PropertyDescription("scalding")
+    def getLineNumber: String = metaData
   }
 
   class FlatMapFunction[S, T](@transient fn: S => TraversableOnce[T], fields: Fields,
--- a/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
@@ -34,11 +34,11 @@ package com.twitter.scalding {
       super.prepare(flowProcess, operationCall)
     }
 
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -32,6 +33,12 @@ package com.twitter.scalding {
       RuntimeStats.addFlowProcess(flowProcess)
       super.prepare(flowProcess, operationCall)
     }
+
+    val descriptions: scala.collection.mutable.Set[String] = scala.collection.mutable.Set()
```
johnynek: can this be `private[scalding]` or something?

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -82,9 +82,17 @@ object RichPipe extends java.io.Serializable {
...
 
   def setPipeDescriptions(p: Pipe, descriptions: Seq[String]): Pipe = {
+    val combinedDescriptions = getPipeDescriptions(p) ++ descriptions
     p.getStepConfigDef().setProperty(
       Config.PipeDescriptions,
-      encodePipeDescriptions(getPipeDescriptions(p) ++ descriptions))
+      encodePipeDescriptions(combinedDescriptions))
+
+    p match {
+      case pp: Operator => if (pp.getOperation.isInstanceOf[ScaldingPrepare[_]]) {
```
johnynek: minor style. can we do:

```scala
case pp: Operator =>
  pp.getOperation match {
    case sp: ScaldingPrepare[_] => sp.descriptions ++= combinedDescriptions
    case _ => ()
  }
  case _ => ()
```

cwensel:
In a near future version (2.7.x) I think we can get away with adding `getDescription()` and `setDescription()` methods directly on `Pipe`. 

But we will need to update the `Property` annotation to have an *ignore null properties* flag so we don't store `description=null` a billion times when it is unused.

This will also require a minor/maintenance release of Driven to honor the annotation, and to look for annotations directly on all `Pipe` instances. but all that is just logistics, nothing complicated.

johnynek:
going to close this as it seems to have gone stale.

PR #1428, opened 2015-08-24
Use JobConf input files list for input size computation used by ReducersEstimators

amazari:
Instead of globing in Tap.getPath, use the fully resolved input paths
found in JobConf.
It fixes the issue with some (badly behaving?) Taps, like PailTap, where
Tap.getPath/\* != [actual list of paths].

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
@@ -5,7 +5,7 @@ import com.twitter.algebird.Monoid
 import com.twitter.scalding.{ StringUtility, Config }
 import cascading.tap.{ Tap, CompositeTap }
 import cascading.tap.hadoop.Hfs
-import org.apache.hadoop.mapred.JobConf
+import org.apache.hadoop.mapred.{ FileInputFormat, InputFormat, JobConf }
 import org.slf4j.LoggerFactory
 import java.util.{ List => JList }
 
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -28,37 +28,14 @@ object EstimatorConfig {
...
-  def inputSizes(step: FlowStep[JobConf]): Seq[(String, Long)] = {
-    val conf = step.getConfig
-    unrollTaps(step).flatMap {
-      case tap: Hfs => Some(tap.toString -> size(tap, conf))
-      case _ => None
-    }
-  }
-
-  def totalInputSize(step: FlowStep[JobConf]): Long = inputSizes(step).map(_._2).sum
+  def inputSizes(conf: JobConf) =
```
rubanm: Can you add back the return type here

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -28,30 +28,26 @@ class InputSizeReducerEstimator extends ReducerEstimator {
    * Figure out the total size of the input to the current step and set the number
    * of reducers using the "bytesPerReducer" configuration parameter.
    */
-  override def estimateReducers(info: FlowStrategyInfo): Option[Int] =
-    Common.inputSizes(info.step) match {
-      case Nil =>
-        LOG.warn("InputSizeReducerEstimator unable to estimate reducers; " +
-          "cannot compute size of:\n - " +
-          Common.unrollTaps(info.step).filterNot(_.isInstanceOf[Hfs]).mkString("\n - "))
-        None
```
rubanm: I think we still need the Nil case. There can be steps with non-Hfs source taps only. How about keeping the `unrollTaps` implementation for logging purposes here and using the new `inputSizes` for computing input sizes?

PR #1432, opened 2015-08-26
Fix Readme travis link

ianoc:
Was reporting that the build was failing if the most recent build was red, regardless of what branch


Gabriel439:
:ship:

PR #1434, opened 2015-09-03
Don't squash the exception in history service when there's a failure

isnotinvain:
exception's .toString only gives you the message, passing the exception to the logger will surface the nested errors as well which is what you care about in this case.


rubanm:
LGTM

PR #1436, opened 2015-09-04
make getBytesPerReducer support human readable values like 128m and 1g

epishkin:
Supported formats are <code>long</code> or human readable format.
For human readable format you can use the following suffix (case insensitive): k(kilo), m(mega), g(giga), t(tera), p(peta), e(exa).

Examples: 1024, 128m, 1g.

@bholt @sid-kap @sagemintblue @rubanm guys could you review?


rubanm:
@epishkin looks good. Thanks.

PR #1439, opened 2015-09-16
Fix `readPathsFor` to use the `tz` argument

Gabriel439:
`readPathsFor` is given a timezone argument, but the default implementation internally ignores the argument and hard codes `DateOps.UTC`.  This change updates it to thread the time zone argument to `stepSize` and `allPathsWithDuration`


rubanm:
+1

benpence:
+1

PR #1440, opened 2015-09-22
Scalding viz options

MansurAshraf:
Continuation of https://github.com/twitter/scalding/pull/1426 with published Chill version

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.scalding
 
 import cascading.flow.hadoop.HadoopFlow
-import cascading.flow.{ FlowDef, Flow }
+import cascading.flow.{ Flow, FlowDef, FlowListener, FlowStepListener }
 import cascading.flow.planner.BaseFlowStep
 import cascading.pipe.Pipe
 import com.twitter.scalding.reducer_estimation.ReducerEstimatorStepStrategy
@@ -91,6 +91,23 @@ trait ExecutionContext {
         case _: HadoopMode =>
           config.get(Config.ReducerEstimators)
             .foreach(_ => flow.setFlowStepStrategy(ReducerEstimatorStepStrategy))
+
+          config.get(Config.FlowListeners).foreach { clsNames: String =>
+            println(s"XXX adding FlowListeners $clsNames")
+            val clsLoader = Thread.currentThread.getContextClassLoader
+            StringUtility.fastSplit(clsNames, ",")
+              .map { clsLoader.loadClass(_).newInstance.asInstanceOf[FlowListener] }
+              .map { flow.addListener(_) }
+          }
+
+          config.get(Config.FlowStepListeners).foreach { clsNames: String =>
+            println(s"XXX adding FlowStepListeners $clsNames")
+            val clsLoader = Thread.currentThread.getContextClassLoader
+            StringUtility.fastSplit(clsNames, ",")
+              .map { clsLoader.loadClass(_).newInstance.asInstanceOf[FlowStepListener] }
+              .map { flow.addStepListener(_) }
+          }
+
         case _ => ()
       }
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
@@ -25,6 +25,7 @@ import org.apache.hadoop.mapred.JobConf
 import scala.collection.JavaConverters._
 import scala.concurrent.Future
 import scala.util.{ Failure, Success, Try }
+import org.slf4j.{ Logger, LoggerFactory }
 
 /*
  * This has all the state needed to build a single flow
@@ -35,6 +36,8 @@ trait ExecutionContext {
   def config: Config
   def flowDef: FlowDef
   def mode: Mode
+
+  import ExecutionContext._
 
   private def getIdentifierOpt(descriptions: Seq[String]): Option[String] = {
     if (descriptions.nonEmpty) Some(descriptions.distinct.mkString(", ")) else None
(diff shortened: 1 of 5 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -521,11 +521,11 @@ trait CounterVerification extends Job {
   }
 }
 
-private[scalding] object FlowStepStrategies {
+private[scalding] case class FlowStepStrategies[A]() extends Semigroup[FlowStepStrategy[A]] {
```
rubanm: Should this be a Semigroup? Looks like a side-effecting function here. Are we guaranteed associativity?

MansurAshraf: I think strategies will be applied in a sequence so it is associative (and not commutative)? /cc @ianoc  as he wrote this.

Gabriel439: Yep, this is a semigroup because `plus` is assocative

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -89,8 +92,29 @@ trait ExecutionContext {
...
+            t(mode, configWithId)
+          }
+
+          val optionalFinalStrategy = FlowStepStrategies().sumOption(reducerEstimatorStrategy ++ otherStrategies)
+
+          optionalFinalStrategy.foreach { strategy =>
+            flow.setFlowStepStrategy(strategy)
+          }
+
+          config.getFlowListeners.foreach { tTry: Try[(Mode, Config) => FlowListener] =>
```
johnynek: I prefer the style:
```scala
.foreach {
  case Success(fn) => flow.addListener(fn(mode, configWithId))
  case Failure(err) => throw new Exception(s"Failed to decode flow listener: $err")
}
```

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -89,8 +92,29 @@ trait ExecutionContext {
...
+            val t: (Mode, Config) => FlowListener = tTry.getOrElse(throw new Exception(s"Failed to decode flow listener $tTry when submitting job"))
+            flow.addListener(t(mode, configWithId))
+          }
+
+          config.getFlowStepListeners.foreach { tTry: Try[(Mode, Config) => FlowStepListener] =>
```
johnynek: similar, can we use the case approach here?

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
johnynek: It's not side-effecting. Nothing happens in plus. It is just creating a new strategy. Looks correct to me.

Gabriel439: Making an analogy to Haskell, it's equivalent to the semigroup over this type:

```haskell
newtype FlowStepStrategy a =
    FlowStepStrategy (Flow a -> JList (FlowStep a) -> FlowStep a -> IO ())

instance Monoid (FlowStepStrategy a) where
    mempty = FlowStepStrategy (\_ _ _ -> return ())

    mappend (FlowStepStrategy f1) (FlowStepStrategy f2) = FlowStepStrategy f
      where
        f x y z = do
            f1 x y z
            f2 x y z
```

... and that's guaranteed to satisfy the `Monoid` laws in both Haskell and Scala.

The most common source of this thing failing in Scala specifically is if side effects are triggered prematurely by evaluating things in the wrong order but that does not appear to be the case for this code.

rubanm: Ah ok. What I was confused about was if plus(a, plus(b, c)).apply and plus(plus(a, b), c).apply will have the same effect.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -89,8 +92,29 @@ trait ExecutionContext {
       // which instantiates and runs them
       mode match {
         case _: HadoopMode =>
-          config.get(Config.ReducerEstimators)
-            .foreach(_ => flow.setFlowStepStrategy(ReducerEstimatorStepStrategy))
+          val reducerEstimatorStrategy: Seq[FlowStepStrategy[JobConf]] = config.get(Config.ReducerEstimators).toList.map(_ => ReducerEstimatorStepStrategy)
+
+          val otherStrategies: Seq[FlowStepStrategy[JobConf]] = config.getFlowStepStrategies.map { tTry: Try[(Mode, Config) => FlowStepStrategy[JobConf]] =>
+            val t: (Mode, Config) => FlowStepStrategy[JobConf] = tTry.getOrElse(throw new Exception(s"Failed to decode flow step strategy $tTry when submitting job"))
```
johnynek: why didn't you use the case style here?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -89,8 +92,29 @@ trait ExecutionContext {
...
+          }
+
+          config.getFlowListeners.foreach {
+            case Success(fn) => flow.addListener(fn(mode, configWithId))
+            case Failure(e) => throw new Exception(s"Failed to decode flow listener: $e")
+          }
+
+          config.getFlowStepListeners.foreach {
+            case Success(fn) => flow.addStepListener(fn(mode, configWithId))
+            case f @ Failure(_) => new Exception(s"Failed to decode flow step listener $f when submitting job")
```
johnynek: inconsistent with above (you only put the exception, here you put the Failure wrapper).

I think it is better to do:
```scala
case Failure(err) => new Exception("Failed to decode flow step listener", err)
```
so we can see that this exception was caused by a previous exception.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
```diff
@@ -89,8 +92,29 @@ trait ExecutionContext {
...
+
+          val optionalFinalStrategy = FlowStepStrategies().sumOption(reducerEstimatorStrategy ++ otherStrategies)
+
+          optionalFinalStrategy.foreach { strategy =>
+            flow.setFlowStepStrategy(strategy)
+          }
+
+          config.getFlowListeners.foreach {
+            case Success(fn) => flow.addListener(fn(mode, configWithId))
+            case Failure(e) => throw new Exception(s"Failed to decode flow listener: $e")
```
johnynek: think we should use the caused by constructor, see below.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala
johnynek: also, use caused-by constructor (see below).

johnynek:
Lgtm. What's up with the version number? What does exec mean here?

rubanm:
+1

PR #1443, opened 2015-09-24
No Source validation when running in TestMode

MansurAshraf:
No Source validation when running in TestMode. https://github.com/twitter/scalding/pull/1441 broke bunch of tests internally.


MansurAshraf:
running internal integration suits, dont merge

Gabriel439:
Is this safe to disable?  I don't know a whole lot about the tests, but I want to make sure that this isn't disabling validations that other tests are using to detect regressions

Gabriel439:
It also seems like a really weird corner case.  If we do include this as a temporary fix there needs to be an explanatory comment explaining why it's there and what work needs to be done to remove it.  Otherwise somebody with no context won't know if it's safe to tamper with that.

MansurAshraf:
@Gabriel439 this was the exact behavior before https://github.com/twitter/scalding/pull/1441 got merged. As in we were not calling Validate in JobTEst

Gabriel439:
Alright.  There should still probably be a comment explaining why these specific test validations are being blacklisted

PR #1446, opened 2015-10-21
[DO NOT MERGE] Trying cascading3

cchepelov:
A quick-n-dirty attempt at using Cascading3 in scalding, at the source level. 

At this point, this will choke quickly based on a few API that either disappeared or moved within Cascading; specifically, FlowStep#getSources and FlowStep#getGraph. 

Really only a quick RFC at this point, will need to investigate how to best to support the "missing" APIs 
(suspecting this is really a matter of adapting to the new API made just to help building ExecutionContext.getDesc in a nicer way)


scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -101,15 +103,21 @@ trait Config extends Serializable {
   def setMapSpillThreshold(count: Int): Config =
     this + (SpillableProps.MAP_THRESHOLD -> count.toString)
 
+  @deprecated("deprecated in Cascading 2.7 and dropped in Cascading 3.0, use setMapSideAggregationCapacity", "cascading 2.7")
+  def setMapSideAggregationThreshold(count: Int): Config = {
```
ianoc: This isn't really correct for scalding, the comment mentions the method is realted to the associated operations which wires down to https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala#L126 which uses the same key as cascading. Not sure we shouldn't just be keeping the method/key as it was just defining the constant in scalding instead

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -134,7 +134,7 @@ package com.twitter.scalding {
     val boxedSemigroup = Externalizer(commutativeSemigroup)
 
     val DEFAULT_CACHE_SIZE = 100000
-    val SIZE_CONFIG_KEY = AggregateBy.AGGREGATE_BY_THRESHOLD
+    val SIZE_CONFIG_KEY = AggregateByProps.AGGREGATE_BY_CAPACITY
```
ianoc: This is going to break lots of users, so lets try not do this. any usage of -D on the commandline will be broken

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/OrderedBufferableProviderImpl.scala
```diff
@@ -16,7 +16,7 @@
 package com.twitter.scalding.serialization.macros.impl
 
 import scala.language.experimental.macros
-import scala.reflect.macros.Context
+import scala.reflect.macros.whitebox.Context
```
ianoc: Please undo non scala 2.10 compatible changes, they don't seem like they should be in this PR anyway

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
johnynek: +1 to what @ianoc said. I'd rather us take the max value of the two keys in Operations.scala, and here set the non-deprecated internally, but not deprecate our API (instead explain what it means to the user, not necessarily in cascading terms).

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
johnynek: I'd rather take the max in cacheSize (maybe logging a warning if two are set).

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/OrderedBufferableProviderImpl.scala
johnynek: we can possibly use https://github.com/milessabin/macro-compat if we want to escape any 2.11/2.12 warnings here, but I'd really like to keep 2.10 support as long as we can easily do so (even though twitter is actually on 2.11 these days).

cchepelov:
Thanks for the prompt feedback, @ianoc and @johnynek ! Will address the various points now, then proceed with the rest

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/OrderedBufferableProviderImpl.scala
cchepelov: Indeed, this was reckless. Reverting now.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
cchepelov: Trick is, AggregateBy.AGGREGATE_BY_THRESHOLD has disappeared from the Cascading API. We can of course read it and carry that value forward if AggregateByProps.AGGREGATE_BY_CAPACITY has been left unset by the user; is that what you'd like me to do?

Do you want me to bring that constant back (within scalding only), take AggregateByProps.AGGREGATE_BY_CAPACITY if defined or else take AggregateBy.AGGREGATE_BY_THRESHOLD (logging a warning) ?

@johnynek just to make sure I understand, when you write " (instead explain what it means to the user, not necessarily in cascading terms)", do you mean logging a warning, or leaving a comment in the code?

cchepelov:
… Right now, busy updating quite a few dependencies to Cascading3 (cascading-parquet, cascading-avro, elephant-bird-cascading3, etc. 
To be continued!

cchepelov:
@ianoc @johnynek Done for now. Trouble is, this requires that code from a few dependencies' pull requests be available at compile time, and I'm not quite sure how best to convince Travis to do this:

* https://github.com/ScaleUnlimited/cascading.avro/pull/44
* https://github.com/apache/parquet-mr/pull/284
* https://github.com/nathanmarz/dfs-datastores/pull/53
* https://github.com/twitter/elephant-bird/pull/454

For that reason, the code currently fails to run on Travis.

cchepelov:
Now working off binaries dumped onto conjars.org for patched dependencies.

There is an outstanding problem at test stage: some Test Jobs seem to deadlock, apparently within the Cascading test driver. Investigating this week (hopefully).

cchepelov:
synced with 0.15.1-RC1+ at 55f03f7e2ba363b26556b5390c77220fdf8ee637

cchepelov:
The deadlock in SkewJoinPipeTest should hopefully be fixed by https://github.com/cwensel/cascading/pull/47 (unless that one breaks more things elsewhere ☺).

cchepelov:
Update: some tests fixed with respect to .WithDescription("hi there")
(9896ed9 will not build on Travis for now, awaiting for successful integration of https://github.com/cwensel/cascading/pull/47 )

cc @simondumas

project/Build.scala
```diff
@@ -45,15 +45,18 @@ object ScaldingBuild extends Build {
...
+  /* NOTE: the temp.cchepelov.* groupIds are to let the scalding build machine access the patched upstream dependencies until they get merged. 
+    This *must* be removed before proceeding. */
 
   val printDependencyClasspath = taskKey[Unit]("Prints location of the dependencies")
 
   val sharedSettings = Project.defaultSettings ++ assemblySettings ++ scalariformSettings ++ Seq(
     organization := "com.twitter",
 
-    scalaVersion := "2.10.6",
+    scalaVersion := "2.10.5",
```
johnynek: any reasons for this?

project/Build.scala
```diff
@@ -319,15 +324,16 @@ object ScaldingBuild extends Build {
 
   lazy val scaldingCommons = module("commons").settings(
     libraryDependencies ++= Seq(
-      "com.backtype" % "dfs-datastores-cascading" % dfsDatastoresVersion,
-      "com.backtype" % "dfs-datastores" % dfsDatastoresVersion,
+      "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
+      "temp.cchepelov.com.backtype" % "dfs-datastores-cascading3" % dfsDatastoresVersion, // FIXME: https://github.com/nathanmarz/dfs-datastores/pull/53
```
johnynek: #1473 should fix this

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
johnynek: Why not call the capacity variant?

scalding-core/src/main/scala/com/twitter/scalding/ConfigBridge.scala
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Created by cchepelov on 13/01/16.
+ */
+object ConfigBridge {
+  /**
+   * This adapter handles the various vessel types that can be used to configure properties of a a
+   * FlowStep[_] since Cascading 3.0
+   */
+  class FlowStepAdapter(flowStep: FlowStep[_]) {
```
johnynek: extends Any

project/Build.scala
cchepelov: #1473 sounds like a good idea, do you want me to pull https://github.com/rubanm/scalding/tree/rubanm/dfs_datastores into this?

project/Build.scala
cchepelov: no, oversight. reverting.

project/Build.scala
rubanm: @cchepelov  I'm running some tests against that branch, so it should get merged to develop soonish. Go ahead and merge that branch if that keeps this PR moving though. Thanks.

johnynek:
I think #1586 may actually be getting close to merge... I'm closing this in favor of that.

If that is not right. Reopen. Thanks a ton @cchepelov for kicking this all off. I'm sorry it has taken so long.

cchepelov:
You're most welcome. I'm relieved and excited this moves forwards.

Will check what's in #1586 and what might be missing and'll forward port 
what I need to switch. Glad to be soon out of the corner I'm painted in ;)

Issue #1218, opened 2015-03-05
java.lang.IllegalArgumentException: n must be positive in Sketched.scala


fdong-coursera:
Could `maxReplicas ` be ZERO? If this is ZERO, the following code 

`maxReplicas.ceil.toInt.min(numReducers)` will be zero and it triggers IllegalArgumentException when calling `rand.nextInt(n)` . 

https://github.com/twitter/scalding/blob/c8313b402a23dbe3cdaa103baac6622c7a043309/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala#L107

Error stack as follows:
[error] Caused by: java.lang.IllegalArgumentException: n must be positive
[error]     at java.util.Random.nextInt(Random.java:300)
[error]     at scala.util.Random.nextInt(Random.scala:66)
[error]     at com.twitter.scalding.typed.SketchJoined$$anonfun$2.apply(Sketched.scala:115)
[error]     at com.twitter.scalding.typed.SketchJoined$$anonfun$2.apply(Sketched.scala:115)
[error]     at com.twitter.scalding.typed.SketchJoined$$anonfun$flatMapWithReplicas$1.apply(Sketched.scala:109)
[error]     at com.twitter.scalding.typed.SketchJoined$$anonfun$flatMapWithReplicas$1.apply(Sketched.scala:102)
[error]     at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:396)
[error]     at scala.collection.Iterator$class.foreach(Iterator.scala:750)
[error]     at scala.collection.AbstractIterator.foreach(Iterator.scala:1202)
[error]     at com.twitter.scalding.FlatMapFunction.operate(Operations.scala:47)
[error]     at cascading.flow.stream.FunctionEachStage.receive(FunctionEachStage.java:99)
[error]     ... 34 more

Thanks.

johnynek:
Thanks for the bug report. Yes the frequency could be zero if the key does not appear in the other side, and that means it is truly absent, so we don't need to send the key at all.

Can you reproduce this in a test? With such a test, it would be easy to fix (I can give you some pointers).

oeddyo:
@johnynek I took a bit of time looking into this. I'm creating a test where a key only appears on one side but the test still passed. 

I must be missing something. Could you give me a pointer please :)

class TypedSketchJoinJobForEmptyKeys(args: Args) extends Job(args) {
  // Deal with when a key appears in left but not right
[...]

PR #1450, opened 2015-10-31
Add test for sketchJoin bug #1218

### LINKED ISSUE #1449: Add failing test for sketchJoin bug #1218
See the repro example for sketchJoin failing in #1218.

To close this issue, we should have a PR with a failing test with that case.

wolfika:
fixes #1449

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/typed/TypedSketchJoinJobForEmptyKeysTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/typed/TypedSketchJoinJobForEmptyKeysTest.scala
@@ -1,0 +1,24 @@
+class TypedSketchJoinJobForEmptyKeys(args: Args) extends Job(args) {
+  // Deal with when a key appears in left but not right
+  val leftTypedPipe = TypedPipe.from(List((1, 1111)))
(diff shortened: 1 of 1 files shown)


johnynek:
great!

Now bonus points: I think the code here:

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala#L119

Should be changed to:
```scala
val lhs = flatMapWithReplicas(left.pipe){
  case 0 => Nil
  case n => List(rand.nextInt(n) + 1)
}
```
So, if we know for sure the key does not appear in the other side, we don't need to replicate it. We can double check the logic here: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala#L105

johnynek:
actually, I'm confused about this bug now. The random only happens on the left side. The count for all the items on the left side should be > 0 by definition (the CMS can overestimate, but not underestimate the count). This requires a bit more thought. Without an explanation of the bug, we shouldn't just merge something, even if it seems to fix things.

johnynek:
Dammit. `==` in Array is biting us here.

https://github.com/twitter/algebird/issues/497

johnynek:
I'm going to close this, but I used your code in #1451. Thanks for the contribution!

PR #1451, opened 2015-11-04
Fix bug with sketch joins and single keys

johnynek:
This fixes the issue in #1218 and #1449 based on the contribution in #1450 

The fix requires a minor change to the API, which should be source compatible for virtually everyone.

It also exposes two design flaws in algebird: 1) that we assume == works, so having a CMSHasher on Arrays is broken. 2) implicits for CMSHasher are not in the companion object, so they are annoying to use.

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/TypedSketchJoinJobForEmptyKeysTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/TypedSketchJoinJobForEmptyKeysTest.scala
@@ -1,0 +1,35 @@
+package com.twitter.scalding
+
+import org.scalatest.{ WordSpec, Matchers }
+
+class TypedSketchJoinJobForEmptyKeys(args: Args) extends Job(args) {
+  // Deal with when a key appears in left but not right
+  val leftTypedPipe = TypedPipe.from(List((1, 1111)))
+  val rightTypedPipe = TypedPipe.from(List((3, 3333), (4, 4444)))
+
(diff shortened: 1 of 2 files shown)


johnynek:
@isnotinvain can you review?

avibryant:
LGTM, FWIW.

scalding-core/src/test/scala/com/twitter/scalding/TypedSketchJoinJobForEmptyKeysTest.scala
```diff
@@ -0,0 +1,35 @@
...
+      case (a, (b, c)) =>
+        (a, b, c.getOrElse(-1))
+    }
+    .write(TypedTsv("output"))
+}
+
+class TypedSketchJoinJobForEmptyKeysTest extends WordSpec with Matchers {
+  import Dsl._
+  "A TypedSketchJoinJobForEmptyKeysTest" should {
+    "Sketch leftJoin should be the same an Join which is empty" in {
```
Gabriel439: I don't understand this explanation of the test.  What is this testing?

johnynek: I didn't edit it properly. There was originally a normal join and a sketch join in the job, but I removed the normal join since it was just easier to assert the output was a particular thing. This particular input and output caused an exception before, so the fact that it is green now is a guard against a particular regression, but the message is not good. I'll update it.

PR #1455, opened 2015-11-19
Fix incorrect usage of `percent'.

### LINKED ISSUE #1448: Documentation of sample in TypedPipe
The sample method of TypedPipe is documented as taking "percent" as an argument. But percent is typically in [0,100], while if I understand the code correctly, the argument is really supposed to be a fraction, or a value in [0,1.0] . Are people open to changing the documentation for this?

vikasgorur:
Rename the argument to various sampling methods in TypedPipe and RichPipe
to 'fraction' instead of 'percent' as they accept values in the range [0, 1]
and not [1, 100].

Fixes #1448

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
@@ -556,20 +556,20 @@ class RichPipe(val pipe: Pipe) extends java.io.Serializable with JoinAlgorithms 
   def limit(n: Long): Pipe = new Each(pipe, new Limit(n))
 
   /**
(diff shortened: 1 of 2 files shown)


vikasgorur:
Would it be appropriate for these methods to throw an `IllegalArgumentException` if the fraction is not in the range [0, 1]?

Motivation: I ran a job `input.sample(10).save(output)` intending to copy a fraction of the data but didn't realize my mistake until the job finished running. An early failure might be helpful in such cases.

johnynek:
Thanks.  Yes please add:
```scala
require(0.0 <= fraction && fraction <=1.0, s"got $fraction which is an invalid fraction")
```

johnynek:
Thank you!

Issue #1418, opened 2015-08-12
Matrix2 does not support OrderedSerialization.


johnynek:
Ofter `Ordering`s are composed in here:

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala

We need to do that using:
https://github.com/twitter/scalding/blob/develop/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala#L61

if all Orderings are OrderedSerializations. We probably need some method like:

```scala
implicit def withOrderedSerialization[A: Ordering, B: Ordering]: Ordering[(A, B)] =
  // if both Orderings are OrderedSerialization, use OrderedSerialization2, else Ordering.tuple2 
```
and use that in internal methods.

reconditesea:
@johnynek Here I think if the individual ordering for row/col is an OrderedSerialization, it will just work. But for the composed ordering of the two, we will need to add OrderedSerialization support. Is this correct?

johnynek:
Yes, that's right. Sometimes we group on the pair, and that needs to be composed properly.

PR #1457, opened 2015-12-15
Add OrderedSerialization2 support in Matrix2.

reconditesea:
To resolve issue #1418

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
@@ -18,6 +18,7 @@ package com.twitter.scalding.mathematics
 import cascading.flow.FlowDef
 import cascading.pipe.Pipe
 import cascading.tuple.Fields
+import com.twitter.scalding.serialization.{ OrderedSerialization, OrderedSerialization2 }
 import com.twitter.scalding.TDsl._
 import com.twitter.scalding._
 import com.twitter.scalding.typed.{ ValuePipe, EmptyValue, LiteralValue, ComputedValue }
@@ -332,9 +333,8 @@ case class Product[R, C, C2, V](left: Matrix2[R, C, V],
     if (isSpecialCase) {
       joined
     } else {
-      val ord2: Ordering[(R, C2)] = Ordering.Tuple2(rowOrd, colOrd)
       val localRing = ring
-      joined.groupBy(w => (w._1, w._2))(ord2).mapValues { _._3 }
+      joined.groupBy(w => (w._1, w._2)).mapValues { _._3 }
         .sum(localRing)
         .filter { kv => localRing.isNonZero(kv._2) }
         .map { case ((r, c), v) => (r, c, v) }
@@ -359,6 +359,13 @@ case class Product[R, C, C2, V](left: Matrix2[R, C, V],
 
   implicit override val rowOrd: Ordering[R] = left.rowOrd
   implicit override val colOrd: Ordering[C2] = right.colOrd
+  implicit def withOrderedSerialization: Ordering[(R, C2)] = {
+    (rowOrd, colOrd) match {
+      case (ros: OrderedSerialization[_], cos: OrderedSerialization[_]) =>
+        new OrderedSerialization2(ros.asInstanceOf[OrderedSerialization[R]], cos.asInstanceOf[OrderedSerialization[C2]])
+      case _ => Ordering.Tuple2(rowOrd, colOrd)
+    }
+  }
   override lazy val transpose: Product[C2, C, R, V] = Product(right.transpose, left.transpose, ring)
   override def negate(implicit g: Group[V]): Product[R, C, C2, V] = {
     if (left.sizeHint.total.getOrElse(BigInt(0L)) > right.sizeHint.total.getOrElse(BigInt(0L))) {
@@ -423,10 +430,9 @@ case class Sum[R, C, V](left: Matrix2[R, C, V], right: Matrix2[R, C, V], mon: Mo
     if (left.equals(right)) {
       left.optimizedSelf.toTypedPipe.map(v => (v._1, v._2, mon.plus(v._3, v._3)))
     } else {
-      val ord: Ordering[(R, C)] = Ordering.Tuple2(left.rowOrd, left.colOrd)
       collectAddends(this)
         .reduce((x, y) => x ++ y)
-        .groupBy(x => (x._1, x._2))(ord).mapValues { _._3 }
+        .groupBy(x => (x._1, x._2)).mapValues { _._3 }
         .sum(mon)
         .filter { kv => mon.isNonZero(kv._2) }
         .map { case ((r, c), v) => (r, c, v) }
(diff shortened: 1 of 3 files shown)


reconditesea:
@johnynek Is this good to go?

johnynek:
Sorry. Didn't see this. Thanks for bringing to my attention. I'll review
Monday.

On Saturday, December 19, 2015, Kevin Lin <notifications@github.com> wrote:

> @johnynek <https://github.com/johnynek> Is this good to go?
>
> —
> <https://github.com/twitter/scalding/pull/1457#issuecomment-166011244>.
>


-- 

reconditesea:
Thanks, happy holidays!

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
```diff
@@ -18,6 +18,7 @@ package com.twitter.scalding.mathematics
 import cascading.flow.FlowDef
 import cascading.pipe.Pipe
 import cascading.tuple.Fields
+import com.twitter.scalding.serialization.{ OrderedSerialization, OrderedSerialization2 }
 import com.twitter.scalding.TDsl._
```
johnynek: can you remove TDsl while you are at it?

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Matrix2.scala
```diff
@@ -359,6 +359,13 @@ case class Product[R, C, C2, V](left: Matrix2[R, C, V],
 
   implicit override val rowOrd: Ordering[R] = left.rowOrd
   implicit override val colOrd: Ordering[C2] = right.colOrd
+  implicit def withOrderedSerialization: Ordering[(R, C2)] = {
```
johnynek: Can we not repeat this code 3 times? What if we have this function in the companion object of `OrderedSerialization2` in this file:

https://github.com/twitter/scalding/blob/develop/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala#L61

```scala
def maybeOrderedSerialization[A, B](implicit ordA: Ordering[A], ordB: Ordering[B]): Ordering[(A, B)] = ...
```

johnynek:
Need a test for this that some Matrix2 job works with the option: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Config.scala#L118 `setRequiredOrderedSerialization(true)`.

In fact, we should write the test first and find an example that fails (I guess adding two matrices)

reconditesea:
Moved the method implementation into OrderedSerialization2 companion object.
Will add new test cases shortly.

reconditesea:
I had troubles in importing serialization test's intOrderedSerialization (https://github.com/twitter/scalding/blob/398545991c63019714d80ac22de6494adfd798f2/scalding-serialization/src/test/scala/com/twitter/scalding/serialization/SerializationProperties.scala#L46) into Matrix2 test, so pasted the code into Matrix2 test. 

BTW, feel like we can use slack channel for discussions like this more frequently.

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
```diff
@@ -40,6 +44,37 @@ class Matrix2Sum(args: Args) extends Job(args) {
...
+    def write(o: OutputStream, t: Int) = Try(Writer.write[Int](o, t))
+    def hash(t: Int) = t.hashCode
+    def compare(a: Int, b: Int) = java.lang.Integer.compare(a, b)
+    def compareBinary(a: InputStream, b: InputStream) =
+      OrderedSerialization.readThenCompare(a, b)(this)
+    val staticSize = Some(4)
+    def dynamicSize(i: Int) = staticSize
+  }
+
+  val p1: Pipe = Tsv("mat1", ('x1, 'y1, 'v1)).read
```
johnynek: this style is not recommended any more (.toTypedPipe), so let's just use TypedText.tsv the whole way:
```scala
val tp1 = TypedPipe.from(TypedText.tsv[(Int, Int, Double)]("mat1"))
```

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
```diff
@@ -293,6 +328,21 @@ class Matrix2Test extends WordSpec with Matchers {
     }
   }
 
+  "A MatrixSum job with Orderedserialization" should {
+    TUtil.printStack {
+      JobTest(new Matrix2SumOrderedSerialization(_))
```
johnynek: did you verify that this test fails if you don't have the intOS in scope?

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
```diff
@@ -40,6 +44,37 @@ class Matrix2Sum(args: Args) extends Job(args) {
   sum.write(TypedText.tsv[(Int, Int, Double)]("sum"))
 }
 
+class Matrix2SumOrderedSerialization(args: Args) extends Job(args) {
+
+  import Matrix2._
+  import cascading.pipe.Pipe
```
johnynek: let's not import Pipe, Fields, or TDsl.

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
```diff
@@ -40,6 +44,37 @@ class Matrix2Sum(args: Args) extends Job(args) {
...
+  import cascading.tuple.Fields
+  import com.twitter.scalding.TDsl._
+
+  override def config = super.config + (Config.ScaldingRequireOrderedSerialization -> "true")
+  implicit val intOS: OrderedSerialization[Int] = new OrderedSerialization[Int] {
```
johnynek: can you add a comment that this is just for a test, normally you would use: 
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/serialization/RequiredBinaryComparators.scala#L21

```scala
import RequiredBinaryComparators.orderedSerialization
```

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Serialization2.scala
```diff
@@ -58,6 +58,16 @@ class Serialization2[A, B](val serA: Serialization[A], val serB: Serialization[B
   } yield a + b
 }
 
+object OrderedSerialization2 {
+  def maybeOrderedSerialization[A, B](implicit ordA: Ordering[A], ordB: Ordering[B]): Ordering[(A, B)] = {
```
johnynek: can we actually rename this `maybeOrderedSerialization2`. Sorry, in scala you can't have implicits with colliding names (but you can have normal methods with colliding names as long as their parameters are different). We might want to add `maybe...[2-22]` at some point in the future.

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
reconditesea: Yup, it failed if I don't declaim intOS in scope.

johnynek: Cool. Just making sure. :)

reconditesea: Of source! :)
I will have another update to address other comments shortly.

scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala
reconditesea: I changed intOS as ```implicit val intOS = orderedSerialization[Int] ```.
I guess it's better to use the normal macro here.

PR #1458, opened 2015-12-17
Add InvalidSourceTap to catch all cases for no good path.

### LINKED ISSUE #1454: Add separate InvalidSourceTap for dealing with no valid paths in FileSource
The following can lead to edge cases:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala#L264

The assumption is that validateTaps throws if validation fails, but there are codepaths like source.toIterator where validate is not called.

Adding separate InvalidSourceTap that throws in openForRead can cover both the regular read and toIterator cases I think.

reconditesea:
To close #1454.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -261,7 +261,7 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
         // This case is going to result in an error, but we don't want to throw until
         // validateTaps, so we just put a dummy path to return something so the
         // Job constructor does not fail.
-        CastHfsTap(createHfsTap(hdfsScheme, hdfsPaths.head, sinkMode))
+        new InvalidSourceTap(CastHfsTap(createHfsTap(hdfsScheme, hdfsPaths.head, sinkMode)))
       }
       case 1 => taps.head
       case _ => new ScaldingMultiSourceTap(taps)
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -261,7 +261,7 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
         // This case is going to result in an error, but we don't want to throw until
         // validateTaps, so we just put a dummy path to return something so the
         // Job constructor does not fail.
-        new InvalidSourceTap(CastHfsTap(createHfsTap(hdfsScheme, hdfsPaths.head, sinkMode)))
+        new InvalidSourceTap(goodHdfsPaths(hdfsMode))
       }
       case 1 => taps.head
       case _ => new ScaldingMultiSourceTap(taps)
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -259,9 +259,10 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
     taps.size match {
       case 0 => {
         // This case is going to result in an error, but we don't want to throw until
-        // validateTaps, so we just put a dummy path to return something so the
-        // Job constructor does not fail.
-        new InvalidSourceTap(goodHdfsPaths(hdfsMode))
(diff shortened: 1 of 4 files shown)


rubanm:
We have an internal version of this Tap with the same name. It handles some of the quirks required to work with cascading's MultiInputFormat and also unique identifiers. If you could consolidate the two, that'd be great.

scalding-core/src/main/scala/com/twitter/scalding/Source.scala
```diff
@@ -39,6 +39,21 @@ import scala.collection.JavaConverters._
  */
 class InvalidSourceException(message: String) extends RuntimeException(message)
 
+/**
+ * Represents an invalid tap which throws when openForRead is called
+ */
+class InvalidSourceTap(tap: Tap[JobConf, RecordReader[_, _], OutputCollector[_, _]])
+  extends Tap[JobConf, RecordReader[_, _], OutputCollector[_, _]]() {
```
rubanm: This can extend just SourceTap.

reconditesea:
@rubanm  Didn't know we already have an internal version. Will take a look and consolidate them.

benpence:
Do we have a test yet that hits this code path in `FileSource`?

reconditesea:
@benpence Added a unit-test. Thanks!

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -261,7 +261,7 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
         // This case is going to result in an error, but we don't want to throw until
         // validateTaps, so we just put a dummy path to return something so the
         // Job constructor does not fail.
-        CastHfsTap(createHfsTap(hdfsScheme, hdfsPaths.head, sinkMode))
+        new InvalidSourceTap(goodHdfsPaths(hdfsMode))
```
rubanm: goodHdfsPaths will return Nil in this case right? We need to pass hdfsPaths here I think so that openForRead throws with the right message. Can you also update the comment above.

scalding-core/src/test/scala/com/twitter/scalding/FileSourceTest.scala
```diff
@@ -155,6 +155,15 @@ class FileSourceTest extends WordSpec with Matchers {
     }
 
   }
+
+  "invalid source input" should {
+    import TestEmptyFileSource.hdfsPaths
```
rubanm: This one seems unused

scalding-core/src/test/scala/com/twitter/scalding/FileSourceTest.scala
```diff
@@ -186,3 +195,14 @@ object TestSuccessFileSource extends FileSource with SuccessFileSource {
 
   def pathIsGood(p: String) = super.pathIsGood(testfsPathRoot + p, conf)
 }
+
+object TestEmptyFileSource extends FileSource {
+
+  override def hdfsPaths: Iterable[String] = Iterable.empty
+  override def localPaths: Iterable[String] = Iterable.empty
+
```
rubanm: A more realistic test case could be we set some dummy paths here, and override def pathIsGood to always return false. What do you think?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
reconditesea: Good catch!

scalding-core/src/test/scala/com/twitter/scalding/FileSourceTest.scala
reconditesea: Yeah, makes sense.

rubanm:
LGTM. @johnynek could you take a look as well? Thanks.

reconditesea:
@johnynek @rubanm Any suggestions on this PR? Thanks and happy holidays!

reconditesea:
Gentle ping.

rubanm:
LGTM merging this. @reconditesea thanks!

The internal version of this at Twitter has been working fine so I don't foresee any problems. Also with this change, opening an Iterator via toIterator on an invalid source will throw InvalidSourceException (instead of returning an iterator on an invalid hdfs datapath, which in turn, throws hadoop's InvalidInputException).

Issue #1459, opened 2015-12-17
Scalding Execution API memory leak / OOM


vineethvarghese:
We are running into OOM issues due to https://github.com/twitter/scalding/pull/1057

We have some `Execution` jobs that takes an hour or more to run and lift some heavy lists through the `Execution` monad assuming gc will take care after we are done using them. However, we see that all the object we raise through `Execution` are held in the cache well after their use.

I have created a test case for this which you can find at https://github.com/CommBank/execution-memory. Details of how to run the test is in the README and you don't need Hadoop for it. While the example might seem a tad aggressive in its object creation, it is pretty much a dumbed down version of our job.

johnynek:
Execution.run definitely is currently biased now to absolutely minimize work at the expense of memory. It is probably not a good general IO monad for that reason, even though it might look sort of like that.

We could address this at least two three ways:

1. disable all caching. I think this is a non-starter because it can result in hadoop jobs being run over and over again needlessly (and sadly, not all hadoop jobs are idempotent).

2. we could only cache infinitely hadoop writes: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L226 but we could use a weak or soft reference on the general cache: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L224

3. We could only cache the writes, but not cache at all the non-writes, so `Execution.fromFuture` which is ultimately how `Execution.from` is implemented, would not be cached.

What are your thoughts?

johnynek:
@vineethvarghese Also I note you are using a very old version of scalding in the test. If we fix it, are you going to be able to consume it?

johnynek:
Also, could you use another monad and then lift it into Execution as a work around? For instance you could use something like scalaz's Task? You could write the code that is creating millions of operations inside of a scalaz.Task, and then write a method `def liftTask[T](t: Task[T]): Execution[T]` which would `unsafePerformSync` inside `Execution.fromFuture { implicit cec => Future(task.unsafePerformSync) }`.

Might be a work-around until we get the trade-offs correctly here.

vineethvarghese:
@johnynek Thanks for your quick response. Let me try to answer your questions here.

We were thinking along the lines of item 2. Item 1 is definitely a no. Item 3 also won't work if you look at the code below. 

We are ok to upgrade to the latest version. 

[...]

PR #1460, opened 2015-12-18
Weaken the cache to address #1459

johnynek:
This puts everything other than the hadoop operation cache as SoftReferences, which can be garbage collected.

Seems to help #1459 for me, but it is probably not a complete solution. We should probably clean the keys out periodically (maybe every N evaluations, do a complete scan of the ConcurrentHashMap looking for dead references).

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -221,7 +221,7 @@ object Execution {
    * as it is evaluating.
    */
   private[scalding] class EvalCache {
-    private[this] val cache = new FutureCache[Execution[Any], (Any, ExecutionCounters)]
+    private[this] val cache = new FutureSoftCache[Execution[Any], (Any, ExecutionCounters)]
 
     private[this] val toWriteCache = new FutureCache[ToWrite, ExecutionCounters]
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
@@ -43,3 +43,61 @@ class FutureCache[-K, V] {
     }
   }
 }
+
+class FutureSoftCache[-K, V] {
+  import java.lang.ref.SoftReference
+
+  private[this] val cache = new ConcurrentHashMap[K, SoftReference[Future[V]]]()
+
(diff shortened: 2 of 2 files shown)


johnynek:
@vineethvarghese take a look.

vineethvarghese:
Sure

rowandavies:
With soft values in the key, some executions will be re-executed, changing the semantics, right?  Also,
programs which depend on this to avoid OOM will tend to keep their heap full, which may not be desirable, and the choice of what to remove from the cache will be based on what values still have references, rather than what keys do.

It may be better instead to change the cache to a weak hash map that also concurrent - i.e., with the keys weak and the cache entry garbage collected when the key no longer has any references to it like (hence no possibility of being looked up).  E.g., you can do this with:   http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/collect/MapMaker.html

I think such a concurrent weak hash map will better address the specific issue motivating #1459 (I work with Vineeth).  In the more general case, it wouldn't change the semantics at all, but I guess that wouldn't address the situation where execution objects have other references to them so that they can be used multiple calls to `run` on `Execution`.  But, I'd suggest not doing that anyway, plus I'd expect it's rare and easy to fix in specific situations.

johnynek:
It's not really the case that the weak key hashmap is safe either. The internal private classes which implement Execution are case classes. You could have a flatMap that have a function that will create identical case classes, so they are referentially distinct, but equivalent. I *think* the weak key hashmap cannot deal with this case, so would result in recomputation then.

RE: guava, we really hate to add dependencies on the classpath. This is a big pain for monorepos (such as Twitter's) where adding a constraint on a common library makes dependencies much harder to satisfy.

rowandavies:
Ah, of course you're right.  To match the current behaviour, I don't see any other way than to keep all results in the cache forever.  Given that, and a way to prevent caching of some `Execution`s such as those arising via `Execution.from`, I think it's more transparent to fail fast with an OOM error so the code can be adapted than to see a less repeatable evacuation of the `EvalCache` even if the latter is more "robust" in each particular run.  

Re: guava, I think `java.util.Collections.synchronizedMap` combined with `java.util.WeakHashMap` and an `equals` method based on object identity would work just fine here too, with just a tiny loss of concurrency.  (But, the point seems moot in this case.)

johnynek:
#1511 is the new approach here.

PR #1462, opened 2015-12-22
Cluster info and fs shell in repl

gerashegalov:
Often users don't realize what cluster they work with, and why a source has no data. This PR is to display in the console the current mode and the working directory. It also adds the ability to list and print files using FsShell.

### DIFF
--- a/scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
+++ b/scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
@@ -105,29 +105,13 @@ trait BaseReplState {
   }
 
   /**
-   * List files under path
+   * Access to Hadoop FsShell
    *
-   * @param pathStr
+   * @param cmdArgs
    * @return
    */
-  def fsLs(pathStr: String): Int = {
-    val args = Seq("-ls", pathStr).toArray
-    new FsShell(modeHadoopConf).run(args)
-  }
-
-  def fsLs(): Int = {
-    fsLs(".")
-  }
-
-  /**
-   * Run fs -text. Dump text, automatically decompress if needed
-   *
-   * @param pathStr
-   * @return
-   */
-  def fsCat(pathStr: String): Int = {
-    val args = Seq("-text", pathStr).toArray
-    new FsShell(modeHadoopConf).run(args)
+  def fsShell(cmdArgs: String*): Int = {
+    new FsShell(modeHadoopConf).run(cmdArgs.toArray)
   }
 
   /**
--- a/scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
+++ b/scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
@@ -17,8 +17,10 @@ package com.twitter.scalding
 
 import cascading.flow.FlowDef
 import cascading.pipe.Pipe
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{ FsShell, FileSystem }
 import typed.KeyedListLike
-import scala.util.{ Failure, Success }
+import scala.util.{ Try, Failure, Success }
 import scala.concurrent.{ Future, ExecutionContext => ConcurrentExecutionContext }
 
 /**
@@ -43,8 +45,15 @@ trait BaseReplState {
   private[scalding] var storedHdfsMode: Option[Hdfs] = None
 
   /** Switch to Local mode */
-  def useLocalMode() { mode = Local(false) }
-  def useStrictLocalMode() { mode = Local(true) }
+  def useLocalMode() {
+    mode = Local(false)
+    modeBanner
+  }
+
+  def useStrictLocalMode() {
+    mode = Local(true)
+    modeBanner
+  }
 
   /** Switch to Hdfs mode */
   private def useHdfsMode_() {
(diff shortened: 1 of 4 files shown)


johnynek:
looks like some repl tests failed.

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -58,12 +67,62 @@ trait BaseReplState {
...
   }
 
   def useHdfsLocalMode() {
     useHdfsMode_()
     customConfig += mr1Key -> mrLocal
     customConfig += mr2Key -> mrLocal
+    modeBanner
+  }
+
+  def modeBanner: Unit = {
```
johnynek: scala style in to use `def modeBanner(): Unit` to have `()` for methods that have side effects (strange style, but the compiler will warn you otherwise).

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -58,12 +67,62 @@ trait BaseReplState {
...
+    println(s"#### User home: ${scaldingMode._2}${Console.RESET}")
+  }
+
+  /**
+   * List files under path
+   *
+   * @param pathStr
+   * @return
+   */
+  def fs_ls(pathStr: String) = {
```
johnynek: Let's not use `_` style mixed. Can we just do `fsLs`here? Similarly below?

johnynek: What's the return type here? Looks like Any to me. Can you put a type and make it `Try[FsShell]` I guess?

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -58,12 +67,62 @@ trait BaseReplState {
...
+    }
+  }
+
+  /**
+   * Run fs -text. Dump text, automatically decompress if needed
+   *
+   * @param pathStr
+   * @return
+   */
+  def fs_cat(pathStr: String) = {
```
johnynek: similar comment here about return type (which we want on all public methods).

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
johnynek: also can we call this something like `printModeBanner` to make it clear what the side effect is?

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -58,12 +67,62 @@ trait BaseReplState {
...
 
   def useHdfsLocalMode() {
     useHdfsMode_()
     customConfig += mr1Key -> mrLocal
     customConfig += mr2Key -> mrLocal
+    modeBanner
+  }
+
+  def modeBanner: Unit = {
+    val scaldingMode = mode match {
```
johnynek: can we do

`val (modeString, homeDir) =`

so it is more clear what is in each part of the tuple?

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
```diff
@@ -53,13 +53,6 @@ class ScaldingILoop
    */
   private val scaldingCommands: List[LoopCommand] = List()
 
-  /**
-   * Change the shell prompt to read scalding&gt;
-   *
-   * @return a prompt string to use for this REPL.
-   */
-  override def prompt: String = Console.BLUE + "\nscalding> " + Console.RESET
```
johnynek: is this change due to using a different version of jline?

johnynek:
@ianoc I remember some jline issue in source. Do you remember what it was? I'm not that familar with Jline, actually.

@gerashegalov thanks for submitting the PR!

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingILoop.scala
gerashegalov: I undid the change with the jline dependency. It belongs to a different patch I tried for intercepting "Ctrl-C" to stop the last submitted scalding flow without exiting the REPL.

This change was originally in connection to the first version where I was changing the prompt. I discarded the idea. I kept this particular change because method prompt is not really necessary since you can call setPrompt on base.

gerashegalov: that's actually what's failing on travis but not locally:
```
[error] /home/travis/build/twitter/scalding/scalding-repl/src/main/scala/com/twitter/scalding/ScaldingShell.scala:98: value setPrompt is not a member of scala.tools.nsc.interpreter.ILoop
[error]     scaldingREPL.get.setPrompt(Console.BLUE + "\nscalding> " + Console.RESET)
[error]                      ^
[error] one error found
[error] (scalding-repl/compile:compileIncremental) Compilation failed
```

johnynek:
Looks good to me. @ianoc should probably double check to get CDL's signoff.

gerashegalov:
Thanks for reviewing @johnynek !

richwhitjr:
We could instead have a generic command in the REPL to run Hadoop fs commands.  So instead of having one for ls, another for cat etc we could do:

def hadoopFs(args: String*) = new FsShell(modeHadoopConf).run(args)

gerashegalov:
@richwhitjr thanks for suggestions. Any opinion what feels more natural UX;

`hadoopFs("-ls /path1 /path2") `
and do splitting into array
or force the user to write 
`hadoopFs("-ls", "/path1",  "/path2") `

The former seems good for the standard case but if you already have args in the REPL session you might want the latter:
`hadoopFs("-ls", res0) `

gerashegalov:
of course what I want least is parsing edge cases. Asking the user to parse manually is easy :)

richwhitjr:
Yep I could imagine some edge cases even though a single string seems more natural.

gerashegalov:
@richwhitjr How about this?
```
scalding> fsShell("-ls",  "version.sbt", "spaces are not a problem")
-rw-r--r--   1 gshegalov staff         42 2015-07-20 15:19 version.sbt
-rw-r--r--   1 gshegalov staff          0 2016-01-03 23:45 spaces are not a problem
res2: Int = 0
```

ianoc:
This looks good to me too. @gerashegalov if you just ping me to know when you and Richard are happy with the fsShell semantics i'll merge. Either way seems fine to me.

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
```diff
@@ -58,12 +67,51 @@ trait BaseReplState {
...
+      }
+      case hdfsMode: Hdfs => {
+        val defaultFs = FileSystem.get(hdfsMode.jobConf)
+        val m = customConfig.get(mr2Key) match {
+          case Some("local") =>
+            s"${hdfsMode.getClass.getSimpleName}Local(${hdfsMode.strict})"
+          case _ =>
+            s"${hdfsMode.getClass.getSimpleName}(${hdfsMode.strict})"
+        }
+        (m, defaultFs.getWorkingDirectory.toString)
```
benpence: Is the working directory for the default `FileSystem` always the user's home directory?

richwhitjr:
Looks good to me.

scalding-repl/src/main/scala/com/twitter/scalding/ReplImplicits.scala
gerashegalov: It is typically for cluster FileSystem such as HDFS. For the local fs used in hdfsLocal and local modes, the working dir is the $PWD ("user.home" system property) of the OS process (JVM).

ianoc:
Merged thanks!

gerashegalov:
Thanks everyone!

PR #1463, opened 2015-12-27
Update Scala version to 2.10.6

zaneli:


benpence:
Would you please talk about why you updated scala? It's not clear from your commit.

zaneli:
Sorry, there is no clear reason in this...
but generally, I'd like to keep the latest version as much as possible.

I guess https://github.com/twitter/scalding/commit/512eca502069c928c30a29ace4739bb3812fd1c7, https://github.com/twitter/scalding/commit/9510d5145d5099eb705d0cc18081b54a305098a7 look the similar reason.

johnynek:
Seems fine to me. This does not change the dependencies (since scala publishes with `_2.10` so it should be binary compatible).

zaneli:
Thank you!

PR #1464, opened 2015-12-29
Fix median estimation

zaneli:
I confirmed the past change,
https://github.com/twitter/scalding/commit/539fcce1a60fed7fb8529d3ca222dfb3683468ae#diff-cce8f848668405e685fa11b989199fb6R98
https://github.com/twitter/scalding/commit/3bd578c80cc447da4f0bf0c6d11a14cc8913f4d2#diff-cce8f848668405e685fa11b989199fb6L98
but I couldn't make sense whether reasonable that `MedianEstimationScheme.estimateJobTime(times)` returns `mean(times)`...

Should this return `median(times)`?

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/package.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/package.scala
@@ -1,3 +1,5 @@
+package com.twitter.scalding
+
 package object reducer_estimation {
   def median(xs: Seq[Double]): Option[Double] = xs.sorted.lift(xs.length / 2)
   def mean(xs: Seq[Double]): Option[Double] = if (xs.isEmpty) None else Some(xs.sum / xs.length)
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/package.scala
```diff
@@ -1,3 +1,5 @@
+package com.twitter.scalding
```
zaneli: I think `reducer_estimation` package object should be in `com.twitter.scalding` package.
http://www.scala-lang.org/docu/files/packageobjects/packageobjects.html
http://docs.scala-lang.org/style/naming-conventions.html#packages

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/RuntimeReducerEstimatorTest.scala
```diff
@@ -158,11 +157,11 @@ class RuntimeReducerEstimatorTest extends WordSpec with Matchers with HadoopShar
           //
           // We don't scale by input size.
           //
-          // The mean of these is 3600 ms, so we anticipate
-          // that the job will take 3600 ms total.
+          // The median of these is 3000 ms, so we anticipate
+          // that the job will take 3000 ms total.
```
zaneli:
```scala
scala> com.twitter.scalding.reducer_estimation.mean(Seq(1000.0D, 200.0D, 2400.0D).map(_ * 3))
res0: Option[Double] = Some(3600.0)

scala> com.twitter.scalding.reducer_estimation.median(Seq(1000.0D, 200.0D, 2400.0D).map(_ * 3))
res1: Option[Double] = Some(3000.0)
```

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/package.scala
johnynek: Yes, but I wonder why the compile didn't fail before. Ahh, because it was a root package. :/ Good catch.

johnynek:
Thanks for the cleanup. This looks good.

zaneli:
It's my pleasure.

PR #1467, opened 2016-01-06
Moves the default to 2.11

ianoc:
This mostly comes in if you want to do some local publishing of just one version. Right now publishing for 2.11 wouldn't publish maple. We can only wire it in for one to avoid publishing it twice when we need to do everything it seems. Given 2.11 is the way forward seems more common we would want to do it only.


johnynek:
This test failed (it is a random test: https://github.com/twitter/scalding/blob/develop/scalding-serialization/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala#L284 )

Restarted.

Merge when green.

PR #1468, opened 2016-01-08
Adds Error Message to REPL when Current Directory Not Readable

benpence:
Normally, if the current directory is not readable, `new File(currentDirectory).listFiles()` will lead to a `NullPointerException` in `findAllUpPath`. This patch adds a better error message.

Alternatively, I could be persuaded to return an empty `List` instead of throwing a `RuntimeException` from `findAllUpPath` because I'm not aware of any contract that says you have to run the `hadoop` command from a readable directory.


johnynek:
merge when green

PR #1469, opened 2016-01-12
allow Executions to set Config values

lclementi:
This patch allow setting Config values on a per Execution basis.
The values will be pushed down to the corresponding flowSteps generated
by the Execution.

This allows monitoring execution using the addFlowStepListener as shown
in the unit test and switching based on the values present in the
Configuration object

### DIFF
--- a/scalding-commons/src/test/scala/com/twitter/scalding/ExecutionTest.scala
+++ b/scalding-commons/src/test/scala/com/twitter/scalding/ExecutionTest.scala
@@ -15,10 +15,12 @@ limitations under the License.
 */
 package com.twitter.scalding.typed
 
+import cascading.flow.{ FlowStep, FlowStepListener }
 import org.scalatest.{ Matchers, WordSpec }
 
 import com.twitter.scalding._
 import com.twitter.algebird.monad.Reader
+import org.apache.hadoop.conf.Configuration
 
 // Need this to flatMap Future
 import scala.concurrent.ExecutionContext.Implicits.global
@@ -66,6 +68,61 @@ class WordCountEc(args: Args) extends ExecutionJob[Unit](args) {
     def execute(r: Runnable) = r.run
     def reportFailure(t: Throwable) = ()
   }
+}
+
+/**
+ * Run to two execution in paralel (zip) setting different config option
+ * and verify that Execution.setExtraConfig are preserved in each flowStep
+ *
+ */
+class ConfTestExecution(args: Args) extends ExecutionJob[Unit](args) {
+
+  override def config: Map[AnyRef, AnyRef] = {
+    //addFlowStepListerner to verify that Execution.setExtraConfig works
+    val newconf = Config.empty.addFlowStepListener{ (m: Mode, c: Config) =>
+      (m, c) match {
+        case (_, _) => new TestExecutionFlowListener()
+      }
+    }
+    super.config ++ newconf.toMap
+  }
+
+  def execution = {
+    val exec1 = TypedPipe.from(TypedTsv[(String, Int)](args("input1"))).writeExecution(TypedTsv[(String, Int)](args("output1")))
+    val exec2 = TypedPipe.from(TypedTsv[(String, Int)](args("input2"))).writeExecution(TypedTsv[(String, Int)](args("output2")))
+    exec1.setExtraConfig(Config(Map("testconf1" -> "1")))
+    exec2.setExtraConfig(Config(Map("testconf2" -> "2")))
+    val zippedExec = exec1.zip(exec2)
+    zippedExec.setExtraConfig(Config(Map("testconf3" -> "3")))
+    zippedExec.unit
+  }
+
+  private[this] class TestExecutionFlowListener extends FlowStepListener {
+
+    override def onStepCompleted(flowStep: FlowStep[_]): Unit = ()
+    override def onStepRunning(flowStep: FlowStep[_]): Unit = ()
+    override def onStepStopping(flowStep: FlowStep[_]): Unit = ()
+    override def onStepStarting(flowStep: FlowStep[_]): Unit = {
+      flowStep.getConfig() match {
+        case conf: Configuration =>
+          if (conf.get("testconf1") != null)
+            ConfTestExecutionCounter.test1 = ConfTestExecutionCounter.test1 + 1;
+          if (conf.get("testconf2") != null)
+            ConfTestExecutionCounter.test2 = ConfTestExecutionCounter.test2 + 1;
+          if (conf.get("testconf3") != null)
+            ConfTestExecutionCounter.test3 = ConfTestExecutionCounter.test3 + 1;
+        case _ => ()
+      }
+    }
+    override def onStepThrowable(flowStep: FlowStep[_], throwable: Throwable): Boolean = false
+  }
+
+}
+
+object ConfTestExecutionCounter {
+  var test1: Int = 0
+  var test2: Int = 0
+  var test3: Int = 0
 }
 
 class ExecutionTest extends WordSpec with Matchers {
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -475,12 +481,29 @@ object Execution {
...
 
   private trait ToWrite {
+    def setExtraConfig(conf: Config): Config = Config.empty
+
     def write(config: Config, flowDef: FlowDef, mode: Mode): Unit
   }
+
   private case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite {
+
+    var extraConfig = Config.empty
```
ianoc: Generally we avoid mutable stuff like this where at all possible, this config could persist/grow or do different things now depending on re-executions of this execution?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -498,6 +521,13 @@ object Execution {
   private case class WriteExecution[T](head: ToWrite, tail: List[ToWrite], fn: (Config, Mode) => T) extends Execution[T] {
 
     /**
+     * override this so that we can push down the conf value to the current TypedPipes
+     */
+    override def setExtraConfig(conf: Config): Unit = {
+      (head :: tail).foreach(_.setExtraConfig(conf))
```
ianoc: This is only setting step configs on a single step for each ToWrite, that won't set the config options globally for a given run. Its not really clear what it would mean since it won't apply for a full execution.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -238,6 +238,9 @@ trait TypedPipe[+T] extends Serializable {
   /** adds a description to the pipe */
   def withDescription(description: String): TypedPipe[T] = new WithDescriptionTypedPipe[T](this, description)
 
+  /* add a config parameter to this pipe */
+  def withExtraConfig(key: String, value: String): TypedPipe[T] = new WithExtraConfig[T](this, key, value)
```
ianoc: This means adding several configuration options is going to make a deep nested set of typed pipe classes, could we not stick a map in here?

lclementi: You mean instead of:

    def withExtraConfig(key: String, value: String): TypedPipe[T]

something like this:

    def withExtraConfig(extraConf: Config): TypedPipe[T]

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
lclementi: Yes the extraConfig var can grow in different moment see my unit test, we could have 

    val exec1 = TypedPipe.from(TypedTsv[(String, Int)](args("input1"))).writeExecution(TypedTsv[(String, Int)](args("output1")))
    val exec2 = TypedPipe.from(TypedTsv[(String, Int)](args("input2"))).writeExecution(TypedTsv[(String, Int)](args("output2")))
    exec1.setExtraConfig(Config(Map("testconf1" -> "1")))
    exec2.setExtraConfig(Config(Map("testconf2" -> "2")))
    val zippedExec = exec1.zip(exec2)
    zippedExec.setExtraConfig(Config(Map("testconf3" -> "3")))
    zippedExec.unit

So I expect to find testconf1 only in the flowStep generated by exec1, testconf2 only in the flowStep generated by exec2, and testconf3 in all the steps since it was set on the zipped execs.

I think to satisfy this use case we need to make use of that:

    var extraConfig = Config.empty

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
lclementi: Yes I'm not trying to set the Config globally, but only to the stepFlows that will be generated by this execution, see my explanation above ^^

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ianoc: So in general you've a mutable API, this is a no-no in most of our settings. Any call on an execution should return a new execution, not mutate the existing one. Thread safety and other things are all issues once you start down the mutation route, what happens with two competing setExtraConfig's in different threads?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ianoc: But this doesn't push it down to that, it attempts to push it onto the final rich pipe thats doing the write out, which won't effect any intermediate mappers , reducers or other things in this execution.

I'm not really sure this is workable, I don't think any of the API's in cascading allow for accurate targeting of steps correctly. Then there is also what happens if you zip two executions together with competing opposite options? 

@johnynek  thoughts?

isnotinvain:
+1 to converting this to an immutable API, in addition to all the reasons Ian pointed out, it's also consistent / expected w.r.t the rest of the scalding API.

Giving Execution a method `withConfig` that returns a new `Execution` with the config applied inside it sounds like a reasonable API for this to me.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: +1 to the comment of no mutation being a requirement here.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
isnotinvain: Yeah lots of corner cases here. You can already set config options per-source and per-sink, which has at least more easily understood semantics

scalding-commons/src/test/scala/com/twitter/scalding/ExecutionTest.scala
```diff
@@ -169,6 +226,33 @@ class ExecutionTest extends WordSpec with Matchers {
...
   "Executions" should {
+    "allow setting config key values on a per execution basis" in {
+
+      val testInput: List[(String, Int)] = List(("a", 1), ("b", 2))
+
+      // Sources
+      def ourType(str: String) = TypedTsv[(String, Int)](str)
+
+      // A method that runs a JobTest where the sources don't match
+      JobTest(new ConfTestExecution(_))
```
ianoc: Need a platform test here too

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: What is the scope of this? There needs to be really precise contracts here to make a new API. I'm pretty nervous what this will mean, and that users will start coding to accidental implementation details which will be really hard to keep stable.

johnynek:
Why not use `Execution.onSuccess` and flatMap to do monitoring? Not sure hard wiring Configs is the right way to go. Can you share other ideas for using this API?

lclementi:
So I'm trying to monitor the Stats counter _in real time_ while the various jobs run.

The way to do that with execution AFAIK is to implement a FlowStepListener, where I can access the getFlowStepStats() information at runtime.

The problem is that the "user" (who writes the job) can zip multiple executions (where each execution represents a "logical job") and run them in parallel (zip).

So now I have the problem of understanding which flow is which, I know I can use the cascading.flow.step.id but then the "user" need to do the correlation between his job and the step.id (which is also not exposed AFAIK). 

Another idea I had was to expose the RichPipe.name(), but I thought that the "Config" options is more flexible in this case.

So my idea is that the "user" can set somehow a job name on a execution basis so that the FlowStepListener can then understand which flow is dealing when it intercept them in the FlowStepListener and collect the stats associating them to job name selected by the "user".

Does it make sense?

ianoc:
Can you use the withDescription to identify the step your running in? There are 2 resolutions of descriptions included in the jobconfs today. Automatically we add ones about where things were called from(file name, line number), but you can add more

lclementi:
@ianoc I suppose I can use withDescription (that is where I "copied" the code to implement setPipeExtraConfig).

I just thought it was cleaner and more error prone to use a separate withExtraConfig method instead of mixing various content under the same field. 

I have the feeling that if I encode a hypothetical ExecutionDescription field inside the PipeDescription list it might look very messy.

I feel the best way might be to merge the WithDescriptionTypedPipe class and the WithExtraConfig together so both withDescription and withExtraConfig are using the same implementation (less code, cleaner API and lots of free unit test for withExtraConfig).


Luca

ianoc:
Superseeded by #1489

PR #1472, opened 2016-01-14
 Limit task history fields consumed from hraven

rubanm:
Some jobs internally have been running into http content limits set by hraven. We really only need a few of the task history fields that are currently being consumed.

This change uses the new response filters added in hraven client to limit the amount of data being read:
https://github.com/twitter/hraven/blob/master/hraven-core/src/main/java/com/twitter/hraven/rest/client/HRavenRestClient.java#L356

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -31,7 +31,7 @@ object ScaldingBuild extends Build {
   val hadoopLzoVersion = "0.4.19"
   val hadoopVersion = "2.5.0"
   val hbaseVersion = "0.94.10"
-  val hravenVersion = "0.9.16"
+  val hravenVersion = "0.9.17.t05"
   val jacksonVersion = "2.4.2"
   val json4SVersion = "3.2.11"
   val paradiseVersion = "2.0.1"
--- a/scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
+++ b/scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
@@ -176,9 +176,31 @@ object HRavenHistoryService extends HistoryService {
         keys = FlowStepKeys(step.getJobName, step.getUser, step.getPriority, step.getStatus, step.getVersion, "")
         // update HRavenHistoryService.TaskDetailFields when consuming additional task fields from hraven below
         tasks = step.getTasks.asScala.map { t => Task(t.getType, t.getStatus, t.getStartTime, t.getFinishTime) }
-      } yield FlowStepHistory(keys, step.getSubmitTime, step.getLaunchTime, step.getFinishTime, step.getTotalMaps, step.getTotalReduces, step.getFinishedMaps, step.getFinishedReduces, step.getFailedMaps, step.getFailedReduces, step.getMapFileBytesRead, step.getMapFileBytesWritten, step.getReduceFileBytesRead, step.getHdfsBytesRead, step.getHdfsBytesWritten, step.getMapSlotMillis, step.getReduceSlotMillis, step.getReduceShuffleBytes, 0, tasks)
+      } yield toFlowStepHistory(keys, step, tasks)
     }
 
+  private def toFlowStepHistory(keys: FlowStepKeys, step: JobDetails, tasks: Seq[Task]) =
+    FlowStepHistory(
+      keys = keys,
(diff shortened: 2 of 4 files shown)


project/Build.scala
```diff
@@ -31,7 +31,7 @@ object ScaldingBuild extends Build {
   val hadoopLzoVersion = "0.4.19"
   val hadoopVersion = "2.5.0"
   val hbaseVersion = "0.94.10"
-  val hravenVersion = "0.9.16"
+  val hravenVersion = "0.9.17.t05"
```
ianoc: Is t05 a public release?

ianoc:
Could you talk about if we will loose compatibility with older versions of hraven or if its a fall back. Have we manually tested this in real jobs? 

(Does anyone outside twitter even use hraven?)

project/Build.scala
rubanm: Yep it's on maven central and sonatype 
https://oss.sonatype.org/content/repositories/releases/com/twitter/hraven/hraven-core/

rubanm:
We will lose compatibility with older versions. The newer version is already deployed internally at twitter. I've run a test cron job to verify hraven history gets picked up correctly. cc @vrushalivc 

I'm not aware of anyone outside of twitter using the hraven estimators, so we're probably fine here.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -165,7 +174,8 @@ object HRavenHistoryService extends HistoryService {
       for {
         step <- history
         keys = FlowStepKeys(step.getJobName, step.getUser, step.getPriority, step.getStatus, step.getVersion, "")
-        tasks = step.getTasks.asScala.map{ t => Task(t.getTaskId, t.getType, t.getStatus, t.getSplits.toSeq, t.getStartTime, t.getFinishTime, t.getTaskAttemptId, t.getTrackerName, t.getHttpPort, t.getHostname, t.getState, t.getError, t.getShuffleFinished, t.getSortFinished) }
+        // update HRavenHistoryService.TaskDetailFields when consuming additional task fields from hraven below
+        tasks = step.getTasks.asScala.map { t => Task(t.getType, t.getStatus, t.getStartTime, t.getFinishTime) }
       } yield FlowStepHistory(keys, step.getSubmitTime, step.getLaunchTime, step.getFinishTime, step.getTotalMaps, step.getTotalReduces, step.getFinishedMaps, step.getFinishedReduces, step.getFailedMaps, step.getFailedReduces, step.getMapFileBytesRead, step.getMapFileBytesWritten, step.getReduceFileBytesRead, step.getHdfsBytesRead, step.getHdfsBytesWritten, step.getMapSlotMillis, step.getReduceSlotMillis, step.getReduceShuffleBytes, 0, tasks)
```
johnynek: can this be factored out to a method call? The line is super long.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -38,6 +38,15 @@ object HRavenHistoryService extends HistoryService {
 
   private val LOG = LoggerFactory.getLogger(this.getClass)
 
+  // List of fields that we consume from fetchTaskDetails api.
+  // This is sent to hraven service to filter the response data
+  // and avoid hitting http content length limit on hraven side.
+  private val TaskDetailFields = List(
```
rubanm: We could make this overridable as Option[List[String]] so any external cases still work with the older hraven version.

johnynek:
This means a version bump if we are strict (not sure we should be for a submodule that maybe no one uses).

rubanm:
Refactored the long line. Not bumping the version for now (we'll need to go from 0.15 to 0.16 to be strict?).

ianoc:
@johnynek We sort of have been on RC's for too long around 0.16 i guess anyway. But thoughts in general about the backwards incomaptible for external users here. I'm not sure it really matters since I doubt anyone external to twitter uses this. I'd almost be inclined to say in future this is penciled for something that could just go back internal to twitter and not be in the OSS project

johnynek:
Yeah, the specific Hraven calls should probably be internal to Twitter. LGTM otherwise.

PR #1475, opened 2016-01-16
ScaldingILoop should enable one to pass in in/out

sriramkrishnan:
I had to fork the ScaldingILoop to get it to work with Zeppelin because it didn't provide me a way to pass thru the in/out that it needs. See:
https://github.com/apache/incubator-zeppelin/blob/master/scalding/src/main/scala/org/apache/zeppelin/scalding/ScaldingILoop.scala and https://github.com/apache/incubator-zeppelin/pull/561.

I want to add that into Scalding proper so I can kill that fork. Once we release the next version of Scalding, I can create a PR to kill that code from Zeppelin. The Scalding REPL tests pass, and I already have a branch that works at:
https://github.com/sriramkrishnan/incubator-zeppelin/tree/scalding_cleanup

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -113,7 +113,7 @@ object ScaldingBuild extends Build {
     testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-oF"),
 
     // Uncomment if you don't want to run all the tests before building assembly
-    test in assembly := {},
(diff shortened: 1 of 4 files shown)


sriramkrishnan:
Thanks @ianoc. Will update Zeppelin after the next Scalding release.

PR #1478, opened 2016-01-21
Add two functions that assist in testing a TypedPipe

richwhitjr:
Many times when testing a scalding job you just want to test a function that takes a TypedPipe and returns another TypedPipe.  This can be difficult using JobTest.  TypedPipeChecker makes use of Executions to pull the given TypedPipe into memory so that can asserts can be run against the resulting List.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
@@ -9,7 +9,7 @@ object TypedPipeChecker {
    * Takes a List and a transform function.
    * The resulting TypedPipe form the transform will be run through asserts
    */
-  def checkOutputTransform[T, U](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => Unit) =
+  def checkOutputTransform[T, U, R](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => R): R =
     assertions(checkOutputInline(transform(TypedPipe.from(input))))
 
   /*
@@ -17,7 +17,7 @@ object TypedPipeChecker {
    * a list and run it through a function that makes arbitrary
    * assertions on it.
    */
-  def checkOutput[T](output: TypedPipe[T])(assertions: List[T] => Unit) =
+  def checkOutput[T, R](output: TypedPipe[T])(assertions: List[T] => R): R =
     assertions(checkOutputInline(output))
 
   /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
@@ -10,7 +10,7 @@ object TypedPipeChecker {
    * The resulting TypedPipe form the transform will be run through asserts
    */
   def checkOutputTransform[T, U, R](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => R): R =
-    assertions(checkOutputInline(transform(TypedPipe.from(input))))
+    assertions(inMemoryToList(transform(TypedPipe.from(input))))
 
   /*
    * Execute a TypedPipe in memory, convert the resulting Iterator to
@@ -18,12 +18,12 @@ object TypedPipeChecker {
    * assertions on it.
    */
   def checkOutput[T, R](output: TypedPipe[T])(assertions: List[T] => R): R =
-    assertions(checkOutputInline(output))
+    assertions(inMemoryToList(output))
 
   /**
    * Execute a TypedPipe in memory and return the result as a List
    */
-  def checkOutputInline[T](output: TypedPipe[T]): List[T] =
+  def inMemoryToList[T](output: TypedPipe[T]): List[T] =
     output
       .toIterableExecution
       .waitFor(Config.unitTestDefault, Local(strictSources = true))
--- a/scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
@@ -1,0 +1,24 @@
+package com.twitter.scalding
+
+/**
+ * This class is used to assist with testing a TypedPipe
+ */
+object TypedPipeChecker {
+  /*
+   * Execute a TypedPipe in memory, convert the resulting Iterator to
+   * a list and run it through a function that makes arbitrary
+   * assertions on it.
+   */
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * This class is used to assist with testing a TypedPipe
+ */
+object TypedPipeChecker {
+  /*
+   * Execute a TypedPipe in memory, convert the resulting Iterator to
+   * a list and run it through a function that makes arbitrary
+   * assertions on it.
+   */
+  def checkOutput[T](output: TypedPipe[T])(assertions: List[T] => Unit) =
```
ianoc: Is it worth having a flavor like
```
def checkOutputTransform[T, U](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[T] => Unit) =
  assertions(checkOutputInline(transform(TypedPipe.from(input))))
```

Or doing TypedPipe.from we should leave to the test writer?

richwhitjr: Sure that actually may be nice to have.

richwhitjr: Function added

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * This class is used to assist with testing a TypedPipe
+ */
+object TypedPipeChecker {
+
+  /*
+   * Takes a List and a transform function.
+   * The resulting TypedPipe form the transform will be run through asserts
+   */
+  def checkOutputTransform[T, U](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => Unit) =
```
johnynek: can we put a return type on all methods?

also, what about mre general:
```scala
def checkOutputTransform[T, U, R](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => R): R =
```
then you could use methods that return other than unit.

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
johnynek: same comment here: make the return generic please.

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
```diff
@@ -0,0 +1,32 @@
...
+   * a list and run it through a function that makes arbitrary
+   * assertions on it.
+   */
+  def checkOutput[T, R](output: TypedPipe[T])(assertions: List[T] => R): R =
+    assertions(checkOutputInline(output))
+
+  /**
+   * Execute a TypedPipe in memory and return the result as a List
+   */
+  def checkOutputInline[T](output: TypedPipe[T]): List[T] =
```
johnynek: any reason not to call this `inMemoryToList` or something?

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
```diff
@@ -0,0 +1,32 @@
...
+   */
+  def checkOutputTransform[T, U, R](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => R): R =
+    assertions(checkOutputInline(transform(TypedPipe.from(input))))
+
+  /*
+   * Execute a TypedPipe in memory, convert the resulting Iterator to
+   * a list and run it through a function that makes arbitrary
+   * assertions on it.
+   */
+  def checkOutput[T, R](output: TypedPipe[T])(assertions: List[T] => R): R =
```
johnynek: maybe ` def mapAll` is a name closer to what we are doing?

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * This class is used to assist with testing a TypedPipe
+ */
+object TypedPipeChecker {
+
+  /*
+   * Takes a List and a transform function.
+   * The resulting TypedPipe form the transform will be run through asserts
+   */
+  def checkOutputTransform[T, U, R](input: List[T])(transform: TypedPipe[T] => TypedPipe[U])(assertions: List[U] => R): R =
```
johnynek: `def transformMapAll`?

johnynek:
ready to merge, just some name bikeshedding.

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
richwhitjr: I like this name for this function.  I kind of like idea of keeping the names for the others though since it really is just a nice to have utility function for running some asserts on a typed pipe and the name indicates that it probably should only be used for unit testing.

If you think it could be used outside of tests though we can try to think of some better names.

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
richwhitjr: Mentioned above, I don't mind this name too much since it indicates that it should probably only be used for tests.  With mapAll we should also warn that it pulls everything into memory.  mapAllAsList?  If we can imagine people using this function in normal jobs it may be worthwhile to think about other names.

scalding-core/src/main/scala/com/twitter/scalding/TypedPipeChecker.scala
richwhitjr: Same point as checkOutput.

isnotinvain:
Is there any need for the checkOutputTransform part?
Why not just keep the inMemoryToList part, and then let users use it directly, eg:

```
val expected = List(1,2,3)
val found = inMemoryToList(myTestedMethod(TypedPipe.from(List("a", "b", "c"))))
assert(found === expected)
```

richwhitjr:
Mostly just a simple utility function.  Not strictly needed.  You could imagine have a function that checks a  TypedPipe that you may want reuse on different tests so it could be a little cleaner to use the checkOutput function.

Sometimes just having the function where you are doing the testing can make the test a little more readable.  

I can see the transform being useful when you have some function of TypedPipe[T] -> TypedPipe[U] and you want to test this.  This is common in many of my real world tests.

PR #1479, opened 2016-01-21
Makes permission failures non-fatal when looking for .scalding_repl files

benpence:
It seems that on unixy systems, given the following permission tree structure:

    d--x--x--x  dir/
    drwxr-xr-x  dir/file

anybody can still read `dir/file`. When we're looking for `.scalding_repl` files, we don't take this into consideration. Now, instead of failing in this situation, we `println` the problem and ignore it.

BTW is there a reason we don't use `System.getenv("HOME")` here?


ianoc:
Non $HOME is because you could have more local overrides to the setting in different working paths you invoke the repl from. Or so goes the thinking anyway

johnynek:
looks fine to me

PR #1483, opened 2016-01-28
Utility for expanding libjars

ianoc:
Hadoop's -libjars doesn't support wildcards, with large class paths its easy to exhaust the max arg length for linux/os x when running commands. This acts as a filter above our interaction with the generic options parser to expand wildcards

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/ExpandLibJarsGlobsTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/ExpandLibJarsGlobsTest.scala
@@ -12,12 +12,17 @@ class ExpandLibJarsGlobsTest extends WordSpec with Matchers {
     f.getAbsolutePath
   }
 
+  def getTmpRoot = {
+    val tmpRoot = new File(System.getProperty("java.io.tmpdir"), scala.util.Random.nextInt(Int.MaxValue).toString)
+    require(tmpRoot.mkdirs(), "Failed to make temporary directory")
+    tmpRoot.deleteOnExit()
+    tmpRoot
+  }
+
   "ExpandLibJarsGlobs" should {
     "expand entries" in {
-      val tmpRoot = new File(System.getProperty("java.io.tmpdir"), System.currentTimeMillis.toString)
-      require(tmpRoot.mkdirs(), "Failed to make temporary directory")
-      tmpRoot.deleteOnExit()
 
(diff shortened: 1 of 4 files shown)


scalding-core/src/test/scala/com/twitter/scalding/ExpandLibJarsGlobsTest.scala
```diff
@@ -0,0 +1,51 @@
...
+        touch(tmpRoot, s"myF_${idx}.jar")
+      }
+
+      val resultingLibJars1 = ExpandLibJarsGlobs(Array("-libjars", s"${tmpRoot.getAbsolutePath}/*.jar"))(1).split(",")
+      assert(resultingLibJars1.sorted.toList == jars.sorted.toList)
+
+      val resultingLibJars2 = ExpandLibJarsGlobs(Array("-libjars", s"${tmpRoot.getAbsolutePath}/"))(1).split(",")
+      assert(resultingLibJars2.sorted.toList == jars.sorted.toList)
+
+      val resultingLibJars3 = ExpandLibJarsGlobs(Array("-libjars", s"${tmpRoot.getAbsolutePath}/*"))(1).split(",")
```
rubanm: needs another test for multiple libjar dirs?

rubanm:
LGTM, just one comment about the tests

scalding-core/src/test/scala/com/twitter/scalding/ExpandLibJarsGlobsTest.scala
ianoc: Added

johnynek:
Doesn't the shell already expand out wildcards? does this actually work?

ianoc:
Shell only expands something that 'appears like a path', what that means is if you for instance do something like:

java -cp :libs/* it will not be expanded
java -cp libs/* would be expanded however

Skipping the shell expansion is the goal here since a large number of files in the libs folder hits the args max limit for your shell/kernel causing it to fail. Java's support of wildcards for the classpath in the example above continues to work, in hadoop land that effectively means HADOOP_CLASSPATH works but libjars does not

johnynek:
to be clear the use case here would be to prefix with `,`: `-libjars ,foo/*` should work right?

ianoc:
Yep, if you had a -libjars foo/* it actually would fail I believe since that would be whitespace separated. (similar trick required for java classpath env variables)

PR #1484, opened 2016-01-28
Adds a pass at an alternative lazy parquet reader implementation

ianoc:
Needs some more work before merging around cleaning it up, more benchmarks and seeing what does/doesn't need to be in lui.

TBD are real world tests against jobs with manual projection.


isnotinvain:
One thing I'm working on is seeing if any of the classes that are mostly scala re-writes of parquet classes can either use the original parquet classes, or even better be skipped entirely. Parts of Parquet's class hierarchy can be pretty confusing, I'd rather not mirror that all over again here, especially the classes with non trivial code in them (which have tests in parquet but not here). So for the stuff we need to keep, I'm also working on adding some tests.

PR #1485, opened 2016-01-29
Shouldn't skip hidden files, user can decide such things with their glob

ianoc:


johnynek:
I guess this caused an issue with an internal test?

ianoc:
Yep :/ worked under a bunch of things... then not some others. pants generated targets for things like scrooge use hidden file names for their jars

johnynek:
weird. LGTM.

benpence:
+1

PR #1487, opened 2016-01-30
Fix FileSystem.get issue

### LINKED ISSUE #1486: FileSystem.get(conf) is not safe
This always gets the default FS and precludes using something like s3. See:

https://forums.aws.amazon.com/thread.jspa?threadID=30945

The correct pattern appears to be `FileSystem.get(uri, conf)` where `uri` points to the FileSystem in question.

This happens three times in this file:
https://github.com/twitter/scalding/blob/develop/scalding-commons/src/main/java/com/twitter/scalding/commons/tap/VersionedTap.java#L154

johnynek:
fixes #1486 

Tested on S3. Seems to work.


ianoc:
LGTM, merge when green

PR #1488, opened 2016-02-01
Remove dependency on parquet-cascading

rubanm:
This PR drops our dependency on parquet-cascading for easier cascading upgrades in the future (decoupled from apache's parquet release process). 

As a follow-up to this change, I will copy over @cchepelov's cascading3 changes from https://github.com/apache/parquet-mr/pull/284 and file an issue on parquet-mr to possibly remove their version of this code (presumably, no one outside of scalding consumes this).

I copied over the java code instead of translating to scala so that
1. We do not lose anything in translation
2. Cyrille's already done work for cascading3 can be easily ported
3. A majority of the code here is just wiring things down to cascading's layer which imo wouldn't give us much in the scala version.

But let me know if anyone feels strongly about creating a scala version of these schemes instead.

cc @ianoc @isnotinvain @johnynek @julienledem


project/Build.scala
```diff
@@ -347,11 +349,39 @@ object ScaldingBuild extends Build {
...
+
   lazy val scaldingParquet = module("parquet").settings(
     libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
+      "org.apache.parquet" % "parquet-column" % parquetVersion,
+      "org.apache.parquet" % "parquet-hadoop" % parquetVersion,
+      "org.apache.parquet" % "parquet-pig" % parquetVersion,
+      "org.apache.parquet" % "parquet-thrift" % parquetVersion
       // see https://issues.apache.org/jira/browse/PARQUET-143 for exclusions
-      "org.apache.parquet" % "parquet-cascading" % parquetVersion
-        exclude("org.apache.parquet", "parquet-pig")
```
johnynek: why don't we need this exclude?

rubanm: Added back the exclude. There was an unused pig import in the original code which we don't need.

julienledem:
@rubanm sounds good to me.

rubanm:
This passes Twitter internal e2e tests.

johnynek:
:+1: basically this is becoming a scalding mono-repo...

rubanm:
Hah, with monobuild first this time around.

PR #1489, opened 2016-02-01
Add withConfig api to allow running an execution with a transformed config

ianoc:
Adds the the ability to transform Config's (and hence args) for sub-sections of Execution flows. This can be useful to override hadoop or source level options in subsections. This lets the user have more control over things like split sizes, memory used in mappers/reducers, combining small files, etc..


johnynek:
I think the cache key needs to include the config if we do this.

Also, if we go this route, I would say we change `Execution` from a `Reader[(Config Mode), T]` to `State[(Config, Mode), T]` and even allow mode changes in the flow.

richwhitjr:
I like this idea, will help in many cases where we have to go wide on a source using splits map side but want to write larger files to hdfs at the end of a job.  Currently there is not much choice aside from using .shard

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -209,6 +209,9 @@ object Execution {
     override def join[T, U](t: Execution[T], u: Execution[U]): Execution[(T, U)] = t.zip(u)
   }
 
+  def withConfig[T](ex: => Execution[T])(c: Config => Config): Execution[T] =
```
johnynek: if this is lazy, the arg in the constructor needs to be, otherwise this should not be lazy.

johnynek:
Looks great! (1 minor issue)

ianoc:
Killed the lazy, not useful I don't think here

ianoc:
Good to go now @johnynek ?

PR #1490, opened 2016-02-01
Call validateTaps in toIterator codepath

rubanm:
https://github.com/twitter/scalding/pull/1458 was meant to ensure we throw InvalidSourceException for all cases, but caused RuntimeException to be thrown in the toIterator codepath (instead of hadoop's InvalidInputException as was the case before that change). I looked into fixing it so that it throws InvalidSourceException.

As a more uniform solution though, I propose we call validateTaps in the toIterator codepath for all sources. Currently, toIterator can throw a few different exception types (hadoop's InvalidInputException, scalding' InvalidSourceException, RuntimeException) depending on whether we are in strict vs non strict mode, etc. This change will help ensure we throw scalding's InvalidSourceException for all cases.


johnynek:
LGTM. others?

ianoc:
LGTM also

reconditesea:
LGTM.

PR #1492, opened 2016-02-02
Arg Descriptions/Help for Execution Apps

richwhitjr:
Many times I think it would be helpful to have described Args for a Scalding Job so you can run a help command.  This is difficult though because of the lazy arg model used by Scalding.  I really didn't want to change Arg parsing in Scalding so instead this review does the simple thing and lets users add descriptions to Args and see the descriptions with a --help command option.


scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,73 @@
...
+
+object ArgOptions extends Enumeration {
+  type ArgOptions = Value
+  val RequiredArg, OptionalArg, ListArg, BooleanArg = Value
+}
+
+class DescriptionArgs(override val m: Map[String, List[String]]) extends Args(m){
+  import ArgOptions._
+
+  private val map = new scala.collection.mutable.HashMap[String, (String, ArgOptions)]()
```
ianoc: Does this need to be mutable? seems unfortunate to stick a mutable thing in place here

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,73 @@
+package com.twitter.scalding
+
+object ArgOptions extends Enumeration {
```
johnynek: can we use `sealed abstact class` and `object` subclasses instead? `Enumeration has a lot of issues and many consider it non-idiomatic.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
johnynek: +1 no mutability here. Have the methods below return new instances.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,73 @@
...
+  type ArgOptions = Value
+  val RequiredArg, OptionalArg, ListArg, BooleanArg = Value
+}
+
+class DescriptionArgs(override val m: Map[String, List[String]]) extends Args(m){
```
johnynek: why not put this in scalding-args package?

johnynek:
There is an ugly hack with Args that are passed to `Job` to hide a `Mode` in there so we didn't break source compatibility when we fixed a design flaw long ago. This, I think, is safe because it is never used with Job, but if a user did so, it would give a runtime error.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
richwhitjr: Removed the mutable, instead went with a List of case classes.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,57 @@
+package com.twitter.scalding
+
+sealed abstract class ArgOptions
+
+object RequiredArg extends ArgOptions
```
ianoc: case object and sealed abstract trait are normally used here.

Should these all be private?

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,57 @@
...
+object OptionalArg extends ArgOptions
+object ListArg extends ArgOptions
+object BooleanArg extends ArgOptions
+
+case class DescribedArg(key: String, description: String, argOption: ArgOptions)
```
ianoc: Would using a builder style pattern not be cleaner/closer to what you had before?

DescribedArgs(args).withRequired("key"...).withRequired(...)

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,57 @@
+package com.twitter.scalding
+
+sealed abstract class ArgOptions
+
+object RequiredArg extends ArgOptions
+object OptionalArg extends ArgOptions
+object ListArg extends ArgOptions
```
ianoc: If going this route I think the API would be cleaner if these were case classes extending ArgOptions, taking key and description.

richwhitjr: Good point, I wasn't a huge fan of exposing that parameter.  I also thought about the builder pattern but I liked the List because I can share a list of Described Args more easily between jobs.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,57 @@
...
+      str + s"--$key($opt) :: $description \n"
+    } + "--help :: Show this help message."
+  }
+}
+
+/**
+  * Extending this Trait allows for descriptions to be added to Jobs
+  * If you run a job with --help the descriptions are output
+  */
+trait DescribedExecutionApp { this: ExecutionApp =>
```
johnynek: rather than a mixin style could we do:
```
object ArgHelp {
  def with(describedArgs: Seq[DescribedArg], ex: Execution[Unit]): Execution[Unit] = 
    // your code here?
}
```
Then it could be used with other Execution combinators.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,57 @@
...
+sealed abstract class ArgOptions
+
+object RequiredArg extends ArgOptions
+object OptionalArg extends ArgOptions
+object ListArg extends ArgOptions
+object BooleanArg extends ArgOptions
+
+case class DescribedArg(key: String, description: String, argOption: ArgOptions)
+
+case class ArgHelp(describedArgs: Seq[DescribedArg]) {
```
johnynek: still like moving this to `scalding-args` project

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
richwhitjr: Good idea, I was imaging some cases where the mixin style could be odd for users.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,64 @@
...
+case class RequiredArg(key: String, description: String) extends DescribedArg
+case class OptionalArg(key: String, description: String) extends DescribedArg
+case class ListArg(key: String, description: String) extends DescribedArg
+case class BooleanArg(key: String, description: String) extends DescribedArg
+
+case class ArgHelp(describedArgs: Seq[DescribedArg]) {
+  def argString: String = {
+    describedArgs.foldLeft("") { case (str, describedArg) =>
+      val msg = describedArg match {
+        case RequiredArg(key, _) => s"--$key VALUE "
```
johnynek: could this be a method on `DescribedArg`?

johnynek:
:+1:

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+  }
+
+  /**
+    * Command line arg string given the Described Args
+    * @param describedArgs List of Argument Descriptions
+    * @return Command Line Parameters
+    */
+  def argString(describedArgs: Seq[DescribedArg]): String = {
```
ianoc: This isn't used now, is it needed with the help string? is it maybe the usage string?

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
```diff
@@ -0,0 +1,69 @@
...
+
+object ArgHelp {
+  /**
+    * Describe the Arguments of this Execution.  By running --help the args will output
+    * and the execution will end
+    * @param describedArgs List of Argument Descriptions
+    * @param ex Input Execution
+    * @return Output Execution
+    */
+  def describe(describedArgs: Seq[DescribedArg], ex: Execution[Unit]): Execution[Unit] = {
```
ianoc: Should we be doing validation in here? we don't validate at all anymore right?

ianoc:
I'd love to see an example/tutorial added or a test to act as documentation if nothing else.

johnynek:
+1 for test. Sorry I missed that.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
richwhitjr: Is is still output for the command line option.  I wanted the output to look like

--key required [--key optional] 

Then output the detailed help below.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
richwhitjr: Maybe a second method that does validation and leave it up to the user to decide if they want to describe all the arguments or not?

richwhitjr:
Yep tests coming, actually much easier now without the mixin.  Wanted to nail down the API first.

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
ianoc: A second method sounds reasonable, or a flag here. None of the information is passed onto the output execution here so chaining is likely difficult.

ianoc: Also maybe describe should be typed on T and passthrough that type T rather than Unit?

scalding-core/src/main/scala/com/twitter/scalding/DescribedExecutionApp.scala
ianoc: Ok cool, lets make it private[this] if we just expect it to be used above, since we won't have to worry about binary changes later

richwhitjr: Made it package private so I could override it for the unit tests.

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,106 @@
+package com.twitter.scalding
+
+sealed abstract class DescribedArg {
```
ianoc: Why abstract class than abstract trait? seems an unusual choice. No real problem with it...just an unusual pattern

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,106 @@
...
+case class BooleanArg(key: String, description: String) extends DescribedArg
+
+trait ArgHelper {
+  /**
+   * Similar to describe but validate all args are described
+   * @param describedArgs List of Argument Descriptions
+   * @param ex Input Execution
+   * @return Output Execution
+   */
+  def validatedDescribe(describedArgs: Seq[DescribedArg], ex: Execution[Unit]): Execution[Unit] = {
```
ianoc: Any reason not to put a type T on the Executions so you can pass through your state here?

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
richwhitjr: Didn't give it too much thought.  I made the second one a trait so I can easily extend it for my unit tests.  I could make it an abstract class though.

ianoc: Well more usually we have these all be sealed traits. Given its sealed I don't actually know if it makes a difference really.

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
richwhitjr: Unsure of how to handle help.  I could mark the Execution as a failure when help is given but that doesn't seem correct and the user would have to handle the Help Error upstream.  I could also output the T wrapped by Option but that always seem awkward.

Any suggestion?

ianoc: Mmm I think maybe an exception is the way to go here. I think this API will be a bit funky otherwise. Since if you continue on from the returned execution from this in a chain it will throw away the fact the help was printed since its not represented in the type system that you should exit here. This would wind up really only being suitable for the root entry point from Execution App? It seems like you should be able to zip two of these together or flatMap them. 

@johnynek ^ thoughts?

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,108 @@
...
+    val describedKeys = describedArgs.map(_.key).toSet
+    Execution.getArgs.flatMap { args =>
+      val missingKeys = args.m.keySet.filter(_.nonEmpty).diff(describedKeys)
+
+      if (args.boolean("help")) {
+        helpRequest(describedArgs)
+      } else {
+        if (missingKeys.nonEmpty) {
+          val msg = missingKeys.mkString(", ")
+          Execution.failed(throw new DescriptionValidationException(s"Must describe missing keys : $msg"))
```
ianoc: throw here is wrong, this is actually throwing inside the outer execution I believe, the failed method never gets called. Just drop throw and should be good

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  def helpRequest[T](describedArgs: Seq[DescribedArg]): Execution[T] = {
+    val top = "\n###########################################################################\n\n"
+    val usage = s"Command Line Args :: ${argString(describedArgs)}\n\n\n"
+    val bottom = "\n\n###########################################################################\n"
+
+    println(top + usage + help(describedArgs) + bottom)
+
+    Execution.failed(throw new HelpException())
```
ianoc: as above

scalding-core/src/test/scala/com/twitter/scalding/ArgHelpTest.scala
```diff
@@ -0,0 +1,79 @@
+package com.twitter.scalding
+
+import org.scalatest.{ Matchers, WordSpec }
+
+case class ArgHelperTest(testFn: Seq[DescribedArg] => Unit) extends ArgHelper {
+  override def helpRequest[T](describedArgs: Seq[DescribedArg]): Execution[T] = {
+    testFn(describedArgs)
+    Execution.failed(throw new HelpException())
```
ianoc: also here

ianoc:
With the fixed exception handling, lgtm.

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,108 @@
...
+    Execution.getArgs.flatMap { args =>
+      if (args.boolean("help")) {
+        helpRequest(describedArgs)
+      } else {
+        ex
+      }
+    }
+  }
+
+  def helpRequest[T](describedArgs: Seq[DescribedArg]): Execution[T] = {
```
johnynek: this should return `Execution[Nothing]` and not have a `T` anywhere.

johnynek: actually it should return `Nothing`.

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,108 @@
...
+   * and the execution will end
+   *
+   * @param describedArgs List of Argument Descriptions
+   * @param ex Input Execution
+   * @return Output Execution
+   */
+  def describe[T](describedArgs: Seq[DescribedArg], ex: Execution[T]): Execution[T] = {
+    Execution.getArgs.flatMap { args =>
+      if (args.boolean("help")) {
+        helpRequest(describedArgs)
```
johnynek: can we factor this bit into a method that could be used with Job?
```scala
def describe(describedArgs: Seq[DescribedArg], args: Args): Unit =
  if (args.boolean("help")) helpRequest(describedArgs)
  else ()
```
then call that in your code above?

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
richwhitjr: Not sure what you mean here.  No longer pass in the Execution in the method or just wildcard it and return Nothing?

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
richwhitjr: So keep describe as above but call this new method for help?

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
ianoc: I think what he means is just a function signature like:
```
def helpRequest(describedArgs: Seq[DescribedArg]): Nothing = {
```

Since you never actually return an Execution

richwhitjr: Ah for helpRequest, was looking at describe.

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
ianoc: For both describe and validate we can have functions that go Seq[DescribedArg] => Nothing, Then we can call those from job's or anywhere else I think would be his idea. But can still have the execution variants too

scalding-core/src/main/scala/com/twitter/scalding/ArgHelp.scala
```diff
@@ -0,0 +1,122 @@
...
+   * Describe a set of Args given Descriptions
+   *
+   * @param describedArgs List of Argument Descriptions
+   * @param args Job Arguments
+   */
+  def describe(describedArgs: Seq[DescribedArg], args: Args): Unit =
+    if (args.boolean("help")) helpRequest(describedArgs)
+    else ()
+
+  def helpRequest[T](describedArgs: Seq[DescribedArg]): Nothing = {
```
ianoc: drop the [T] here, looks unused

ianoc:
Think this looks by me, I had one small fix similar to one of those other orphan type parameters you've found before :)

PR #1493, opened 2016-02-02
Fix issue #1429

### LINKED ISSUE #1429: Specialize return types of `TypedText.{tsv,csv,osv}`
Currently the `TypedText` object provides the following helper functions:

```scala
def tsv[T: TypeDescriptor](path: String*): TypedTextDelimited[T] = new FixedTypedText[T](TAB, path: _*)
def osv[T: TypeDescriptor](path: String*): TypedTextDelimited[T] = new FixedTypedText[T](ONE, path: _*)
def csv[T: TypeDescriptor](path: String*): TypedTextDelimited[T] = new FixedTypedText[T](COMMA, path: _*)
```

The type annotation on the return values (`TypedTextDelimited`) is more general than the right-hand side (`FixedTypedText`), which means that they will fail for any function expecting a `FileSource` super-class, such as `TypedPipe.make`:

```scala
scala> execute(someTypedPipe.make[Int](TypedText.tsv[Int]("temp.tsv")))
<console>:27: error: type mismatch;
 found   : com.twitter.scalding.source.TypedTextDelimited[Int]
 required: com.twitter.scalding.FileSource with com.twitter.scalding.typed.TypedSink[Int] with com.twitter.scalding.typed.TypedSource[Int]
                                      execute(someTypedPipe.make[Int](TypedText.tsv[Int]("temp.tsv")))
```

... and they would succeed if they had the more specific `FixedTypedText` annotation instead of `TypedTextDelimited` because `FixedTypedText` extends `FileSource` whereas `TypedTextDelimited` does not.

Is there any reason for the `TypedTextDelimited` annotations, or could they be changed to `FixedTypedText`?

johnynek:
closes #1429

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/source/TypedText.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/source/TypedText.scala
@@ -18,9 +18,12 @@ object TypedText {
   val ONE = TypedSep("\u0001")
   val COMMA = TypedSep(",")
 
-  def tsv[T: TypeDescriptor](path: String*): TypedTextDelimited[T] = new FixedTypedText[T](TAB, path: _*)
-  def osv[T: TypeDescriptor](path: String*): TypedTextDelimited[T] = new FixedTypedText[T](ONE, path: _*)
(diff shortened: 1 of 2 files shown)


ianoc:
LGTM merge when green

isnotinvain:
+1

Issue #1494, opened 2016-02-04
Improve counter usage


johnynek:
We could get the reporter:
https://github.com/Cascading/cascading/blob/2.5.6/cascading-hadoop/src/main/shared/cascading/flow/hadoop/HadoopFlowProcess.java#L182

and then cache the relevant `Counter`s which should be close to zero cost.

Also we could cache the Counter for `Stats` as well, which should improve perf there.

johnynek:
see: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala#L172

PR #1495, opened 2016-02-04
Cache counters for stat updates

ianoc:
If we are in hadoop mode we can keep a handle to the Counter itself, rather than doing a lookup for every increment. So we should do it

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
@@ -123,6 +123,10 @@ package com.twitter.scalding {
    * to deal with that.  Since this does many fewer allocations, and has a smaller code-path it may be faster for
    * the typed-API.
    */
+  object MapsideReduce {
+    val COUNTER_GROUP = "MapsideReduce"
+  }
+
   class MapsideReduce[V](
     @transient commutativeSemigroup: Semigroup[V],
     keyFields: Fields, valueFields: Fields,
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -211,35 +207,50 @@ package com.twitter.scalding {
...
+        new SummingMapsideCache(flowProcess, new SummingWithHitsCache(size))
     }
   }
 
-  class SummingMapsideCache[V](summingCache: SummingWithHitsCache[Tuple, V])
+  class SummingMapsideCache[V](flowProcess: FlowProcess[_], summingCache: SummingWithHitsCache[Tuple, V])
     extends MapsideCache[V] {
+    private[this] val misses = CounterImpl(flowProcess, StatKey(MapsideReduce.COUNTER_GROUP, "misses"))
+    private[this] val hits = CounterImpl(flowProcess, StatKey(MapsideReduce.COUNTER_GROUP, "hits"))
+    private[this] val evicitions = CounterImpl(flowProcess, StatKey(MapsideReduce.COUNTER_GROUP, "evicitions"))
```
johnynek: evictions is misspelled.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -162,18 +165,11 @@ package com.twitter.scalding {
...
-      //iterator for performance
-      val it = stats.iterator
-      while (it.hasNext) {
-        val (key, value) = it.next
-        flowProcess.increment(COUNTER_GROUP, key, value)
-      }
-
       if (evicted.isDefined)
-        flowProcess.increment(COUNTER_GROUP, "evictions", evicted.get.size)
+        flowProcess.increment(MapsideReduce.COUNTER_GROUP, "evictions", evicted.get.size)
```
johnynek: this is duplicated.

johnynek:
Looks great!

closes #1494

johnynek:
:+1:

PR #1496, opened 2016-02-04
Improve type safety of internal `Boxed` logic

Gabriella439:
I've been applying `wartremover` [(Link)](https://github.com/puffnfresh/wartremover/tree/latest-release/core/src/main/scala/wartremover/warts) to the `scalding` code base to try to
improve type safety.  One of the first things singled out was the `Boxed` logic,
for three reasons:
1. The use of `Any`
2. The use of `asInstanceOf`
3. The exception when exhausting the finite supply of `Boxed{n}` classes

I was able to fix (1), which required factoring out the functions into
object-level `def`s, since AFAICT Scala does not permit polymorphic anonymous
functions.

I was not able to fix (2) because the `Class` class is invariant in its
type parameter.

I didn't understand the motivation behind the `Boxed` work-around to assess
whether or not the exception could be eliminated through more type safety.  All
I did was improve the error message to be more helpful.

Things that I would like feedback on:
- Will factoring out the anonymous functions into object-level `def`s have an impact on peformance?  Do we have a benchmark suite to detect performance regressions?
- Why is this `Boxed` workaround necessary?  I spent a lot of time trying to reverse engineer this from the code and comments in both `serialization/Boxed.scala` and `typed/Grouped.scala` (where it is used) and it still wasn't completely clear to me
- Why is `Class` invariant in its type parameter?  Is there a good reason for that?
- Are further type-safety refactors like this welcome?

### DIFF
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
@@ -543,268 +543,540 @@ case class BoxedOrderedSerialization[K](box: K => Boxed[K],
 }
 
 object Boxed {
-  private[this] val allBoxes = List(
-    ({ t: Any => new Boxed0(t) }, classOf[Boxed0[Any]]),
-    ({ t: Any => new Boxed1(t) }, classOf[Boxed1[Any]]),
-    ({ t: Any => new Boxed2(t) }, classOf[Boxed2[Any]]),
-    ({ t: Any => new Boxed3(t) }, classOf[Boxed3[Any]]),
-    ({ t: Any => new Boxed4(t) }, classOf[Boxed4[Any]]),
-    ({ t: Any => new Boxed5(t) }, classOf[Boxed5[Any]]),
-    ({ t: Any => new Boxed6(t) }, classOf[Boxed6[Any]]),
-    ({ t: Any => new Boxed7(t) }, classOf[Boxed7[Any]]),
-    ({ t: Any => new Boxed8(t) }, classOf[Boxed8[Any]]),
-    ({ t: Any => new Boxed9(t) }, classOf[Boxed9[Any]]),
-    ({ t: Any => new Boxed10(t) }, classOf[Boxed10[Any]]),
-    ({ t: Any => new Boxed11(t) }, classOf[Boxed11[Any]]),
-    ({ t: Any => new Boxed12(t) }, classOf[Boxed12[Any]]),
-    ({ t: Any => new Boxed13(t) }, classOf[Boxed13[Any]]),
-    ({ t: Any => new Boxed14(t) }, classOf[Boxed14[Any]]),
-    ({ t: Any => new Boxed15(t) }, classOf[Boxed15[Any]]),
-    ({ t: Any => new Boxed16(t) }, classOf[Boxed16[Any]]),
-    ({ t: Any => new Boxed17(t) }, classOf[Boxed17[Any]]),
-    ({ t: Any => new Boxed18(t) }, classOf[Boxed18[Any]]),
-    ({ t: Any => new Boxed19(t) }, classOf[Boxed19[Any]]),
-    ({ t: Any => new Boxed20(t) }, classOf[Boxed20[Any]]),
-    ({ t: Any => new Boxed21(t) }, classOf[Boxed21[Any]]),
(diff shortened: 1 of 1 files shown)


scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -543,268 +543,540 @@ case class BoxedOrderedSerialization[K](box: K => Boxed[K],
...
+  @SuppressWarnings(Array("org.brianmckenna.wartremover.warts.IsInstanceOf", "org.brianmckenna.wartremover.warts.Throw"))
+  def next[K]: (K => Boxed[K], Class[Boxed[K]]) = boxes[K].get match {
     case list @ (h :: tail) if boxes.compareAndSet(list, tail) =>
       h.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
     case (h :: tail) => next[K] // Try again
-    case Nil => sys.error("Exhausted the boxed classes")
+    case Nil => sys.error(
+      """|Scalding's ordered serialization logic exhausted the finite supply of boxed classes.
+           |
+           |Resolution: Report this error to the Scalding maintainers
```
ianoc: Its not really something to report to the scalding maintainers

ianoc:
Boxed gives us a single set of types that can be registered to Cascading/Scalding to support more efficient serialization/comparasions. Effectively its for around the I/O boundary where maintaining type safety isn't really doable.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
Gabriel439: What about in the case where they actually need 251+ `Boxed` types and need a code change here?  Or is there some other mechanism that guarantees that will never happen

ianoc: We can't add an infinite amount here, the approach here can only scale so far. Its sort of a your out of luck with ordered serialization sorry sort of situation -- this occurs in some internal tests @ twitter where lots and lots of JobTest's run in one JVM. In the real world i've never seen an exhausting of these boxed classes

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -543,268 +543,540 @@ case class BoxedOrderedSerialization[K](box: K => Boxed[K],
...
+      (f243[K](_), classOf[Boxed243[K]]),
+      (f244[K](_), classOf[Boxed244[K]]),
+      (f245[K](_), classOf[Boxed245[K]]),
+      (f246[K](_), classOf[Boxed246[K]]),
+      (f247[K](_), classOf[Boxed247[K]]),
+      (f248[K](_), classOf[Boxed248[K]]),
+      (f249[K](_), classOf[Boxed249[K]]),
+      (f250[K](_), classOf[Boxed250[K]]))
+
+  private[this] def boxes[K]: AtomicReference[List[(K => Boxed[K], Class[_ <: Boxed[K]])]] =
```
johnynek: you are changing a val to a def. This breaks it, right?

johnynek:
I think this does not improve type safety (give an example of a bit of code we disallow now that we would have accepted before). Also, it breaks the code since the AtomicRef can't be a def.

Gabriel439:
Alright, then I'll close this pull request since I wasn't aware of that limitation of `AtomicReference`.  However, would you still like me to open a separate pull request for improving the error message when exhausting the supply of `Boxed` types?

johnynek:
I would love to see more type safe code, but we need to make sure we don't add bugs to get it.

In my view: type safety is a tool for correct programs, not the goal itself. Some code we accept that we can prove correct but due to limitations in the scala compiler, it cannot.

Gabriella439:
I feel that it's okay for code to be non-type-safe, but you should at least document why in the code to save time for new contributors.  Otherwise what happens is that you discourage new contributors because they're not sure what is safe to modify and what isn't without running the very time-consuming test suite.

A good example of this is that `AtomicReference` limitation, which is not obvious at all to me, and (even worse) it compiles, so that's an excellent candidate for something that warrants an explanatory comment in the source code.

Gabriella439:
One last question: does the number of `Boxed` types that can be registered have a hard limit or could the limit be lifted by a pull request to add more `Boxed` classes?  The reason I'm asking is to fix up the error message.

PR #1499, opened 2016-02-04
Add a liftToTry function to Execution

richwhitjr:


ianoc:
LGTM, merge when green

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -48,6 +48,13 @@ sealed trait Execution[+T] extends java.io.Serializable {
   import Execution.{ EvalCache, FlatMapped, GetCounters, ResetCounters, Mapped, OnComplete, RecoverWith, Zipped }
 
   /**
+   * Lift an Execution into a Try
+   * @return Execution with object wrapped in a Try
+   */
+  def liftToTry: Execution[Try[T]] =
+    map(e => Try(e)).recoverWith{ case throwable => Execution.from(Failure(throwable)) }
```
johnynek: can we use `Execution.failed` here?

johnynek: also, should be `map(e => Success(e))` no need for `Try`.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -48,6 +48,13 @@ sealed trait Execution[+T] extends java.io.Serializable {
   import Execution.{ EvalCache, FlatMapped, GetCounters, ResetCounters, Mapped, OnComplete, RecoverWith, Zipped }
 
   /**
+   * Lift an Execution into a Try
```
johnynek, line 51: can you explain that this Execution should never be failed? Only the Try might be failed.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
richwhitjr: Can't actually use Execution.failed here because it will fail the Execution itself when what we really want is the Failure object inside of a good Execution. 

Success was a good catch though and is fixed.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -48,6 +48,15 @@ sealed trait Execution[+T] extends java.io.Serializable {
...
+   *
+   * When this function is called the Execution should never be failed
+   * instead only the Try.
+   */
+  def liftToTry: Execution[Try[T]] =
+    map(e => Success(e)).recoverWith{ case throwable => Execution.from(Failure(throwable)) }
```
johnynek, line 57: @richwhitjr I made a second request (use Exception.failed). Was there a reason you didn't take that suggestion?

richwhitjr, line 57: You mentioned Execution.failed.  The issue is that doing that will fail the Execution itself.  On an exception being thrown we want a non-failure execution with a Failure Try inside of it.

johnynek, line 57: Ahh, yes. Thank you for correcting me there.

:+1:

isnotinvain:
+1

johnynek:
@richwhitjr this needs a merge resolve.

richwhitjr:
Cool, the core vs commons thing was confusing.  Kept running ./sbt "project scalding-core" test and couldn't figure out why my test didn't run.

johnynek:
merge when green

PR #1500, opened 2016-02-04
Small improvements to the `Boxed.scala` module

Gabriella439:
- Improve the error message when exhausting the supply of `Boxed{n}` types
- Add an explanatory comment for why the code does not use a more type-safe
  approach


johnynek:
:+1:

PR #1501, opened 2016-02-04
Cache boxed classes

ianoc:
This should make it far harder to exhaust boxed classes. The macros now inherit from a class providing a sane equals and hash code. They supply the class name they operate on as their ID to this parent class.  Includes test that runs through a few primitive types and a custom case class.

### DIFF
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
@@ -809,8 +809,8 @@ object Boxed {
         val untypedRes = Option(boxedCache.get(cls)) match {
           case Some(r) => r
           case None =>
-            val r = next[K]()
-            boxedCache.putIfAbsent(cls, r.asInstanceOf[(Any => Boxed[Any], Class[Boxed[Any]])])
+            val r = next[Any]()
+            boxedCache.putIfAbsent(cls, r)
             r
         }
         untypedRes.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
@@ -820,7 +820,7 @@ object Boxed {
   def next[K](): (K => Boxed[K], Class[Boxed[K]]) = boxes.get match {
     case list @ (h :: tail) if boxes.compareAndSet(list, tail) =>
       h.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
-    case (h :: tail) => next[K] // Try again
+    case (h :: tail) => next[K]() // Try again
     case Nil => sys.error("Exhausted the boxed classes")
   }
 }
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
@@ -803,25 +803,24 @@ object Boxed {
 
   private[this] val boxedCache = new java.util.concurrent.ConcurrentHashMap[AnyRef, (Any => Boxed[Any], Class[Boxed[Any]])]()
 
-  def next[K](cacheKey: Option[AnyRef]): (K => Boxed[K], Class[Boxed[K]]) =
+  private[scalding] def nextCached[K](cacheKey: Option[AnyRef]): (K => Boxed[K], Class[Boxed[K]]) =
     cacheKey match {
       case Some(cls) =>
         val untypedRes = Option(boxedCache.get(cls)) match {
           case Some(r) => r
           case None =>
-            val r = fetchNextBoxed[K]()
+            val r = next[K]()
             boxedCache.putIfAbsent(cls, r.asInstanceOf[(Any => Boxed[Any], Class[Boxed[Any]])])
             r
         }
         untypedRes.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
-      case None =>
-        fetchNextBoxed[K]()
+      case None => next[K]()
     }
 
-  private[scalding] def fetchNextBoxed[K](): (K => Boxed[K], Class[Boxed[K]]) = boxes.get match {
+  def next[K](): (K => Boxed[K], Class[Boxed[K]]) = boxes.get match {
     case list @ (h :: tail) if boxes.compareAndSet(list, tail) =>
       h.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
-    case (h :: tail) => fetchNextBoxed[K] // Try again
+    case (h :: tail) => next[K] // Try again
     case Nil => sys.error("Exhausted the boxed classes")
   }
 }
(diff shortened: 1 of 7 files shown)


scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/runtime_helpers/MacroEqualityOrderedSerialization.scala
```diff
@@ -0,0 +1,31 @@
+/*
+ Copyright 2014 Twitter, Inc.
```
johnynek: back dating it? :)

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -801,10 +801,27 @@ object Boxed {
...
+            val r = fetchNextBoxed[K]()
+            boxedCache.putIfAbsent(cls, r.asInstanceOf[(Any => Boxed[Any], Class[Boxed[Any]])])
+            r
+        }
+        untypedRes.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
+      case None =>
+        fetchNextBoxed[K]()
+    }
+
+  private[scalding] def fetchNextBoxed[K](): (K => Boxed[K], Class[Boxed[K]]) = boxes.get match {
```
johnynek: can we keep this called `next` for binary compat and have a `def nextCached(o: EquivSerialization[_]): ...` which handles the cache logic?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -34,6 +34,12 @@ trait OrderedSerialization[T] extends Ordering[T] with Serialization[T] {
   def compareBinary(a: InputStream, b: InputStream): OrderedSerialization.Result
 }
 
+/**
+ * In order to cache OrderedSerializations having equality and hashes can be useful.
+ * Extend this trait when those two properties can be satisfied
+ */
+trait EquivOrderedSerialization[T] extends OrderedSerialization[T]
```
johnynek: what about just `EquivSerialization[T]` so we could do this for non-ordered later?

ianoc:
Addressed all the comments i think

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -801,7 +801,23 @@ object Boxed {
...
-  def next[K]: (K => Boxed[K], Class[Boxed[K]]) = boxes.get match {
+  private[this] val boxedCache = new java.util.concurrent.ConcurrentHashMap[AnyRef, (Any => Boxed[Any], Class[Boxed[Any]])]()
+
+  private[scalding] def nextCached[K](cacheKey: Option[AnyRef]): (K => Boxed[K], Class[Boxed[K]]) =
+    cacheKey match {
+      case Some(cls) =>
+        val untypedRes = Option(boxedCache.get(cls)) match {
+          case Some(r) => r
+          case None =>
+            val r = next[K]()
```
johnynek: if you did the `Any` here instead of `K` you would not need the cast in putIfAbsent, right?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
```diff
@@ -55,6 +55,12 @@ trait Serialization[T] extends Equiv[T] with Hashing[T] with Serializable {
   def dynamicSize(t: T): Option[Int]
 }
 
+/**
+ * In order to cache Serializations having equality and hashes can be useful.
+ * Extend this trait when those two properties can be satisfied
+ */
+trait EquivSerialization[T] extends OrderedSerialization[T]
```
johnynek: can this extend Serialization?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -801,7 +801,23 @@ object Boxed {
...
+            r
+        }
+        untypedRes.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
+      case None => next[K]()
+    }
+
+  def next[K](): (K => Boxed[K], Class[Boxed[K]]) = boxes.get match {
     case list @ (h :: tail) if boxes.compareAndSet(list, tail) =>
       h.asInstanceOf[(K => Boxed[K], Class[Boxed[K]])]
     case (h :: tail) => next[K] // Try again
```
johnynek: do we need the `()` to call now (because we added, probably rightly since there is a side-effect here)

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
```diff
@@ -228,10 +228,12 @@ object TreeOrderedBuf {
     val lenB = freshT("lenB")
 
     t.ctx.Expr[OrderedSerialization[T]](q"""
-      new _root_.com.twitter.scalding.serialization.OrderedSerialization[$T] {
+      new _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.MacroEqualityOrderedSerialization[$T] {
         // Ensure macro hygene for Option/Some/None
         import _root_.scala.{Option, Some, None}
 
+        override val uniqueId: String = ${T.tpe.toString}
```
johnynek, line 235: is this the full type? Can you test for that in the test? I have anxiety about it not being the fully qualified name.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
ianoc: Sorry yes my typo.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
ianoc, line 235: I can't see an easy way to say it must be the FQN, it seems in manual testing to be the FQN when its not a built in type

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
ianoc: Yep, that works

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/TreeOrderedBuf.scala
ianoc, line 235: I have a test in there that for a custom type it is the FQN now

johnynek:
:+1: MWG

PR #1502, opened 2016-02-04
Fix unnecessary use of `.get` in `Globifier.scala`

Gabriella439:
This is just fixing a gratuitous use of `.isEmpty`/`.get` to use safer pattern matching.  I believe that this is not performance-critical code, so there's no need to use the `.isEmpty`/`.get` pattern here


isnotinvain:
+1

johnynek:
:+1:

PR #1505, opened 2016-02-06
Fix unnecessary uses of `Option.get`

Gabriella439:
More minor fixes identified by `wartremover`.  I'm still in search to see if there are genuine bugs that this tool will catch.  The worst thing I've found so far are invariants that are enforced purely by documentation, but I'm skipping over those for now since fixing them would require breaking binary compatibility


scalding-date/src/main/scala/com/twitter/scalding/Globifier.scala
```diff
@@ -56,27 +56,24 @@ class BaseGlobifier(dur: Duration, val sym: String, pattern: String, tz: TimeZon
     val estr = format(dr.end)
     if (dr.end < dr.start) {
       Nil
+    } else if (child.isEmpty) {
```
Gabriella439, line 59: Whoops.  This reversion is a mistake due to some `git` mistakes on my part.  Let me fix this.

PR #1506, opened 2016-02-06
Fix unnecessary uses of `Option.get`

Gabriella439:


scalding-core/src/main/scala/com/twitter/scalding/GroupBuilder.scala
```diff
@@ -379,11 +379,7 @@ class GroupBuilder(val groupFields: Fields) extends FoldOperations[GroupBuilder]
 class ScanLeftIterator[T, U](it: Iterator[T], init: U, fn: (U, T) => U) extends Iterator[U] with java.io.Serializable {
   protected var prev: Option[U] = None
   def hasNext: Boolean = { prev.isEmpty || it.hasNext }
-  def next = {
-    prev = prev.map { fn(_, it.next) }
-      .orElse(Some(init))
-    prev.get
-  }
+  def next = prev.map { fn(_, it.next) }.getOrElse(init)
```
ianoc: This i think might actually be performance critical code

johnynek: that is not the same logic. Note `prev` is a var that is updated.

scalding-core/src/main/scala/com/twitter/scalding/Tool.scala
```diff
@@ -38,17 +38,17 @@ class Tool extends Configured with HTool {
...
-    } else {
-      val jobName = args.positional(0)
-      // Remove the job name from the positional arguments:
-      val nonJobNameArgs = args + ("" -> args.positional.tail)
-      Job(jobName, nonJobNameArgs)
-    }
+  protected def getJob(args: Args): Job = rootJob match {
+    case Some(job) => job.apply(args)
+    case None =>
+      if (args.positional.isEmpty) {
```
johnynek: put these in the case:

``` scala
case None if args.positional.isEmpty =>
 ...
case None => // has at least one arg
```

scalding-core/src/main/scala/com/twitter/scalding/GroupBuilder.scala
johnynek: Iterators are always mutable.

scalding-core/src/main/scala/com/twitter/scalding/Tool.scala
johnynek: also write `job(args)` rather than `job.apply(args)`.

johnynek:
@Gabriel439 any chance we could wire wart remover into the build? Start with lax rules in CI and tighten them (or add exceptions as needed). Ideally we could also run a task with the strict options to list all the issues.

Gabriella439:
Yeah, I'm actually trying to get scalding `wartremover` clean for one of the rules as an initial step (the one that looks for `Option.get`) and if I get that working then I'll try adding it to the build.  This commit is one of the steps along the way to that.

Sorry for all the little mistakes in this pull request!  I'll fix them ASAP

Gabriella439:
Alright, this is fixed

isnotinvain:
LGTM! +1

PR #1507, opened 2016-02-08
Utility methods for running Executions in parallel

richwhitjr:
Pulling this over from the Twitter world.  I decided against pulling it the twitter promise library here and instead use the java concurrency library for the locking.  Looking at the documents the methods look thread safe. 

I am also open to a better name for the for the class.  I was going to them on the Execution object but that didn't seem correct.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
@@ -1,0 +1,92 @@
+package com.twitter.scalding
+
+import java.util.concurrent.Semaphore
+
+import com.twitter.algebird.Monoid
+import scala.util.{ Failure, Success }
+
+object ExecutionUtil {
+  /**
+   * Run a sequence of executions but only permitting parallelism amount to run at the
+   * same time.
+   *
+   * @param executions List of executions to run
+   * @param parallelism Number to run in parallel
+   * @return Execution Seq
+   */
+  def withParallelism[T](executions: Seq[Execution[T]], parallelism: Int): Execution[Seq[T]] = {
+    require(parallelism > 0, s"Parallelism must be > 0: $parallelism")
+
+    val lock = new Semaphore(parallelism)
+
+    def waitRun(e: Execution[T]): Execution[T] = {
+      Execution.from(lock.acquire())
+        .flatMap(_ => e)
+        .onComplete {
+          case Success(_) => lock.release()
+          case Failure(_) => ()
+        }
+    }
+
+    Execution.sequence(executions.map(waitRun))
+  }
+
+  /**
+   * Generate a list of executions from a date range
+   *
+   * @param duration Duration to split daterange
+   * @param parallelism How many jobs to run in parallel
+   * @param fn Function to run a execution given a date range
+   * @return Sequence of Executions per Day
+   */
+  def executionsFromDates[T](duration: Duration, parallelism: Int = 1)(fn: DateRange => Execution[T])
+  (implicit dr: DateRange): Seq[Execution[T]] =
+    dr.each(duration).map(fn).toSeq
+
+  /**
+   * Split a DateRange and allow for max parallel running of executions
+   *
+   * @param duration Duration to split daterange
+   * @param parallelism How many jobs to run in parallel
+   * @param fn Function to run a execution given a date range
+   * @return Seq of Dates split by Duration with corresponding execution result
+   */
+  def runDatesWithParallelism[T](duration: Duration, parallelism: Int = 1)(fn: DateRange => Execution[T])
+  (implicit dr: DateRange): Execution[Seq[(DateRange, T)]] = {
+
+    val dates = dr.each(duration).toSeq
+    withParallelism(dates.map(fn), parallelism).map(e => dates.zip(e))
+  }
+
+  /**
+   * Split a DateRange and allow for max parallel running of executions
+   *
+   * @param duration Duration to split daterange
+   * @param parallelism How many jobs to run in parallel
+   * @param fn Function to run a execution given a date range
+   * @return Execution of Sequences
+   */
+  def runDateRangeWithParallelism[T](duration: Duration, parallelism: Int = 1)(fn: DateRange => Execution[T])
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -0,0 +1,92 @@
...
+  def withParallelism[T](executions: Seq[Execution[T]], parallelism: Int): Execution[Seq[T]] = {
+    require(parallelism > 0, s"Parallelism must be > 0: $parallelism")
+
+    val lock = new Semaphore(parallelism)
+
+    def waitRun(e: Execution[T]): Execution[T] = {
+      Execution.from(lock.acquire())
+        .flatMap(_ => e)
+        .onComplete {
+          case Success(_) => lock.release()
```
ianoc: Should we not be releasing on success or failure?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -0,0 +1,92 @@
...
+   * @param parallelism Number to run in parallel
+   * @return Execution Seq
+   */
+  def withParallelism[T](executions: Seq[Execution[T]], parallelism: Int): Execution[Seq[T]] = {
+    require(parallelism > 0, s"Parallelism must be > 0: $parallelism")
+
+    val lock = new Semaphore(parallelism)
+
+    def waitRun(e: Execution[T]): Execution[T] = {
+      Execution.from(lock.acquire())
```
johnynek: doesn't this block? We don't want that. The twitter one provides a Future for when we can act.

We want something like async semphore using scala futures.

ianoc: That deserves a good test, the Execution.from implementation is such I'm not sure if it does, there is a lot of lazy creation of things that its possible its deferred into a future pool. -- However its still not great since it will block a thread in the future pool rather than behaving like a Promise would

isnotinvain:
Why not use scala futures?

johnynek:
@isnotinvain I guess you mean Twitter futures. Answer: the deps are already heavy adding util just to get this method is not worth it. Mo deps mo problems.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
richwhitjr: Let me see what I can come up with without the complexity and added dependencies of pulling in twitter futures etc.  I was trying to avoid that above but you are correct about the blocking.  I agree about the unit tests, hard to think of a simple fast test that can exercise the parallelism portion of the code.

richwhitjr:
Let me know what you guys think, pulled in AsyncSemaphore and made it work with scala futures.  Made it private also since I have really only tested it for this use case.

I also hopefully came up with a test that stresses the parallelism.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+    val sem = new AsyncSemaphore(parallelism)
+
+    def waitRun(e: Execution[T]): Execution[T] = {
+      Execution.fromFuture(_ => sem.acquire())
+        .flatMap(p => e.liftToTry.map((_, p)))
+        .onComplete {
+          case Success((_, p)) => p.release()
+          case Failure(_) => ()
+        }
+        .map(_._1.get)
```
johnynek: rather than `map(... get)` can we call `flatMap(... fromTry( )...)`

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+  def withParallelism[T](executions: Seq[Execution[T]], parallelism: Int): Execution[Seq[T]] = {
+    require(parallelism > 0, s"Parallelism must be > 0: $parallelism")
+    val sem = new AsyncSemaphore(parallelism)
+
+    def waitRun(e: Execution[T]): Execution[T] = {
+      Execution.fromFuture(_ => sem.acquire())
+        .flatMap(p => e.liftToTry.map((_, p)))
+        .onComplete {
+          case Success((_, p)) => p.release()
+          case Failure(_) => ()
```
johnynek: this should never happen right? because acquire returns a non-failed future. Can we do `Failure(e) => throw e // should never happen or there is a logic bug`.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Run a sequence of executions but only permitting parallelism amount to run at the
+   * same time.
+   *
+   * @param executions List of executions to run
+   * @param parallelism Number to run in parallel
+   * @return Execution Seq
+   */
+  def withParallelism[T](executions: Seq[Execution[T]], parallelism: Int): Execution[Seq[T]] = {
```
johnynek: I don't see why not to put this in `object Execution`. Seems almost as general as sequence.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -0,0 +1,122 @@
...
+   * a TypedPipe[U] and you want to independently compute
+   * the pipes on each day and union them into a
+   * single TypedPipe at the end.
+   *
+   * Another possible use case would be if the executions were created by
+   * summing intermediate monoids (e.g. T was a Map[String,HLL] since
+   * algebird supports monoids for maps and hll) and you wanted to do a
+   * final aggregation of the Monoids computed for each duration.
+   */
+  def runDateRangeWithParallelismSum[T](duration: Duration, parallelism: Int = 1)(fn: DateRange => Execution[T])(implicit dr: DateRange, mon: Monoid[T]): Execution[T] =
```
johnynek: can we use a semigroup here since there is no empty DateRange that really makes sense (error if the `Seq` is empty)

PR #1508, opened 2016-02-09
Typed Mapside Reduce

ianoc:


ianoc:
mapper ms
gc ms
cpu ms

Comparative test:

```
import com.twitter.scalding_internal.source._
import com.twitter.scalding.serialization.macros.impl.BinaryOrdering._
val tmpStorage = TypedTempSource[Long]("tmp_data")
TypedPipe.from(1 until 1000).flatMap { _ => (0 until 10000).map { _ => scala.util.Random.nextLong }}.save(tmpStorage)
val oldTP = TypedPipe.from(tmpStorage).flatMap{k => (0 until 10).map { _ => scala.util.Random.nextInt(300) -> k }}.map(_.swap).sumByLocalKeys.group.forceToReducers.sum
val newTP = TypedPipe.from(tmpStorage).flatMap{k => (0 until 10).map { _ => scala.util.Random.nextInt(300) -> k }}.map(_.swap).sumByLocalKeys2.group.forceToReducers.sum
```

ianoc:
Re-did so its off the latest develop

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -169,24 +169,108 @@ package com.twitter.scalding {
...
+
+      def iterator = backingMap.iterator
+
+      def +[B1 >: V](kv: (K, B1)) = backingMap.toMap + kv
+
+      def -(key: K) = backingMap.toMap - key
+    }
+
+    private[this] def mergeTraversableOnce[K, V: Semigroup](items: TraversableOnce[(K, V)]): Map[K, V] = {
+      val mutable = MMap[K, V]()
```
johnynek: I've heard java mutable maps are a little faster, might be worth trying that (can use `.asScala` to return the `CollectionBackedMap`. Just an idea.

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -226,10 +310,20 @@ package com.twitter.scalding {
         evictions.increment(evicted.get.size)
       evicted
     }
+
+    def putAll(kvs: Map[K, V]): Option[Map[K, V]] = {
+      val (curHits, evicted) = summingCache.putWithHits(kvs)
+      misses.increment(1 - curHits)
```
johnynek: is this logic correct? I don't think so. `1` was the number of keys before. I think you need `kvs.size` instead of `1`

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -248,6 +342,20 @@ package com.twitter.scalding {
         evictions.increment(evicted.get.size)
 
       evicted
+
+    }
+
+    def putAll(kvs: Map[K, V]): Option[Map[K, V]] = {
+      val (stats, evicted) = adaptiveCache.putWithStats(kvs)
+      misses.increment(1 - stats.hits)
```
johnynek: same comment on key size.

johnynek:
Looks pretty great!

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
ianoc: Seems scala's OpenHashMap is faster again from a quick look around

https://gist.github.com/pchiusano/1423303

PR #1509, opened 2016-02-09
Use `wartremover` to guard against careless use of `_.get`

Gabriella439:
This is the first step to adding `wartremover` support to the build.  Specifically:
- this adds the `wartremover` plugin to `project/plugins.sbt`
- this enables only one warning in `build.sbt` (the one that guards against `Option.get`)

I then went through every warning and either:
- fixed the use of `_.get` to use pattern matching or some other safe `Option` combinator, or:
- supressed the warning with an explanatory comment

There was one class of warnings that were fixable in principle but required a non-empty list type, and I'm not aware of one in the standard library.  `scalaz` does provide one but I believe `scalding` does not want to take a dependency on `scalaz`.  We could also inline the definition of a non-empty list within the project, but that also seemed overkill.

The main benefit of adding `wartremover` support is to encourage new contributions to avoid needless use of `_.get` and to document clearly the cases where it is necessary


scalding-commons/src/main/scala/com/twitter/scalding/commons/source/VersionedKeyValSource.scala
```diff
@@ -78,12 +78,12 @@ class VersionedKeyValSource[K, V](val path: String, val sourceVersion: Option[Lo
...
   def getTap(mode: TapMode) = {
     val tap = new VersionedTap(path, hdfsScheme, mode).setVersionsToKeep(versionsToKeep)
-    if (mode == TapMode.SOURCE && sourceVersion.isDefined)
-      tap.setVersion(sourceVersion.get)
-    else if (mode == TapMode.SINK && sinkVersion.isDefined)
-      tap.setVersion(sinkVersion.get)
-    else
-      tap
+    sourceVersion.orElse(sinkVersion) match {
+      case Some(version) if mode == TapMode.SOURCE =>
```
johnynek: what about SINK? Also `setVersion` is called in the old code.

scalding-commons/src/main/scala/com/twitter/scalding/commons/extensions/Checkpoint.scala
```diff
@@ -89,26 +89,25 @@ object Checkpoint {
...
-      LOG.info("Checkpoint \"" + checkpointName + "\": reading " + format +
-        " input from \"" + filename.get + "\"")
-      getSource(format, filename.get)
-        .read
-        .mapTo(List.range(0, resultFields.size) -> resultFields)((x: A) => x)(conv, setter)
-    } else {
+    filename match {
+      case Some(name) if hasInput(checkpointName, name) =>
+        // We have checkpoint input; read the file instead of executing the flow.
+        LOG.info("Checkpoint \"" + checkpointName + "\": reading " + format +
```
ianoc: can we change this to be string interpolation instead while I see the diff...

ianoc:
Think this looks good, only issue is the one @johnynek referred to I could see

scalding-commons/src/main/scala/com/twitter/scalding/commons/extensions/Checkpoint.scala
johnynek: +1

Gabriella439:
Actually, `wartremover` is causing huge issues with the test suite because the test suite has a very large number of `_.get`s spread out over many files and `wartremover` does not appear to support wildcard excludes.  Hunting down and excluding all the offending files is very time consuming, even with an automated script to scrape the test failure output and auto-generate excludes, because each level of excludes just opens up a whole new set of failing files and the test suite takes so long to run.

I opened up an issue against `wartremover` here: https://github.com/puffnfresh/wartremover/issues/219.  In the meantime, I can either:
- (A) wait until that `wartremover` issue is resolved
- (B) back out the changes to `build.sbt` and `project/plugins.sbt` or:
- (C) back out all `wartremover`-related changes (including the suppressed warnings) and just keep the fixes to `Option`-related code.

ianoc:
Can we just turn wart remover on for the main compile and not test compile
too ?

johnynek:
@Gabriel439 I think it is hitting these in macro generated code, so I guess you need to put the annotation in the macro itself, or exclude certain subclasses.

There may be other warts to remove, though (aside from Option.get, which at least causes exceptions when wrong).

Gabriella439:
@ianoc I didn't even realize that was possible.  Let me give that a try

@johnynek Right now only the warning for `Option.get` is enabled so other issues won't trigger for now.  I verified that at least `compile` works (modulo the change I need to make to fix `VersionedKeyValSource.scala` like you suggested)

Gabriella439:
So digging deeper into what `_.get`s are auto-generated, it looks like some of them (but not all) belong to `scrooge`-generated Scala code, such as in this template for writing out a field:

https://github.com/twitter/scrooge/blob/99dc4ff600283d8ef40c1b314404bc1828004a0c/scrooge-generator/src/main/resources/scalagen/writeField.scala

That's an example of a safe usage (preceded by `_.isDefined` and necessary for performance reasons) but fixing that would require upstreaming `wartremover` annotations into `scrooge`-generated code, which in turn would require `wartremover` buy-in by a much larger group of people.

I think in this case silencing `wartremover` for tests may be necessary in the short-term.

Gabriella439:
Alright, this should be ready now

scalding-core/src/main/scala/com/twitter/scalding/typed/CoGrouped.scala
```diff
@@ -333,6 +333,8 @@ abstract class CoGroupedJoiner[K](inputSize: Int,
 
   override def getIterator(jc: JoinerClosure) = {
     val iters = (0 until distinctSize).map { jc.getIterator(_).asScala.buffered }
+    // Don't pattern match in performance-critical code
```
johnynek: that's not perf critical as much as the invariant that one of the iters has an item is not communicated in the types by cascading

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -227,6 +227,8 @@ sealed trait ReduceStep[K, V1] extends KeyedPipe[K] {
         mapped
           .toPipe(Grouped.kvFields)(fd, mode, ts)
           .groupBy(fields) { inGb =>
+            // Don't pattern match in performance-critical code
```
johnynek: this it not perf critical at all (planning code). You could fix this with:

``` scala
 val withSort = sortOpt match {
  case Some(s) => inGb.sortBy(s)
  case None => inGb
}
```

johnynek: or: `val withSort = sortOpt.fold(inGb)(inGb.sortBy)`

scalding-core/src/main/scala/com/twitter/scalding/typed/CoGrouped.scala
Gabriel439: Whoops.  When I wrote that I assumed that Scala had some sort of safe way to get the head that returned an `Option` and that you weren't using it for performance reasons, but now I see that there is no such function in the API.  I'll change the wording to your version.

Gabriella439: Ignore my last comment (I was talking about the wrong thing).  I'll implement your suggestion.

scalding-core/src/main/scala/com/twitter/scalding/typed/ValuePipe.scala
```diff
@@ -53,6 +53,8 @@ sealed trait ValuePipe[+T] extends java.io.Serializable {
    * The name here follows the convention of adding
    * Execution to the name so in the repl in is removed
    */
+  // Suppress warning on `_.get` since the behavior is intentionally part of the API
```
johnynek: could use `flatMap` here:

``` scala
def getExecution: Execution[T] = toOptionExecution.flatMap {
  case Some(t) => Execution.from(t)
  // same exception as scala.None.get
  // https://github.com/scala/scala/blob/2.12.x/src/library/scala/Option.scala#L347
  case None => Execution.failed(new java.util.NoSuchElementException("None.get"))
}
```

johnynek:
:+1: Merge when green

PR #1511, opened 2016-02-09
Add in an API around cache isolation

### LINKED ISSUE #1459: Scalding Execution API memory leak / OOM

ianoc:
Fixes #1459


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -338,24 +344,43 @@ object Execution {
...
+      val mutatedConfig = fn(conf)
+      cache.getOrElseInsert(mutatedConfig, this, prev.runStats(mutatedConfig, mode, cache))
+    }
+  }
+
+  private case class WithNewCache[T](prev: Execution[T]) extends Execution[T] {
+    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      val ec = new EvalCache
+      val ret = prev.runStats(conf, mode, ec)
+      // When the final future in complete we stop the submit thread
```
johnynek: I forgot about this. This will create two concurrent cascading threads, which is not safe. We need to share that cascading thread between the two, I think.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -209,6 +209,12 @@ object Execution {
     override def join[T, U](t: Execution[T], u: Execution[U]): Execution[(T, U)] = t.zip(u)
   }
 
+  def withConfig[T](ex: Execution[T])(c: Config => Config): Execution[T] =
+    TransformedConfig(ex, c)
+
+  def withNewCache[T](ex: Execution[T]): Execution[T] =
```
johnynek, line 261: can we add some documentation on the purpose of this?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -209,6 +209,12 @@ object Execution {
     override def join[T, U](t: Execution[T], u: Execution[U]): Execution[(T, U)] = t.zip(u)
   }
 
+  def withConfig[T](ex: Execution[T])(c: Config => Config): Execution[T] =
```
johnynek, line 247: can we add docs on the semantics of this and expected use cases?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ianoc: Good catch... will add something to share

johnynek:
looks good to me. Want to get the ToWrite caching in a new PR?

ianoc:
Have it here, so will just add

rowandavies:
I'll try this with some of our code that previously ran into an oom, just to be sure we don't end up with any references from the outer `EvalCache`.

johnynek:
@rowandavies Let us know if this does not work for you. I'll leave it open for a bit to give you time to test.

ianoc:
@rowandavies Any luck getting a chance to test this to see if it resolves your issues?

@johnynek If we don't merge this soon we should rip out the EvalCache changes into their own PR I guess so they don't get lost. (the ToWrite cache fix)

rowandavies:
@ianoc @johnynek It works.  Confirming that our test that previously hit an oom runs fine with the scalding version in this PR, allocating 64GB with a 1.75GB heap.

johnynek:
Excellent! @ianoc can you fix the merge conflicts?

ianoc:
Should all be good to go now i believe

PR #1512, opened 2016-02-11
Add implicit Ordering[RichDate]

johnynek:
see scala design issue: https://issues.scala-lang.org/browse/SI-8541
that causes the following for tuples with RichDate:

```
/Users/oscar/oss/scalding/scalding-date/src/test/scala/com/twitter/scalding/DateTest.scala:139:
diverging implicit expansion for type ((String, com.twitter.scalding.RichDate)) =>
Comparable[(String, com.twitter.scalding.RichDate)]
[error] starting with method $conforms in object Predef
[error]       implicitly[Ordering[(String, RichDate)]]
[error]                 ^
[error] one error found
[error] (scalding-date/test:compileIncremental) Compilation failed
[error] Total time: 5 s, completed Feb 10, 2016 2:27:14 PM
```

### DIFF
--- a/scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala
+++ b/scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala
@@ -61,6 +61,10 @@ object RichDate {
   }
 
   def now: RichDate = RichDate(System.currentTimeMillis())
+
+  implicit def richDateOrdering: Ordering[RichDate] = new Ordering[RichDate] {
+    def compare(a: RichDate, b: RichDate) = java.lang.Long.compare(a.timestamp, b.timestamp)
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Gabriella439:
+1

scalding-date/src/main/scala/com/twitter/scalding/RichDate.scala
```diff
@@ -61,6 +61,10 @@ object RichDate {
   }
 
   def now: RichDate = RichDate(System.currentTimeMillis())
+
+  implicit def richDateOrdering: Ordering[RichDate] = new Ordering[RichDate] {
+    def compare(a: RichDate, b: RichDate) = java.lang.Long.compare(a.timestamp, b.timestamp)
```
ianoc, line 66: Should we call the compare function inside the richdate's instead? just so as we avoid any potential future drift.

(Curious I didn't think defining both Ordered and Ordering would get around the bug totally, but good!)

PR #1513, opened 2016-02-11
Fix MultipleTextLineFiles source in JobTest

danosipov:
Using MultipleTextLineFiles source in JobTest fails to locate the source. This PR should fix it.

The underlying problem is highlighted by this comment:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/JobTest.scala#L63

If you don't think this approach is good, and would rather fix the equality issue, I'd need a pointer as to how a String key can be generated.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -421,7 +421,7 @@ abstract class FixedPathSource(path: String*) extends FileSource {
 
   // `toString` is used by equals in JobTest, which causes
   // problems due to unstable collection type of `path`
-  override def toString = getClass.getName + path.mkString(",")
+  override def toString = getClass.getName + path.mkString("(", ",", ")")
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -419,6 +419,7 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   def localPaths = path.toList
   def hdfsPaths = path.toList
 
+  // The collection type of `path` causes equality issues in JobTest
   override def toString = getClass.getName + path.mkString(",")
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -419,7 +419,8 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   def localPaths = path.toList
   def hdfsPaths = path.toList
 
-  // The collection type of `path` causes equality issues in JobTest
+  // `toString` is used by equals in JobTest, which causes
+  // problems due to unstable collection type of `path`
   override def toString = getClass.getName + path.mkString(",")
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -419,7 +419,7 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   def localPaths = path.toList
   def hdfsPaths = path.toList
 
-  override def toString = getClass.getName + path
+  override def toString = getClass.getName + path.mkString(",")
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -419,7 +419,7 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   def localPaths = path.toList
   def hdfsPaths = path.toList
 
-  override def toString = getClass.getName + path
+  override def toString = getClass.getName + path.mkString(",")
```
ianoc: The comment mentions that equality could be buggy (I guess jvm classes and what have you). I don't see why this would fix that?  -- in the sense, I just don't see how this fixes the bug?

danosipov: @ianoc Before this change, the toString would generate MultipleTextLineFilesListinputname for "inputname" path. It doesn't match when JobTest is given a source, which outputs MultipleTextLineFilesWrappedArrayinputname

Gabriella439: I recommend adding what you just said as a comment since it's non-obvious from just reading the code why `mkString` is being used here

rubanm: The test you added passes for me on develop. Is there a better one that currently fails on develop and this branch fixes it? That might help clarify what this fix does (I don't quite understand it either).

danosipov: @rubanm I'll double check it - maybe if it only has one input source it fails
@Gabriel439 good call, I'll add that info. Not sure why WrappedArray is used though.

danosipov:
@rubanm I changed the test to better represent the job that caused me to encounter the issue. It now fails on develop.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
jnievelt: It's a Scala thing:

```
scala> def foo(path: String*) = path.toString
foo: (path: String*)String
scala> foo("a", "b", "c")
res5: String = WrappedArray(a, b, c)
```

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -419,7 +419,8 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   def localPaths = path.toList
   def hdfsPaths = path.toList
 
-  override def toString = getClass.getName + path
+  // The collection type of `path` causes equality issues in JobTest
+  override def toString = getClass.getName + path.mkString(",")
```
jnievelt: Does the order of paths on this source matter?  If not, should we also sort this before concatenation?

danosipov: I don't think so - since since both List & WrappedArray preserve order - but I may be mistaken.

danosipov:
CI failure appears to be unrelated

johnynek:
:+1: thanks! merge when green

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -419,7 +419,8 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   def localPaths = path.toList
   def hdfsPaths = path.toList
 
-  override def toString = getClass.getName + path
+  // The collection type of `path` causes equality issues in JobTest
```
ianoc: I hadn't noticed what was doing this before, its really that the equals relies on toString (usually toString really is just for human's and doesn't matter.). The change looks good to me, but you change this to mention that equals is relying on toString and path.toString isn't stable ordinarily without this.

I suspect the underlying problem here is that the type of the var arg's string is probably something like a Seq[String], and the 2 means of creation in your test example here provide underlying different classes. toString uses toString on the elements which winds up calling getClass, so is printing out the actual implementations passed.

danosipov: @ianoc That's my understanding as well, and yes, I've confirmed that toString is the primary means of [doing equality check](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala#L424) (hence the [comment here](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/JobTest.scala#L63)). 

So, you'd like me to append that to the comment?

ianoc: Yes, I think we should include the note that the equals method calls this. The comment reads strange otherwise, since ordinarily toString won't be used for equality

danosipov:
@ianoc LMK if this is better

danosipov:
@johnynek done

johnynek:
shipit! (when green)

PR #1514, opened 2016-02-11
Move parquet cascading schemes to subprojects

rubanm:
This moves all the parquet schemes to separate sub-projects than the parquet sources. This is mainly for easier upgrade to cascading3 and future versions perhaps. (As things stand now, we need to upgrade scalding-core in our cascading3 branch, before scalding-parquet for example.)


build.sbt
```diff
@@ -394,17 +395,25 @@ lazy val scaldingParquet = module("parquet").settings(
       exclude("com.twitter.elephantbird", "elephant-bird-pig")
       exclude("com.twitter.elephantbird", "elephant-bird-core"),
     "org.apache.thrift" % "libthrift" % "0.7.0",
```
johnynek: can we move this raw version string with the rest at the top of this file?

build.sbt
```diff
@@ -443,11 +451,25 @@ lazy val scaldingParquetScrooge = module("parquet-scrooge")
         exclude("com.twitter.elephantbird", "elephant-bird-core"),
        "com.twitter" %% "scrooge-serializer" % scroogeVersion,
       "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
-      "com.novocode" % "junit-interface" % "0.11" % "test",
       "junit" % "junit" % junitVersion % "test"
+    )
+).dependsOn(scaldingParquetCascading % "compile->compile;test->test", scaldingParquetScroogeFixtures % "test->test")
 
+lazy val scaldingParquetScrooge = module("parquet-scrooge")
```
johnynek, line 461: shouldn't we merge this with `scaldingParquetScroogeCascading`? Same dependencies, same use case, right?

rubanm, line 461: I separated these so that the cascading schemes go in a different sub-project. parquet-scrooge, after this change would contain only the ParquetScrooge trait and related time pathed sources (which need scalding-core). If this is too thin to be its own sub-project, then I can merge back the two.

build.sbt
```diff
@@ -384,8 +385,8 @@ lazy val scaldingParquetFixtures = module("parquet-fixtures")
      )
    )
 
-lazy val scaldingParquet = module("parquet").settings(
-  libraryDependencies <++= (scalaVersion) { scalaVersion => Seq(
+lazy val scaldingParquetCascading = module("parquet-cascading").settings(
```
johnynek, line 389: ok, so the point here is that this target only depends on cascading + parquet, not scalding. Can we add a comment so that is not overlooked as to the purpose?

build.sbt
```diff
@@ -432,10 +441,9 @@ lazy val scaldingParquetScroogeFixtures = module("parquet-scrooge-fixtures")
   )
 )
 
-lazy val scaldingParquetScrooge = module("parquet-scrooge")
+lazy val scaldingParquetScroogeCascading = module("parquet-scrooge-cascading")
```
johnynek, line 446: again, this only depends on cascading, parquet and scrooge, right? Comment?

build.sbt
rubanm: Updated. The global thriftVersion is already present and set to 0.5.0, which I think should be okay.

johnynek:
:+1:

ianoc:
This seems useful for the cascading 3 upgrade, and makes sense there. Do we want it in the 0.16 release cycle ? Or keep it closer to the big painful cascading one? 

@rubanm would this be hard to take as an update at twitter? (just as an exemplar of issues elsewhere possibly).

ianoc:
With the aliases and deps in place it should be handled mostly by ivy resolves I think? 

Can we have the new types/classes be package private to scalding? They seem like an internal implementation detail as far as scalding is concerned, would be nice to not have them in our ABI if we could? thoughts @johnynek ?

rubanm:
Re cascading3, I could cherry pick the sbt update commit (moves from Build.scala to build.sbt) to cascading3 branch, and apply this on top of it. If we don't merge this to develop, it will be one more thing to look at in the diff in the eventual cascading3 -> develop merge.

Yes given the pre-existing sub-projects will still pull in classes in the corresponding new *-cascading ones, the upgrade should be fine I think. I'm in favor of merging to develop (and cherry picking to cascading3 branch) if this seems reasonable as a general thing to do.

rubanm:
Some of Schemes are consumed at Twitter in multiformat sources, for example. The rest could be made package private I think.

johnynek:
@rubanm we will need help from you and @isnotinvain to make sure we keep twitter up to date and not let any forks happen!

So, if you think you can get the merge in at Twitter, I'm fine with a merge for 0.16.

ianoc:
Yep, good way to put it. Merge when green from me if the hassle level is ok
for you guys to handle to merge.

On Friday, February 12, 2016, P. Oscar Boykin notifications@github.com
wrote:

> @rubanm https://github.com/rubanm we will need help from you and
> @isnotinvain https://github.com/isnotinvain to make sure we keep
> twitter up to date and not let any forks happen!
> 
> So, if you think you can get the merge in at Twitter, I'm fine with a
> merge for 0.16.
> 
> —
> https://github.com/twitter/scalding/pull/1514#issuecomment-183558818.

## 

Sent from Gmail Mobile

rubanm:
@ianoc @johnynek Makes sense. I'll wait for a green sandbox internally before merging. (So we block this on the new algebird version to release internally at Twitter first, with the related breaking tests fixed.) I'll just cherry pick this to the cascading3 branch meanwhile to keep that going.

rubanm:
Sorry, closed by mistake.

johnynek:
should we merge this now? I kind of lost track of it.

PR #1520, opened 2016-02-19
Update FixedPathSource to strip out '*' in paths ending with '/*' for writes

piyushnarang:
Currently the paths we seem to pass in for reads end up terminating in a "/*". When we end up using the same for our write paths, data is written out under a subdirectory (for example: testdata/*/part-0000.lzo instead of testdata/part-0000.lzo). 

Updated the FixedPathSource to behave similarly to the TimePathedSource and trim the trailing '*' in these cases.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -422,17 +422,20 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   // `toString` is used by equals in JobTest, which causes
   // problems due to unstable collection type of `path`
   override def toString = getClass.getName + path.mkString("(", ",", ")")
-  override def hdfsWritePath = stripTrailingStar(super.hdfsWritePath)
+  override def hdfsWritePath = stripTrailing(super.hdfsWritePath)
 
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
 
   /**
-    * Strip trailing '*' from the path string.
+    * Similar in behavior to {@link TimePathedSource.writePathFor}.
+    * Strip out the trailing slash star.
     */
-  protected def stripTrailingStar(path: String): String = {
+  protected def stripTrailing(path: String): String = {
+    assert(path != "*", "Path must not be *")
+    assert(path != "/*", "Path must not be /*")
     if(path.takeRight(2) == "/*") {
-      path.dropRight(1)
+      path.dropRight(2)
     } else {
       path
     }
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -422,8 +422,21 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   // `toString` is used by equals in JobTest, which causes
   // problems due to unstable collection type of `path`
   override def toString = getClass.getName + path.mkString("(", ",", ")")
+  override def hdfsWritePath = stripTrailingStar(super.hdfsWritePath)
+
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
+
+  /**
+    * Strip trailing '*' from the path string.
+    */
+  protected def stripTrailingStar(path: String): String = {
+    if(path.takeRight(2) == "/*") {
+      path.dropRight(1)
+    } else {
+      path
+    }
+  }
 }
 
 /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
@@ -56,6 +56,7 @@ object TimePathedSource {
    * Gives the write path based on daterange end.
    */
   def writePathFor(pattern: String, dateRange: DateRange, tz: TimeZone): String = {
+    assert(pattern != "/*", "Pattern must not be /*")
     assert(pattern.takeRight(2) == "/*", "Pattern must end with /* " + pattern)
     val stripped = pattern.dropRight(2)
     toPath(stripped, dateRange.end, tz)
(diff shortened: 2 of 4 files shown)


danosipov:
I haven't checked, but I believe this change would break reading from S3

ulyssepence:
> for example: testdata//part-0000.lzo instead of testdata/part-0000.lzo

So these are not equivalent under HDFS?

Also, do we have a test case that fails under the old code, but succeeds with the change?

rubanm:
@benpence I think github markup is hiding the \* between // there.

Gabriella439:
I fixed his original comment

ulyssepence:
@Gabriel439 Gotcha

piyushnarang:
@Gabriel439 thanks for updating the comment :-). Missed that when I put it up

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -419,9 +419,22 @@ abstract class FixedPathSource(path: String*) extends FileSource {
...
+
   override def toString = getClass.getName + path
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
+
+  /**
+    * Strip trailing '*' from the path string.
+    */
+  protected def stripTrailingStar(path: String): String = {
+    if(path.takeRight(2) == "/*") {
```
Gabriella439, line 437: What about the case where the path is just "*" (i.e. a relative path pointing to all files within the current directory)?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -419,9 +419,22 @@ abstract class FixedPathSource(path: String*) extends FileSource {
...
 
+  override def hdfsWritePath = stripTrailingStar(super.hdfsWritePath)
+
   override def toString = getClass.getName + path
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
+
+  /**
+    * Strip trailing '*' from the path string.
+    */
```
johnynek, line 433: can you add a link to the `TimePathedSource` method `writePathFor`?

It seems like that code is dropping the trailing `/`. Should we not do the same thing here? I think that was done to accomodate local mode which writes to one file. Hadoop seems to create a directory in any case.

Also, can we make that code use the same style `.dropRight(2)` in that case?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
johnynek, line 437: What should we do there? It is a weird path. I'm not sure what the intent would be. Maybe error in that case?

piyushnarang, line 437: Yeah I was wondering the same as well. Getting rid of the `*` would seems like it wouldn't work. We could potentially replace the `*` with `.` or error out like @johnynek suggested.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
piyushnarang, line 433: @johnynek I guess you're referring to adding a link to that method in the source? If yes, I can make that change. 

I can make this code consistent with TimePathedSource and also drop the `/`.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
johnynek, line 437: I'd say error.

piyushnarang:
@benpence - Added a test that would fail without these changes - it checks the hdfsWritePath
@Gabriel439 - updated to handle the `*`
@johnynek - Updated to also strip out the `/` and your other comments

piyushnarang:
@johnynek - Updated based on your comments. Added the assertion (and corresponding unit tests) + reworked the other test.

johnynek:
:+1: I'll let another commenter merge to make sure they get a chance to check.

Thanks a lot for your help!

piyushnarang:
@johnynek sounds good. Thanks for taking such a detailed look :-)

piyushnarang:
@benpence / @Gabriel439 - can you guys take a look at the rework and let me know what you think?

scalding-core/src/main/scala/com/twitter/scalding/TimePathedSource.scala
```diff
@@ -56,9 +56,9 @@ object TimePathedSource {
    * Gives the write path based on daterange end.
    */
   def writePathFor(pattern: String, dateRange: DateRange, tz: TimeZone): String = {
+    assert(pattern != "/*", "Pattern must not be /*")
     assert(pattern.takeRight(2) == "/*", "Pattern must end with /* " + pattern)
-    val lastSlashPos = pattern.lastIndexOf('/')
-    val stripped = pattern.slice(0, lastSlashPos)
+    val stripped = pattern.dropRight(2)
```
Gabriella439, line 61: Could you use `stripTrailing` here to avoid code duplication?

Gabriella439, line 61: Never mind.  On closer reading this behavior is not the same

Gabriella439:
Looks good to me!

ulyssepence:
LGTM

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -422,8 +422,24 @@ abstract class FixedPathSource(path: String*) extends FileSource {
...
+
   override def hashCode = toString.hashCode
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
+
+  /**
+    * Similar in behavior to {@link TimePathedSource.writePathFor}.
+    * Strip out the trailing slash star.
+    */
+  protected def stripTrailing(path: String): String = {
+    assert(path != "*", "Path must not be *")
```
isnotinvain, line 435: should these asserts be requires instead?

piyushnarang, line 435: Yeah that might be a good idea as a followup. Ended up choosing assert as the existing code in FileSource & TimePathedSource was already using it and I thought it might be nice to keep it consistent.

Issue #1465, opened 2016-01-04
Cascading 3 upgrade strategy


ianoc:
So I believe cascading 3 change won't be backwards compatible with 2.6. Though we obviously don't want to have too long running a PR. Thoughts?

1) I think we should probably have a cascading3 branch for now, and do several PR's into it. Its not ideal since we will need to be careful about incompatible changes to develop. But hopefully it wouldn't live long. I'm not sure how well it would work, but one more code involved but maybe easier stepping stone approach would be to stick an API/set of files to act as a bridge layer and have one for cascading 2/3. Outside of commons most of the changes look pretty isolated. (I'm in favor of this if it works, since I think we could get onto develop sooner). 

2) Should we cut the dependencies of this down by eliminating some dependencies? The most obvious here is dfs-datastores for this. 

3) What should our merge to develop criteria be?
(1) All unit/integration tests pass
(2) Some usage at Twitter/Stripe/Other ? 
(3) Should Tez working be a requirement? Or just having cascading 3 for hadoop being good enough to make it develop.
(4) Versioning/ANN for scalding. Just next incompatible release sounds probably ok as per our current numbering. When is reasonable for us to aim for this to be in? 

//cc @johnynek @cchepelov @rubanm

sriramkrishnan:
I am OK with Tez being a non-requirement if it makes it easier to make it to develop. Tez would be great to have of course, but it could wait till after that?

johnynek:
+1 to a short lived branch.

+1 to cascading 3 working on Hadoop for merge to develop/master. In my view, Tez is still somewhat experimental, so I wouldn't gate on that. And moving forward to the latest cascading is still progress.

I would do the merge to develop when Twitter's CI is green and integration tests pass on a snapshot published from the cascading3 branch. I'll try to coordinate similar at stripe, but I'm just getting ramped up here.

/cc @avibryant

cchepelov:
(agreeing with 99% of what's being said here, especially not gating on Tez or Flink, since willing folks can now switch the fabric at runtime anyway)

re. backwards compatibility: indeed, it really can't be 100% backwards compatible. It seems most user code won't be affected at the *source* level (few Cascading concepts spilled above scalding, and fewer of those underwent interface chances), but there's always someone somewhere doing something. Safer to deny compatibility☺and binaries will break anyway.

cchepelov:
Question with respect to testing the various fabrics:
[...]

PR #1521, opened 2016-02-19
[cascading3] Migrate core, commons and related

rubanm:
part of https://github.com/twitter/scalding/issues/1465
based on Cyrille's work in https://github.com/twitter/scalding/pull/1446

Most of the interesting changes are in:
- Operations.scala -- to handle both old and new cascading aggregate by thresholds
- PlatformTest.scala -- some updated tests, hashjoining and then merging the result with one side of the same join is no longer supported in cascading3

Cascading fabric selection changes will be sent in a separate PR.


johnynek:
@cchepelov take a look?

build.sbt
```diff
@@ -340,7 +336,7 @@ lazy val scaldingCommons = module("commons").settings(
     "com.twitter" %% "bijection-core" % bijectionVersion,
     "com.twitter" %% "algebird-core" % algebirdVersion,
     "com.twitter" %% "chill" % chillVersion,
-    "com.twitter.elephantbird" % "elephant-bird-cascading2" % elephantbirdVersion,
+    "com.twitter.elephantbird" % "elephant-bird-cascading3" % elephantbirdVersion,
```
johnynek: can we move this to where the versions are? `elephant-bird-artifact` so we can keep all these switches in one place?

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -134,15 +134,23 @@ package com.twitter.scalding {
...
+    val CASCADING3_SIZE_CONFIG_KEY = AggregateByProps.AGGREGATE_BY_CAPACITY
 
     def cacheSize(fp: FlowProcess[_]): Int =
       cacheSize.orElse {
-        Option(fp.getStringProperty(SIZE_CONFIG_KEY))
-          .filterNot { _.isEmpty }
-          .map { _.toInt }
-      }
-        .getOrElse(DEFAULT_CACHE_SIZE)
+        val cascading2Property = Option(fp.getStringProperty(CASCADING2_SIZE_CONFIG_KEY)).filterNot(_.isEmpty).map(_.toInt)
```
johnynek:
``` scala
def getInt(k: String): Option[Int] = Option(fp.getStringProperty(k)).filterNot(_.isEmpty).map(_.toInt)
val cascading2Property = getInt(CASCADING2_SIZE_CONFIG_KEY)
val cascading3Property = getInt(CASCADING3_SIZE_CONFIG_KEY)
```

scalding-core/src/main/scala/com/twitter/scalding/Operations.scala
```diff
@@ -134,15 +134,23 @@ package com.twitter.scalding {
...
+        val cascading3Property = Option(fp.getStringProperty(CASCADING3_SIZE_CONFIG_KEY)).filterNot(_.isEmpty).map(_.toInt)
+        // we support both old and new properties for backward compatibility
+        // and pick the max of the two, when both exist
+        (cascading2Property, cascading3Property) match {
+          case (Some(a), Some(b)) if a >= b => Some(a)
+          case (Some(a), Some(b)) if b > a => Some(b)
+          case (None, None) => None
+          case (Some(a), _) => Some(a)
+          case (_, Some(b)) => Some(b)
+        }
```
johnynek, line 152:
``` scala
case (Some(a), Some(b)) => Some(Ordering[Int].max(a, b))
```

rather than the two cases.

scalding-core/src/main/scala/com/twitter/scalding/TemplateSource.scala
```diff
@@ -1,121 +0,0 @@
...
-import cascading.tap.local.FileTap
-import cascading.tap.local.{ TemplateTap => LTemplateTap }
-import cascading.tap.SinkMode
-import cascading.tap.Tap
-import cascading.tuple.Fields
-
-/**
- * This is a base class for template based output sources
- */
-abstract class TemplateSource extends SchemedSource with HfsTapProvider {
```
johnynek, line 29: is this used at Twitter?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -46,13 +49,43 @@ object TinyJoinAndMergeJob {
...
+class TinyJoinAndMergeUnsupportedJob(args: Args) extends Job(args) {
+  import TinyJoinAndMergeJob._
+
+  val input1 = joinInput1.read.mapTo(0 -> 'id) { v: Int => v }
+
+  val joined = joinInput2.read
+    .mapTo(0 -> 'id) { v: Int => v }
+    .joinWithTiny('id -> 'id, input1)
+
+  // merging the output of a hashjoin with one of its inputs is
```
johnynek, line 72: this is probably going to break some jobs. What is the work around? Would a `.forceToDisk` fix it? Would you need `.forceToDiskExecution` to put it in a new flow? That is a bummer, because it is obviously plannable. /cc @cwensel

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -46,13 +49,43 @@ object TinyJoinAndMergeJob {
...
+  import TinyJoinAndMergeJob._
+
+  val input1 = joinInput1.read.mapTo(0 -> 'id) { v: Int => v }
 
-  val messages = messageInput.read
+  val joined = joinInput2.read
     .mapTo(0 -> 'id) { v: Int => v }
-    .joinWithTiny('id -> 'id, people)
+    .joinWithTiny('id -> 'id, input1)
+    .forceToDisk // workaround for cascading3
```
johnynek, line 85: ahh, here is the workaround. we probably need to add this to the comments on `joinWithTiny` and `hashJoin`

johnynek, line 85: I wonder if we could automatically put this in the graph inside the merge? I hate for syntactically legal things to be errors. Compiles should me "works" as much as possible.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
```diff
@@ -140,14 +140,14 @@ class TypedParquetTupleScheme[T](val readSupport: ParquetReadSupport[T], val wri
   type SourceCallType = SourceCall[Array[AnyRef], Reader]
   type SinkCallType = SinkCall[Array[AnyRef], Output]
 
-  override def sourceConfInit(flowProcess: FlowProcess[JobConf], tap: TapType, jobConf: JobConf): Unit = {
+  override def sourceConfInit(flowProcess: FlowProcess[_ <: JobConf], tap: TapType, jobConf: JobConf): Unit = {
```
johnynek, line 143: I see both `_ <: Configuration` and `_ <: JobConf` what is the difference?

johnynek:
Does this pass e2e tests or CI at Twitter?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
rubanm, line 85: For the fields api, we could probably encode this in the Pipe name and automatically add a forceToDisk. In the typed api, we could just encode it in a new typeclass? I'm wondering how frequent such operations are and if it's worth adding code to handle this vs adding a comment as a warning. Handling in the graph would be more ideal, so I'll take a stab at this anyway.

johnynek, line 85: The `HashJoin` is a subclass of `Pipe`. I think we could walk up the nodes in a merge, see if there is this case, and add a `Checkpoint` node.

Pretty much (or exactly) the same code could be done in the Typed case. The only challenge is inserting an exta node in the graph. Summingbird has some code for this we could use, or we could possibly use something in cascading3 which I understand makes it easier to add rules.

@cwensel

rubanm, line 85: Looks like there is no way to access the underlying Pipes in a Merge right now. The closest is getPipePos which returns (name, position) pairs:
https://github.com/cwensel/cascading/blob/wip-3.1/cascading-core/src/main/java/cascading/pipe/Splice.java#L944

I've sent an email to cascading-user list to look into adding a getPipes method. We may have to rely on the name otherwise.

cwensel:
If we can get an isolated Cascading3 test case we can take a stab at promoting this from 'no longer supported' to 'bug' and then to 'resolved'.

cchepelov:
Hi @posco @rubanm 
Great to see a lot of progress! Will have to come back to this next week (away from keyboard this week).

Re. the spurious ".forceToDisk"; indeed, the code should do the right thing without. The transform facility @cwensel wrote about looks like the correct place to put the necessary Boundaries in place.

  -- Cyrille 

Le 19 févr. 2016 19:26, à 19:26, "P. Oscar Boykin" notifications@github.com a écrit:

johnynek:
@cwensel about the repro: It should be as easy as a cascading `HashJoin` followed by `Merge` followed by `GroupBy`. Sorry kind of swamped...

rubanm:
@johnynek  This branch now passes e2e tests at Twitter (with a related EB change https://github.com/twitter/elephant-bird/pull/465). I'm working on piloting some user jobs.

scalding-core/src/main/scala/com/twitter/scalding/TemplateSource.scala
rubanm, line 29: There was one instance of TemplateSource being used, which I've migrated to PartitionedSource.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
rubanm, line 85: I'll post a follow-up PR against cascading3 branch for this. I think @piyushnarang's hashjoin + forceToDisk change already has most of the code to deal with this.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/tuple/scheme/TypedParquetTupleScheme.scala
rubanm, line 143: JobConf is the old hadoop1 api and Configuration is hadoop2. In cascading3, Hfs taps and descendents were moved to the hadoop2 api. Cascading's MultiInputFormat still looks at the hadoop1 key for the input formats though (see https://github.com/twitter/elephant-bird/pull/465). So broadly, the changes involved moving anything that's closely tied to the Hfs tap to use Configuration, leaving the remaining to use JobConf, and still using the Deprecated wrappers for the EB input and output formats.

sriramkrishnan:
@rubanm this is pretty amazing work!

johnynek:
Amazing!

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
johnynek, line 85: yeah, I think so too, but probably not for the fields API. Ideally we could use something like that here so we don't break any super old jobs.

johnynek:
looks good to me to merge into cascading3 branch.

does this have all the changes from current develop branch?

rubanm:
@johnynek Thanks for the review! RC6 is currently being released to twitter source. I plan to merge develop once that release is done so it's in tandem, with the joinWithTiny fix to follow.

johnynek:
@rubanm sounds good. Way to push through on this!

cchepelov:
Wow, @rubanm, this is great, many thanks for doing the hard work!

With apologies for the lack of testing on my side, been litterally 
swamped, and no real end in sight… but using cascading3+tez (yes, my old 
build) on a daily basis. Really looking forward to adopting your QA'd 
branch.

PR #1523, opened 2016-02-23
ExecutionUtil API cleanup and Completion Hook

richwhitjr:
Cleaned up the ExecutionUtil code a little so there is less duplication and the base class for splitting a date range into executions now zips in the dates.

I also added a function to help in cases where you want to run an Execution job that runs steps in parallel.  For example lets say you have a list of arguments say and convert each to an execution and lift.  This will prevent one execution failure from killing the rest.  The added function here will make it easy to log which executions failed at the end and why.


scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -2,17 +2,27 @@ package com.twitter.scalding
 
 import com.twitter.algebird.Semigroup
 
+import scala.util.{Failure, Success, Try}
+
 object ExecutionUtil {
   /**
+   * Useful for when using parallelism and want to output the final state of the job
+   */
+  def liftedSequenceUnit[T](executions: Execution[Seq[Try[T]]], handleComplete: Try[T] => Unit): Execution[Unit] =
```
johnynek: this API seems really easy to misuse. Not sure it should be public in scalding-core.
1. assumes executions can't fail.
2. uses side effects, and applies the function an unknown number of times (once per item).
3. replaces a somewhat uncommon pattern with only 3 lines fewer of code.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
```diff
@@ -2,17 +2,27 @@ package com.twitter.scalding
...
    * Generate a list of executions from a date range
    *
    * @param duration Duration to split daterange
-   * @param parallelism How many jobs to run in parallel
    * @param fn Function to run a execution given a date range
    * @return Sequence of Executions per Day
    */
-  def executionsFromDates[T](duration: Duration, parallelism: Int = 1)(fn: DateRange => Execution[T])(implicit dr: DateRange): Seq[Execution[T]] =
-    dr.each(duration).map(fn).toSeq
+  def executionsFromDates[T](duration: Duration)(fn: DateRange => Execution[T])(implicit dr: DateRange): Seq[Execution[(DateRange, T)]] =
```
johnynek: you can't change signatures like this in general. This breaks both source and binary compatibility. If you want the date inside, return in with fn. Seems not worth the change.

Dropping parallelism seems good since it was unused.

Maybe this should be private. Is is really helpful? If we are making incompatibilities, I'd rather be conservative and drop APIs we are not sure of.

richwhitjr: Yea I worried about that, wasn't sure if this made it out on a release or not.  Thought the release was held up by some dependency problems.  

Let me refactor a bit to avoid breaking the API.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
richwhitjr: It really is a side effect function, there are a few cases where I can see it being useful for alerting/logging failures especially on adhoc jobs. 

I can drop if you find the API confusing and move it internally into my code base.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionUtil.scala
johnynek: 0.16.0 final is not out yet, but several RCs have gone out.

Ideally, we never remove an API, but in this case if we want to tighten it a bit (and given the consumers can reimplement on their side).

The parallelism feature was the biggest win here, in my view. The rest is pretty easy to add.

richwhitjr:
Remove the callback.  Also made that simple function private, pretty easy to do as an end user but it helps cleanup the other functions in the object.

johnynek:
:+1:

johnynek:
was this fixed by #1562

johnynek:
ping @richwhitjr is this safe to close?

Issue #1430, opened 2015-08-25
Optimizing TypedPipe graphs


johnynek:
Optimizing scalding TypedPipe graphs at the scalding layer rather than cascading makes a lot of sense. For one, we don't have to box and unbox into cascading Tuples between operations. Secondly, the semantics and laws of many operations in scalding are quite clear, when we put them into general cascading Operations, that is either lost or we have to duplicate the structure.

Some optimizations we would like:
1. Make TypedPipe.fork a no-op. Users should not have to explicitly hint to the planner when they are branching out. Now if you don't do that, in some cases you might get recomputations you don't want.
2. user puts .toTypedPipe but then immediately joins or reduces on the same keys. They inserted an extra map/reduce step which could have been removed. If the really mean that, they should add .forceToDisk.
3. remove the use of cascading Merge in most cases. Right now, it is not a no-op on Tez as it is in mapreduce. We put merges in front of GroupBy, but they don't need to be there (this could and probably should be a cascading optimization, but it could done here too).
4. express laws like: `(a ++ b).join(c) == (a.join(c) ++ b.join(c))` or `a.flatMap(f) ++ a.flatMap(g) == a.flatMap(x => f(x) ++ g(x))`. There are several more laws like this. We could apply them to minimize the number of map/reduce steps a job takes.

A win of scalding is that we have been able to add features without breaking (too many) old users. This is mostly due to how many existing scalding jobs there are in production at Twitter. We don't want to break the TypedPipe API or drop support for the Fields API, so how can we do this optimization?

One idea is to check the flowDef right before we run. If all the tails are due to TypedPipe writes, we can just take that like of `(TypedPipe[T], TypedSink[T])` pairs do a whole graph optimization step, and render it into a FlowDef. If there are any tails that are not due to a TypedPipe write, we just run the flow as is. How do we know if a tail is due to a TypedPipe write? We can match a TypedSink to a pipe using the name: the pipe name in the getTails should match the sourceId for the TypedSink (which are always subclasses of Source though not strictly required by the type system). Then we could add to the FlowState each write that do by having some object like: `case class Write[T](pipe: TypedPipe[T], sink: TypedSink[T])` and keeping a `List[Write[_]]`. If all the writes account for all the tails, you are in business: optimize the graph and rewrite the FlowDef.

Gabriel439:
[...]

PR #1529, opened 2016-03-01
WIP: Add forceToDisk parameter to hashJoin in TypedPipe

piyushnarang:
Putting out a WIP PR for these changes to start some early feedback / conversations. 

We've seen a couple of incidents over the last few months of Scalding jobs causing issues on Hadoop if they had a large dataset which they filtered down and passed to the RHS of a hashJoin and didn't manually call forceToDisk. These changes make a call to forceToDisk on the RHS of a hashJoin by default. Users can opt out by setting it to false. If the user already has a forceToDisk on their RHS, this will result in a redundant extra forceToDisk on the small RHS data(as Scalding / Cascading don't currently optimize it away). 

Explored a couple of other options before coming back to this. As Scalding and Cascading TypedPipes / Pipes are source & Tap agnostic it doesn't seem to be possible to apply this selectively using that information. Another idea would be to potentially walk the Pipe graph and only add the forceToDisk if that data hasn't already been checkpointed. This is probably a bit more involved an option and seems like a one off optimization. 

Testing: Added a unit test to verify that opt-out behavior works fine. Also tested this out with a Scalding job. Shall update this PR with some results on the overhead of additional forceToDisk calls.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -359,13 +359,13 @@ trait Config extends Serializable {
   def setHRavenHistoryUserName: Config =
     this + (Config.HRavenHistoryUserName -> System.getProperty("user.name"))
 
-  def setSkipForceHashJoinRHS(b: Boolean): Config =
-    this + (SkipForceHashJoinRHS -> (b.toString))
+  def setHashJoinAutoForceRight(b: Boolean): Config =
+    this + (HashJoinAutoForceRight -> (b.toString))
 
-  def getSkipForceHashJoinRHS: Boolean =
-    get(SkipForceHashJoinRHS)
+  def getHashJoinAutoForceRight: Boolean =
+    get(HashJoinAutoForceRight)
       .map(_.toBoolean)
-      .getOrElse(true)
+      .getOrElse(false)
 
   override def hashCode = toMap.hashCode
   override def equals(that: Any) = that match {
@@ -412,11 +412,11 @@ object Config {
 
   /**
    * Parameter that can be used to determine behavior on the rhs of a hashJoin.
-   * If true, we skip force to disk on MapFn and FlatMapFn transforms as long as they're
-   * followed by a Pipe we think has been persisted (e.g. Checkpoint).
-   * Else we forceToDisk in those two scenarios (along with potentially others like FilterFn)
+   * If true, we try to guess when to auto force to disk before a hashJoin
+   * else (the default) we don't try to infer this and the behavior can be dictated by the user manually
+   * calling forceToDisk on the rhs or not as they wish.
    */
-  val SkipForceHashJoinRHS: String = "scalding.hashjoin.skipforceright"
+  val HashJoinAutoForceRight: String = "scalding.hashjoin.autoforceright"
 
   val empty: Config = Config(Map.empty)
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -359,6 +359,14 @@ trait Config extends Serializable {
   def setHRavenHistoryUserName: Config =
     this + (Config.HRavenHistoryUserName -> System.getProperty("user.name"))
 
+  def setSkipForceHashJoinRHS(b: Boolean): Config =
+    this + (SkipForceHashJoinRHS -> (b.toString))
+
+  def getSkipForceHashJoinRHS: Boolean =
+    get(SkipForceHashJoinRHS)
+      .map(_.toBoolean)
+      .getOrElse(true)
+
   override def hashCode = toMap.hashCode
   override def equals(that: Any) = that match {
     case thatConf: Config => toMap == thatConf.toMap
@@ -401,6 +409,14 @@ object Config {
   /** Manual description for use in .dot and MR step names set using a `withDescription`. */
   val PipeDescriptions = "scalding.pipe.descriptions"
   val StepDescriptions = "scalding.step.descriptions"
+
+  /**
+   * Parameter that can be used to determine behavior on the rhs of a hashJoin.
+   * If true, we skip force to disk on MapFn and FlatMapFn transforms as long as they're
+   * followed by a Pipe we think has been persisted (e.g. Checkpoint).
+   * Else we forceToDisk in those two scenarios (along with potentially others like FilterFn)
+   */
+  val SkipForceHashJoinRHS: String = "scalding.hashjoin.skipforceright"
 
   val empty: Config = Config(Map.empty)
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
@@ -69,14 +69,17 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
   private def getMappedPipe(fd: FlowDef, mode: Mode): Pipe = {
     val mappedPipe = mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter)
 
-    if (isSafeToSkipForceToDisk(mappedPipe, mode)) mappedPipe
+    // if the user has turned off auto force right, we fall back to the old behavior and
+    //just return the mapped pipe
+    if (!getHashJoinAutoForceRight(mode) || isSafeToSkipForceToDisk(mappedPipe, mode)) mappedPipe
     else mappedPipe.forceToDisk
   }
 
   /**
-   * Computes if it is safe to skip a force to disk. If we know the pipe is persisted,
-   * we can safely skip. If the Pipe is an Each operator, we check if the function it holds
-   * can be skipped and we recurse to check its parent pipe.
+   * Computes if it is safe to skip a force to disk (only if the user hasn't turned this off using
+   * Config.HashJoinAutoForceRight).
+   * If we know the pipe is persisted,we can safely skip. If the Pipe is an Each operator, we check
+   * if the function it holds can be skipped and we recurse to check its parent pipe.
    * Recursion handles situations where we have a couple of Each ops in a row.
    * For example: pipe.forceToDisk.onComplete results in: Each -> Each -> Checkpoint
    */
@@ -108,21 +111,21 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
         f.getFunction match {
           case _: Converter[_] => true
           case _: FilteredFn[_] => false //we'd like to forceToDisk after a filter
-          case _: MapFn[_, _] => getSkipForceHashJoinRHS(mode)
-          case _: FlatMappedFn[_, _] => getSkipForceHashJoinRHS(mode)
-          case _ => true // empty fn
+          case _: MapFn[_, _] => false // todo: figure out what to do for this
+          case _: FlatMappedFn[_, _] => false // todo: figure out what to do for this
+          case _ => false // treat empty fn as a Filter all so forceToDisk
         }
       case _: CleanupIdentityFunction => true
       case _ => false
     }
   }
 
-  private def getSkipForceHashJoinRHS(mode: Mode): Boolean = {
+  private def getHashJoinAutoForceRight(mode: Mode): Boolean = {
     mode match {
       case h: HadoopMode =>
         val config = Config.fromHadoop(h.jobConf)
-        config.getSkipForceHashJoinRHS
-      case _ => true //default to true
+        config.getHashJoinAutoForceRight
+      case _ => false //default to false
     }
   }
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.scalding.typed
 
 import cascading.flow.FlowDef
-import cascading.pipe.{ Pipe, HashJoin }
+import cascading.pipe._
 import com.twitter.scalding._
 
 import com.twitter.scalding.TupleConverter.tuple2Converter
@@ -60,15 +60,44 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
     })
 
   /**
-   * Returns a Pipe for the mapped (rhs) pipe. If this hasn't been persisted to disk,
-   * we force a checkpoint. Currently this persistence check is trivial: if the pipe is
-   * a ForceToDiskTypedPipe we skip an additional checkpoint.
+   * Returns a Pipe for the mapped (rhs) pipe with checkpointing applied if needed.
+   * Currently we skip checkpointing if:
+   * 1) If the pipe is a ForceToDiskTypedPipe (and thus the Cascading pipe is a Checkpoint)
+   * 2) If the Cascading pipe is a GroupBy / CoGroup
+   * 3) If while walking the pipe's parents we hit an Every before an Each
    */
-  def getMappedPipe[R, V1](fd: FlowDef, mode: Mode): Pipe = {
+  private def getMappedPipe(fd: FlowDef, mode: Mode): Pipe = {
     val mappedPipe = mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter)
-    mapped match {
-      case ForcedToDiskTypedPipe(_) => mappedPipe
-      case _ => mappedPipe.forceToDisk
+
+    mappedPipe match {
+      case _: Checkpoint => mappedPipe
+      case _: GroupBy => mappedPipe
+      case _: CoGroup => mappedPipe
+      case _ =>
+        val checkParents = checkEveryBeforeEach(mappedPipe)
+        checkParents match {
+          case Some(true) => mappedPipe
+          case _ => mappedPipe.forceToDisk
+        }
     }
+  }
+
+  /**
+   * Check if we encounter an Every before an Each while traversing parents of the pipe
+   */
+  private def checkEveryBeforeEach(pipe: Pipe): Option[Boolean] = {
+    Iterator
+      .iterate(Seq(pipe))(pipes => for (p <- pipes; prev <- p.getPrevious) yield prev)
+      .takeWhile(_.length > 0)
+      .flatten
+      .find {
+        case _: Each => true
+        case _: Every => true
+        case _ => false
+      }
+      .map {
+        case _: Each => false
+        case _: Every => true
+      }
   }
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
@@ -15,7 +15,8 @@ limitations under the License.
 */
 package com.twitter.scalding.typed
 
-import cascading.pipe.HashJoin
+import cascading.flow.FlowDef
+import cascading.pipe.{ Pipe, HashJoin }
 import com.twitter.scalding._
 
 import com.twitter.scalding.TupleConverter.tuple2Converter
@@ -50,11 +51,24 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
       val newPipe = new HashJoin(
         RichPipe.assignName(mapside.toPipe(('key, 'value))(fd, mode, tup2Setter)),
         Field.singleOrdered("key")(keyOrdering),
-        mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter),
+        getMappedPipe(fd, mode),
         Field.singleOrdered("key1")(keyOrdering),
         WrappedJoiner(new HashJoiner(joinFunction, joiner)))
 
       //Construct the new TypedPipe
       TypedPipe.from[(K, R)](newPipe.project('key, 'value), ('key, 'value))(fd, mode, tuple2Converter)
     })
+
+  /**
+   * Returns a Pipe for the mapped (rhs) pipe. If this hasn't been persisted to disk,
+   * we force a checkpoint. Currently this persistence check is trivial: if the pipe is
+   * a ForceToDiskTypedPipe we skip an additional checkpoint.
+   */
+  def getMappedPipe[R, V1](fd: FlowDef, mode: Mode): Pipe = {
+    val mappedPipe = mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter)
+    mapped match {
+      case ForcedToDiskTypedPipe(_) => mappedPipe
+      case _ => mappedPipe.forceToDisk
+    }
+  }
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
@@ -16,6 +16,7 @@ limitations under the License.
 package com.twitter.scalding.typed
 
 import cascading.flow.FlowDef
+import cascading.operation.Operation
 import cascading.pipe._
 import com.twitter.scalding._
 
@@ -60,44 +61,83 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
     })
 
   /**
-   * Returns a Pipe for the mapped (rhs) pipe with checkpointing applied if needed.
-   * Currently we skip checkpointing if:
-   * 1) If the pipe is a ForceToDiskTypedPipe (and thus the Cascading pipe is a Checkpoint)
-   * 2) If the Cascading pipe is a GroupBy / CoGroup
-   * 3) If while walking the pipe's parents we hit an Every before an Each
+   * Returns a Pipe for the mapped (rhs) pipe with checkpointing (forceToDisk) applied if needed.
+   * Currently we skip checkpointing if we're confident that the underlying rhs Pipe is persisted
+   * (e.g. a source / Checkpoint / GroupBy / CoGroup / Every) and we have 0 or more Each operator Fns
+   * that are not doing any real work (e.g. Converter, CleanupIdentityFunction)
    */
   private def getMappedPipe(fd: FlowDef, mode: Mode): Pipe = {
     val mappedPipe = mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter)
 
-    mappedPipe match {
-      case _: Checkpoint => mappedPipe
-      case _: GroupBy => mappedPipe
-      case _: CoGroup => mappedPipe
-      case _ =>
-        val checkParents = checkEveryBeforeEach(mappedPipe)
-        checkParents match {
-          case Some(true) => mappedPipe
-          case _ => mappedPipe.forceToDisk
-        }
+    if (isSafeToSkipForceToDisk(mappedPipe, mode)) mappedPipe
+    else mappedPipe.forceToDisk
+  }
+
+  /**
+   * Computes if it is safe to skip a force to disk. If we know the pipe is persisted,
+   * we can safely skip. If the Pipe is an Each operator, we check if the function it holds
+   * can be skipped and we recurse to check its parent pipe.
+   * Recursion handles situations where we have a couple of Each ops in a row.
+   * For example: pipe.forceToDisk.onComplete results in: Each -> Each -> Checkpoint
+   */
+  private def isSafeToSkipForceToDisk(pipe: Pipe, mode: Mode): Boolean = {
+    pipe match {
+      case eachPipe: Each =>
+        if (canSkipEachOperation(eachPipe.getOperation, mode)) {
+          //need to recurse down to see if parent pipe is ok
+          getPreviousPipe(eachPipe).exists(prevPipe => isSafeToSkipForceToDisk(prevPipe, mode))
+        } else false
+      case _: Checkpoint => true
+      case _: GroupBy => true
+      case _: CoGroup => true
+      case _: Every => true
+      case p if isSourcePipe(p) => true
+      case _ => false
     }
   }
 
   /**
-   * Check if we encounter an Every before an Each while traversing parents of the pipe
+   * Checks the transform to deduce if it is safe to skip the force to disk.
+   * If the FlatMapFunction is a converter / EmptyFn / IdentityFn then we can skip
+   * For FilteredFn we could potentially save substantially so we want to forceToDisk
+   * For map and flatMap we defer to the user config.
    */
-  private def checkEveryBeforeEach(pipe: Pipe): Option[Boolean] = {
-    Iterator
-      .iterate(Seq(pipe))(pipes => for (p <- pipes; prev <- p.getPrevious) yield prev)
-      .takeWhile(_.length > 0)
-      .flatten
-      .find {
-        case _: Each => true
-        case _: Every => true
-        case _ => false
-      }
-      .map {
-        case _: Each => false
-        case _: Every => true
-      }
+  private def canSkipEachOperation(eachOperation: Operation[_], mode: Mode): Boolean = {
+    eachOperation match {
+      case f: FlatMapFunction[_, _] =>
+        f.getFunction match {
+          case _: Converter[_] => true
+          case _: FilteredFn[_] => false //we'd like to forceToDisk after a filter
+          case _: MapFn[_, _] => getSkipForceHashJoinRHS(mode)
+          case _: FlatMappedFn[_, _] => getSkipForceHashJoinRHS(mode)
+          case _ => true // empty fn
+        }
+      case _: CleanupIdentityFunction => true
+      case _ => false
+    }
+  }
+
+  private def getSkipForceHashJoinRHS(mode: Mode): Boolean = {
+    mode match {
+      case h: HadoopMode =>
+        val config = Config.fromHadoop(h.jobConf)
+        config.getSkipForceHashJoinRHS
+      case _ => true //default to true
+    }
+  }
+
+  private def getPreviousPipe(p: Pipe): Option[Pipe] = {
+    if (p.getPrevious != null && p.getPrevious.length == 1) p.getPrevious.headOption
+    else None
+  }
+
+  /**
+   * Return true if a pipe is a source Pipe (has no parents / previous) and isn't a
+   * Splice.
+   */
+  private def isSourcePipe(pipe: Pipe): Boolean = {
+    pipe.getParent == null &&
+      (pipe.getPrevious == null || pipe.getPrevious.isEmpty) &&
+      (!pipe.isInstanceOf[Splice])
   }
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
@@ -52,7 +52,7 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
       val newPipe = new HashJoin(
         RichPipe.assignName(mapside.toPipe(('key, 'value))(fd, mode, tup2Setter)),
         Field.singleOrdered("key")(keyOrdering),
-        getMappedPipe(fd, mode),
+        getForceToDiskPipeIfNecessary(fd, mode),
         Field.singleOrdered("key1")(keyOrdering),
         WrappedJoiner(new HashJoiner(joinFunction, joiner)))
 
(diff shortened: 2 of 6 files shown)


johnynek:
I'm not crazy about this approach. I don't like adding options that most users don't want to have to think about.  What about this:

``` scala
case class ForcedToDisk[T](pipe: TypedPipe[T]) extends TypedPipe[T] {
  override def forceToDisk: TypedPipe[T] = this
  override def asPipe ... // move the code from TypedPipe.forceToDisk
}
```

The we can recursively look up the pipe to see if there is already a ForceToDisk between us and any filter. If not, we add it.

Would that work?

piyushnarang:
@johnynek I think that might work. Let me try prototyping that.

isnotinvain:
@johnynek yeah, we discussed this approach too. I think it will work, it just seemed more involved of a change when as you've pointed out before, there's lots of auto-optimizations we could be doing and currently aren't mostly due to how scalding is designed preventing us from having a generic dag optimizer. But I think this will work, though I'm not sure for how much gain. If the RHS is supposed to fit in memory, then worst case we are double-writing something very small.

If we go the route of `ForcedToDisk` then I think the logic for whether to add a `forceToDisk` in `hashJoin` needs to be something like: "if the rhs is a ForcedToDisk, or the result of a join, or any other operation we happen to know has just been persisted, then don't add the forceToDisk, otherwise add it." Does that sound right? If we get a `Grouped` then can we assume it's been written to disk? Not really right because there's some optimizing that goes on there if no aggregations are asked for?

isnotinvain:
A similar approach is that the cascading `Pipe` actually can probably tell us if it's been checkpointed as well, maybe with more accuracy? How do you feel about just walking the `Pipe`'s dag?

johnynek:
@isnotinvain yeah, I hear you. I keep thinking about making scalding's dag more like summingbird. If we could solve the issues here: #1430 this becomes easy. Maybe in 0.17?

It should be easy to check if you have something persisted: look at the pipe. If it is a subtype of one of a few cascading types, you skip, else add the Checkpoint.

piyushnarang:
@johnynek reworked the implementation to use the ForceToDisk pipe. 
Tried this out with a test hash Join job and I see that if there isn't a forceToDisk on the RHS it gets added. The code does handle the trivial case of a user already calling .forceToDisk on the RHS (I see that only one checkpoint is taken and not two). 

I haven't yet put the more sophisticated checks in place in HashJoinable. Was a bit unclear on your last comment. 

Do you know which cascading types indicate that something has been persisted? Checkpoint I guess we can say has been. So if the rhs in HashJoinable is a checkpoint we know we're good. What I'm not sure really sure about:
1) Do you know which other types we should check? 
2) Should we also be recursing down the rhs' previous Pipes? Or are we good for now just checking the rhs pipe?

Also, what do you think of adding more sophisticated checks as a follow up PR?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1131,6 +1132,28 @@ class MappablePipeJoinEnrichment[T](pipe: TypedPipe[T]) {
...
+  import Dsl._
+
+  override def forceToDisk: TypedPipe[T] = this
+
+  override def asPipe[U >: T](fieldNames: Fields)(implicit flowDef: FlowDef, mode: Mode, setter: TupleSetter[U]): Pipe = {
+    val pipe = typedPipe.toPipe[U](fieldNames)(flowDef, mode, setter)
+    pipe.forceToDisk
+  }
+
+  override def cross[U](tiny: TypedPipe[U]): TypedPipe[(T, U)] =
```
johnynek: I don't think we want these two, do we? We are breaking the contract that the forceToDisk happens before the cross, right?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1131,6 +1132,28 @@ class MappablePipeJoinEnrichment[T](pipe: TypedPipe[T]) {
...
+    ForcedToDiskTypedPipe(typedPipe.cross(tiny))
+
+  override def flatMap[U](f: T => TraversableOnce[U]): TypedPipe[U] =
```
johnynek: same.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +51,24 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
       //Construct the new TypedPipe
       TypedPipe.from[(K, R)](newPipe.project('key, 'value), ('key, 'value))(fd, mode, tuple2Converter)
     })
+
+  /**
+   * Returns a Pipe for the mapped (rhs) pipe. If this hasn't been persisted to disk,
+   * we force a checkpoint. Currently this persistence check is trivial: if the pipe is
+   * a ForceToDiskTypedPipe we skip an additional checkpoint.
+   */
+  def getMappedPipe[R, V1](fd: FlowDef, mode: Mode): Pipe = {
```
johnynek: can this be private?

johnynek: what is `R, V1` here? They are not needed right?

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +51,24 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+    val mappedPipe = mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter)
+    mapped match {
```
johnynek: this is confusing. Do the above in the match:

``` scala
val unwrapped = mapped match {
  case ForceToDiskTypedPipe(p) => p
  case p => p
}
unwrapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter).forceToDisk
```

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang: Update to make this private & remove the [R, V1]. Missed removing the generic args after refactoring some stuff.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
piyushnarang: Hmm yeah we do end up breaking the contract. I guess the problem is we do need to override these two methods if we want our class to be concrete (as they're not defined in the parent - TypedPipe). 
I think I'll need a way to force the checkpoint on the 'typedPipe' object and then call cross / flatMap on it.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
johnynek: the better solution looks something like:

``` scala
def forceBeforeFlatMap[T](tp: TypedPipe[T]): TypedPipe[T] = tp match {
  case f@ForceToDiskTypedPipe(_) => f // can't move something the user did
  case tpi : TypedPipeInst[_] =>
    // need to clean up this API to avoid the cast:
   val tpit = tpi.asInstanceOf[TypedPipeInst[T]]
   tpit.flatMapFn match {
     case FlatMappedFn(prev, f) =>
       val upstream = new TypePipeInst(pipe, ... , prev)
       val forced = forceBeforeFlatMap(upstream).
       // apply the flatMap
       forced.flatMap(f)
    case Converter(_) => tpit // this is just a source, already forced
    case Empty => TypedPipe.empty // may never happen, but for completeness
    case _ => tpit.forceToDisk
   }
   case p => p.forceToDisk
}
```

The goal here is to not materialize a big flatMap on disk. We do want to materialize after a filter or even a map (which is often a projection). We don't want a forceToDisk on a source, that is a waste.

to be more correct, until we have a better AST, at the point of calling `toPipe` we should check if that pipe is an instance of:

https://github.com/Cascading/cascading/blob/2.7/cascading-core/src/main/java/cascading/pipe/GroupBy.java#L74

or:

https://github.com/Cascading/cascading/blob/2.7/cascading-core/src/main/java/cascading/pipe/CoGroup.java#L114

in which case we don't need to checkpoint.

See https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala#L701

you can call `getPrevious` and if you don't hit any `Each` before hitting an `Every` you are okay.

To answer your question about going iteratively, I am nervous about that. This is a change that could hurt a lot of people (and without exaggeration cost millions of dollars) if done poorly. Let's take a little time and do it right.

I can help.

piyushnarang: updated

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: see `TypedPipeFactory` for this use case then.

Do wish we had a nice AST here.  :/

piyushnarang:
@johnynek made some updates based on your comments:
1) In HashJoinable, I check if the pipe is a Checkpoint, Groupby, CoGroup (or among the parents an Every shows up before an Each) - In those cases, I skip calling forceToDisk. Else I do.
2) In the ForceToDiskTypedPipe, I ensure we `forceBefore` the cross & flatmap. If the underlying TypedPipe is a TypedPipeInst, I skip forcing if it wraps a Checkpoint, CoGroup or Every pipe. In other cases I trigger a forceToDisk. 

Currently for TypedPipeFactories in `forceBefore` I end up calling `physicalForceToDisk`. Looked at the TypedPipeFactory API and it wasn't clear to me how we'd want to hook that into the forceBefore method as well as ensure the force to disk happens(unless we want to potentially introspect and optimize similar to TypedPipeInst?). Thought I'd put up what I had at this point as it would make feedback and continuing forward easier.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +51,53 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+          case Some(true) => mappedPipe
+          case _ => mappedPipe.forceToDisk
+        }
+    }
+  }
+
+  /**
+   * Check if we encounter an Every before an Each while traversing parents of the pipe
+   */
+  private def checkEveryBeforeEach(pipe: Pipe): Option[Boolean] = {
```
johnynek: can this just return Boolean?

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +51,53 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+    Iterator
+      .iterate(Seq(pipe))(pipes => for (p <- pipes; prev <- p.getPrevious) yield prev)
```
johnynek: I'm worried that if there are multiple previous this logic does not hold.

Can we stop in that case or `.filter(_.size == 1)` below?

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +51,53 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+      .flatten
+      .find {
+        case _: Each => true
+        case _: Every => true
+        case _ => false
+      }
+      .map {
+        case _: Each => false
+        case _: Every => true
+      }
```
johnynek:
``` scala
.collect {
  case (_: Each) => true
  case (_: Every) => false
}
.toList
.headOption // should be on iterator, grrrrrrr
.getOrElse(false) // assume we are not checkpointed if we something other that `Every`
```

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang: Yeah I'll add the filter for size

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
johnynek: actually, isn't all this logic just:

``` scala
pipe match {
  case _: Every => true
  case _ => false
}
```

what is the difference?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -928,7 +930,7 @@ class TypedPipeFactory[T] private (@transient val next: NoStackAndThen[(FlowDef,
 /**
  * This is an instance of a TypedPipe that wraps a cascading Pipe
  */
-class TypedPipeInst[T] private[scalding] (@transient inpipe: Pipe,
+class TypedPipeInst[T] private[scalding] (@transient val inpipe: Pipe,
```
johnynek: let's not add the val, but a `private[scalding] def getPipe: Pipe` to be 100% clear this is not a public method.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1131,6 +1133,47 @@ class MappablePipeJoinEnrichment[T](pipe: TypedPipe[T]) {
...
+  override def flatMap[U](f: T => TraversableOnce[U]): TypedPipe[U] =
+    forceBefore(typedPipe).flatMap(f)
+
+  private def forceBefore(pipe: TypedPipe[T]): TypedPipe[T] = pipe match {
+    case fToDiskPipe: ForceToDiskTypedPipe[_] => fToDiskPipe // can't move something the user did
+    case typedPipeInst: TypedPipeInst[_] =>
+      if (isMaterialized(typedPipeInst.inpipe))
+        typedPipeInst
+      else
+        physicalForceToDisk
```
johnynek: what about `case TypedPipeFactory(next) => TypedPipeFactory(next.andThen(forceBefore))` (or something)

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1131,6 +1133,47 @@ class MappablePipeJoinEnrichment[T](pipe: TypedPipe[T]) {
...
+    case _ => physicalForceToDisk
+  }
+
+  private def isMaterialized(pipe: Pipe): Boolean = pipe match {
```
johnynek: another materialized case is a Pipe with no parent. This is a source.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -342,15 +342,16 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest
       HadoopPlatformJobTest(new TypedPipeJoinWithDescriptionJob(_), cluster)
         .inspectCompletedFlow { flow =>
           val steps = flow.getFlowSteps.asScala
-          steps should have size 1
+          steps should have size 2
```
johnynek: can you add a comment as to why we expect two steps here?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
piyushnarang: sounds good, done

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
johnynek: actually, I expect 1 step still. This is a good unit test because we should not be adding an extra step in the hashjoin here, as far as I can tell.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1131,6 +1133,47 @@ class MappablePipeJoinEnrichment[T](pipe: TypedPipe[T]) {
...
+
+  override def cross[U](tiny: TypedPipe[U]): TypedPipe[(T, U)] =
+    forceBefore(typedPipe).cross(tiny)
+
+  override def flatMap[U](f: T => TraversableOnce[U]): TypedPipe[U] =
+    forceBefore(typedPipe).flatMap(f)
+
+  private def forceBefore(pipe: TypedPipe[T]): TypedPipe[T] = pipe match {
+    case fToDiskPipe: ForceToDiskTypedPipe[_] => fToDiskPipe // can't move something the user did
+    case typedPipeInst: TypedPipeInst[_] =>
```
johnynek:
``` scala
case typedPipeInst: TypedPipeInst[_] if isMaterialized(typedPipeInst.inpipe) => pipe
case _ => physicalForceToDisk
```

is clearer

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1131,6 +1133,47 @@ class MappablePipeJoinEnrichment[T](pipe: TypedPipe[T]) {
...
+  override def asPipe[U >: T](fieldNames: Fields)(implicit flowDef: FlowDef, mode: Mode, setter: TupleSetter[U]): Pipe = {
+    val pipe = typedPipe.toPipe[U](fieldNames)(flowDef, mode, setter)
+    pipe.forceToDisk
+  }
+
+  override def cross[U](tiny: TypedPipe[U]): TypedPipe[(T, U)] =
+    forceBefore(typedPipe).cross(tiny)
+
+  override def flatMap[U](f: T => TraversableOnce[U]): TypedPipe[U] =
+    forceBefore(typedPipe).flatMap(f)
```
johnynek: actually, I think we don't want forceBefore at all. Don't we always want: `physicalForceToDisk.flatMap(f)`?

If the user has a `groupBy`, that is materialized, but `groupBy.map(f)` will glue the `f` into the reducer. That is not what they mean when they do a `.forceToDisk`.

I'm starting to wonder if this whole `ForceToDisk` class is a red-herring.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +51,53 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
       val newPipe = new HashJoin(
         RichPipe.assignName(mapside.toPipe(('key, 'value))(fd, mode, tup2Setter)),
         Field.singleOrdered("key")(keyOrdering),
-        mapped.toPipe(('key1, 'value1))(fd, mode, tup2Setter),
+        getMappedPipe(fd, mode),
```
johnynek: I think maybe this is the only change we want to make.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang: In this case if we encounter any other Pipe before an Every we'll return false and hence forceToDisk. The prior logic was to see between Each and Every which one we saw first. If it was an Every, skip forceToDisk and if it was an each write to disk.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang: Yeah, I've updated this based on your suggestion below with the .collect.toList ...

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
piyushnarang: ah nice. To confirm, this is indicated getPrevious array being empty right? (and not the `parent` field on the Pipe :-) )

On stepping through using a debugger it seems like the parent field is always null. I see previous seems to have data in most cases (apart from situations like when the Pipe is a CoGroup / HashJoin / .. in those cases, previous is empty but `pipes` is filled up)

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
piyushnarang: Yeah so what's happening here as of now is that the rhs is a Pipe wrapping an IterableSource. Since we don't currently have a check for Pipe without parent, we end up falling back to checking if there's an Every before an Each and that ends up defaulting to false and we checkpoint (which is the extra step).

I can add the Pipe with no parent check in this location as well.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
piyushnarang: agree, done

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
piyushnarang: So if I understand correctly, the `ForceToDiskTypedPipe` seems to be giving us:
1) Protection against a user calling multiple forceToDisk operations successively by mistake
2) Type encapsulation which we could use maybe in the future if we have a fancier optimizer to prune away some of these ForceToDisk calls that are redundant.
3) Current scenario: If a force2Disk is called followed by a cross / flatMap we check the prior pipe / operations to potentially optimize things. I think in this case if we do a physicalForceToDisk.flatMap or physicalForceToDisk.cross always, that shouldn't make things worse than they currently are. 

We could drop the `ForceToDiskTypedPipe` and just do the checking in `HashJoinable.getMappedPipe`. That's similar to @isnotinvain 's suggestion earlier of walking the DAG in HashJoinable. 

Let me know if this sounds good, I'll update the PR accordingly.

johnynek: yeah, I think these are two issues the users care about:
1) as you say: `t.forceToDisk.forceToDisk == t.forceToDisk`, which is: make `forceToDisk` idempotent. Nice to have, not a huge deal.

2) make sure all `hashJoin` calls happen with a right that has been forced (maybe except the case of a flatMap). But don't force when it is not needed.

Is that right? To address 2), we don't need the new `ForceToDiskTypedPipe` type, right? I think 2) is the main feature. We can come back to the others.

piyushnarang: Yeah for 2, we don't need a new ForceToDiskTypedPipe. We're already ensuring that the rhs has either been forcedToDisk prior to the Cascading hashJoin call or we end up explicitly calling it (when we can verify that).

piyushnarang:
@johnynek - Been working on the unit tests and trying to get them to work and that seems to throw up a few interesting questions.

Here's the skeleton of a unit test that tries to pass a hashJoin rhs that has already been checkpointed:

```
class TypedPipeHashJoinWithForceToDiskJob(args: Args) extends Job(args) {
  val x = TypedPipe.from[(Int, Int)](List((1, 1)))
  val y = TypedPipe.from[(Int, String)](List((1, "first")))
  val z = TypedPipe.from[(Int, Boolean)](List((2, true))).group

  //trivial transform and forceToDisk on the rhs
  val yMap = y.map( p => (p._1, p._2.toUpperCase)).forceToDisk

  x.hashJoin(yMap) 
      .withDescription("hashJoin")
      .leftJoin(z)
      .withDescription("leftJoin")
      .values
      .write(TypedTsv[((Int, String), Option[Boolean])]("output"))
}
```

Our expectation is that since this rhs has been Checkpointed, we'd end up skipping it in [HashJoinable.getMapped()](https://github.com/piyushnarang/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala#L69)

However, what ends up happening is that due to the first line in that function where we call toPipe() on the rhs, we end up with a pipe setup that looks like this: Each -> Checkpoint (Each which has a parent of Checkpoint). This is because toPipe ends up calling [TypedPipeInst.asPipe()](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala#L983) which calls a flatMapTo on the underlying RichPipe. 

End result of this is our plain mappedPipe.isInstanceOf[Checkpoint] will not work. (I've seen the same behavior for GroupBy & CoGroup). 

One option is to check if the mappedPipe is a Each -> Checkpoint (or Each -> GroupBy or Each -> CoGroup). I've been poking around with this and it seems like if we pass an RHS such as:  
`val yMap = y.map( p => (p._1, p._2.toUpperCase)).forceToDisk.flatMap( p => List((p._1, p._2)))` (basically a filter / map / flatMap after the forceToDisk) 
that also ends up resulting in an Each -> Checkpoint. 

I'm not sure if we can go ahead and assume that we don't need the Checkpoint in those kinds of scenarios as well? Or can we assume that typically the transforms following the HashJoin's rhs.forceToDisk shouldn't expand the size too much?

piyushnarang:
@johnynek - Pushed some updates & unit tests to tackle the issues I brought up above based on our conversation.  
In the HashJoinable code we now look at the rhs pipe (and its parents) to infer if we need a forceToDisk or not. 

If the pipe is an Each we check the type of the operation it holds to decide whether that is something we can skip if the following Pipe has been persisted (like Checkpoints / GroupBy / ...). Stuff like [Converter](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala#L36) is skipped and we end up forcing on [Filter](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala#L55). 

Also noticed we could run into scenarios such as:
`pipe.forceToDisk.onComplete` which results in `Each -> Each -> Checkpoint`. To handle those and potentially redundant Each operations in a chain, we walk the Each operations until we either reach a different Pipe which helps us determine if its persisted OR until we get to know we need a forceToDisk earlier (e.g. finding a Filter).

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +52,92 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+   */
+  private def canSkipEachOperation(eachOperation: Operation[_], mode: Mode): Boolean = {
+    eachOperation match {
+      case f: FlatMapFunction[_, _] =>
+        f.getFunction match {
+          case _: Converter[_] => true
+          case _: FilteredFn[_] => false //we'd like to forceToDisk after a filter
+          case _: MapFn[_, _] => getSkipForceHashJoinRHS(mode)
+          case _: FlatMappedFn[_, _] => getSkipForceHashJoinRHS(mode)
+          case _ => true // empty fn
```
johnynek: I don't think this ever comes up, but `Empty` is basically like filter all. I think we should treat like `FilterFn`.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -402,6 +410,14 @@ object Config {
   val PipeDescriptions = "scalding.pipe.descriptions"
   val StepDescriptions = "scalding.step.descriptions"
 
+  /**
+   * Parameter that can be used to determine behavior on the rhs of a hashJoin.
+   * If true, we skip force to disk on MapFn and FlatMapFn transforms as long as they're
+   * followed by a Pipe we think has been persisted (e.g. Checkpoint).
+   * Else we forceToDisk in those two scenarios (along with potentially others like FilterFn)
+   */
+  val SkipForceHashJoinRHS: String = "scalding.hashjoin.skipforceright"
```
johnynek: let's change this to `scalding.hashjoin.autoforceright` : if true, scalding will guess when to try to force to disk before a hashjoin, if false (the default) do exactly what the user says to do.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +52,92 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+
+  /**
+   * Computes if it is safe to skip a force to disk. If we know the pipe is persisted,
+   * we can safely skip. If the Pipe is an Each operator, we check if the function it holds
+   * can be skipped and we recurse to check its parent pipe.
+   * Recursion handles situations where we have a couple of Each ops in a row.
+   * For example: pipe.forceToDisk.onComplete results in: Each -> Each -> Checkpoint
+   */
+  private def isSafeToSkipForceToDisk(pipe: Pipe, mode: Mode): Boolean = {
+    pipe match {
```
johnynek, line 87: I think we first check that `! autoForceRight` at this point so we have a way to totally trigger the old behavior.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +52,92 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+   * For FilteredFn we could potentially save substantially so we want to forceToDisk
+   * For map and flatMap we defer to the user config.
+   */
+  private def canSkipEachOperation(eachOperation: Operation[_], mode: Mode): Boolean = {
+    eachOperation match {
+      case f: FlatMapFunction[_, _] =>
+        f.getFunction match {
+          case _: Converter[_] => true
+          case _: FilteredFn[_] => false //we'd like to forceToDisk after a filter
+          case _: MapFn[_, _] => getSkipForceHashJoinRHS(mode)
```
johnynek: we could look back in the `MapFn` and see if there are only filter and map parents. If there is a flatMapParent don't force, else do?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -154,6 +154,94 @@ class TypedPipeJoinWithDescriptionJob(args: Args) extends Job(args) {
...
+  val coGroupTuplePipe = coGroupTypedPipe.map{ case(a,b,c) => (a, (b,c)) }
+  x.hashJoin(coGroupTuplePipe)
+   .withDescription("hashJoin")
+   .write(TypedTsv[(Int, (Int, (Int, Int)))]("output"))
+}
+
+class TypedPipeHashJoinWithEveryJob(args: Args) extends Job(args) {
+  val x = TypedPipe.from[(Int, String)](Tsv("input1", ('x1, 'y1)), Fields.ALL)
+  val y = Tsv("input2", ('x2, 'y2)).groupBy('x2) {
+    _.foldLeft('y2 -> 'y2)(0){ (b: Int, a: Int) => b + a}
```
johnynek, line 264: does

``` scala
TypedPipe.from(TypedTsv[(Int, Int)]("input1").group.foldLeft(0)(_ + _)
```

not work here? I don't care much about the fields API. Let's test the typed API.

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang: yes, that makes sense.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
piyushnarang: sounds good, done

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang, line 87: done

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
piyushnarang, line 264: Hmm seems like the line you've listed does not create an `Every` pipe. I end up with an Each -> source Pipe. Not super familiar with this but from what I can make out debugging, seems like when we do typedPipe.group we end up with a Grouped object instead of a GroupBuilder. The foldLeft on GroupBuilder is the only one that creates the Every (and not the one on Grouped). 

Seems like all the calls to create a new Every seem to be in GroupBuilder and that GroupBuilder is being used in RichPipe..

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
piyushnarang: Actually, I’m wondering now that we have the top level autoForceRight toggle, if it makes sense in HashJoinable.canSkipEachOperation to just say forceToDisk for map & flatmap. That way we end up skipping force to disk when we definitively know the work / transform functions have been trivial (Converter / Identity) after a persisted pipe (Checkpoint / GroupBy). In others - filter, map,flatMap we end up forcing to disk. 

Think the idea originally was to end up skipping force to disk when we know it's redundant / not of any use (e.g. Each(converter) -> Checkpoint). Seems like we can't say that for sure for flatMap (it could be reducing / expanding the dataset size). As we now have an opt out, we could make flatMap end up forcing to Disk.

I'm a bit unsure about map. We could also treat it similarly to Converter - check the parents of it till we can make a decision (filter = forceToDisk, persisted pipe = no forceToDisk). I guess the unknown is that map could also be potentially expanding / reducing the size of each object (number of records staying same).

johnynek:
:+1: when it is green.

Can you do a trial of the e2e tests at twitter with this branch and TwitterJob/TwitterConfig turning it on?

Thanks for pushing through!

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +52,95 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
+   * if the function it holds can be skipped and we recurse to check its parent pipe.
+   * Recursion handles situations where we have a couple of Each ops in a row.
+   * For example: pipe.forceToDisk.onComplete results in: Each -> Each -> Checkpoint
+   */
+  private def isSafeToSkipForceToDisk(pipe: Pipe, mode: Mode): Boolean = {
+    pipe match {
+      case eachPipe: Each =>
+        if (canSkipEachOperation(eachPipe.getOperation, mode)) {
+          //need to recurse down to see if parent pipe is ok
+          getPreviousPipe(eachPipe).exists(prevPipe => isSafeToSkipForceToDisk(prevPipe, mode))
```
rubanm, line 91: If the parent pipe does not exist, this will return false. We still want to skip in that case though right? Should this be `forall`?

johnynek, line 91: Good question. This relies on the fact that there is at least on `Each` on all pipes in a TypedPipe to do the work of pulling from cascading Tuple to scala type.

@piyushnarang when do you see more that one previous? Is that only on a splice?

piyushnarang, line 91: Based on my limited understanding of this, it seems as though we shouldn't end up with an Each without an previous Pipe. Can we say that's true (or am I missing something)? 

Regarding the case with two previous pipes - I don't think I've encountered a Pipe with more than one previous. I've seen us handling it in other places in the code but I'm not sure if it was just safe programming (as `previous` is an array and could in theory have multiple elements) or due to a specific scenario. If anyone knows of a situation where this occurs, do chime in. 

I have seen `previous` end up being null though and that is in the case where it is a Splice (the parents there are in the [pipes](https://github.com/twitter/cascading/blob/2.1/cascading-core/src/main/java/cascading/pipe/Splice.java#L62) field, `previous` and `parent` are null. 

Guess, I'm wondering if we do end up forcing to disk here - is that bad? We're not able to infer that the pipe is persisted, so we just end up forcing to disk. We have a config option that allows users to opt out.

johnynek, line 91: That sounds good to me.

rubanm, line 91: :+1:

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -50,11 +52,96 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
       TypedPipe.from[(K, R)](newPipe.project('key, 'value), ('key, 'value))(fd, mode, tuple2Converter)
     })
+
+  /**
+   * Returns a Pipe for the mapped (rhs) pipe with checkpointing (forceToDisk) applied if needed.
+   * Currently we skip checkpointing if we're confident that the underlying rhs Pipe is persisted
+   * (e.g. a source / Checkpoint / GroupBy / CoGroup / Every) and we have 0 or more Each operator Fns
+   * that are not doing any real work (e.g. Converter, CleanupIdentityFunction)
+   */
+  private def getMappedPipe(fd: FlowDef, mode: Mode): Pipe = {
```
isnotinvain: maybe call this `forceToDiskIfNecessary` or something like that?

piyushnarang: sure, will update

piyushnarang:
@johnynek thanks for the detailed review and all the help on this :-). Looks like the end to end tests went through successfully with autoForce turned on. :tada:

johnynek:
@piyushnarang Exciting! Nice work on this. Will be a nice feature (I expect we will default it on at Stripe too).

Issue #1247, opened 2015-04-09
Deprecate or warn against using the weakly typed `source` function in `JobTest`


Gabriel439:
`JobTest` provides two `source` methods:

```scala
-- Weakly typed version
﻿def source(s: Source, iTuple: Iterable[Product]): JobTest

-- Strongly typed version
def source[T](s: Source, iTuple: Iterable[T])(implicit setter: TupleSetter[T]): JobTest
```

I just debugged a failing test with @nathantchan where he was accidentally using the weakly typed `source` method, causing a `ClassCastException` and then when we switched to the strongly typed `source` method the problem disappeared.

I'm recommending that we minimally deprecate or warn when the user accidentally picks up the weakly typed `source` method to avoid these sorts of errors and eventually remove it.

johnynek:
I say delete it.

I can't imagine anyone being burned by this. There is a
TupleSetter[Product] that they can manually pass to get the old behavior.

On Thu, Apr 9, 2015 at 9:15 AM, Gabriel Gonzalez <notifications@github.com>
wrote:

> JobTest provides two source methods:
>
> -- Weakly typed version
> ﻿def source(s: Source, iTuple: Iterable[Product]): JobTest
> -- Strongly typed versiondef source[T](s: Source, iTuple: Iterable[T])(implicit setter: TupleSetter[T]): JobTest
>
> I just debugged a failing test with @nathantchan
> <https://github.com/nathantchan> where he was accidentally using the
> weakly typed source method, causing a ClassCastException and then when we
> switched to the strongly typed source method the problem disappeared.
>
> I'm recommending that we minimally deprecate or warn when the user
> accidentally picks up the weakly typed source method to avoid these sorts
> of errors and eventually remove it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twitter/scalding/issues/1247>.
>



-- 
Oscar Boykin :: @posco :: http://twitter.com/posco

johnynek:
I'd say this is a good candidate for removal in 0.14. It's mechanical to fix, unlikely to hit many anyway, and only in tests: so by construction tests will catch issues.

PR #1531, opened 2016-03-01
Remove weakly typed Source

tglstory:
Addresses https://github.com/twitter/scalding/issues/1247

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -428,13 +428,13 @@ abstract class FixedPathSource(path: String*) extends FileSource {
   override def equals(that: Any): Boolean = (that != null) && (that.toString == toString)
 
   /**
-    * Similar in behavior to {@link TimePathedSource.writePathFor}.
-    * Strip out the trailing slash star.
-    */
+   * Similar in behavior to {@link TimePathedSource.writePathFor}.
+   * Strip out the trailing slash star.
+   */
   protected def stripTrailing(path: String): String = {
     assert(path != "*", "Path must not be *")
     assert(path != "/*", "Path must not be /*")
-    if(path.takeRight(2) == "/*") {
+    if (path.takeRight(2) == "/*") {
       path.dropRight(2)
     } else {
       path
--- a/scalding-core/src/main/scala/com/twitter/scalding/JobTest.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/JobTest.scala
@@ -100,9 +100,6 @@ class JobTest(cons: (Args) => Job) {
   def ifSource[T](fn: PartialFunction[Source, Iterable[T]])(implicit setter: TupleSetter[T]): JobTest =
     source(fn.lift)
 
-  def source(s: Source, iTuple: Iterable[Product]): JobTest =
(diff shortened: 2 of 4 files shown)


johnynek:
I'm +1. @rubanm @isnotinvain @jnievelt are you folks willing to help fix any tests this breaks at Twitter?

rubanm:
I'll run a Twitter sandbox with this change so we know just how many targets it affects (hopefully not many).

johnynek:
@rubanm thanks a ton!

rubanm:
@johnynek 22 failing targets, not too bad. So I think we can fix those.

@tglstory There's one failing test on travis, could you fix that so it's green to merge. Thanks

```
[error] /home/travis/build/twitter/scalding/scalding-core/src/test/scala/com/twitter/scalding/mathematics/Matrix2Test.scala:399: ambiguous implicit values:
[error]  both value UnitSetter in object TupleSetter of type => com.twitter.scalding.TupleSetter[Unit]
[error]  and value CTupleSetter in object TupleSetter of type => com.twitter.scalding.TupleSetter[cascading.tuple.Tuple]
[error]  match expected type com.twitter.scalding.TupleSetter[T]
[error]         .source(Tsv("mat2", ('x2, 'y2, 'v2)), List())
[error]                ^
[error] one error found
[error] (scalding-core/test:compileIncremental) Compilation failed
```

tglstory:
updated test.

tglstory:
is this break expected? I resolved the test mentioned by @rubanm, but apparently travis is still failing.

johnynek:
No, this is not expected. These are probably cases where the lack of type-safety of the Fields API is biting us. Those tests probably need an explicit type parameter added.

Can you run `./sbt scalding-core/test` and see if the tests pass on your machine?

tglstory:
Thanks. I'll run that and update the branch when all the tests are fixed.

tglstory:
The BddDsl had some utilization of the weakly typed source.

isnotinvain:
From a binary compatibility standpoint, shouldn't we deprecate this first and then remove in the release that comes after the next release? At some point we were being very careful with our binary compatibility. Even if this isn't too hard to migrate to at Twitter and Stripe there are other users out there too :p

isnotinvain:
I guess we can also just be accurate with our version numbers and let this trigger a new major release

johnynek:
I am kind of bearish on deprecation.

Seems approximately no one acts on that.

A better model might be: use the old code until you want to do the (tiny) work to upgrade.

So, tiny changes (like < 1% or jobs have a mechanical change), is okay if it keeps many users from hitting bad cases.

Something that hits almost everyone, we should not (maybe ever) change.

johnynek:
I'd be +1 this change if @rubanm is still willing to fix the breakage (that fix can happen without shipping w new scalding, just fix the old tests to point to the right new method).

rubanm:
I have the affected tests fixed in a branch, so we can merge this if there's consensus.

johnynek:
Nice, thanks @rubanm !

I want to merge as that method was dangerous and confusing.

jnievelt:
Yeah +1 I've been bitten by this myself

johnynek:
@isnotinvain want to do the honors?

tglstory:
:joy: happy that this is helpful!

isnotinvain:
Sorry, was out of town. +1

PR #1534, opened 2016-03-04
Because, because... fun, the scala compiler has special naming rules …

### LINKED ISSUE #1528: bug in thrift ordered serialization macros
Got an error:

```
 found   : Option[Boolean]
 required: Option[String]
```

in generated code. Can't see the code in a compiled binary, so not sure what the deal is. Just don't want to forget.

I think we need more tests here to call this a success.

ianoc:
…it appears when there are leading underscores

fixes #1528

/cc @johnynek

### DIFF
--- a/scalding-thrift-macros-fixtures/src/test/resources/test.thrift
+++ b/scalding-thrift-macros-fixtures/src/test/resources/test.thrift
@@ -78,11 +78,28 @@ struct TestTypes{
  9: required binary a_binary
 }
 
+struct TestOptionTypes{
+ 1: optional bool a_bool,
+ 2: optional byte a_byte,
+ 3: optional i16 a_i16,
+ 4: optional i32 a_i32,
+ 5: optional i64 a_i64,
+ 6: optional double a_double,
+ 7: optional string a_string,
+ 8: optional TestEnum a_enum,
+ 9: optional binary a_binary
+}
+
 union TestUnion{
  1:  list<i32> a_i32_list,
  2:  set<double> a_double_set,
  3:  TestStruct a_struct
 }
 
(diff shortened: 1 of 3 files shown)


johnynek:
Nice catch. Why the toLowerCase?

ianoc:
We do this to try work out the methods which are the field names for the trait, since they aren't case class accessor, we sort of have a guessing game.

The reason for the toLowerCase is because the compiler does capitalization of the first letter of the field name, except if the leading characters are underscores, then its the first non underscore character it seems.

So to just try sidestep the whole mess I went with the toLowerCase, the risk there becomes i guess of a false positive, but I couldn't manage to figure out how to generate something where it doesn't work that is valid thrift.

johnynek:
Can we add an example with a collision in the upper and lower casing? a and A as fields? Is that allowed?

johnynek:
Could we just call the Scrooge method that is doing this mangling? We already depend on Scrooge, right?

ianoc:
well sorry i mean i you can do it if you put long enough field names to avoid the auto-casing. But if you add them as fields, then they are actually fields... so they will be there anyway. They are just there twice.

ianoc:
Its different mangling, this is the scala compiler mangling, not thrift -> scrooge

johnynek:
Ahh. I see.

Well, this is an improvement for sure. Merge when green.

On Thursday, March 3, 2016, ianoc notifications@github.com wrote:

> Its different mangling, this is the scala compiler mangling, not thrift ->
> scrooge
> 
> —
> https://github.com/twitter/scalding/pull/1534#issuecomment-192119618.

## 

PR #1535, opened 2016-03-05
Scala implicit ordering fun -- have 2 forms of sketch, defaulting to …

ianoc:
…the existing

@johnynek  as we were discussing for the sketch method. More of a consideration than hard PR

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -38,7 +38,7 @@ import scala.util.Try
  * This object is here rather than in the typed package because a lot of code was written using
  * the functions in the object, which we do not see how to hide with package object tricks.
  */
-object TypedPipe extends Serializable with LowerPriorityTypedPipeImplicits {
+object TypedPipe extends Serializable {
   import Dsl.flowDefToRichFlowDef
 
   /**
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -119,6 +119,67 @@ object TypedPipe extends Serializable {
       override def hash(x: Int): Int = x
     }
   }
+  implicit class HigherPriorityTypedPipeMethods[T, K, V](val tp: TypedPipe[T])(implicit ev: TypedPipe[T] <:< TypedPipe[(K, V)],
```
ianoc: Can only enter an implicit class if you meet all of its criteria

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -119,6 +119,67 @@ object TypedPipe extends Serializable {
...
+   * to shard the heavy keys across many reducers.
+   * This increases communication cost in order to reduce the maximum time needed
+   * to complete the join.
+   *
+   * {@code pipe.sketch(100).join(thatPipe) }
+   * will add an extra map/reduce job over a standard join to create the count-min-sketch.
+   * This will generally only be beneficial if you have really heavy skew, where without
+   * this you have 1 or 2 reducers taking hours longer than the rest.
+   */
+  def sketch[K, V](reducers: Int,
```
ianoc: Once you do, your bound to its method, so the lower priority TypedPipe only being the requirement here means this one is always available. So the errors remain the same if users don't have the implicits for this method for instance

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: can we drop `T` entirely here and just do `(K, V)`?

ianoc: Looks like the tests here pass with that yep

johnynek: can we just do `SketchJoinMethods` similar to https://github.com/twitter/scalding/blob/ianoc/addSketchOrderedSerializationInterface/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala#L92 so we can fix `(K, V)` and simpify a bit

johnynek:
looks good to me modulo the naming (I would use names for classes more specific to SketchJoin)

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
ianoc: We can put the classes somewhere else, but I think to keep the same sort of source feel/compatibility we need this low priority implicit stuff to fall back to the old one. If we would be happy to make it disappear in the event of no matching signature i think we can just do 2 implicit methods in the object pointing at 2 classes (in sketched.scala) containing the sketch method of the 2 signatures.

johnynek:
:+1: I like this. You can use the OrderedSerialization macros, not have an implicit conversion in scope _and_ do a more efficient join at the same time. Seems like a win.

johnynek:
what do you think of this @rubanm @isnotinvain ? I like not having to set up an implicit conversion and the ability to use compiler generated serializers.

scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
```diff
@@ -1367,6 +1386,46 @@ class TypedSketchLeftJoinJob(args: Args) extends Job(args) {
...
+    .write(TypedText.tsv[(Int, Int, Int)]("output-sketch"))
+
+  zero
+    .group
+    .leftJoin(one.group)
+    .map{ case (k, (v0, v1)) => (k, v0, v1.getOrElse(-1)) }
+    .write(TypedText.tsv[(Int, Int, Int)]("output-join"))
+}
+
+class BothAvailableSketchMethodsTypedSketchLeftJoinJob(args: Args) extends Job(args) {
```
rubanm, line 1408: This job looks unused in the test specs below.

rubanm:
@johnynek @ianoc  Sorry late to this. Just one unit test comment. LGTM!

PR #1537, opened 2016-03-21
Fixes Config to accommodate spaces in argument values

ulyssepence:
I was hitting some macro issues with `OrderedSerialization` when trying to generate a `Serialization[Map[String, List[String]]]`. I think they may have been related to @Gabriel439's recent wart remover work because the only error message I could see was related to using `Option.fold` instead of `Option.get`.

Anyway, I ended up using chill for the serialization.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -198,17 +198,19 @@ trait Config extends Serializable {
     if (toMap.contains(ConfiguredInstantiator.KEY)) Some((new ConfiguredInstantiator(ScalaMapConfig(toMap))).getDelegate)
     else None
 
-  def getArgs: Args = get(Config.ScaldingJobArgs) match {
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -200,11 +200,15 @@ trait Config extends Serializable {
...
+    case Some(str) => argsSerializer
+      .invert(str)
+      .map(new Args(_))
+      .getOrElse(throw new RuntimeException(
+        s"""Could not deserialize Args from Config. Maybe "$ScaldingJobArgs" was modified without using Config.setArgs?"""))
   }
 
   def setArgs(args: Args): Config =
-    this + (Config.ScaldingJobArgs -> args.toString)
+    this + (Config.ScaldingJobArgs -> argsSerializer(args.m))
```
johnynek: can we put this in a new key and keep the old one for people that might what to read those keys?

It it nice for human debugging and cheap to add a new key.

Issue #1539, opened 2016-03-23
FileSource InvalidSourceTap Race Condition


ulyssepence:
[`FileSoure.createHdfsReadTap`](https://github.com/twitter/scalding/blob/7759d9c399896fc4c54be0f2406e047e030d8586/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala#L291), when strict mode is `false` and there are no good paths, will create an `InvalidSourceTap`. [`FileSource.validateTaps`](https://github.com/twitter/scalding/blob/7759d9c399896fc4c54be0f2406e047e030d8586/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala#L251), if a good path has appeared since `.createHdfsReadTap`, will not throw. This sounds like a problem, albeit one without using strict mode.

jnievelt:
Yeah here's a small repro, which can be run in non-strict Hdfs mode.  This overrides validateTaps to simulate the case that one of the paths became usable.

The effect is amplified for Twitter because our internal source definitions can trigger this in strict mode as well.

``` scala
import com.twitter.scalding.source._
val input = new FixedTypedText[String](TypedText.TAB, "foo_bad_path") { override def validateTaps(mode: Mode): Unit = {} }
TypedPipe.from(input).write(NullSink)
```

```
Mar 24, 2016 2:17:53 AM cascading.flow.hadoop.util.HadoopUtil findMainClass
INFO: resolving application jar from found main method on: com.twitter.scalding.BaseScaldingShell$class
Mar 24, 2016 2:17:53 AM cascading.flow.hadoop.planner.HadoopPlanner initialize
INFO: using application jar: /opt/scalding/309/libs/624.jar
Mar 24, 2016 2:17:53 AM cascading.property.AppProps getAppID
INFO: using app.id: 29548BACD55B4ECC8AD015FA158A8AAF
Mar 24, 2016 2:17:53 AM cascading.util.Version printBanner
INFO: Concurrent, Inc - Cascading 2.5.6-t-7
Mar 24, 2016 2:17:53 AM cascading.flow.BaseFlow logInfo
INFO: [ScaldingShell] starting
Mar 24, 2016 2:17:53 AM cascading.flow.BaseFlow logInfo
INFO: [ScaldingShell]  source: InvalidSourceTap["NullScheme"]["InvalidSourceTap-a47ddbbe-d92f-4513-b5a3-334835f4190c"]
[...]

PR #1540, opened 2016-03-28
Adds late tap validation for cases where race conditions cause it to fail

ulyssepence:
Better failure message for #1539. If we write a `MultiInputFormat` via `InvalidSourceTap` and that `InvalidSourceTap` gets by tap validation, we won't ever get to `.openForRead(...)` to fail because `MultiInputFormat.getSplits` will lead to a stack overflow or OOM, depending on your situation. See the issue.

This fixes `InvalidSourceTap` to blow up instead while the job is running.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
@@ -87,10 +87,10 @@ class InvalidSourceTap(val hdfsPaths: Iterable[String]) extends SourceTap[JobCon
  * Better error messaging for the occassion where an InvalidSourceTap does not
  * fail in validation.
  */
-class InvalidInputFormat extends InputFormat[Unit, Unit] {
-  override def getSplits(conf: JobConf, i: Int): Array[InputSplit] =
+class InvalidInputFormat extends InputFormat[Nothing, Nothing] {
+  override def getSplits(conf: JobConf, numSplits: Int): Nothing =
     throw new InvalidSourceException("getSplits called on InvalidInputFormat")
-  override def getRecordReader(split: InputSplit, conf: JobConf, reporter: org.apache.hadoop.mapred.Reporter): RecordReader[Unit, Unit] =
+  override def getRecordReader(split: InputSplit, conf: JobConf, reporter: org.apache.hadoop.mapred.Reporter): Nothing =
     throw new InvalidSourceException("getRecordReader called on InvalidInputFormat")
 }
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
@@ -87,7 +87,7 @@ class InvalidSourceTap(val hdfsPaths: Iterable[String]) extends SourceTap[JobCon
  * Better error messaging for the occassion where an InvalidSourceTap does not
  * fail in validation.
  */
-class InvalidInputFormat extends InputFormat[Nothing, Nothing] {
+private[scalding] class InvalidInputFormat extends InputFormat[Nothing, Nothing] {
   override def getSplits(conf: JobConf, numSplits: Int): Nothing =
     throw new InvalidSourceException("getSplits called on InvalidInputFormat")
   override def getRecordReader(split: InputSplit, conf: JobConf, reporter: org.apache.hadoop.mapred.Reporter): Nothing =
--- a/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
@@ -28,6 +28,8 @@ import cascading.tuple.{ Fields, Tuple => CTuple, TupleEntry, TupleEntryCollecto
 
 import cascading.pipe.Pipe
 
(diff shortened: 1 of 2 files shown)


scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -702,4 +723,21 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest
...
+  "An InvalidSourceTap that gets past validation" should {
+    "throw an InvalidSourceException" in {
+      val result = Try {
+        HadoopPlatformJobTest(new ReadPathJob(_), cluster)
+          .arg("input", "/sploop/boop/doopity/doo/")
+          .run
+      }
+
+      result match {
+        case Failure(fe: FlowException) if Option(fe.getCause).forall(_.isInstanceOf[InvalidSourceException]) => ()
```
rubanm: .forall passes if the Option is empty. Should this be .exists?

scalding-core/src/main/scala/com/twitter/scalding/Source.scala
```diff
@@ -76,11 +78,22 @@ class InvalidSourceTap(val hdfsPaths: Iterable[String]) extends SourceTap[JobCon
...
+    conf.setInputFormat(classOf[InvalidInputFormat])
     super.sourceConfInit(flow, conf)
   }
 }
 
+/**
+ * Better error messaging for the occassion where an InvalidSourceTap does not
+ * fail in validation.
+ */
+class InvalidInputFormat extends InputFormat[Nothing, Nothing] {
```
rubanm: Maybe make this private [scalding] ?

rubanm:
LGTM otherwise

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
ulyssepence: Oh wow good point. Will change.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -702,4 +723,21 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest
     }
   }
 
+  "An InvalidSourceTap that gets past validation" should {
+    "throw an InvalidSourceException" in {
+      val result = Try {
```
isnotinvain: you can use the intercept helper here, eg:

```
val e = intercept[FlowException] {  
  HadoopPlatformJobTest....run
}
```

johnynek: do you want this change, or just a note? To me, using the standard Try here is not so bad, and requires less scalatest knowledge.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -702,4 +723,21 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest
...
+        HadoopPlatformJobTest(new ReadPathJob(_), cluster)
+          .arg("input", "/sploop/boop/doopity/doo/")
+          .run
+      }
+
+      result match {
+        case Failure(fe: FlowException) if Option(fe.getCause).exists(_.isInstanceOf[InvalidSourceException]) => ()
```
johnynek: can we have an assert in here? otherwise scalatest won't count it in the test count.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
isnotinvain: well, it changes it from:

```
val result = Try {
  HadoopPlatformJobTest(new ReadPathJob(_), cluster)
    .arg("input", "/sploop/boop/doopity/doo/")
    .run
}

result match {
  case Failure(fe: FlowException) if Option(fe.getCause).exists(_.isInstanceOf[InvalidSourceException]) => ()
  case Failure(t) =>
    throw new RuntimeException("Expected InvalidSourceException wrapped in FlowException.", t)
  case _ =>
    throw new RuntimeException("Expected InvalidSourceException wrapped in FlowException, but the job succeeded")
}
```

to:

```
val result: FlowException = intercept[FlowException]
  HadoopPlatformJobTest(new ReadPathJob(_), cluster)
    .arg("input", "/sploop/boop/doopity/doo/")
    .run
}

assert(Option(result.getCause).exists(_.isInstanceOf[InvalidSourceException])
```

So, it saves you the case match and all the exception throwing and so on.
The intercept helper is itself an assertion that an exception should happen, so that's handled for you. I think it's fairly common to use, just as the `@Test(expected=FooException.class` junit annotation. Though intercept is much better because you get to inspect the exception.

ulyssepence:
I don't really understand how adding an import was a merge conflict...

johnynek:
:shipit:

Issue #1541, opened 2016-03-31
Rounding Bug in RatioBasedEstimator


isnotinvain:
Currently the formula we use is:

```
val baseEstimate: Int = delegateToInputSizeEstimator(...)
val e = (baseEstimate * reducerRatio).ceil.toInt max 1
```

However, if the baseEstimate was initially tiny (say we need about 0.01 reducers) and then that gets rounded up to 1, and now our reducerRatio is rather large, when multiplied together we get something large. Instead of `.01 * reducerRatio` we have `ceil(.01) * reducerRatio`

So I think the fix is here:
https://github.com/twitter/scalding/blob/bb1c4d2d454eec85d65dd4d27d5399e17cb7c619/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RatioBasedEstimator.scala#L72

We need to add a static helper method to InputSizeReducerEstimator that returns a Double instead of an Int and call that instead of using the already `ceil`ed estimate.

johnynek:
So the issue is that `Int` does not compose well due to this `ceil` effect. I guess using `Double` is a fine way to signal when you need much less that 1 reducer.

Seems fine to me, but is obviously binary incompatible.

isnotinvain:
I don't think it's binary incompatible. Fixing the trat to use Doubles would be. But we can give InputSizeReducerEstimator a method that returns double in addition to the one that returns int w/o breaking compatibility. PR on the way

gerashegalov:
@isnotinvain Are you also adding some reasonable cap? I would make it configurable with a default 5000.

isnotinvain:
Yeah we could add a cap too.
This isn't done, but just to illustrate:
https://github.com/twitter/scalding/pull/1542

PR #1542, opened 2016-03-31
Fix Rounding Bug in RatioBasedEstimator

isnotinvain:
Fixes the rounding bug mentioned in #1541, and adds a maximum for reducer estimation as well.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
@@ -191,14 +191,15 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
         .map(clsLoader.loadClass(_).newInstance.asInstanceOf[ReducerEstimator])
       val combinedEstimator = Monoid.sum(estimators)
 
-      // try to make estimate
       val info = FlowStrategyInfo(flow, preds.asScala, step)
 
-      // if still None, make it '-1' to make it simpler to log
+      // get estimate
       val estimatedNumReducers = combinedEstimator.estimateReducers(info)
 
+      // apply cap if needed
       val cappedNumReducers = estimatedNumReducers.map { n =>
         val configuredMax = conf.getInt(EstimatorConfig.maxEstimatedReducersKey, EstimatorConfig.defaultMaxEstimatedReducers)
+
         if (n > configuredMax) {
           LOG.warn(
             s"""
@@ -206,10 +207,11 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
                |Will use $configuredMax instead.
              """.stripMargin)
         }
+
         n.min(configuredMax)
       }
 
-      // save the estimate in the JobConf which should be saved by hRaven
+      // save the estimate and capped estimate in the JobConf which should be saved by hRaven
       conf.setInt(EstimatorConfig.estimatedNumReducers, estimatedNumReducers.getOrElse(-1))
       conf.setInt(EstimatorConfig.cappedEstimatedNumReducersKey, cappedNumReducers.getOrElse(-1))
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
@@ -23,6 +23,12 @@ object EstimatorConfig {
   /** Maximum number of history items to use for reducer estimation. */
   val maxHistoryKey = "scalding.reducer.estimator.max.history"
 
+  /** If we estimate more than this number of reducers, consider it a fatal error */
+  val maxEstimatedReducersKey = "scalding.reducer.estimator.max.estimated.reducers"
+
+  // TODO: what's a reasonable default? Int.maxValue? 5k? 100k?
+  val defaultMaxEstimatedReducers = 100 * 1000
+
   def getMaxHistory(conf: JobConf): Int = conf.getInt(maxHistoryKey, 1)
 
 }
@@ -167,7 +173,7 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
     // whether we should override explicitly-specified numReducers
     val overrideExplicit = conf.getBoolean(Config.ReducerEstimatorOverride, false)
 
-    Option(conf.get(Config.ReducerEstimators)).map { clsNames =>
+    Option(conf.get(Config.ReducerEstimators)).foreach { clsNames =>
 
       val clsLoader = Thread.currentThread.getContextClassLoader
 
@@ -180,6 +186,13 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
 
       // if still None, make it '-1' to make it simpler to log
       val numReducers = combinedEstimator.estimateReducers(info)
+
+      numReducers.foreach { n =>
+        val configuredMax = conf.getInt(EstimatorConfig.maxEstimatedReducersKey, EstimatorConfig.defaultMaxEstimatedReducers)
+        if (n > configuredMax) {
+          throw new IllegalArgumentException(s"Reducer estimator estimated $n reducers, which is more than the configured maximum of $configuredMax")
+        }
+      }
 
       // save the estimate in the JobConf which should be saved by hRaven
       conf.setInt(EstimatorConfig.estimatedNumReducers, numReducers.getOrElse(-1))
--- a/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
@@ -17,17 +17,27 @@ object EstimatorConfig {
   /** Output param: what the Reducer Estimator recommended, regardless of if it was used. */
   val estimatedNumReducers = "scalding.reducer.estimator.result"
 
+  /**
+   * Output param: same as estimatedNumReducers but with the cap specified by maxEstimatedReducersKey
+   * applied. Can be used to determine whether a cap was applied to the estimated number of reducers
+   * and potentially to trigger alerting / logging.
+   */
+  val cappedEstimatedNumReducersKey = "scalding.reducer.estimator.result.capped"
+
   /** Output param: what the original job config was. */
   val originalNumReducers = "scalding.reducer.estimator.original.mapred.reduce.tasks"
 
   /** Maximum number of history items to use for reducer estimation. */
   val maxHistoryKey = "scalding.reducer.estimator.max.history"
 
-  /** If we estimate more than this number of reducers, consider it a fatal error */
+  /**
+   * If we estimate more than this number of reducers,
+   * we will use this number instead of the estimated value
+   */
   val maxEstimatedReducersKey = "scalding.reducer.estimator.max.estimated.reducers"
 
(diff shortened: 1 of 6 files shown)


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -18,22 +25,15 @@ object InputSizeReducerEstimator {
...
-   * Figure out the total size of the input to the current step and set the number
-   * of reducers using the "bytesPerReducer" configuration parameter.
+   * Same as estimateReducers, except doesn't round or ceil the result.
+   * This is useful for composing with other estimation strategies that
+   * don't want to lose the fractional number of reducers. Especially
+   * helpful for when less than 1 reducer is needed, but this fraction
+   * will be multiplied by a scaling factor later.
    */
-  override def estimateReducers(info: FlowStrategyInfo): Option[Int] =
+  def estimateExactReducers(info: FlowStrategyInfo): Option[Double] = {
```
isnotinvain: probably a bad name

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -59,4 +59,15 @@ class InputSizeReducerEstimator extends ReducerEstimator {
 
         Some(nReducers)
     }
+  }
+
 }
+
+// TODO: can we delete this? Just make the object above extend ReducerEstimator?
```
johnynek: we use reflection to create these. Why not just leave it as is? Sometimes people get confused with reflection and objects because their name (due to mangling) is not always clear.

johnynek:
Can you have a test that shows this issue and that this fixes it?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
isnotinvain: yeah that's fine.

isnotinvain:
@johnynek yes:

> Needs tests, and some discussion about a hard cap as well

isnotinvain:
Running the new test on the develop branch yields:

```
[info] - should handle mapper output explosion over small data correctly *** FAILED ***
[info]   1000 did not equal 2 (RatioBasedEstimatorTest.scala:216)
```

isnotinvain:
I'm adding a cap as well @gerashegalov

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -23,6 +23,12 @@ object EstimatorConfig {
   /** Maximum number of history items to use for reducer estimation. */
   val maxHistoryKey = "scalding.reducer.estimator.max.history"
 
+  /** If we estimate more than this number of reducers, consider it a fatal error */
+  val maxEstimatedReducersKey = "scalding.reducer.estimator.max.estimated.reducers"
+
+  // TODO: what's a reasonable default? Int.maxValue? 5k? 100k?
+  val defaultMaxEstimatedReducers = 100 * 1000
```
sriramkrishnan: 100K seems way too many, no? Pig's max is 999.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -181,6 +187,13 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
       // if still None, make it '-1' to make it simpler to log
       val numReducers = combinedEstimator.estimateReducers(info)
 
+      numReducers.foreach { n =>
+        val configuredMax = conf.getInt(EstimatorConfig.maxEstimatedReducersKey, EstimatorConfig.defaultMaxEstimatedReducers)
+        if (n > configuredMax) {
+          throw new IllegalArgumentException(s"Reducer estimator estimated $n reducers, which is more than the configured maximum of $configuredMax")
```
sriramkrishnan: Do you want to throw here or just cap it at the max? I would prefer to cap it to the max, or at least add a flag to change that behavior.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
isnotinvain: Yeah I have no idea. I guess I can spot check some jobs in our hadoop cluster.
But if you can have 30k mappers, why not 30k reducers? and if 30k is somewhat normal, then 100k isn't that far off in terms of being "way too much"

I don't know whether pig's max of 999 was chosen with much thought or what size of cluster it was chosen for.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
isnotinvain: We had talked about this internally I think. Pig just caps at 999. The result is that almost all jobs ran with 999. The job that stumbled across this bug and got us to think about fixing things was found sort of by accident, if it had thrown an exception we would have found out sooner that there was a bug in the reducer estimator. It would be easy to just apply the cap, but generally I prefer to fail fast when expectations are not met.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.scalding.platform
 
 import cascading.pipe.joiner.{ JoinerClosure, InnerJoin }
-import cascading.tuple.{Fields, Tuple}
+import cascading.tuple.{ Fields, Tuple }
```
isnotinvain, line 19: you can ignore this file, it got hit by the auto-formatter.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
sriramkrishnan: One reason I can think of - too many reducers equals too many files. 

Yeah, I don't know the historic reason for 999 reducer. I think Hive has the same too.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
sriramkrishnan: Yeah, although it is not ideal behavior from the standpoint of a user. So if there is a bug, it ends up paging the user, not the developers of the reducer estimator. In this case, failing fast would be a bad user experience IMO.

vrushalivc: Yes, let's not throw an exception to the user.

I don't think all kinds of cluster sizes can converge to one suitable default value for max number of reducers. As such I believe, we should have max number of reducers <= max number of mappers, presuming mappers are not exploding data.  For hadoop2 clusters, the config param mapreduce.jobtracker.split.metainfo.maxsize controls the number of mappers that can be launched. 
Not sure if it's possible in this code, but could it be possible to arrive at a suitable max value based on cluster config setting? For example, 1MB of mapreduce.jobtracker.split.metainfo.maxsize translates to roughly 10K map tasks. So we should set the number of reducers <= number of mappers (presuming mappers are not exploding input data). 
cc @lohitvijayarenu cc @gerashegalov

gerashegalov: I too prefer not to fail jobs. As for max, I suggest theoretical max mappers / 5. We could set  it 10000 for clusters with 50K-split jobs. Anyways I am not worried too much of the default in the code here because I expect we'll configure it explicitly for important pipelines.

The layering of estimator in Scalding is not as ideal. If we deferred to mapreduce JobSubmitter we could run it after getSplits and actually know the number of mappers. But this discussion is beyond the scope of this issue.

vrushalivc: Thanks Gera! In the context of layering of estimator, it is done at a higher level like Scalding or Pig since they can query for the historical runs and thereby better estimate the number of reducers for this run, which as such is right on the money. But we also want to take care of these corner cases where we see an explosive number of reducers being launched. Hence this PR.

isnotinvain:
I can make the max non-fatal, though I'm not entirely convinced.

Having the max act as a cap means that it's another tuning dimension that users need to worry about, that might be causing them to have an unexpected number of reducers (if the cap is set too low).

On the other hand, if we set the cap relatively high, as in a value where we think "anything that estimates needing this is clearly broken" and treat it as exceptional, this becomes a faster feedback mechanism for users. I actually think that in many cases fail-fast is a much better user experience than "just try to make it work and hope". The grey area here is whether this is one of those cases or not. If the estimator estimates needing a ton of reducers, they are either needed, and capping will probably not be great, or the estimator is broken.

That's my pitch, but I don't feel too strongly if everyone else want the cap to not be exceptional.

johnynek:
thinking as a user, I think I would not want scalding to (at least by default) fail to run if it uses too many reducers.

If I set N as the max, and it runs slow, I should have alerts that I notice, and I can rewrite my job or turn up the max. If it runs too fast because of a bug (and I waste resources) I should have a system to watch that too, and if not, set a lower max so the waste is not _that_ bad.

I have a hard time imagining making the default to fail the job.

I can see adding an option to fail if we exceed the max (since we expect a bug in that case, or a totally giant input), and some users may want it, but I would err on the side of running the job and not changing default behavior.

isnotinvain:
I'll update to make it non-fatal. I will also add a property to the hadoop config so that tools that monitor hadoop config properties can warn users that the max has been applied.

isnotinvain:
OK, this should be good to go. LMK if 5k is not a good default for the cap

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -179,14 +195,27 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
       val info = FlowStrategyInfo(flow, preds.asScala, step)
 
       // if still None, make it '-1' to make it simpler to log
```
sriramkrishnan: Is this comment still valid? Move it down maybe before you log it?

isnotinvain: I guess this could be moved to below where it become -1

sriramkrishnan:
LGTM - may want @rubanm take a look at it too.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -175,18 +191,33 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
...
+
+        if (n > configuredMax) {
+          LOG.warn(
+            s"""
+               |Reducer estimator estimated $n reducers, which is more than the configured maximum of $configuredMax.
+               |Will use $configuredMax instead.
+             """.stripMargin)
+        }
+
+        n.min(configuredMax)
```
rubanm: This could go in an if-else block above?

rubanm:
One minor comment, LGTM!

isnotinvain:
I will merge once the tests run

PR #1545, opened 2016-04-15
Make types of localPath & hdfsPaths explicit

piyushnarang:
While merging Scalding release 0.16.0-RC6 and trying to hook up FixedPathSource to our internal sources, noticed that the types of localPaths & hdfsPaths seemed to default to `List[String]` instead of `Iterable[String]` (which is what it is in FileSource). Making it explicit to make it consistent.


rubanm:
+1 While you're at it, could you also set the type on `def hdfsWritePath` and `def localWritePath`?

piyushnarang:
@rubanm - done

ulyssepence:
+2

johnynek:
Nice!

PR #1546, opened 2016-04-19
Update the WordCountJob example to use TypedText.tsv

joshualande:
I feel that TypedTsv is now legacy and TypedText.tsv is the way to go. The original code works, but might be cargo-culted into a situation where it is doesn't work correctly, so it would be better to use TypedText.tsv in the example.


rubanm:
Thanks!

joshualande:
@rubanm thanks so much!!!!

PR #1547, opened 2016-04-22
Trampoline Execution runs

### LINKED ISSUE #1544: Avoid nested/recursive computations in Execution.sequence
Today, calling Execution.sequence results in a linear chain of Zipped and Mapped Executions.  Due to the case class nesting, computations on these are also nested.  In particular, even calling hashCode to use the EvalCache can cause a stack overflow.

It may be better to just have a directly implemented SequenceExecution, which stores a sequence directly, and does iterative computations instead of nested ones.

jnievelt:
Resolves #1544

Two essential parts to this.  First, we eagerly compute and save the `hashCode`.  This means we'll compute as we go, instead of on-demand at the end.

Second, `runStats` returns a `Trampoline` of its previous return type.  This should be safe since a) the base `Execution#runStats` is protected and b) the concrete `Execution` types which expose public `runStats` are private classes.  I've also made these methods protected across the board as part of this change.

Note that there is a little bit of a different style of computation as a result.  For example, with `Mapped`, we need to trampoline through the previous `Execution` before we're ready to call into `getOrElseInsert`.  As it stands, a `run` through a large `Execution` graph will always traverse the immediate graph and generate new `Trampoline`s in preparation for computing the top-level `Future` (though it will hit the cache before actually computing it).

I've also written up a version that also caches the `Trampoline`s themselves, but it's not needed for functional correctness, and I didn't think it was necessarily justified from a performance standpoint.

Finally, I'm moving `ExecutionTest` into the `scalding` package, to match `Execution` itself.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -15,7 +15,7 @@ limitations under the License.
  */
 package com.twitter.scalding
 
-import com.twitter.algebird.monad.Reader
+import com.twitter.algebird.monad.{ Reader, Trampoline }
 import com.twitter.algebird.{ Monoid, Monad, Semigroup }
 import com.twitter.scalding.cascading_interop.FlowListenerPromise
 import com.twitter.scalding.Dsl.flowDefToRichFlowDef
@@ -25,6 +25,7 @@ import scala.util.{ Failure, Success, Try }
 import scala.util.control.NonFatal
 import cascading.flow.{ FlowDef, Flow }
 import scala.collection.mutable
+import scala.runtime.ScalaRunTime
 
 /**
  * Execution[T] represents and computation that can be run and
@@ -45,7 +46,7 @@ import scala.collection.mutable
  * in some libraries) composes two Executions is parallel. Prefer
  * zip to flatMap if you want to run two Executions in parallel.
  */
-sealed trait Execution[+T] extends java.io.Serializable {
+sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
   import Execution.{ EvalCache, FlatMapped, GetCounters, ResetCounters, Mapped, OnComplete, RecoverWith, Zipped }
 
   /**
@@ -144,7 +145,8 @@ sealed trait Execution[+T] extends java.io.Serializable {
   final def run(conf: Config, mode: Mode)(implicit cec: ConcurrentExecutionContext): Future[T] = {
     val ec = new EvalCache
     val confWithId = conf.setScaldingExecutionId(java.util.UUID.randomUUID.toString)
-    val result = runStats(confWithId, mode, ec)(cec).map(_._1)
+    // get on Trampoline
+    val result = runStats(confWithId, mode, ec)(cec).get.map(_._1)
     // When the final future in complete we stop the submit thread
     result.onComplete { _ => ec.finished() }
     // wait till the end to start the thread in case the above throws
@@ -161,7 +163,7 @@ sealed trait Execution[+T] extends java.io.Serializable {
    */
   protected def runStats(conf: Config,
     mode: Mode,
-    cache: EvalCache)(implicit cec: ConcurrentExecutionContext): Future[(T, ExecutionCounters)]
+    cache: EvalCache)(implicit cec: ConcurrentExecutionContext): Trampoline[Future[(T, ExecutionCounters)]]
 
   /**
    * This is convenience for when we don't care about the result.
@@ -192,6 +194,10 @@ sealed trait Execution[+T] extends java.io.Serializable {
    */
   def zip[U](that: Execution[U]): Execution[(T, U)] =
     Zipped(this, that)
+
+  override val hashCode: Int = ScalaRunTime._hashCode(self)
+  override def equals(other: Any): Boolean =
+    hashCode == other.hashCode && ScalaRunTime._equals(self, other)
 }
 
 /**
@@ -381,49 +387,56 @@ object Execution {
       res: => Future[(T, ExecutionCounters)]): Future[(T, ExecutionCounters)] =
       getOrElseInsertWithFeedback(cfg, ex, res)._2
   }
+
   private case class FutureConst[T](get: ConcurrentExecutionContext => Future[T]) extends Execution[T] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      cache.getOrElseInsert(conf, this,
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
+      Trampoline(cache.getOrElseInsert(conf, this,
         for {
           futt <- toFuture(Try(get(cec)))
           t <- futt
-        } yield (t, ExecutionCounters.empty))
+        } yield (t, ExecutionCounters.empty)))
 
     // Note that unit is not optimized away, since Futures are often used with side-effects, so,
     // we ensure that get is always called in contrast to Mapped, which assumes that fn is pure.
   }
   private case class FlatMapped[S, T](prev: Execution[S], fn: S => Execution[T]) extends Execution[T] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      cache.getOrElseInsert(conf, this,
-        for {
-          (s, st1) <- prev.runStats(conf, mode, cache)
-          next = fn(s)
-          fut2 = next.runStats(conf, mode, cache)
-          (t, st2) <- fut2
-        } yield (t, Monoid.plus(st1, st2)))
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { fut1 =>
+        cache.getOrElseInsert(conf, this,
+          for {
+            (s, st1) <- fut1
+            next = fn(s)
+            (t, st2) <- Trampoline.call(next.runStats(conf, mode, cache)).get
+          } yield (t, Monoid.plus(st1, st2)))
+      }
   }
 
   private case class Mapped[S, T](prev: Execution[S], fn: S => T) extends Execution[T] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      cache.getOrElseInsert(conf, this,
-        prev.runStats(conf, mode, cache)
-          .map { case (s, stats) => (fn(s), stats) })
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { fut =>
+        cache.getOrElseInsert(conf, this,
+          fut.map { case (s, stats) => (fn(s), stats) })
+      }
   }
   private case class GetCounters[T](prev: Execution[T]) extends Execution[(T, ExecutionCounters)] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      cache.getOrElseInsert(conf, this,
-        prev.runStats(conf, mode, cache).map { case tc @ (t, c) => (tc, c) })
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { fut =>
+        cache.getOrElseInsert(conf, this,
+          fut.map { case tc @ (t, c) => (tc, c) })
+      }
   }
   private case class ResetCounters[T](prev: Execution[T]) extends Execution[T] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      cache.getOrElseInsert(conf, this,
-        prev.runStats(conf, mode, cache).map { case (t, _) => (t, ExecutionCounters.empty) })
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { fut =>
+        cache.getOrElseInsert(conf, this,
+          fut.map { case (t, _) => (t, ExecutionCounters.empty) })
+      }
   }
 
   private case class TransformedConfig[T](prev: Execution[T], fn: Config => Config) extends Execution[T] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
       val mutatedConfig = fn(conf)
-      cache.getOrElseInsert(mutatedConfig, this, prev.runStats(mutatedConfig, mode, cache))
+      Trampoline.call(prev.runStats(mutatedConfig, mode, cache))
     }
   }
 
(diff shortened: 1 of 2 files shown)


scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
```diff
@@ -417,5 +415,15 @@ class ExecutionTest extends WordSpec with Matchers {
 
       assert(result.waitFor(Config.default, Local(true)).get == 0.to(10).reverse)
     }
+
+    "can hashCode, compare, and run a long sequence" in {
+      val execution = Execution.sequence((1 to 100000).toList.map(Execution.from(_)))
```
johnynek, line 419: did this fail without this patch?

jnievelt, line 419: Yeah the breaking point varied (and could change within the life of a process), but hits rather consistently.

Against clean develop, all actually die on `hashCode`:

```
[info] Exception encountered when attempting to run a suite with class name: com.twitter.scalding.ExecutionTest *** ABORTED ***
[info]   java.lang.StackOverflowError:
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at com.twitter.scalding.Execution$Zipped.hashCode(Execution.scala:547)
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3.productHash(MurmurHash3.scala:64)
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at com.twitter.scalding.Execution$Mapped.hashCode(Execution.scala:406)
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3.productHash(MurmurHash3.scala:64)
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at com.twitter.scalding.Execution$Zipped.hashCode(Execution.scala:547)
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3.productHash(MurmurHash3.scala:64)
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at com.twitter.scalding.Execution$Mapped.hashCode(Execution.scala:406)
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3.productHash(MurmurHash3.scala:64)
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at com.twitter.scalding.Execution$Zipped.hashCode(Execution.scala:547)
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3.productHash(MurmurHash3.scala:64)
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at com.twitter.scalding.Execution$Mapped.hashCode(Execution.scala:406)
...
```

If we patch just that part, we see this on the `run`:

```
[info] Exception encountered when attempting to run a suite with class name: com.twitter.scalding.ExecutionTest *** ABORTED ***
[info]   java.lang.StackOverflowError:
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3$$anonfun$unorderedHash$1.apply(MurmurHash3.scala:92)
[info]   at scala.util.hashing.MurmurHash3$$anonfun$unorderedHash$1.apply(MurmurHash3.scala:91)
[info]   at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:221)
[info]   at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:428)
[info]   at scala.util.hashing.MurmurHash3.unorderedHash(MurmurHash3.scala:91)
[info]   at scala.util.hashing.MurmurHash3$.mapHash(MurmurHash3.scala:222)
[info]   at scala.collection.GenMapLike$class.hashCode(GenMapLike.scala:35)
[info]   at scala.collection.AbstractMap.hashCode(Map.scala:59)
[info]   at com.twitter.scalding.Config$class.hashCode(Config.scala:376)
[info]   at com.twitter.scalding.Config$$anon$1.hashCode(Config.scala:458)
[info]   at scala.runtime.ScalaRunTime$.hash(ScalaRunTime.scala:206)
[info]   at scala.util.hashing.MurmurHash3.productHash(MurmurHash3.scala:64)
[info]   at scala.util.hashing.MurmurHash3$.productHash(MurmurHash3.scala:211)
[info]   at scala.runtime.ScalaRunTime$._hashCode(ScalaRunTime.scala:168)
[info]   at scala.Tuple2.hashCode(Tuple2.scala:20)
[info]   at java.util.concurrent.ConcurrentHashMap.putVal(ConcurrentHashMap.java:1012)
[info]   at java.util.concurrent.ConcurrentHashMap.putIfAbsent(ConcurrentHashMap.java:1535)
[info]   at com.twitter.scalding.FutureCache.getOrPromise(FutureCache.scala:40)
[info]   at com.twitter.scalding.FutureCache.getOrElseUpdateIsNew(FutureCache.scala:21)
[info]   at com.twitter.scalding.Execution$EvalCache.getOrElseInsertWithFeedback(Execution.scala:379)
[info]   at com.twitter.scalding.Execution$EvalCache.getOrElseInsert(Execution.scala:384)
[info]   at com.twitter.scalding.Execution$FutureConst.runStats(Execution.scala:388)
[info]   at com.twitter.scalding.Execution$Zipped$$anonfun$runStats$9.apply(Execution.scala:552)
[info]   at com.twitter.scalding.Execution$Zipped$$anonfun$runStats$9.apply(Execution.scala:551)
[info]   at com.twitter.scalding.FutureCache.getOrElseUpdateIsNew(FutureCache.scala:24)
[info]   at com.twitter.scalding.Execution$EvalCache.getOrElseInsertWithFeedback(Execution.scala:379)
[info]   at com.twitter.scalding.Execution$EvalCache.getOrElseInsert(Execution.scala:384)
[info]   at com.twitter.scalding.Execution$Zipped.runStats(Execution.scala:551)
...
```

johnynek, line 419: To be clear: if we override hashCode to be eager (e.g. a `val`), would this solve the problem?

Is that what you meant by fix just that problem?

jnievelt, line 419: It won't solve the problem, and it is indeed what I meant for the last part.  We (still) see the latter error if we (only) override hashCode to be eager.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -545,27 +555,28 @@ object Execution {
...
   }
   private case class UniqueIdExecution[T](fn: UniqueID => Execution[T]) extends Execution[T] {
-    def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      cache.getOrElseInsert(conf, this, {
-        val (uid, nextConf) = conf.ensureUniqueId
-        fn(uid).runStats(nextConf, mode, cache)
-      })
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      val (uid, nextConf) = conf.ensureUniqueId
+      Trampoline.call(fn(uid).runStats(nextConf, mode, cache))
```
johnynek: I don't think you are caching here. A second run, if the user makes an Execution that is different would not be cached.

Can we do:

``` scala
Trampoline(cache.getOrElseInsert(conf, this, {
   val (uid, nextConf) = conf.ensureUniqueId
   Trampoline.call(fn(uid).runStats(nextConf, mode, cache)).get
  }))
```

johnynek:
I'm inclined to merge this with the minor fixes to make sure we cache as aggressively as we did before.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
jnievelt: Ah, of course.  I guess it would be possible then to still blow the stack via nested `withId`.

edit: I think I'd need to go back to caching Trampolines to solve this one completely.  I'm in favor of leaving it for now, since we'd anyway need these changes anyhow.  I'll write up a test to make sure that we don't repeatedly call to fn on the same config.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -192,6 +193,8 @@ sealed trait Execution[+T] extends java.io.Serializable {
    */
   def zip[U](that: Execution[U]): Execution[(T, U)] =
     Zipped(this, that)
+
+  override val hashCode = super.hashCode
```
johnynek: can we add a test that this is still working like normal case class equality and hashCode?

For instance, `val e1 = Execution.from(1); val e2 = Execution.from(2); e1.zip(e2).hashCode == e1.zip(e2).hashCode`.

I'm worried this could break caching if somehow this picked up `hashCode` from `java.lang.Object`.

Other than that, this looks good to me.

johnynek:
@rubanm @isnotinvain any concerns other than the above?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
jnievelt: You're correct that this currently defers to `Object`, so it breaks caching.  In theory we could build custom `hashCode`s on each subclass and make it a val, but I don't feel great about that approach.

johnynek: can't we just manually add a strict hashCode on all the subclasses that take `Executions`? That will stop it from recursing forever, right?

Of course, it does not address equality checks (but this PR didn't do that anyway).

johnynek: so, `Mapped, FlatMapped, Zipped, OnComplete` would have strict `hashCode` that calls `p, but the others would not need it? We could just call this: https://github.com/scala/scala/blob/2.12.x/src/library/scala/runtime/ScalaRunTime.scala#L145 and store it in a val?

jnievelt: Yeah, that's what I meant (not sure if "strict" means something other than eager/val).

And `ScalaRunTime._hashCode` seems to be exactly what we want here.  All of the subclasses are `Product` anyway; so maybe it could stay in the base trait too.

johnynek: yeah, using `lazy val hashCode = ScalaRunTime._hashCode(this)` might work. I'm nervous about putting a `val` there due to init order bugs.

Yes, by `strict` I meant eager/val or even lazy val (which means strict is bad terminology to use). Just don't want recursion is all.

jnievelt: Looking into this further, there are actually a few other obstacles to having this work.  For example, even `Execution.from(5)` is going to create a closure object, and multiple calls won't make things that are "equal"; on develop:

```
scala> Execution.from(5) == Execution.from(5)
res0: Boolean = false

scala> Execution.from(5).hashCode
res1: Int = 765470035

scala> Execution.from(5).hashCode
res2: Int = -1375044982
```

I wouldn't mind fixing up that sort of thing.  From what I can tell it would involve making a few more case classes that wrap the functions we use here.  For example:

```
  private case class FutureTryWrapper[T](t: Try[T]) extends Function1[ConcurrentExecutionContext, Future[T]] {
    def apply(context: ConcurrentExecutionContext) = toFuture(t)
  }
  def fromTry[T](t: => Try[T]): Execution[T] = fromFuture(FutureTryWrapper(t))
```

Alternately I can just make sure we preserve the consistency we have today.

On a side note, any reason not to have `ReaderExecution`-based executions be lazy vals?

johnynek: @jnievelt any time we have call-by-name, equality is not going to work, but we only use call-by-name in a couple of places. In your example, we are just forcing the `Try` which, if we are going to do that, why make it call by name?

The idea was that call-by-name is expensive in general, but in `Execution` the real cost is running data flows, and call-by-name is nothing compared to that.

That said, it does not kill the caching. Consider:

``` scala
val a = loadStart()
val b = a.flatMap(foo)
val c = a.flatMap(bar)
val d = b.zip(c)
```

in this case, `a` will only be executed once. I realize this case also works with the `identityHashCode`, but I still think we should seek to maximize where the cache can work: which means, yes any function is a black box that uses reference equality, but there are many non-function cases.

jnievelt: Ah, of course.  It's perhaps an argument for const semantics, but I suppose that's getting well out of the immediate scope.

jnievelt: Okay, reworked this with `ScalaRunTime`, also to preserve the `hashCode`/`equals` contract.  Note that I've introduced a few more bits to expand the cases where we can reconstruct and achieve equality.

Also added a bunch of tests around reconstruction and the handling of mutable `Execution`s.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -856,14 +874,24 @@ object Execution {
    * these executions are executed in parallel: run is called on all at the
    * same time, not one after the other.
    */
+  private case class SequencingFn[T]() extends Function1[(T, List[T]), List[T]] {
```
johnynek, line 877: what's the motivation here?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -856,14 +874,24 @@ object Execution {
...
   def sequence[T](exs: Seq[Execution[T]]): Execution[Seq[T]] = {
     @annotation.tailrec
     def go(xs: List[Execution[T]], acc: Execution[List[T]]): Execution[List[T]] = xs match {
       case Nil => acc
-      case h :: tail => go(tail, h.zip(acc).map { case (y, ys) => y :: ys })
+      case h :: tail => go(tail, h.zip(acc).map(SequencingFn()))
     }
     // This pushes all of them onto a list, and then reverse to keep order
-    go(exs.toList, from(Nil)).map(_.reverse)
+    go(exs.toList, nil).map(ReversingFn())
```
johnynek, line 894: a note, you could, like you have with `nil` make `val cons = { p: ((Any, List[Any])) => (p._1) :: (p_.2) }` then cast it to `Function1[(T, List[T]), T]` with `def seqFn[T] = cons.asInstanceOf[((T, List[T])) => List[T]]`. This won't fail on the jvm due to type erasure, and it will keep the functions referentially equivalent.

The same could be done with `reverse`

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
jnievelt, line 877: If we leave the anonymous closure, we won't get equality/etc.  So even for a fixed sequence, we won't have `Execution.sequence(a) == Execution.sequence(a)`.  As I'm sure you'd imagine, the same principle is behind `ReversingFn` and `nil`.

This does have the property that all instances will be equal (e.g., `SequencingFn[String]() == SequencingFn[Long]()`) but I don't think that's necessarily a problem since they really all do perform the same operation.

I don't feel very strongly about it, but I wanted to at least illustrate what would be necessary for that equality to work, since we could do it without changing the runtime semantics.

johnynek, line 877: I see, yours actually does the same as mine without casting due to the way equals works on case classes.

That's better (casts are ugly).

johnynek:
LGTM.

Anyone else?

rubanm:
+1!

PR #1548, opened 2016-04-27
Skip invoking reducer estimators when reducers are set explicitly

gerashegalov:
I noticed that our execution jobs that are setting reducers explicitly were flooded with hraven estimator warnings, but there is no point calling an estimator to begin with.


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -172,17 +172,22 @@ object ReducerEstimatorStepStrategy extends FlowStepStrategy[JobConf] {
...
     val overrideExplicit = conf.getBoolean(Config.ReducerEstimatorOverride, false)
 
+    if (setExplicitly && !overrideExplicit) {
+      LOG.info(
+        s"""
+            |Flow step ${step.getName} was configured with reducers
+            |set explicitly (${setExplicitly}) and estimator explicit override
+            |(${overrideExplicit}). Skipping reducer estimation.
+         """.stripMargin)
+      return
```
johnynek: explicit return is generally avoided in scala (makes the code harder to read).

Can we just add an `else { }` for the rest of the code or put that block into a private method and call it there?

johnynek:
👍 merge when green

PR #1551, opened 2016-05-03
Fixes a bug that could cause limit to be run multiple times

ianoc:
Adds a test and fixes a bug around limits.

Case is that a TypedPipeFactory could sweep up a limit as a deferred operation. This effectively hides the operation from cascading until its used. This means if we use the limit in 2 places it will run twice despite having its own I/O.

### DIFF
--- a/scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
+++ b/scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
@@ -506,8 +506,8 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest 
     }
   }
 
-  "A TypedPipeJoinWithDescriptionPipe" should {
-    "A limit should not fan out into consumers" in {
+  "A limit" should {
+    "not fan out into consumers" in {
       // This covers a case where limit was being swept into a typed pipe factory
       // so each consumer was re-running the limit independently
       // which makes it usage unstable too.
--- a/scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
+++ b/scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
@@ -290,6 +290,27 @@ class TypedPipeForceToDiskWithDescriptionJob(args: Args) extends Job(args) {
     .write(TypedTsv[(Int, Long)]("output"))
 }
 
+class GroupedLimitJobWithSteps(args: Args) extends Job(args) {
+  val writeWords =
+    TypedPipe.from[String](List("word1 word2", "word1", "word2"))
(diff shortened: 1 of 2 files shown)


johnynek:
good call. After we made limit precise, this became a bug I think.

johnynek:
+1

jnievelt:
Should we test for "running twice" more directly?

ianoc:
This seems pretty direct to me for the issue. I'm not sure how another test
should look here, but if you want to add one or have some sample code happy
to take it.

On Monday, May 2, 2016, Joe Nievelt notifications@github.com wrote:

> Should we test for "running twice" more directly?
> 
> —
> You are receiving this because you authored the thread.
> https://github.com/twitter/scalding/pull/1551#issuecomment-216432843

## 

Sent from Gmail Mobile

jnievelt:
Possible approaches I toyed around with:
- write a parallel job without `limit`, and make sure that the `limit` adds only one more step
- put `withDescription` on all the steps, and make sure that the `limit` only appears once

But I don't feel strongly, after looking at it a bit.  Per my understanding this should be purely a performance issue, and not any case of job code getting multiply executed.

jnievelt:
Actually, I get a surprising result from the second approach.  When I do:

```
  .limit(3)
  .withDescription("limit")
```

This description doesn't make it into the limit step.  Instead, it gets mixed in with both of the write steps.  This seems like it would be a bug/regression?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -485,6 +506,20 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest
     }
   }
 
+  "A TypedPipeJoinWithDescriptionPipe" should {
```
jnievelt: This should be updated (or removed)

ianoc:
Descriptions are unfortunately flaky/best effort only :/ So i'd be slow to want to use them for anything correctness related. We can add a job with/without limit to compare numbers if you think it'll likely catch something this won't?

BTW --> This actually came up as a correctness issue, limit is non-deterministic as an operation. the operation someone was doing:
1. Join several large datasets to figure out intsersection keys
2. Take limit from joined dataset as a sample
3. Write samples out to provide samples of the input datasets that can be used for quicker test operations.

With the limit running for each write operation the results cannot be joined back together.

johnynek:
@jnievelt should we not merge this? To me this looks like strictly an improvement, and the test is doing something. I don't want to block such an easy fix on raising the test bar even higher.

jnievelt:
I am okay merging this; I'm not too concerned about the test.  But I would not say that it's strictly an improvement, given the labeling regression.

johnynek:
yeah, if we can improve labeling, great. It has always been a bit of a neglected aspect of scalding.

PR #1552, opened 2016-05-03
jdbc - fix reads for nullable columns

rubanm:
The bug was that java.sql.ResultSet getters for numeric types return 0 when the column value is NULL. This means that on the scalding side, we read the column value as Some(0) instead of None. (likewise for boolean, we get back Some(false) instead of None)

This change uses the following fix:

```
val v = resultset.getInt
if (resultset.wasNull) null else v // this then gets wrapped in an Option
```

### DIFF
--- a/scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
+++ b/scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
@@ -230,18 +230,30 @@ object ColumnDefinitionProviderImpl {
       case cf: ColumnFormat[_] => {
         val fieldName = cf.fieldName.toStr
         // java boxed types needed below to populate cascading's Tuple
-        cf.fieldType match {
-          case "VARCHAR" | "TEXT" => q"""$rsTerm.getString($fieldName)"""
-          case "BOOLEAN" | "TINYINT" => q"""_root_.java.lang.Boolean.valueOf($rsTerm.getBoolean($fieldName))"""
-          case "DATE" | "DATETIME" => q"""Option($rsTerm.getTimestamp($fieldName)).map { ts => new java.util.Date(ts.getTime) }.orNull"""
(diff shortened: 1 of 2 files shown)


scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
```diff
@@ -230,18 +230,30 @@ object ColumnDefinitionProviderImpl {
...
+          case "INT" | "SMALLINT" =>
+            (Some(q"""_root_.java.lang.Integer.valueOf"""), q"""$rsTerm.getInt($fieldName)""")
+          case f =>
+            (None, q"""sys.error("Invalid format " + $f + " for " + $fieldName)""")
         }
         // note: UNSIGNED BIGINT is currently unsupported
+        val valueTerm = newTermName(c.fresh("colValue"))
+        val boxed = box.map { b => q"""$b($valueTerm)""" }.getOrElse(q"""$valueTerm""")
+        q"""
+          { val $valueTerm = $primitiveGetter; if ($rsTerm.wasNull) null else $boxed }
```
johnynek, line 257: can we change this to:

``` scala
if ($rsTerm.wasNull) null else { val $valueTerm = $primitiveGetter; $boxed }
```

do we need the `primitiveGetter` in the null case?

rubanm, line 257: The contract here is a little weird. The java.sql.ResultSet getter needs to be called first for the wasNull call to then say correctly if the value we just read was null in the database or not. Added a comment to clarify that.

johnynek, line 257: ahh, that is confusing.

Glad we have some clean scala macros to hide all this! :) Nice work.

rubanm:
This is green. Any other comments? Thanks!

PR #1553, opened 2016-05-03
Merge release branch 0.16.0

piyushnarang:
Here are the draft release notes: https://github.com/twitter/scalding/releases/tag/untagged-24299adacd5d20387139. Updated scala docs have been pushed as well (https://github.com/twitter/scalding/commit/7c9105e933c76904087ce3d6accbea37b938d7b5).
Can mark the draft as the latest when we merge this.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.16.0-SNAPSHOT"
+version in ThisBuild := "0.16.0"
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "0.16.0"
+version in ThisBuild := "0.17.0-SNAPSHOT"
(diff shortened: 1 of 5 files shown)


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.16.0-SNAPSHOT"
\ No newline at end of file
+version in ThisBuild := "0.17.0-SNAPSHOT"
```
ianoc: i think this should just be 0.16.1 ?

ianoc:
Minor nit, but other than that LGTM

version.sbt
johnynek: Agree with Ian maybe (0.16.1-SNAPSHOT)

johnynek:
+1 on different version string, other than that :shipit:

version.sbt
piyushnarang: ok, updated due to popular demand.

isnotinvain:
+1

PR #1554, opened 2016-05-03
Switch to https based remote repo for DocGen

piyushnarang:
Was running into permission denied issues while trying to run `sbt ghpages-push-site`. Noticed that the remote repo address we're using is the ssh-style address. Switching to the https address worked for me and I thought that might be a better default to have here?


isnotinvain:
I have no idea why this matters, but it seems OK to me. Any thoughts @johnynek @ianoc

johnynek:
Don't know either, but if it works...

PR #1555, opened 2016-05-04
Cleanup cleanup (in VersionedStore)

fgcallari:
Friends don't let friends use a HashSet just to traverse a sorted list's tail ....

### DIFF
--- a/scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
+++ b/scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
@@ -128,18 +128,12 @@ public class VersionedStore {
     }
 
     public void cleanup(int versionsToKeep) throws IOException {
+        if (versionsToKeep < 0) return;
         final List<Long> versions = getAllVersions();
-        final HashSet<Long> keepers;
-        if(versionsToKeep >= 0) {
-            keepers = new HashSet<Long>(versions.subList(0, Math.min(versions.size(), versionsToKeep)));
-        } else {
-            keepers = new HashSet<Long>(versions);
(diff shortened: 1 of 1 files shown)


scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
```diff
@@ -128,18 +128,12 @@ public void cleanup() throws IOException {
...
-            keepers = new HashSet<Long>(versions);
-        }
-
-        for(Long v : versions) {
-            if(v!=null && !keepers.contains(v)) {
-                    deleteVersion(v);
-            }
+        int numExisting = versions.size(); 
+        if (numExisting <= versionsToKeep) return;
+        for (Long v : versions.subList(versionsToKeep, numExisting)) {
```
johnynek, line 135: generally looks good, only anxiety is lack of unit tests on this and the fact this has been here for years (and it is not in the critical path of perf).

fgcallari, line 135: Not sure I follow - isn't this code already covered by [these unit tests](https://github.com/fgcallari/scalding/blob/develop/scalding-commons/src/test/java/com/twitter/scalding/commons/datastores/VersionedStoreTest.java)? One of them even uses a HashSet as above, so it already compares for consistency of new code with old one.

johnynek, line 135: Well, I'll be. :) We just copied this code in and I assumed like much of the source code it lacked tests. :)

Looks good to me. 

Thanks for the clean up!

PR #1561, opened 2016-05-13
Making side-effecting functions non-nullary

jnievelt:
This is to address some warnings here and some that can arise from overloading these functions.  Scala will like to give one of these warnings:
- "side-effecting nullary methods are discouraged"
- "non-nullary method overrides nullary method"

This way folks can override non-nullary methods with non-nullary methods and avoid the warnings.  AFAICT it will be source- and binary-compatible.

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/TypedFieldsTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/TypedFieldsTest.scala
@@ -26,7 +26,7 @@ class TypedFieldsTest extends WordSpec with Matchers {
     // the Opaque class has no comparator
 
     "throw an exception if a field is not comparable" in {
-      val thrown = the[FlowException] thrownBy untypedJob
+      val thrown = the[FlowException] thrownBy untypedJob()
       thrown.getMessage shouldBe "local step failed"
     }
 
@@ -44,20 +44,20 @@ class TypedFieldsTest extends WordSpec with Matchers {
           outMap("bar") shouldBe 6
         }
         .run
-        .finish
+        .finish()
 
     }
 
   }
 
-  def untypedJob {
+  def untypedJob(): Unit = {
     JobTest(new UntypedFieldsJob(_))
       .arg("input", "inputFile")
       .arg("output", "outputFile")
       .source(TextLine("inputFile"), List("0" -> "5,foo", "1" -> "6,bar", "2" -> "9,foo"))
       .sink[(Opaque, Int)](Tsv("outputFile")){ _ => }
       .run
-      .finish
+      .finish()
   }
 
 }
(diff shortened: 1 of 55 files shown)


scalding-core/src/test/scala/com/twitter/scalding/TypedFieldsTest.scala
```diff
@@ -44,20 +44,20 @@ class TypedFieldsTest extends WordSpec with Matchers {
           outMap("bar") shouldBe 6
         }
         .run
-        .finish
+        .finish()
```
johnynek, line 47: PS: i'm actually not even sure we need `finish` anymore. That was just a type trick that I think is not needed with Scalatest. ? Can to try just removing a few and seeing that it still compiles and tests?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/RatioBasedEstimatorTest.scala
```diff
@@ -187,7 +187,7 @@ class RatioBasedReducerEstimatorTest extends WordSpec with Matchers with HadoopS
           val conf = steps.head.getConfig
           conf.getNumReduceTasks should equal (2)
         }
-        .run
+        .run()
```
johnynek, line 190: note this has no `finish`.

johnynek:
Thanks for doing this, @jnievelt . My only comment is maybe we can just remove the uses of `finish` (though still fix the site in case others are calling it).

That was only there because specs failed somehow if you didn't return unit (I don't know it it had some implicit that converted, and was only defined for `Unit` or what), but we don't use Specs anymore.

jnievelt:
Ah so it seems that the `should` in the tests for `"abcdef" should { ... }` is the same one that can be used for, e.g., `"abcdef" should endWith "def"`, or something like that.  So it doesn't know that the `{ ... }` should resolve to `Unit`, and then the compiler doesn't add the `()` automatically.

It is still an issue (TypedPipeTest, etc.), though only for cases where the `JobTest` running ends the test case.  Instead of using `finish()` we could explicitly just put `()` there ourselves.  Thoughts?

jnievelt:
On closer inspection, I'm realizing we're doing a rather odd thing.  Many tests are declared like:

```
"xyz" should {
  JobTest(...)
    .source(...)
    .sink(...) { outputBuffer =>
      "have the right output" in {
        outputBuffer should ...
      }
    }
    .run
}
```

This means the job runs as part of the test construction, not as part of the test execution.  If we changed them to look like:

```
"xyz" should {
  "have the right output" in {
    JobTest(...)
      .source(...)
      .sink(...) { outputBuffer =>
        outputBuffer should ...
      }
      .run
  }
}
```

I think it would "just work," since the ambiguity doesn't exist for `in`.  But it's tricky because many tests contain multiple `in` statements within the `sink` clause.  We could, e.g., do something like this:

```
"xyz" should {
  var output = _ // TODO: type
  JobTest(...)
    .source(...)
    .sink(...) { outputBuffer =>
      output = outputBuffer
    }
    .run

  "have the right output" in {
    output should ...
  }
}
```

Or we could just duplicate the job in each test execution.

johnynek:
honestly I really hate this `should..in` style, but that's a bigger change.

I hate crazy hacks to make code sort of read like english. Code is precise, english is not, it's the wrong direction (we should make english more like code).

I don't think we need to make a big change as part of this, so I'm +1 on this unless anyone sees any issues.

PR #1562, opened 2016-05-31
removed unused parameter parallelism

oranda:
Somebody might call this method with the parameter `parallelism`, imagining they were getting functionality that they are not.


johnynek:
Thank you!

Actually we should use https://github.com/HairyFotr/linter and we might have caught this.

Issue #1560, opened 2016-05-12
add a linter


johnynek:
https://github.com/HairyFotr/linter

This would be good along with adding more warnings, and cleaning up the warnings we have.

PR #1564, opened 2016-06-02
Adds scala linter for scalding

johnynek:
This adds: https://github.com/HairyFotr/linter

and fixes many errors it complains about.

We still have a lot of warnings (too many, in my opinion) which are generally caused by:
1. macro deprecations between scala 2.10 and 2.11. Perhaps we can use: https://github.com/milessabin/macro-compat to deal with these.
2. hadoop/cascading related deprecations that are beyond our control. We may be able to use: https://github.com/scala/scala/pull/3955 to turn off deprecation warnings but turn on fatal warnings otherwise.

This didn't seem to catch any bugs (maybe 1 or 2) but it did find many fishy things.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/mathematics/Combinatorics.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/mathematics/Combinatorics.scala
@@ -172,7 +172,8 @@ object Combinatorics {
     // create as many single-column pipes as the number of weights
     val pipes = allColumns.zip(weights).map(x => {
       val (name, wt) = x
-      IterableSource((0.0 to result by wt), name).read
+      val points = Stream.iterate(0.0) { _ + wt }.takeWhile(_ <= result)
+      IterableSource(points, name).read
     }).zip(allColumns)
 
     val first = pipes.head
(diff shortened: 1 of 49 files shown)


johnynek:
closes #1560

johnynek:
@isnotinvain @rubanm ?

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Combinatorics.scala
```diff
@@ -172,7 +172,8 @@ object Combinatorics {
     // create as many single-column pipes as the number of weights
     val pipes = allColumns.zip(weights).map(x => {
       val (name, wt) = x
-      IterableSource((0.0 to result by wt), name).read
+      val points = Stream.iterate(0.0) { _ + wt }.takeWhile(_ <= result)
```
rubanm, line 175: Just curious why this was flagged?

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -94,7 +94,7 @@ object CascadingBinaryComparator {
     else {
       val badSteps = missing.size
       val msg = missing.zipWithIndex.map { case (msg, idx) => s"<step$idx>$msg</step$idx>" }.mkString
-      error(s"There are $badSteps missing OrderedSerializations: $msg")
+      sys.error(s"There are $badSteps missing OrderedSerializations: $msg")
```
rubanm: Looks like the return type is Try[Unit] and def error returns a Failure. Should this stay unchanged?

rubanm:
Just a couple of comments. LGTM otherwise. Thanks!

scalding-core/src/main/scala/com/twitter/scalding/mathematics/Combinatorics.scala
johnynek, line 175: the linter does not like ranges with floats. I think this is the same logic, but maybe side steps some confusing cases with float ranges (not that anyone uses that code).

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
johnynek: ahhh This was a bug!!!! Our error was out of scope here so it was hitting `Predef.error` which is not what we intended!  :)

PR #1566, opened 2016-06-04
Never double count stats in Execution

johnynek:
Problem:

`a.zip(b)` naively adds the counters from left and right. This leads to very confusing results since both sides may include the same parent jobs, which means minor refactorings can duplicate counters.

Solution:
Rather than using a monoid on `ExecutionCounters` we note that counters are only created in one place, and that place is serialized. So we keep a `Map[Long, ExecutionCounters]` so we can de-dup the counters by only counting each id once. The API is not changed, but people relying on this bug will obviously be broken. Since the current behaviour makes the counters very difficult to use correctly, I strongly feel this is worth it.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -163,7 +163,7 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
    */
   protected def runStats(conf: Config,
     mode: Mode,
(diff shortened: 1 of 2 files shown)


johnynek:
@rubanm we'd love to publish an RC/snapshot fix for this. Can you review?

johnynek:
Actually, I realized, @jnievelt has a lot of `Execution` experience. Can you review, sir?

rubanm:
@johnynek LGTM! Just published 0.16.1-RC1 last week btw. So this could go in RC2.

PR #1569, opened 2016-06-10
case class implicit children

ianoc:
Use case here is that you have some class like

case class MyClass[A](myfield: A)

So, say a List/Set/Map , today we will handle those if the A is resolved at the call site. But we don't need that restriction if there is already an implicit OrderedSerialization in scope for A.

### DIFF
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ImplicitOrderedBuf.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ImplicitOrderedBuf.scala
@@ -27,13 +27,11 @@ import com.twitter.scalding.serialization.macros.impl.ordered_serialization._
   type. This is for the case where its an opaque class to our macros where we can't figure out the fields
 */
 object ImplicitOrderedBuf {
-  val macroMarker = "MACROASKEDORDEREDSER"
 
   def dispatch(c: Context): PartialFunction[c.Type, TreeOrderedBuf[c.type]] = {
     import c.universe._
-
     val pf: PartialFunction[c.Type, TreeOrderedBuf[c.type]] = {
-      case tpe if !tpe.toString.contains(macroMarker) => ImplicitOrderedBuf(c)(tpe)
+      case tpe => ImplicitOrderedBuf(c)(tpe)
     }
     pf
   }
@@ -45,10 +43,9 @@ object ImplicitOrderedBuf {
     val variableID = (outerType.typeSymbol.fullName.hashCode.toLong + Int.MaxValue.toLong).toString
     val variableNameStr = s"orderedSer_$variableID"
     val variableName = newTermName(variableNameStr)
-    val typeAlias = newTypeName(c.fresh("MACROASKEDORDEREDSER"))
+
     val implicitInstanciator = q"""
-      type $typeAlias = $outerType
-      implicitly[_root_.com.twitter.scalding.serialization.OrderedSerialization[$typeAlias]]"""
+      implicitly[_root_.com.twitter.scalding.serialization.OrderedSerialization[${outerType}]]"""
 
     new TreeOrderedBuf[c.type] {
       override val ctx: c.type = c
--- a/scalding-thrift-macros/src/main/scala/com/twitter/scalding/thrift/macros/impl/ScroogeInternalOrderedSerializationImpl.scala
+++ b/scalding-thrift-macros/src/main/scala/com/twitter/scalding/thrift/macros/impl/ScroogeInternalOrderedSerializationImpl.scala
@@ -23,13 +23,18 @@ import com.twitter.scalding.thrift.macros.impl.ordered_serialization.{ ScroogeEn
 import scala.language.experimental.macros
 import scala.reflect.macros.Context
 
+// The flow here is that we start with the outer dispatcher. Outer dispatcher is the only one allowed to recurse into a thrift struct `ScroogeOrderedBuf.dispatch`.
+// However it cannot use implicits at the top level. Otherwise this would always be able to return once with an implicit. We will use implicits for members inside
+// that struct as needed however.
+
+// The inner ones can recurse into Enum's and Union's, but will use the `ScroogeOuterOrderedBuf` to ensure we drop an implicit down to jump back out.
 object ScroogeInternalOrderedSerializationImpl {
-  // The inner dispatcher
+  // The base dispatcher
   // This one is able to handle all scrooge types along with all normal scala types too
   // One exception is that if it meets another thrift struct it will hit the ScroogeOuterOrderedBuf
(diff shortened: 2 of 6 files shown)


ianoc:
@rubanm , review? you've probably done the most with macros other than me I think. Its pretty short, relies on the `accessorMethod.returnType.asSeenFrom(outerType, outerType.typeSymbol.asClass)` magic we need for other things too. Despite the abstract type itself being unresolved at macro invocation time the inner A type on the field that we might use in an implicit doesn't seem to be the same one that the class is parameterized on.

isnotinvain:
Would this quietly change behavior? Like if you used to have a in implicit X in scope that was getting ignored, now it'd get picked up if a child of the case class is an X ?

ianoc:
Well, i'm not sure quietly, previously it just wouldn't compile. So what didn't compile before would compile now.

The only way it could change behavior in an unexpected way I can think of is that if the macros were imported implicitly, and hadn't been able to work on a type it would have been using some fallback, now it wouldn't be. (Though thats true of all feature extensions to these macros... but something I haven't really thought much about as we've added stuff)

isnotinvain:
I see so previously it would have been too many implicits in scope? I thought we let people override what ordering is used by putting one in scope?

ianoc:
So previously the behavior here was that the macro would fail, it actually tried to get an implicit for your type, so you'd see some error about unable to find implicit for MACROORDERSER with some random numbers after it. Since the case class one would just refuse to work with anything having an abstract type.

ianoc:
Previously from my read of this the macros actually would stick in an OrderedSerialization shim ontop of something it didn't understand, so if you supplied an implicit OrderedSerialization, they would wrap it. Wasteful/random but not behavior impacting. I changed it so we won't use the implicit code at the outer-most level anymore, just on inner fields we can't operate on.

isnotinvain:
OK

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ImplicitOrderedBuf.scala
```diff
@@ -27,13 +27,11 @@ import com.twitter.scalding.serialization.macros.impl.ordered_serialization._
   type. This is for the case where its an opaque class to our macros where we can't figure out the fields
 */
 object ImplicitOrderedBuf {
-  val macroMarker = "MACROASKEDORDEREDSER"
```
rubanm, line 30: So this was being used just to form type aliases before? Was this some macro workaround that's no longer needed?

scalding-thrift-macros/src/main/scala/com/twitter/scalding/thrift/macros/impl/ScroogeInternalOrderedSerializationImpl.scala
```diff
@@ -51,16 +62,14 @@ object ScroogeInternalOrderedSerializationImpl {
...
-
-    val scroogeDispatcher = ScroogeOrderedBuf.dispatch(c)(buildDispatcher)
+    import c.universe._
 
-    OrderedSerializationProviderImpl.normalizedDispatcher(c)(buildOuterDispatcher)
-      .orElse(scroogeDispatcher)
-      .orElse(innerDisp)
+    OrderedSerializationProviderImpl.normalizedDispatcher(c)(ScroogeInternalOrderedSerializationImpl.outerDispatcher(c))
+      .orElse(ScroogeOrderedBuf.dispatch(c)(baseScroogeDispatcher(c)))
+      .orElse(baseScroogeDispatcher(c))
```
rubanm, line 68: innerDispatcher seems unused. Should that be chained here?

rubanm, line 68: Ah nevermind. I see it being call from baseScroogeDispatcher

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ImplicitOrderedBuf.scala
ianoc, line 30: It was being used to block infinite recursion on the implicit. So instead of asking for type X, you as for the alias, and then the macros could see you were looking for an alias and would nop out of it. But if we just only use the implicit code paths on inner fields it becomes unnecessary.

scalding-thrift-macros/src/main/scala/com/twitter/scalding/thrift/macros/impl/ScroogeInternalOrderedSerializationImpl.scala
```diff
@@ -41,6 +46,12 @@ object ScroogeInternalOrderedSerializationImpl {
       .orElse(scroogeUnionDispatcher)
       .orElse(scroogeOuterOrderedBuf)
       .orElse(OrderedSerializationProviderImpl.scaldingBasicDispatchers(c)(buildDispatcher))
+  }
+
+  private def innerDispatcher(c: Context): PartialFunction[c.Type, TreeOrderedBuf[c.type]] = {
+    import c.universe._
+    val scroogeDispatcher = ScroogeOrderedBuf.dispatch(c)(ScroogeInternalOrderedSerializationImpl.innerDispatcher(c))
+    baseScroogeDispatcher(c)
```
rubanm, line 53: Is val scroogeDispatcher consumed anywhere?

ianoc, line 53: Not that I can see, killing

rubanm:
LGTM

johnynek:
So, this still does not allow the user to override, say `OrderedSerialization[Int]` and have it used in `case class Foo(get: Int)` right?

The fact that we don't pass those down is one of my main concerns with our approach (I think it can be a source of confusion).

I had to work around a case that our macros did not support, but then I had to plumb it manually all the way through because all classes that contained my case class tried to generate their own inner version.

ianoc:
This doesn't change those behaviors. That would be pretty meaningful for performance of these things, and size based... so if anyone wants to take that on it'll be a whole bunch of benchmarking....

I'm not sure it won't be a non-trivial hit with sets of simple case classes, how we handle length calculation/writing could be an issue too,  I think we could end up writing it more often and doing more buffer transfers.

However you could much easier now to just handle the type the macros won't work on yourself and have everything just work around it I believe... once the macros don't crash of course :)

Future work there someone could do would be:
1) Doing the tests for abstract traits/classes in the selector logic instead so that we don't hit those errors and cause the whole macro to error. That would mean for a given field/trait it would drop the implicit instead.
2) Docs/advice on 'cloaking' your types, if you for instance wrap the value that isn't happy in a single value class (not case class), it will block the macro. Or if the field type is an unknown trait of some sort.

PR #1571, opened 2016-06-17
fix incorrect usage of the Project constructor

fommil:
https://github.com/ensime/ensime-sbt/issues/210

@ktonga

Note that your version of sbt is quite old, I recommend upgrading to 0.13.11

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,7 @@ val junitVersion = "4.10"
 
 val printDependencyClasspath = taskKey[Unit]("Prints location of the dependencies")
 
-val sharedSettings = Project.defaultSettings ++ assemblySettings ++ scalariformSettings ++ Seq(
+val sharedSettings = assemblySettings ++ scalariformSettings ++ Seq(
   organization := "com.twitter",
 
   scalaVersion := "2.11.7",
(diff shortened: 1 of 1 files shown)


fommil:
failures look unrelated to this change

isnotinvain:
Hmm, those failures don't look like transient ones. I'm not sure why this change triggers them but we'll need to get to the bottom of that.

fommil:
Might be that other sbt plugins were misconfigured.

isnotinvain:
Could be, I don't know too much about sbt unfortunately.

We're open to any suggestions on how to refactor / cleanup the build here (including upgrading).
Out of curiosity, why does scalding's build cause issues in the project you linked? It doesn't consume scalding binaries via maven dependencies?

fommil:
One of our users was wanting to hack on scalding and experienced problems. I'm afraid I don't have time to look into this any further. It seems it would probably take quite some time to understand your intricate build and I only sent a PR because it was low hanging fruit to do so.

fommil:
anyway, since I'll not be continuing this PR ... it seems to fidgety ... I'll close the PR.

PR #1573, opened 2016-06-21
Make Writeable sources Mappable to get toIterator

johnynek:
TypedSource lacks the `toIterator` method. `Mappable` a subclass, has it.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/WritableSequenceFile.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/WritableSequenceFile.scala
@@ -73,7 +73,7 @@ object MultipleWritableSequenceFiles {
 }
 
 /**
- * This is only a TypedSource (which is a subclass of Mappable) as sinking into multiple directories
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/WritableSequenceFile.scala
```diff
@@ -73,14 +73,15 @@ object MultipleWritableSequenceFiles {
 }
 
 /**
- * This is only a TypedSource as sinking into multiple directories is not well defined
+ * This is only a TypedSource (which is a subclass of Mappable) as sinking into multiple directories
```
rubanm: Did you mean superclass?

rubanm:
👍

scalding-core/src/main/scala/com/twitter/scalding/WritableSequenceFile.scala
johnynek: yes.

Issue #1565, opened 2016-06-02
Slow TypedPipe.sketch performance with typed mapside reduce operation


spapin:
The change #1508 introduces mapside reduce operation that generally have a performance gain. However in the case of TypedPipe.sketch there is actually a slowdown.

Sketched creates a CounMinSketch monoid and sums it. The creation of the CounMinSketch is expensive. The mapside reduce operations creates comparatively more instance of CountMinSketch and the overall performance is degraded for large datasets.

johnynek:
How much slower is it? Can you share some numbers?

It is hard to make tradeoffs when we have data showing wins in one area. This is a bummer.

avibryant:
It seems plausible that we don't even want to use the standard mapside reduce for `TypedPipe.sketch`. We know there is going to be exactly one key, so most of the code doesn't really apply. We also don't really need all of the generality and immutability of the standard CMS implementation; we might well be better with a small, optimized, mutable CMS that we use directly from the sketch code. I bet even independent of this latest regression we could improve performance considerably.

spapin:
We run a map operation on skewed data over 5,000~20,000 cores and 1B~50B rows. 
The step used to run in about 1 hour. Now we're not even done with half the mappers in 20+ hours.

Trying to add .forceToReducers to the groupAll operation in Sketched.scala but while it fixed the mapper side, the reducer then never completed.

avibryant:
Ouch.

What information have you collected that points to #1508 specifically?

johnynek:
@spapin did the bisect find the exact patch on that regression?

johnynek:
@avibryant +1 to using a more optimized sketching approach. We can use `sumOption` on the CMS with some wrapper tricks (that are almost certainly worth it here).

spapin:
Yes a bisect confirms the origin is #1508 

We noticed that there is higher GC and ~30x more bytes written on the failling mappers. But that is probably related to the different structure being outputted on the mappers.

ningyue:
+1 We also see performance regression on sketch join. Any work around we could apply for now before sketch is fixed? Thanks!

rubanm:
@johnynek @ianoc Any objections to reverting https://github.com/twitter/scalding/pull/1508 on develop? We could bring it back once the ongoing CMS and SketchJoin improvements have been merged. It'll be nice if we can avoid spawning a one-off branch for cutting a RC and getting Twitter jobs using sketch going again.

ianoc:
[...]

PR #1575, opened 2016-06-26
Use Batched to speed up CMS summing on mappers

johnynek:
This uses the Batched type to address #1565.

/cc @rubanm @non @avibryant @ianoc @isnotinvain

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -22,7 +22,7 @@ import cascading.flow.FlowDef
 import cascading.pipe.{ Each, Pipe }
 import cascading.tap.Tap
 import cascading.tuple.{ Fields, TupleEntry }
-import com.twitter.algebird.{ Aggregator, Monoid, Semigroup }
+import com.twitter.algebird.{ Aggregator, Batched, Monoid, Semigroup }
 import com.twitter.scalding.TupleConverter.{ TupleEntryConverter, singleConverter, tuple2Converter }
 import com.twitter.scalding.TupleSetter.{ singleSetter, tup2Setter }
 import com.twitter.scalding._
@@ -471,7 +471,18 @@ trait TypedPipe[+T] extends Serializable {
    * Reasonably common shortcut for cases of total associative/commutative reduction
    * returns a ValuePipe with only one element if there is any input, otherwise EmptyValue.
    */
-  def sum[U >: T](implicit plus: Semigroup[U]): ValuePipe[U] = ComputedValue(groupAll.sum[U].values)
+  def sum[U >: T](implicit plus: Semigroup[U]): ValuePipe[U] = {
+    // every 1000 items, compact.
+    lazy implicit val batchedSG = Batched.compactingSemigroup[U](1000)
+    ComputedValue(map { t => ((), Batched[U](t)) }
+      .sumByLocalKeys
+      // remove the Batched before going to the reducers
+      .map { case (_, batched) => batched.sum }
+      .groupAll
+      .forceToReducers
+      .sum
+      .values)
+  }
 
   /**
    * Reasonably common shortcut for cases of associative/commutative reduction by Key
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ def scalaBinaryVersion(scalaVersion: String) = scalaVersion match {
 }
 def isScala210x(scalaVersion: String) = scalaBinaryVersion(scalaVersion) == "2.10"
 
-val algebirdVersion = "0.12.0"
+val algebirdVersion = "0.12.1"
 val apacheCommonsVersion = "2.2"
 val avroVersion = "1.7.4"
 val bijectionVersion = "0.9.1"
@@ -283,7 +283,7 @@ def youngestForwardCompatible(subProj: String) =
   Some(subProj)
     .filterNot(unreleasedModules.contains(_))
     .map {
-    s => "com.twitter" % ("scalding-" + s + "_2.10") % "0.15.0"
+    s => "com.twitter" %% (s"scalding-$s") % "0.16.0"
   }
 
 def module(name: String) = {
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
@@ -41,19 +41,12 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
   def reducers = Some(numReducers)
 
   private lazy implicit val cms = CMS.monoid[Bytes](eps, delta, seed)
-  lazy val sketch: TypedPipe[CMS[Bytes]] = {
-    // every 1000 items, compact.
-    lazy implicit val batchedSG = Batched.compactingSemigroup[CMS[Bytes]](1000)
+  lazy val sketch: TypedPipe[CMS[Bytes]] =
     pipe
-      .map { case (k, _) => ((), Batched(cms.create(Bytes(serialization(k))))) }
-      .sumByLocalKeys
-      // remove the Batched before going to the reducers
-      .map { case (_, batched) => batched.sum }
-      .groupAll
-      .sum
-      .values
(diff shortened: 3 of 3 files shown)


rubanm:
@johnynek Thanks. I'm going to twitter_release this branch and have one of our jobs run this.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -471,7 +471,18 @@ trait TypedPipe[+T] extends Serializable {
...
    * returns a ValuePipe with only one element if there is any input, otherwise EmptyValue.
    */
-  def sum[U >: T](implicit plus: Semigroup[U]): ValuePipe[U] = ComputedValue(groupAll.sum[U].values)
+  def sum[U >: T](implicit plus: Semigroup[U]): ValuePipe[U] = {
+    // every 1000 items, compact.
+    lazy implicit val batchedSG = Batched.compactingSemigroup[U](1000)
+    ComputedValue(map { t => ((), Batched[U](t)) }
+      .sumByLocalKeys
+      // remove the Batched before going to the reducers
+      .map { case (_, batched) => batched.sum }
```
ianoc, line 480: This i'm far more fuddful about. We can't really know in general how big the in memory objects are for values. (As illustrated by the previous CMS stuff). This is very similar(the same I believe) to how the SB stuff works along these plots, and there we had to be very mindful all the time about exhausting memory or triggering a higher GC burden.

In general i think the 1000 here should be configurable, but also that this behavior should probably be opt in

avibryant: Couldn't this be expressed in a `StatefulSummer` instead (and wouldn't that maybe be a better place for it)? If so, it could be combined with `AdaptiveCache` to ensure that we don't run out of memory. And we already have config there (around the cache size and/or using the adaptive one), so this would be a natural extension of that.

avibryant: I guess actually that enabling `AdaptiveCache` would fix the memory concern even with this code here, so those are separate concerns. But it does feel better to encapsulate in a summer, that's what that interface exists for.

avibryant, line 480: (Deleted my previous comments because I wasn't thinking straight)

It would not be difficult, I think, to write a single-key version of `AdaptiveCache` which used the same trick to tune the batch-size for `Batched` to make sure that we never ran out of memory.

It's also interesting to think about how this optimization could be built, more generally, into `AdaptiveCache` itself - that is, it could try to grow the `Batched` size as well as the number of keys in the cache. Once it starts hitting memory pressure it can decide whether to tradeoff reducing the batch size to grow the number of keys further.

johnynek, line 480: Keep in mind guys, in those instances we already know we only have 1 key.
Next recall that tuple semigroups/monoids already assume they can buffer
1000 items in memory.

This seems very safe to me given the default map-side cache size in
scalding 100K items (keys and values).

Do you think it poses more risk than the above two areas we buffer? (Also
recall sum on TypedPipe is a seldom used API).

On Monday, June 27, 2016, Avi Bryant notifications@github.com wrote:

> In scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
> https://github.com/twitter/scalding/pull/1575#discussion_r68626836:
> 
> > @@ -471,7 +471,18 @@ trait TypedPipe[+T] extends Serializable {
> >     \* Reasonably common shortcut for cases of total associative/commutative reduction
> >     \* returns a ValuePipe with only one element if there is any input, otherwise EmptyValue.
> >     */
> > -  def sum[U >: T](implicit plus: Semigroup[U]): ValuePipe[U] = ComputedValue(groupAll.sum[U].values)
> > -  def sum[U >: T](implicit plus: Semigroup[U]): ValuePipe[U] = {
> > -    // every 1000 items, compact.
> > -    lazy implicit val batchedSG = Batched.compactingSemigroup[U](1000)
> > -    ComputedValue(map { t => ((), Batched[U](t)) }
> > -      .sumByLocalKeys
> > -      // remove the Batched before going to the reducers
> > -      .map { case (_, batched) => batched.sum }
> 
> (Deleted my previous comments because I wasn't thinking straight)
> 
> It would not be difficult, I think, to write a single-key version of
> AdaptiveCache which used the same trick to tune the batch-size for Batched
> to make sure that we never ran out of memory.
> 
> It's also interesting to think about how this optimization could be built,
> more generally, into AdaptiveCache itself - that is, it could try to grow
> the Batched size as well as the number of keys in the cache. Once it
> starts hitting memory pressure it can decide whether to tradeoff reducing
> the batch size to grow the number of keys further.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> https://github.com/twitter/scalding/pull/1575/files/ddc861052d386a5cd3a4f14a368f1bf2cfe5902f..6fcddab5be5c2b183707a75a51703551890ade5f#r68626836,
> or mute the thread
> https://github.com/notifications/unsubscribe/AAEJdiU1AUVHOulADsol404MaSzw5aCrks5qQBIrgaJpZM4I-dkQ
> .

## 

avibryant, line 480: Speaking for myself, I'm good with this change - I agree that 1000 values should be fine in this case where we're limited to a single key.

Separately (and for a later PR), I'm excited about the idea of integrating `Batched` into `AdaptiveCache`, because it seems like it holds some promise of extending this optimization to more general cases (that is, if there is > 1 key, but still a small number of them relative to the size of the values, it would be nice to get the batching automatically).

ianoc, line 480: I'm -0, if its not heavily used API by us, we aren't winning a lot. We've generally erred on the side of stability, this would have edge case failures to do with GC/OOM, so its an area i'd be very slow to change default behavior myself for rare gains. (Most sum's are of Long/Int where this would be slower afaik than just leaving it alone).

isnotinvain, line 480: I guess it would be safer to add `.batchedSum` and use that in sketched above and leave regular `sum` alone. or add a batchSize arg with default value to the `.sum` method?

avibryant, line 480: Adding a batchSize with a default value isn't source-compatible, I think (it would force users to call it like `sum()`). Using `batchedSum` seems like a reasonable compromise (we could also add `batchedSumByKeys` in that case).

johnynek, line 480: adding a method to a trait is not binary compatible (in scala 2.11, in scala 2.12 it will be).

do we really think it is worth adding this complexity given the default for `.sum` already keeps 100k items? Why are we suddenly so conservative?

I don't want to add a second API, personally. It junks things up and confuses users. If people feel keeping 1000 items in RAM is too risky, I'll just revert to putting this only in CMS.

But, I want to reiterate, I don't see how that can rationally be our position given that:
1. `Semigroup.sumOption` on tuples batches with size of 1000 already.
2. default sumByKey keeps 100K keys AND values.
3. the new TypedMapsideReduce _fully_ materializes the full flatMap output into memory (no matter the size) before pushing over to the summing buffer.
4. probably many more cases of in-memory buffering of a constant number of items.

Are we really wanting to set a new policy of something like only 1 item of each type in memory? Or is it 100 would be okay, but not 1000? I'm not sure what people think we should ideally do in these cases.

isnotinvain, line 480: I think 1000 is fine if we are already doing that elsewhere w/o a config option, it seems small enough to me. I thought this was the first non-tunable case, but if we've already been doing it in sumOption and so on, then it seems fine to me.

The 100k in sumByKey is tunable and is fairly often tuned, so if it were that order of magnitude I think it'd need a config setting.

ianoc:
Merged since it seems Alex is onboard with this too

PR #1579, opened 2016-07-09
[parquet tuple macros] listType was deprecated in favor of listOfElements.

jbalogh:
Source: https://github.com/apache/parquet-mr/blob/apache-parquet-1.8.1/parquet-column/src/main/java/org/apache/parquet/schema/ConversionPatterns.java\#L94-L109

/cc @johnynek


johnynek:
👍 

@isnotinvain , let us know if this is not legit, but seems pretty clear win.

isnotinvain:
+1

PR #1581, opened 2016-07-19
Use Caching for FlowDefExecution

johnynek:
When @jnievelt added Trampolines to Execution, this one was left off.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -591,14 +591,17 @@ object Execution {
    * This allows you to run any cascading flowDef as an Execution.
    */
   private case class FlowDefExecution(result: (Config, Mode) => FlowDef) extends Execution[Unit] {
-    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      Trampoline(
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      lazy val future =
         for {
           flowDef <- toFuture(Try(result(conf, mode)))
           _ = FlowStateMap.validateSources(flowDef, mode)
           (id, jobStats) <- cache.runFlowDef(conf, mode, flowDef)
           _ = FlowStateMap.clear(flowDef)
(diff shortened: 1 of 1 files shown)


jnievelt:
Looks like this isn't a trampolining issue but a caching issue, in that we could have multiple runs of a re-used FlowDefExecution?

johnynek:
Right, bad title. We cache to everywhere else. Not sure why this one should not be cached.

johnynek:
The basic idea is that hadoop jobs are so expensive, we should definitely cache them.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -591,14 +591,17 @@ object Execution {
    * This allows you to run any cascading flowDef as an Execution.
    */
   private case class FlowDefExecution(result: (Config, Mode) => FlowDef) extends Execution[Unit] {
-    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      Trampoline(
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      lazy val future =
```
rubanm, line 595: Should this go outside the method? Or does this live across method calls?

johnynek, line 595: it depends on `conf` and `mode` so I don't think it can go outside the method. Being lazy just means we don't actually fire this future unless the cache does not already have a run of this Execution with the same `Config`.

rubanm, line 595: Ahh, right. I missed the getOrElse part below.

rubanm:
👍 So without this fix, it looks like we can have undesirable results if the Execution is not idempotent?

johnynek:
@rubanm exactly. There is also the chance that a job could get run twice (and if our experience in the past is any indication, that happens a lot with Execution without the cache). Note, the cache was added only after bug reports that using Execution resulted in many duplicated jobs.

The main case for duplication is the use or `zip` or `sequence` when both branches have a common ancestor. In that case, each branch recomputes without the cache, which is not what you want.

jnievelt:
👍

PR #1583, opened 2016-07-24
Graph API for Scalding

richwhitjr:
I had been using the GraphX library in Spark and realized that scalding does a much better job with very large graphs with heavy skew.  Some of the features in the GraphX library were really nice though and abstracted away some of the complexity of graph work.  

This review is a first pass at creating a similar API in scalding.  Currently I have only been thinking about directed graphs but undirected graphs should also be supported in a reasonable library.  The tests also need further work but I want to give people a chance to comment on the API.


avibryant:
How does this work relate to https://github.com/twitter/scalding/blob/b1d99378b25b27fe128cb083e46032c83e9e8a88/scalding-core/src/main/scala/com/twitter/scalding/mathematics/TypedSimilarity.scala, which also includes a simple graph abstraction? It might be informative to see what those algorithms look like implemented in terms of this API (or if that's possible).

richwhitjr:
Seems like most of those algorithms could be written in terms of the Graph structure.  The nice thing of this new abstraction is working natively with vertices and minimizing data duplication across edges.  Collecting neighbors has many useful properties for doing efficient graph calculations. 

Let me see what I can come up with.

richwhitjr:
Added an example of doing cosine similarity with the Graph class.  The intersection methods clearly need unit tests but wanted to show an example.

scalding-core/src/main/scala/com/twitter/scalding/graph/Edge.scala
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+case class Edge[T: Ordering, S](source: T, dest: T, attr: S)
```
johnynek: I'm a bit worried about `T: Ordering` here. This will have the `Ordering` serialized with each edge, sadly. Can we move the `T: Ordering` to methods that actually require an ordering?

scalding-core/src/main/scala/com/twitter/scalding/graph/EdgeTriplet.scala
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+case class EdgeTriplet[T: Ordering, S, Q](source: Vertex[T, Q], dest: Vertex[T, Q], edge: Edge[T, S])
```
johnynek: same comment about `T: Ordering`

scalding-core/src/main/scala/com/twitter/scalding/graph/VertexSimilarity.scala
```diff
@@ -0,0 +1,34 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+import com.twitter.scalding.mathematics.SetSimilarity
+
+import scala.reflect.ClassTag
```
johnynek: why do we need this? Seems like spark cargocult.

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+import com.twitter.scalding.TypedPipe
+
+import scala.reflect.ClassTag
+
+class Graph[T: ClassTag, S, Q](inputEdges: TypedPipe[Edge[T, S]], inputVertices: TypedPipe[Vertex[T, Q]])(implicit ord: Ordering[T]) {
```
johnynek: given that this has an `Ordering[T]` I don't think you need each `Edge` or `Vertex` to.

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+    new Graph[T, A, B](
+      newTriplets.map(_.edge),
+      newTriplets.flatMap(trip => List(trip.source, trip.dest)).distinct(Ordering.by(_.id)))
+  }
+
+  /**
+   * Collect only neighbor ids.
+   * Optionally sort by Vertex
+   */
+  def collectNeighborIds(sortNeighbors: Boolean = true): Graph[T, S, Neighbors[T]] = {
```
johnynek: should we not keep the previous label? Make this return `Graph[T, S, (Q, Neighbors[T])]`?

Also, I don't like the sorting. If you want to sort, can't we just mapVertices at effectively the same cost?

scalding-core/src/main/scala/com/twitter/scalding/graph/Vertex.scala
```diff
@@ -0,0 +1,20 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+case class Vertex[T, S](id: T, attr: S)(implicit ord: Ordering[T]) extends Ordered[Vertex[T, _]] {
```
johnynek: should not have an implicit `Ordering`.

You should do:

``` scala
object Vertex {
  implicit def ordering[T: Ordering, S: Ordering]: Ordering[Vertex[T, S]] = ...
}
```

I am not a fan of saying two things are equal in an ordering, but they have different attributes. Do we really need that?

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+      .values
+
+    Graph(edges, nbrs)
+  }
+
+  /**
+   * Collect all Neighbors of an Vertex.
+   * Optionally sort by Vertex
+   */
+  def collectNeighbors(sortNeighbors: Boolean = true): Graph[T, S, Neighbors[Vertex[T, Q]]] = {
```
johnynek: can't the above call this one and just `mapVertices` and a `mapAttr` on `Vertex`?

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+      .values
+
+    Graph(edges, nbrs)
+  }
+
+  /**
+   * Returns each Vertex with all out going edges.
+   * Optionally sort the edges
+   */
+  def collectEdges(sortNeighbors: Boolean = true): TypedPipe[Vertex[T, Neighbors[Edge[T, S]]]] =
```
johnynek: same sort comment. Move that to a method on `Neighbors`.

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  /**
+   * Filter the graph by the Edge and Vertex filters.
+   */
+  def subgraph(epred: EdgeTriplet[T, S, Q] => Boolean = _ => true, vpred: Vertex[T, Q] => Boolean = _ => true): Graph[T, S, Q] = {
+    val newTriplets = triplets.filter(epred)
+
+    new Graph[T, S, Q](
+      newTriplets.map(_.edge),
+      newTriplets.flatMap(trip => List(trip.source, trip.dest)).distinct(Ordering.by(_.id)).filter(vpred))
```
johnynek: why not filter before the distinct? Can't that save time?

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+    new Graph[T, S, Q](fEdges, fVertices)
+  }
+}
+
+/**
+ * Sometimes working just on deges is required, in those cases we don't also want to
+ * take the computational hit of filtering vertices by the updated edges.  In those cases
+ * you get return this subgraph that will only filter the vertices when necessary
+ */
+class GraphUnfilteredVertices[T: Ordering: ClassTag, S, Q](inputEdges: TypedPipe[Edge[T, S]], inputVertices: TypedPipe[Vertex[T, Q]])
```
johnynek: why do we need `ClassTag`?

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
```diff
@@ -0,0 +1,231 @@
...
+      .map(_._1)
+      .groupBy(_.dest)
+      .join(vertices.map(_.id).asKeys)
+      .values
+      .map(_._1)
+
+}
+
+object Graph {
+  def fromEdges[T: Ordering: ClassTag, S](edges: TypedPipe[Edge[T, S]]): Graph[T, S, Unit] = {
```
johnynek: let's not use `ClassTag`.

scalding-core/src/main/scala/com/twitter/scalding/graph/Neighbors.scala
```diff
@@ -0,0 +1,92 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+sealed trait Neighbors[T] {
+  def neighbors: Array[T]
+}
+
+case class SortedNeighbors[T](val neighbors: Array[T])(implicit ord: Ordering[T]) extends Neighbors[T] {
```
johnynek: let's take `Ordering` in the methods that use it, not on the class.

scalding-core/src/main/scala/com/twitter/scalding/graph/Neighbors.scala
```diff
@@ -0,0 +1,92 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.graph
+
+sealed trait Neighbors[T] {
+  def neighbors: Array[T]
```
johnynek, line 19: why an `Array` here and not `Set[T]` or `Vector[T]`? Why do we need the mutation?

scalding-core/src/main/scala/com/twitter/scalding/graph/Neighbors.scala
```diff
@@ -0,0 +1,92 @@
...
+      }
+    }
+
+    intersect
+  }
+
+  /**
+   * Compute Intersection by using binary search O(Mlg(N)) where M is small and N is large
+   */
+  private def binarySearch(bigArray: Array[T], smallArray: Array[T]): Int = {
```
johnynek: this needs legit tests complete with scalacheck.

scalding-core/src/main/scala/com/twitter/scalding/graph/Edge.scala
richwhitjr: Sure, never really thought about the overhead of ordering but it could probably be pretty large.

scalding-core/src/main/scala/com/twitter/scalding/graph/Graph.scala
richwhitjr: Thats a good point.  I put sort there because Neighbors has two subclasses, one called Sorted and Unsorted.  This code is coming from a branch where I was doing lots of work with Neighbor intersection where being sorted has a huge advantage and have a type to confirm sorted was nice check on that.

I can also create a sort function on UnsortedNeighbors which may be more explicit.

scalding-core/src/main/scala/com/twitter/scalding/graph/Neighbors.scala
richwhitjr: Don't need mutation but what is really important here is the least amount of memory overhead for the Neighbors.  Moving from Set -> Array saves a lot of memory.  Does Vector have memory characteristics close to native arrays?

Also about the classtags above, Array needs a classtag so it can be constructed.

johnynek: can we move the `ClassTag` closer to where it is used so we don't pay that cost everywhere?

richwhitjr:
Decided to simply this PR a bit and removed the example vertex similarity code and the neighbor intersection.  I can do a followup PR adding the intersection logic back in.  It becomes a bit tricky to think through the case of mutuals(directed edges in both directions).

richwhitjr:
Also for now I think it might be helpful to keep collectNeighborIds separate from collectNeighbor.  Worried mostly about the memory overhead of first collecting the vertices with the attributes then doing the filtering.  For very large set of neighbors just directly getting ids could be much more efficient.

PR #1585, opened 2016-08-09
[cascading3] use DistCacheTap for HashJoins

rubanm:
part of https://github.com/twitter/scalding/issues/1465

Based on the discussion on https://groups.google.com/forum/#!topic/cascading-user/2zhharIYTjk

This adds a Cascading rule for wrapping the rhs of any HashJoin in a DistCacheTap.

Note this is implemented via a separate FlowConnector and the default behavior has not been changed.

@cwensel a review would be great, thanks.

PR #1586, opened 2016-08-10
Merge cascading3 to develop

rubanm:
This is a single merge consisting of the piecemeal changes listed in https://github.com/twitter/scalding/issues/1465

For testing, I've published `0.16.1-cascading3-RC2` to sonatype.

@johnynek @ianoc would be great if this can be integrated/tested at Stripe before we merge to develop. Thanks!

Note that this does not include the HashJoin DistCacheTap change from https://github.com/twitter/scalding/pull/1585, which can be treated as an add-on change and merged subsequently. We do not have any pre-existing usage outside of Twitter afaik (the corresponding 2.5 fix was made in Twitter's Cascading fork).

cc @piyushnarang @isnotinvain


rubanm:
Just signed it. Glad to see this being merged!

johnynek:
Just FYI others, you don't need to sign this if you were a twitter employee when you did the work (which I think is true).

scalding-parquet-scrooge-cascading/src/main/java/com/twitter/scalding/parquet/scrooge/cascading/ParquetScroogeScheme.java
```diff
@@ -16,13 +16,13 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package com.twitter.scalding.parquet.scrooge;
+package com.twitter.scalding.parquet.cascading.scrooge;
```
oscar-stripe: @piyushnarang looks like the package name does not match the file name. not scrooge/cascading, but we have cascading.scrooge.

piyushnarang: I can push a follow up fix for this tomorrow.

PR #1587, opened 2016-08-25
Use ExecutionContext in Execution.from/fromTry

johnynek:
@isnotinvain pointed out that `Execution.from` does not use the execution context, but we always have access to one. We should probably not just use the thread we happen to be in, and instead actually use the execution context correctly.


isnotinvain:
+1

ulyssepence:
+1

johnynek:
according to this: https://github.com/twitter/analytics-infra-governance I think I can merge now.

thanks.

PR #1589, opened 2016-08-26
Create COMMITTERS.md

isnotinvain:
per https://github.com/twitter/analytics-infra-governance

### DIFF
--- a/COMMITTERS.md
+++ b/COMMITTERS.md
@@ -1,0 +1,19 @@
+# Committers
+
+Please see our [Project Governance](https://github.com/twitter/analytics-infra-governance) page for more details.
+
+## Active
+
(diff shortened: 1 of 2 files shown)


johnynek:
can we add a link in the README about this? No one is likely to check that file.

ulyssepence:
+1

azymnis:
I should probably add myself to emeritus...

isnotinvain:
@azymnis that would be great! You're still listed under Authors as well, I haven't tackled the Authors lists yet. For those, probably the actual original authors (like you) should be emeritus, and the contributors can just be in the contributors graph page.

For this first pass we just wanted to get the active committers (+1-ers) listed.

PR #1590, opened 2016-08-30
Add partitioned sources for Parquet thrift / scrooge

piyushnarang:
Adding a couple of partitioned sources for Parquet thrift / scrooge similar to [PartitionedDelimitedSource](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/PartitionedDelimitedSource.scala#L51).
Usage is similar to the PartitionedDelimitedSource:

```
val myTypedPipe: TypedPipe[ (String, String), MyThriftObject] = ...
myTypedPipe.write(PartitionedParquetThriftSource[(String, String), MyThriftObject](path, "%s/%s"))
// writes out data to: path/%s/%s
```

Added a couple of unit tests and tested these sources out using a couple of test read / write jobs.

### DIFF
--- a/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
+++ b/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
@@ -15,18 +15,18 @@ import com.twitter.scrooge.ThriftStruct
  * {{{
  * val data = MyScroogeObject()
  * IterablePipe(data, flowDef, mode)
- *  .write(PartitionParquetScrooge[(String, String), MyScroogeObject](path, "%s/%s"))
+ *  .write(PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s"))
  * }}}
  *
  * For reading it produces a pair `(P, T)` where `P` is the partition data, `T` is the corresponding
  * scrooge object. Below is an example.
  * {{{
  * val in: TypedPipe[(String, String), MyScroogeObject] =
- * TypedPipe.from( PartitionParquetScrooge[(String, String), MyScroogeObject](path, "%s/%s") )
+ * TypedPipe.from( PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s") )
  * }}}
  *
  */
-case class PartitionParquetScrooge[P, T <: ThriftStruct](
+case class PartitionedParquetScroogeSource[P, T <: ThriftStruct](
   path: String, template: String, fields: Fields = PartitionUtil.toFields(0, implicitly[TupleSetter[T]].arity))(implicit val mf: Manifest[T],
     val valueSetter: TupleSetter[T], val valueConverter: TupleConverter[T], val partitionSetter: TupleSetter[P], val partitionConverter: TupleConverter[P])
   extends FixedPathSource(path) with ParquetThriftBase[T] with PartitionSchemed[P, T] with Serializable {
--- a/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
+++ b/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
@@ -13,9 +13,10 @@ import com.twitter.scrooge.ThriftStruct
  * For writing it expects a pair of `(P, T)`, where `P` is the data used for partitioning and
  * `T` is the scrooge object. Below is an example.
  * {{{
- * val data = MyScroogeObject()
- * IterablePipe(data, flowDef, mode)
- *  .write(PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s"))
+ * val data: TypedPipe[MyScroogeObject] = ???
+ * data.map { obj =>
+ *   ( (obj.country, obj.city), obj)
+ * }.write(PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s"))
  * }}}
  *
  * For reading it produces a pair `(P, T)` where `P` is the partition data, `T` is the corresponding
--- a/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
+++ b/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
@@ -1,7 +1,6 @@
 package com.twitter.scalding.parquet.scrooge
 
 import _root_.cascading.scheme.Scheme
-import _root_.cascading.tuple.Fields
 import com.twitter.scalding._
 import com.twitter.scalding.parquet.thrift.ParquetThriftBase
 import com.twitter.scalding.typed.{ PartitionSchemed, PartitionUtil }
@@ -28,10 +27,11 @@ import com.twitter.scrooge.ThriftStruct
  * }}}
  *
  */
-case class PartitionedParquetScroogeSource[P, T <: ThriftStruct](
-  path: String, template: String, fields: Fields = PartitionUtil.toFields(0, implicitly[TupleSetter[T]].arity))(implicit val mf: Manifest[T],
-    val valueSetter: TupleSetter[T], val valueConverter: TupleConverter[T], val partitionSetter: TupleSetter[P], val partitionConverter: TupleConverter[P])
+case class PartitionedParquetScroogeSource[P, T <: ThriftStruct](path: String, template: String)(implicit val mf: Manifest[T],
+  val valueSetter: TupleSetter[T], val valueConverter: TupleConverter[T], val partitionSetter: TupleSetter[P], val partitionConverter: TupleConverter[P])
   extends FixedPathSource(path) with ParquetThriftBase[T] with PartitionSchemed[P, T] with Serializable {
+
+  override val fields = PartitionUtil.toFields(0, implicitly[TupleSetter[T]].arity)
 
   assert(
     fields.size == valueSetter.arity,
--- a/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
+++ b/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
@@ -6,6 +6,8 @@ import com.twitter.scalding.parquet.thrift.ParquetThriftBase
 import com.twitter.scalding.typed.{ PartitionSchemed, PartitionUtil }
 import com.twitter.scrooge.ThriftStruct
 
+import scala.reflect.ClassTag
+
 /**
  * Scalding source to read or write partitioned Parquet scrooge data.
  *
@@ -27,7 +29,7 @@ import com.twitter.scrooge.ThriftStruct
  * }}}
  *
  */
-case class PartitionedParquetScroogeSource[P, T <: ThriftStruct](path: String, template: String)(implicit val mf: Manifest[T],
+case class PartitionedParquetScroogeSource[P, T <: ThriftStruct](path: String, template: String)(implicit val ct: ClassTag[T],
   val valueSetter: TupleSetter[T], val valueConverter: TupleConverter[T], val partitionSetter: TupleSetter[P], val partitionConverter: TupleConverter[P])
   extends FixedPathSource(path) with ParquetThriftBase[T] with PartitionSchemed[P, T] with Serializable {
 
--- a/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
+++ b/scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
@@ -11,7 +11,8 @@ import com.twitter.scrooge.ThriftStruct
  * Scalding source to read or write partitioned Parquet scrooge data.
  *
  * For writing it expects a pair of `(P, T)`, where `P` is the data used for partitioning and
- * `T` is the scrooge object. Below is an example.
+ * `T` is the scrooge object. `P` must be either a String or a tuple of Strings.
+ * Below is an example.
  * {{{
  * val data: TypedPipe[MyScroogeObject] = ???
  * data.map { obj =>
--- a/scalding-parquet/src/main/scala/com/twitter/scalding/parquet/thrift/ParquetThrift.scala
+++ b/scalding-parquet/src/main/scala/com/twitter/scalding/parquet/thrift/ParquetThrift.scala
@@ -29,16 +29,19 @@ import com.twitter.scalding.source.{ DailySuffixSource, HourlySuffixSource }
 import java.io.Serializable
 import org.apache.thrift.{ TBase, TFieldIdEnum }
 
+import scala.reflect.ClassTag
+
 object ParquetThrift extends Serializable {
   type ThriftBase = TBase[_ <: TBase[_, _], _ <: TFieldIdEnum]
 }
 
 trait ParquetThriftBase[T] extends LocalTapSource with HasFilterPredicate with HasColumnProjection {
 
-  def mf: Manifest[T]
+  implicit def ct: ClassTag[T]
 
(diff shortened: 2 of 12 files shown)


isnotinvain:
Can this be slightly generalized such that any existing source can just "become" a partitioned source? Or maybe a factory that takes a Source and returns a PartitionedSource? or do we need to make one of these for each kind of source (parquet, scrooge, etc)

piyushnarang:
Let me see if I can refactor / tweak things to generalize this. Gets a bit more complicated cause existing sources tend to pull in a set of traits with a specific type [T], when you trying including the PartitionedSchemed trait that pulls in an overlapping set of traits but with a different generic type [P,T]

piyushnarang:
Poked around on this a bit to see if I could get it working. Seems like it's a bit more complicated / involved. Tried setting up a partitioned source which would delegate to another source but it seems like some of the methods on the Source class are protected so Scala's permissions seem to crib on that. 
If possible, would be nice if we can tackle the generalization in a future iteration.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Scalding source to read or write partitioned Parquet scrooge data.
+ *
+ * For writing it expects a pair of `(P, T)`, where `P` is the data used for partitioning and
+ * `T` is the scrooge object. Below is an example.
+ * {{{
+ * val data = MyScroogeObject()
+ * IterablePipe(data, flowDef, mode)
+ *  .write(PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s"))
```
isnotinvain: This isn't quite right is it? The partition key needs to be exposed right?
how about:

```
val data: TypedPipe[MyScroogeObject] = ???
data.map { obj => 
  ((obj.country, obj.city), obj) 
}.write(PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s")
```

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
```diff
@@ -0,0 +1,46 @@
...
+ * }}}
+ *
+ * For reading it produces a pair `(P, T)` where `P` is the partition data, `T` is the corresponding
+ * scrooge object. Below is an example.
+ * {{{
+ * val in: TypedPipe[(String, String), MyScroogeObject] =
+ * TypedPipe.from( PartitionedParquetScroogeSource[(String, String), MyScroogeObject](path, "%s/%s") )
+ * }}}
```
isnotinvain, line 29: for reads, is it required that P be a String or a TupleN of Strings? Seems like it'd have to be?

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/thrift/PartitionedParquetThriftSourceTests.scala
```diff
@@ -0,0 +1,60 @@
...
+    .map { address => (address.getZip, address) }
+    .write(partitionSource)
+}
+
+class PartitionedParquetThriftSourceTests extends WordSpec with Matchers {
+  import PartitionedParquetThriftTestSources._
+
+  def validate(path: Path, expectedAddresses: Address*) = {
+    val parquetReader: ParquetReader[Address] =
+      ThriftParquetReader.build(path).withThriftClass(classOf[Address]).build()
```
isnotinvain, line 32: is this the TBase reader? should be able to use the scrooge reader here right?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
piyushnarang: Yep, missed the map step there. Updated in the partitioned scrooge + thrift src docs.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
piyushnarang: Can also be a string (if its just "%s"). If you see the unit tests, there P ends up being a String.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/thrift/PartitionedParquetThriftSourceTests.scala
piyushnarang: Yeah this is the TBase reader. Can try and hook up the scrooge reader. Do you want to use the scrooge reader in both the partitioned thrift & partitioned scrooge tests?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
piyushnarang: Wonder if we should clarify that in the docs? Could add an example for the %s case cause it isn't a tuple?

isnotinvain: so P must be either: String or tuple of Strings right? That'd be good to clarify.

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/thrift/PartitionedParquetThriftSourceTests.scala
isnotinvain: sorry I misread, didn't realize this was the thrift test not the scrooge test

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
piyushnarang: Ok, I'll update the docs to indicate this and also include an example

scalding-parquet/src/test/scala/com/twitter/scalding/parquet/thrift/PartitionedParquetThriftSourceTests.scala
piyushnarang: Ok, so I updated the scrooge test to use the scrooge reader. Left this one as is

isnotinvain: cool, can you update the docs to explain the requirements on P being a String or tuple of Strings? I guess that might not be true if you provide a different TupleSetter, but at least taht part should be clarified too (that P and the TupleSetter need to match, which is not tracked by the type system)

piyushnarang:
@isnotinvain - updated to clarify the docs to indicate that P must be a String / tuple of Strings. I think there's some code in PartitionSchemed (and some classes it delegates to) that expects the fields to be Strings if I'm understanding right - [TemplatePartition](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/TemplatePartition.scala)

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
johnynek: is it possible, maybe a geek out, to have a macro that checks the format string at compile time?

this would allow it to be really clear that the string matches the tuple type?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
```diff
@@ -0,0 +1,48 @@
...
+ *
+ */
+case class PartitionedParquetScroogeSource[P, T <: ThriftStruct](
+  path: String, template: String, fields: Fields = PartitionUtil.toFields(0, implicitly[TupleSetter[T]].arity))(implicit val mf: Manifest[T],
```
johnynek: can we remove `fields` and `Manifest[T]`? Manifest is deprecated. And we really try to focus on the typed API now.

piyushnarang: Hmm, seems like both were required based on the traits I was mixing in. 
Manifest[T] as we're mixing in [ParquetThriftBase](https://github.com/twitter/scalding/blob/develop/scalding-parquet/src/main/scala/com/twitter/scalding/parquet/thrift/ParquetThrift.scala#L38). 
Fields due to [PartitionSchemed](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/PartitionSchemed.scala#L44)

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
piyushnarang: I can look into this. Not super familiar with Scala macros so this can be a learning opportunity :-). Currently we do check this (at runtime though) - in [TemplatePartition](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/TemplatePartition.scala#L28)

```
case class TemplatePartition(partitionFields: Fields, template: String) extends Partition {
  assert(
    partitionFields.size == "%s".r.findAllIn(template).length,
    "Number of partition fields %s does not correspond to template (%s)".format(partitionFields, template))
...
}
```

piyushnarang: So if you do something like:

```
val partitionSource = PartitionedParquetScroogeSource[String, Address](path, "%s/%s") // should be just %s

you get an error:
assertion failed: Number of partition fields '1' does not correspond to template (%s/%s)
java.lang.AssertionError: assertion failed: Number of partition fields '1' does not correspond to template (%s/%s)
```

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
johnynek: can't we always fix given `Fields` to be `PartitionUtil.toFields(0, implicitly[TupleSetter[T]].arity))`? Also, can we fix `ParquetThriftBase`?

I don't want to confuse users with parameters they should really never change.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
johnynek: yeah, it would be nice to just call the code cascading uses to verify it at compile time.

piyushnarang: Ok, I'll dig into this

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/PartitionedParquetScroogeSource.scala
piyushnarang: Yeah can set Fields to that. Can override it in PartitionedParquetScroogeSource and PartitionedParquetThriftSource unless you'd rather have it set in PartitionSchemed and not exposed to PartitionedParquet{Scrooge|Thrift}Source and PartitionedDelimitedSource?

johnynek: Yeah, any new code should certainly not expose it. How much to remove/lock
down of old code is a question.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/thrift/ParquetThrift.scala
```diff
@@ -29,16 +29,19 @@ import com.twitter.scalding.source.{ DailySuffixSource, HourlySuffixSource }
...
 }
 
 trait ParquetThriftBase[T] extends LocalTapSource with HasFilterPredicate with HasColumnProjection {
 
-  def mf: Manifest[T]
+  implicit def ct: ClassTag[T]
 
-  def config: ParquetValueScheme.Config[T] = {
-    val config = new ParquetValueScheme.Config[T].withRecordClass(mf.runtimeClass.asInstanceOf[Class[T]])
+  def config(implicit ct: ClassTag[T]): ParquetValueScheme.Config[T] = {
```
johnynek: why are we accepting `ct` as a parameter if we already have it on line 40?

johnynek (changes requested):
one minor comment

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/thrift/ParquetThrift.scala
```diff
@@ -29,16 +29,19 @@ import com.twitter.scalding.source.{ DailySuffixSource, HourlySuffixSource }
...
-trait ParquetThriftBase[T] extends FileSource with SingleMappable[T] with TypedSink[T] with LocalTapSource with HasFilterPredicate with HasColumnProjection {
+trait ParquetThriftBase[T] extends LocalTapSource with HasFilterPredicate with HasColumnProjection {
 
-  def mf: Manifest[T]
+  implicit def ct: ClassTag[T]
 
-  def config: ParquetValueScheme.Config[T] = {
-    val config = new ParquetValueScheme.Config[T].withRecordClass(mf.runtimeClass.asInstanceOf[Class[T]])
+  def config(implicit ct: ClassTag[T]): ParquetValueScheme.Config[T] = {
+    val clazz = implicitly[ClassTag[T]].runtimeClass.asInstanceOf[Class[T]]
```
johnynek: can't we just do `val clazz = ct.runtimeClass.asInstanceOf[Class[T]]` here

piyushnarang: Hmm yeah, that's cleaner.

scalding-parquet/src/main/scala/com/twitter/scalding/parquet/thrift/ParquetThrift.scala
piyushnarang: Shall remove this, added this on one of my prior passes over the code when I was seeing some compile errors. Not needed anymore.

piyushnarang:
@johnynek - updated based on your comments today. Have been looking at bit into trying out macros but the best I've got so far ended up being something on the lines of:

```
  def verifyTemplate(partitionFields: Fields, template: String): Unit = macro verifyTemplateImpl

  def verifyTemplateImpl(c: Context)(partitionFields: c.Expr[Fields], template: c.Expr[String]): c.Expr[Unit] = {
    import c.universe._

    reify {
            assert(
              partitionFields.splice.size() == "%s".r.findAllIn(template.splice).length,
              "Number of partition fields %s does not correspond to template (%s)".format(partitionFields.splice, template.splice))
    }
  }
```

This however ends up happening at runtime pretty much (and is identical to what we're doing today). Don't have a good sense of how to do this at compile time (tried a few things out but they seem to fail due to compile / misc errors).

johnynek:
@piyushnarang here is a simple macro that checks that a string is a valid regex at compile time: https://gist.github.com/xuwei-k/4991805

that might give a hint.

We don't have to do it here, but I like this technique. We can punt on that if you like.

johnynek (approved):

piyushnarang:
Thanks @johnynek, that's a pretty useful link. I'd vote for following up with a separate PR for the macro bit. I can write up an issue so that we don't lose track of it and pick it up in the next week or so.

PR #1591, opened 2016-09-09
Support empty directories in success file source

isnotinvain:
There are a important changes here:
1) In FileSource.createTap, use an empty memory tap when there are 0 non-empty paths. Currently this is considered an error. This PR allows for 0 non-empty paths as long as validateTaps() is ok with that.

2) Change SuccessFileSource such that its validateTaps() will not throw when encountering an empty directory that does contain a SUCCESS file (this is a strong signal that the directory is supposed to be empty / legitimately contains no data).

In order to do this, the contract around success files is changed a bit (I think for the better, and not in a breaking way as far as I can tell?), see the comments on SuccessFileSource for more details.

There's not very many tests currently around FileSource / SuccessFileSource / TimePathedSource, so it's hard to know if this change is breaking any current behavior. This could be fixed with some more tests, but the way pathIsGood and friends work gets pretty wacky across the various sources so it is hard to be sure.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -294,10 +294,10 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
         .map { path => CastHfsTap(createHfsTap(hdfsScheme, path, sinkMode)) }
         .toList
 
-    taps.size match {
-      case 0 => new IterableSource[Any](Nil).createTap(Read)(hdfsMode).asInstanceOf[Tap[JobConf, _, _]]
-      case 1 => taps.head
-      case _ => new ScaldingMultiSourceTap(taps)
+    taps match {
+      case Nil => new IterableSource[Any](Nil).createTap(Read)(hdfsMode).asInstanceOf[Tap[JobConf, _, _]]
+      case one :: Nil => one
+      case many => new ScaldingMultiSourceTap(many)
     }
   }
 }
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -385,7 +385,7 @@ trait SuccessFileSource extends FileSource {
    */
   override protected def goodHdfsPaths(hdfsMode: Hdfs): Iterable[String] = {
     super
-     .goodHdfsPaths(hdfsMode)
+      .goodHdfsPaths(hdfsMode)
       // some paths deemed "good" may actually be empty, and hadoop's FileInputFormat
       // doesn't like that. So we filter them away here.
       .filter { p => FileSource.globHasNonHiddenPaths(p, hdfsMode.conf) }
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -275,7 +275,7 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
   /*
    * Get all the set of valid paths based on source strictness.
    */
-  protected def goodHdfsPaths(hdfsMode: Hdfs) = {
+  protected def goodHdfsPaths(hdfsMode: Hdfs): Iterable[String] = {
     hdfsMode match {
       //we check later that all the paths are good
       case Hdfs(true, _) => hdfsPaths
@@ -288,9 +288,6 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
     val taps =
       goodHdfsPaths(hdfsMode)
         .iterator
-        // some paths deemed "good" may actually be empty, and hadoop's FileInputFormat
-        // doesn't like that. So we filter them away here.
-        .filter { p => FileSource.globHasNonHiddenPaths(p, hdfsMode.conf) }
         .map { path => CastHfsTap(createHfsTap(hdfsScheme, path, sinkMode)) }
         .toList
 
@@ -382,6 +379,17 @@ trait SequenceFileScheme extends SchemedSource {
 trait SuccessFileSource extends FileSource {
   override protected def pathIsGood(p: String, conf: Configuration) =
     FileSource.globHasSuccessFile(p, conf)
+
+  /*
+   * Get all the set of valid paths based on source strictness.
+   */
+  override protected def goodHdfsPaths(hdfsMode: Hdfs): Iterable[String] = {
+    super
+     .goodHdfsPaths(hdfsMode)
+      // some paths deemed "good" may actually be empty, and hadoop's FileInputFormat
+      // doesn't like that. So we filter them away here.
+      .filter { p => FileSource.globHasNonHiddenPaths(p, hdfsMode.conf) }
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -137,46 +137,42 @@ object FileSource {
...
-  def globHasSuccessFile(globPath: String, conf: Configuration): Boolean = allGlobFilesWithSuccess(globPath, conf, hiddenFilter = false)
-
-  /**
-   * Determines whether each file in the glob has a _SUCCESS sibling file in the same directory
-   * @param globPath path to check
-   * @param conf Hadoop Configuration to create FileSystem
-   * @param hiddenFilter true, if only non-hidden files are checked
-   * @return true if the directory has files after filters are applied
-   */
-  def allGlobFilesWithSuccess(globPath: String, conf: Configuration, hiddenFilter: Boolean): Boolean = {
```
johnynek, line 151: why are you deleting the method? Generally, unless we really need to, it is nice to keep things around and not break compatibility.

scalding-core/src/test/scala/com/twitter/scalding/FileSourceTest.scala
```diff
@@ -216,13 +216,8 @@ class FileSourceTest extends WordSpec with Matchers {
...
     }
 
-    "reject an empty directory" in {
-      pathIsGood("test_data/2013/05/") shouldBe false
-      pathIsGood("test_data/2013/05/*") shouldBe false
-    }
-
-    "reject a directory with only _SUCCESS when specified as a glob" in {
-      pathIsGood("test_data/2013/06/*") shouldBe false
+    "accept a directory with only _SUCCESS when specified as a glob" in {
```
johnynek, line 219: this is a bit wild. So we are explicitly changing the result of a function that we had tested.

One wonders if anyone will be hit by this. I think that empty with _SUCCESS should be okay, I think. Without _SUCCESS, I would say no.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: We can keep it, but it's not useful. I need to check, but it might be hard to keep it and have it retains it's old behavior.

scalding-core/src/test/scala/com/twitter/scalding/FileSourceTest.scala
isnotinvain: So, SuccessFileSource had a very complicated pathIsGood contract, and we had some (but not complete) tests to verify that contract, including its oddities that were maybe not desirable but were part of the contract, and we did encode them as tests (which is good).

This PR certainly changes that contract, though I think for the better, and it updates the tests to match.

That said, all this "this is only a best effort / sanity check" approach is a broken window, I think longer term we should move towards an explicit and correct implementation of FilleSource that doesn't rely on globbing oddities. Eg the user says "I want to read these files" and if they aren't there, that's an error. Instead, currently, the users says "I want to read whatever matches this gob, and maybe some verification on the results of that glob" which has the issue of not letting us verify things that don't match the glob at all.

johnynek: agree it is a broken window.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -289,17 +285,17 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
...
       goodHdfsPaths(hdfsMode)
-        .toList.map { path => CastHfsTap(createHfsTap(hdfsScheme, path, sinkMode)) }
+        .iterator
+        // some paths deemed "good" may actually be empty, and hadoop's FileInputFormat
+        // doesn't like that. So we filter them away here.
+        .filter { p => FileSource.globHasNonHiddenPaths(p, hdfsMode.conf) }
+        .map { path => CastHfsTap(createHfsTap(hdfsScheme, path, sinkMode)) }
+        .toList
+
     taps.size match {
```
johnynek: can we change this to:

``` scala
taps match {
  case  Nil => new IterableSource[Any](Nil).createTap(Read)(hdfsMode).asInstanceOf[Tap[JobConf, _, _]]
  case one :: Nil => one
  case many => new ScaldingMultiSourceTap(many)
}
```

rather than using size?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -371,30 +367,21 @@ trait SequenceFileScheme extends SchemedSource {
...
+ * Requires that:
+ * 1) Every matched, non-hidden directory contains a _SUCCESS file
+ * 2) Every matched, non-hidden file's parent directory contain a _SUCCESS file
  *
- * On the other hand it will reject an empty output directory of a finished job:
- * <pre>
- *   dir1/_SUCCESS
- * </pre>
+ * pathIsGood should still be considered just a best-effort test. There are still cases where this is
+ * not a sufficient test for correctness. See <ticket>
```
johnynek: what is the `<ticket>` Can we add some examples of how this is broken and even warn people off of the whole design?

Ideally, we want this sanity checking in the mappers, I think. And we fail the tasks if they can't read the data. But this is getting at atomicity in jobs, and that will depend on the underlying fs (we use s3 a lot at Stripe, as I imagine many folks do), so the issue is quite complex. Maybe worth really writing a doc on?

scalding-core/src/test/scala/com/twitter/scalding/FileSourceTest.scala
johnynek: I agree this is a better contract. Basically, `_SUCCESS` should imply the path is good. This whole jank was about cases where there was no `_SUCCESS` and it was a crazy attempt to at least warn on totally missing data (mostly for people that had job schedulers that didn't know about dependencies).

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: sure

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: well, if you supply the glob: `a/b/c/{x,y,z}` it is clear to a human that you expect x, y, and z to all exist and have SUCCESS files. But actually, what happens is:

1) If only one of {x,y,z} exists, and it has a success file, we will call this "good" because the non-existent dirs won't return any files and so we won't validate them
2) if some of the dirs do exist, but have no success files, then it will be "not good"

The only way around this is to try and parse the glob syntax looking for things like {x,y,z}, or instead separating these paths out to 3 separate globs in `hdfsPaths` instead of consoldiating them to 1 glob (which is what TimePathedSource does)

I need to file <ticket> w/ the details and reference it here.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: How strongly do you feel about keeping this method? It probably should not have been public, and I think it's functionality has been removed.

johnynek:
👍

johnynek:
I think it is fine to remove.

isnotinvain:
OK, updated the comments with a reference ticket. I'll merge once the tests have finished.

isnotinvain:
Thanks @johnynek !

johnynek:
👍

Issue #1593, opened 2016-09-11
Strange (or inconsistent) behaviour for GroupBy -> SortBy


siliconluffy:
'Oscar Boykin' via Scalding Development
Can you file an issue if nothing else to track the discussion?

On Sat, Sep 10, 2016 at 18:13 Oscar Boykin oscar@stripe.com wrote:
Wait, sorry. Looking more carefully. Is the bug that originally all data with id 0 was in one reducer but with sorting it winds up on both? That would be a bug. What version of scalding is this?

Can you replicate this bug in a minimal case? Sorting should not change how the keys are paritioned to reducer (which is done by hashCode of the key, which is the same, I suppose).

Basically the test you want to write is that after groupBy with sortBy if you take only the keys in the output each key appears exactly once.

I have a hard time believing there could have been a bug like this that we didn't notice for 5 years but I guess it is possible.
On Sat, Sep 10, 2016 at 17:33 ravi kiran holur vijay ravikiranhv@gmail.com wrote:
Hey Oscar,

Sorry, sounds like I might have misunderstood the semantics of groupBy followed by sortBy.
Is there a way to make sure ALL records having the same key end up at the same reducer (what groupBy does) and within each reducer, have it sorted by value (what sortBy does)?

-Ravi

On Sat, Sep 10, 2016 at 8:06 PM, Oscar Boykin oscar@stripe.com wrote:
Sorry, I don't follow. What did you expect. I don't see a bug.

The data looks sorted within groups, which is all sortBy does.

Note, you don't need forceToReducers here. Sorting can only be done on the reducers.
On Sat, Sep 10, 2016 at 15:12 ravi kiran holur vijay ravikiranhv@gmail.com wrote:
Hello,

I am noticing strange behaviour in my Scalding job which uses groupby and sortby. If I do not have a .sortBy function, each of the reducers are getting all of the values for the same group key. However, if I use .sortBy, each reducer is getting only part of the values for the same group key. I was wondering if any of you have run into a similar issue before or have a hypothesis about what's happening?

Case 1: Observed behaviour = Expected behaviour, without using sortBy

Reducer 1 output:

Processing data for group ... 1
Initializing FM Model with existing parameters ...
Processing model param ... o
Processing model param ... w
Processing model param ... r
Processing model param ... s
Processing model param ... t
Processing model param ... l
Processing model param ... f
Processing model param ... v
Processing model param ... v
Processing model param ... v
Processing model param ... v
Processing model param ... v
[...]

PR #1594, opened 2016-09-12
Add a test for sortBy

johnynek:
This adds a test for the possible bug in #1593 but does not find any issue.

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
@@ -15,7 +15,7 @@ limitations under the License.
 */
 package com.twitter.scalding
 
-import org.scalatest.{ Matchers, WordSpec }
+import org.scalatest.{ FunSuite, Matchers, WordSpec }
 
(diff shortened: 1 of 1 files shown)


piyushnarang:
👍

scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
```diff
@@ -121,6 +121,38 @@ class TypedSumByKeyTest extends WordSpec with Matchers {
   }
 }
 
+class TypedPipeSortByJob(args: Args) extends Job(args) {
+  TypedPipe.from(TypedText.tsv[(Int, Float, String)]("input"))
+    .groupBy(_._1)
+    .sortBy(_._2)
+    .mapValues(_._3)
+    .sum
```
ulyssepence, line 129: `.sum` only requires associativity?

ulyssepence, line 129: Oh right, this uses a `Semigroup`. NVM.

ulyssepence:
+1

PR #1595, opened 2016-09-13
Add fetch size for Mysql

oeddyo:


ulyssepence:
Can you talk about your motivations for doing this? What about the consumers?

gerashegalov:
@oeddyo thanks for tackling this. I wonder if it makes sense to include the other settables from https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html API to be done with it. 
like 

```
StatementEscapeProcessing
StatementFetchDirection
StatementFetchSize
...
```

oeddyo:
Hi @gerashegalov 
Thanks for the suggestion. I'd say let's address these one by one since we expect to checkin our code next week. Thus let's get in fetchSize first. What do you think?

ulyssepence:
@gerashegalov if there are only these additional values then it sounds reasonable to add them, so long as we make them all have defaults.

We will have to file a ticket for Twitter internal to wire this in when we do a release.

@oeddyo we don't continuously release Scalding into Twitter internal, so you likely won't see this code change for weeks/months.

gerashegalov:
I am ok to do it gradually but would be nice to at least think how we want to do it. For instance since some settables are properties of the Connection class, some of the Statement class, do we want to create separate config classes for them or do we want to use appropriate prefixes for case classes?

oeddyo:
@benpence I probably misheard yesterday someone mentioned it's upon request to merge to internal? If not then what shall I do to make sure our code can checkin?

ulyssepence:
@oeddyo We can merge internally whenever we want, but it is not trivial. In general, adding stuff to Scalding open source that you need immediately sounds like a formula to getting blocked. Add the functionality internally and, if it sounds like something others would like, add it in open source.

johnynek:
I'm confused here, this does not seem like it does anything here. Is there some internal code that is looking at this class?

ulyssepence:
@johnynek Yes, for some reason we've forked `JDBCSource` internally and have a utility for reading YAML into `AvailableDatabases`.

Perhaps this should be removed from open source?

johnynek:
Well, it would be nice to have the internal MySQL source released. That might be the best thing, right? But if that is not practical, then I think you should not make parts of it open source, which others cannot use, but only see the shadows of the real system.

isnotinvain:
Yeah, I agree -- if this value is never read by OSS scalding, its configuration doesn't belong here. Lets put this feature into what we've got internally, and when we move that to OSS, we can bring these settings along with it. But adding settings to the OSS API for internal use doesn't sound right to me.

ulyssepence:
Should we close this then? I can create a ticket to remove `AvailableDatabases` from the repo

johnynek:
reopen if you think we should reconsider. Thanks for sending the PR though.

oeddyo:
Thanks!

PR #1597, opened 2016-09-14
Add dependency graph plugin

piyushnarang:
Found this plugin useful while debugging some dependency issues a couple of times. Proposing adding it here to make this easier for folks in the future. Not super familiar with the sbt ecosystem so let me know if we'd rather avoid this :-)


johnynek:
I guess it is fine, but generally unless we are testing it or using for release I tend to say we should have sbt plugins in one's personal `.sbt` directory. This (or say, ide integration) might be more personal in the choice.

If we could think of some test we want to run on the output of this, that could be useful.

piyushnarang:
Yeah that makes sense, I can just move it over to my local sbt. Don't have a test that uses this, found it useful to track down some transitive deps.

PR #1600, opened 2016-09-27
[cascading3] Fix VersionedKeyValueStore bug when used with specific version

piyushnarang:
Noticed this on a failing user job internally that uses VersionedKeyValueStore (works fine on our develop branch). The have a path in hdfs that has 3 versions of data. When they specify VersionedKeyValStore with the hdfs path & version we seem to be pulling files from the path directory instead of path/version as we were doing earlier in v2. 

```
val src = VersionedKeyValSource[String, Foo]("/path/on/hdfs", sourceVersion = Some(version)) }
// now if we iterate over src, we're reading data from /path/on/hdfs/version1, path/on/hdfs/version2 and /path/on/hdfs/version (instead of just the last one)
```

Fix is to unset the Hadoop MRv2 input dir flag (currently we just unset the v1 flag) as that is what is picked up in [InputFileFormat](https://hadoop.apache.org/docs/r2.6.2/api/src-html/org/apache/hadoop/mapreduce/lib/input/FileInputFormat.html). Also reverted back to the SequenceFileInputFormat as it works correctly once this flag is unset. 

Added some validations to our existing VersionedKeyValSourceTest to check that the path for a source with version specified does indeed match the version. When I run this test on the cascading3 branch it ends up failing on the assert due to including paths to files in other version directories.


scalding-commons/src/test/scala/com/twitter/scalding/commons/VersionedKeyValSourceTest.scala
```diff
@@ -89,7 +96,7 @@ class VersionedKeyValSourceTest extends WordSpec with Matchers {
       .sink[(Int, Int)](VersionedKeyValSource[Array[Byte], Array[Byte]]("output")) { outputBuffer: Buffer[(Int, Int)] =>
         "Outputs must be as expected" in {
           assert(outputBuffer.size === input.size)
-          val singleInj = implicitly[Injection[Int, Array[Byte]]]
+          implicitly[Injection[Int, Array[Byte]]]
```
johnynek: what is this doing here?

scalding-commons/src/test/scala/com/twitter/scalding/commons/VersionedKeyValSourceTest.scala
```diff
@@ -146,7 +156,27 @@ class VersionedKeyValSourceTest extends WordSpec with Matchers {
...
+
+  def validateVersionPaths(path: String, version: Option[Long], store: VersionedKeyValSource[_, _], conf: JobConf): Unit = {
+    store.source.sourceConfInit(null, conf) // this sets up the splits needed for input format
+    val fileInputFormat = new TestSequenceFileInputFormat()
+    val paths = fileInputFormat.getPaths(conf)
+    version match {
+      case Some(ver) =>
+        // expect only the part file for the specified version
+        assert(paths.length == paths.count(_.toString.endsWith(ver + "/part-00000")))
+      case _ =>
```
johnynek: can we write `case None` to be explicit?

johnynek:
you have tested other jobs at Twitter with this? I'm nervous of changing this because it feels a bit undertested, but it is also critical code (summingbird, I believe relies on it working).

johnynek (changes requested):

scalding-commons/src/test/scala/com/twitter/scalding/commons/VersionedKeyValSourceTest.scala
piyushnarang: just got rid of the variable here as it wasn't used and was showing up as a warning in intellij

scalding-commons/src/test/scala/com/twitter/scalding/commons/VersionedKeyValSourceTest.scala
piyushnarang: yep, can do that.

piyushnarang:
@johnynek - I'm going to run our E2E suite + Tsar tests (and the failing job). Yeah I'm a bit worried too cause it is core functionality. These changes though switch us back pretty much to the code we have in develop - we use SequenceInputFileFormat there (extra thing is explicitly unsetting the v2 input dir setting before we set it).

scalding-commons/src/test/scala/com/twitter/scalding/commons/VersionedKeyValSourceTest.scala
johnynek: yeah, but `implicitly[Foo]` without keeping the response is not doing anything. Why not just remove the line?

piyushnarang: yeah that's true. Got rid of this one and the other one as the test seems to pass without them.

piyushnarang:
@johnynek - circling back to report on testing. All our internal E2E tests ended up being successful (as well as the user job that failed in the first place to trigger this PR). We have one internal Scalding test that uses VersionedKeyValSrc and I noticed yesterday that for some reason the verification step was turned off on it a year back :-(. I turned it on and ran it with Scalding develop (successful), Cascading3 (fails due to size mismatch) and these changes(successful). Think if we had the verification on in that test earlier we might have caught this issue when we made these changes in the first place.

johnynek (approved):

johnynek:
👍

piyushnarang:
Thanks @johnynek

PR #1601, opened 2016-09-27
REFACTOR: Rename parameter to reflect expectation

tonyjzhou:
A simple first time improvement for code readability.
# Problem

It's very easy to confuse the first parameter of this method as path instead of a glob pattern, which is what it really asks for:
`protected def pathIsGood(p: String, conf: Configuration) = FileSource.globHasNonHiddenPaths(p, conf)`
# Solution

Rename the first parameter from `p` to `globPattern`.

This only result in more readable code.
No behavior change is expected.


johnynek:
👍

isnotinvain:
@johnynek just realized, since scala lets you call methods with named params, does this count as an incompatible change?

johnynek:
@isnotinvain I don't care too much. I care about binary incompatible. Source incompatible that very few users hit are not a big deal in my book.

isnotinvain:
OK, does it wind up being binary compatible? the compiler must ultimately just call the params in order to work w/ the jvm I guess so it probably is binary compatible. 

👍

PR #1605, opened 2016-10-03
When merging multiple pipes, send out a single cascading.Merge rather than a tree

cchepelov:
cascading.Merge accepts a vararg amount of c.Pipe as its argument.

When evaluating this code:

``` scala
  val a: TypedPipe[Foo] = ???
  val b: TypedPipe[Foo] = ???
  val c: TypedPipe[Foo] = ???
  val d: TypedPipe[Foo] = ???

  val combo = a ++ b ++ c ++ d
  combo.write(....)
```

internally, Scalding builds a set of MergedTypedPipe, which it is then able to successfully collapse (exploiting the associativity of the pipe merge operation). **However**, the resulting list of pipes is then repeatedly pairwise folded through nested cascading.Merge pipe operations, which typically leads to wasteful execution on the fabric (Tez, here).

This patch ports the checkpoint-insertion logic to guard against merge-after-hash-join but ensures all merged pipes are sent to a single cascading.Merge step wherever possible.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
@@ -128,7 +128,9 @@ object RichPipe extends java.io.Serializable {
         // because cascading assumes it knows all the Pipe subtypes
         // and fails to match any others (think of it as a sealed trait)
         // So we handle all special types before checking for the assignName case
-        case other @ (_: Checkpoint | _: Operator | _: Splice | _: SubAssembly) =>
+        case other @ (hj: HashJoin) =>
+          collect ++ getJoinedPipeSet(hj)
+        case other @ (_: Checkpoint | _: Operator | _: Splice /* except HashJoin*/ | _: SubAssembly) =>
           collect :+ other
         case renamedPipe: Pipe =>
           // this is the assignName case
@@ -157,6 +159,60 @@ object RichPipe extends java.io.Serializable {
         false
     }
   }
+
+  /**
+   * Special handling for cases where one side of the hashjoin is merged
+   *  with the hashjoin result. Cascading no longer allows it (as of 3.0),
+   *  so we insert checkpoints and/or intermediate merge stages as appropriate
+   *
+   * @param head the first pipe to be merged
+   * @param tail a list of other pipes to be merged within the first
+   *
+   * @return an updated list of pipes, which can safely be merged
+   */
+  private[scalding] def mergeAvoidingHashes(head: Pipe, tail: List[Pipe]): Pipe = {
+
+    // we make use of the fact that the pipe merge operation is not just associative but also commutative
+    val pipes = head :: tail
+    val (colliding, uncolliding) = pipes.partition(p => pipes.exists(o => (o != p) && isHashJoinedWithPipe(p, o)))
+
+    val (innerColliding, innerUncolliding) = colliding.partition(p =>
+      colliding.exists(o => (o != p) && (isHashJoinedWithPipe(p, o) || isHashJoinedWithPipe(o, p))))
+    /* innerUncolliding pipes collide with some pipes in the uncolliding set, but don't collide with one another.
+     It is fine to lump them and merge them together before merging with the 'uncolliding' set.
+
+      innerColliding pipes collide with one another and must each be checkpointed before use in the general merge.
+     */
+    val safedInnerColliding = innerColliding.map(new Checkpoint(_))
+    val safedInnerUncolliding =
+      if (innerUncolliding.isEmpty) Nil
+      else List(new Checkpoint(mergeAvoidingNameClashes(innerUncolliding.head, innerUncolliding.tail)))
+
+    val reassembled = safedInnerColliding ::: safedInnerUncolliding ::: uncolliding
+    mergeAvoidingNameClashes(reassembled.head, reassembled.tail)
+  }
+
+  /**
+   * Cascading Merge does not support having multiple incoming pipes with the same name.
+   * Selectively rename pipes to avoid naming conflicts.
+   *
+   * @param head the first pipe to be merged (or checkpointed)
+   * @param tail a list of other pipes to be merged within the first
+   * @return a Merge of input pipes with any name clashes removed, or the input pipe if there was only one
+   */
+  private[scalding] def mergeAvoidingNameClashes(head: Pipe, tail: List[Pipe]): Pipe = tail match {
+    case Nil => head // avoid generating new Merge(pipes.head)
+    case _ =>
+      val (result, buf) = tail.foldLeft((List[Pipe](head), Set[String](head.getName))) {
+        case ((result, names), p) =>
+          if (names.contains(p.getName))
+            (assignName(p) :: result, names) /* no need to add the new name to names: assignName is guaranteed unique
+                                               and never assigned again */
+          else (p :: result, names + p.getName)
+      }
+      new Merge(result: _*)
+  }
+
 }
 
 /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -1089,24 +1089,11 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
         case (pipe, cnt) => pipe.flatMap(List.fill(cnt)(_).iterator)
       }
       .map(_.toPipe[U](fieldNames)(flowDef, mode, setter)) // linter:ignore
-      .toList
 
-    if (merged.size == 1) {
-      // there is no actual merging here, no need to rename:
-      merged.head
-    } else {
-      merged.reduce[Pipe] {
-        case (left, right) =>
-          // special handling for cases where one side of the hashjoin is merged
-          // with the hashjoin result. Cascading no longer allows it,
-          // so we add a checkpoint to the join result as a workaround
-          if (RichPipe.isHashJoinedWithPipe(left, right))
-            new Merge(RichPipe.assignName(new Checkpoint(left)), RichPipe.assignName(right))
-          else if (RichPipe.isHashJoinedWithPipe(right, left))
-            new Merge(RichPipe.assignName(left), RichPipe.assignName(new Checkpoint(right)))
-          else
-            new Merge(RichPipe.assignName(left), RichPipe.assignName(right))
-      }
+    merged match {
+      case Nil => sys.error("we can never merge to create 0 pipes") // should never happen as we cannot create this without two incoming pipes
+      case h :: Nil => h
+      case h :: tail => RichPipe.mergeAvoidingHashes(h, tail)
     }
   }
 
--- a/scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
@@ -808,6 +808,158 @@ class TypedMergeTest extends WordSpec with Matchers {
   }
 }
 
+class TypedHashAndMergeJob(args: Args) extends Job(args) {
+  val tp = TypedPipe.from(TypedText.tsv[String]("input"))
+  val tp2 = TypedPipe.from(TypedText.tsv[(String, Int)]("mixin"))
+
+  val x = tp.groupBy(x => x)
+    .hashJoin(tp2.group)
+    .values
+    .map { case (s, i) => s"${s}-${i}" }
+
+  (x ++ tp)
+    .write(TypedText.tsv[String]("output"))
+}
+
+class TypedHashAndMergeTest extends WordSpec with Matchers {
+  import Dsl._
+  "A TypedMergeJob" should {
+    var idx = 0
+    JobTest(new TypedHashAndMergeJob(_))
+      .source(TypedText.tsv[String]("input"), List(Tuple1("abc"), Tuple1("def"), Tuple1("ghi")))
+      .source(TypedText.tsv[(String, Int)]("mixin"), List("def" -> 2, "ghi" -> 3))
+      .typedSink(TypedText.tsv[String]("output")) { outBuf =>
+        (idx + ": correctly run despite a hash-merge situation") in { /* which isn't straightforward in Cascading */
+          outBuf.toSet shouldBe Set("def-2", "ghi-3", "abc", "def", "ghi")
+        }
+        idx += 1
+      }
+      .runHadoop
+      .finish()
+  }
+}
+
+trait TypedComplexHashAndMergeJobBase {
+  def ta: TypedPipe[String]
+  def taXb: TypedPipe[String]
+  def tc: TypedPipe[String]
+  def td: TypedPipe[String]
+  def tdXe: TypedPipe[String]
+}
+
+class TypedComplexHashAndMergeJob(args: Args ,
+                                  fieldsToMerge: Seq[(String, TypedComplexHashAndMergeJobBase => TypedPipe[String])])
+  extends Job(args) with TypedComplexHashAndMergeJobBase {
+
+  override def name: String = super.name + " (" + fieldsToMerge.map(_._1).mkString(" ++ ") + ")"
+
+  val ta = TypedPipe.from(TypedText.tsv[String]("a"))
+  val tb = TypedPipe.from(TypedText.tsv[(String, Int)]("b"))
+
+  val tc = TypedPipe.from(TypedText.tsv[String]("c"))
+  val td = TypedPipe.from(TypedText.tsv[String]("d"))
+  val te = TypedPipe.from(TypedText.tsv[(String, Int)]("e"))
+
+  val taXb: TypedPipe[String] = ta.groupBy(x => x).hashJoin(tb.group).values.map { case (s, i) => s"${s}→${i}" }
+  val tdXe: TypedPipe[String] = td.groupBy(x => x).hashJoin(te.group).values.map { case (s, i) => s"${s}⇒${i}" }
+
+  fieldsToMerge.map(_._2(this)).reduce(_ ++ _)
+    .write(TypedText.tsv[String]("output"))
+}
+
+class TypedComplexHashAndMergeTest extends WordSpec with Matchers {
+  import Dsl._
+
+  val fields = Seq[(String, TypedComplexHashAndMergeJobBase => TypedPipe[String])](
+    ("a", _.ta),
+    ("a∩b", _.taXb),
+    ("c", _.tc),
+    ("d",  _.td),
+    ("d∩e", _.tdXe))
+
+  val selection = fields.permutations.take(3) // Take'em all if you need to prove all permutations work equally (kind of slow, and internally we do use commutativity)
+
+  selection.foreach(perm => {
+    val permName = perm.map(_._1).mkString(" ++ ")
+
+    s"A TypedComplexHashAndMergeJob ${permName}" should {
+      var idx = 0
+
+      JobTest(new TypedComplexHashAndMergeJob(_: Args, perm))
+        .source(TypedText.tsv[String]("a"), List(Tuple1("a1"), Tuple1("a2"), Tuple1("a3"), Tuple1("a4")))
+        .source(TypedText.tsv[(String, Int)]("b"), List("a2" -> 2, "a3" -> 3, "a6" -> 6, "d2" -> 7, "d3" -> 8))
+        .source(TypedText.tsv[String]("c"), List(Tuple1("c1"), Tuple1("c2")))
+        .source(TypedText.tsv[String]("d"), List(Tuple1("d1"), Tuple1("d2"), Tuple1("d3"), Tuple1("d4")))
+        .source(TypedText.tsv[(String, Int)]("e"), List("d2" -> 4, "d3" -> 5, "a2" -> 9))
+        .typedSink(TypedText.tsv[String]("output")) { outBuf =>
+          (s"${idx}: correctly run despite a hash-merge situation") in {
+            /* which isn't straightforward in Cascading */
+
+            outBuf.toSet shouldBe Set("a1", "a2", "a3", "a4", "a2→2", "a3→3",
+              "c1", "c2", "d1", "d2", "d3", "d4", "d2⇒4", "d3⇒5")
+          }
+          idx += 1
+        }
+        .runWithoutNext(true) // .runHadoop but we don't want the hadoop thing to run everything at once.
+        .finish()
+    }
+  })
+}
+
+class TypedTwistedHashAndMergeJob(args: Args) extends Job(args) {
+  /* The purpose of this job is to find a complex case where TypedPipe#mergeAvoidingHashes might fail to stop */
+
+  val ta = TypedPipe.from(TypedText.tsv[String]("a"))
+  val tb = TypedPipe.from(TypedText.tsv[(String, Int)]("b"))
+
+  val td = TypedPipe.from(TypedText.tsv[String]("d"))
+  val te = TypedPipe.from(TypedText.tsv[(String, Int)]("e"))
+
+  val taXb: TypedPipe[String] = ta.groupBy(x => x).hashJoin(tb.group).values.map { case (s, i) => s"${s}→${i}" }
+  val tdXe: TypedPipe[String] = td.groupBy(x => x).hashJoin(te.group).values.map { case (s, i) => s"${s}⇒${i}" }
+
+  val taXbXe: TypedPipe[String] = taXb.groupBy(_.split("→").head).hashJoin(te.group).values.map { case (s, i) => s"${s}→${i}" }
+  val tdXeXb: TypedPipe[String] = tdXe.groupBy(_.split("⇒").head).hashJoin(tb.group).values.map { case (s, i) => s"${s}⇒${i}" }
+
+  val twistA = taXbXe.map(x => {
+    val y = x.split("→", 1)
+    (y.head, y.tail)
+  }).group
+
+  val twistB = tdXeXb.map(x => {
+    val y = x.split("⇒", 1)
+    (y.head, y.tail)
+  }).group
+
+  val twistAB: TypedPipe[String] = twistA.hashJoin(twistB).values.map { case (a,b) => a + "≡" + b }
+  val twistBA: TypedPipe[String] = twistB.hashJoin(twistA).values.map { case (a,b) => a + "≢" + b }
+
+  (taXbXe ++ tdXeXb ++ twistAB ++ twistBA)
+    .write(TypedText.tsv[String]("output"))
+}
+
+class TypedTwistedHashAndMergeTest extends WordSpec with Matchers {
+  import Dsl._
+
+  s"A TypedTwistedHashAndMergeTest" should {
+    var idx = 0
+
+    JobTest(new TypedTwistedHashAndMergeJob(_: Args))
+      .source(TypedText.tsv[String]("a"), List(Tuple1("a1"), Tuple1("a2"), Tuple1("a3"), Tuple1("a4")))
+      .source(TypedText.tsv[(String, Int)]("b"), List("a2" -> 2, "a3" -> 3, "a6" -> 6, "d2" -> 7, "d3" -> 8))
+      .source(TypedText.tsv[String]("d"), List(Tuple1("d1"), Tuple1("d2"), Tuple1("d3"), Tuple1("d4")))
+      .source(TypedText.tsv[(String, Int)]("e"), List("d2" -> 4, "d3" -> 5, "a2" -> 9))
+      .typedSink(TypedText.tsv[String]("output")) { outBuf =>
+        (s"${idx}: correctly run despite a hash-merge situation") in {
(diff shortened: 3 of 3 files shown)


cchepelov:
Update: pairwise reduction was not the case on the non-Cascading3 branch. This PR essentially fixes a performance (and intermediate disk space) regression

cchepelov:
(ping @rubanm — apologies for the lack of feedback; and thanks for the hard work you did on making the cascading3+tez stuff solid! We've now switched over to 0.16.1-RC3 (+this) for daily work)

sriramkrishnan:
Adding @piyushnarang

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1095,18 +1095,20 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
-      }
+      val preparedSources = merged.map(p => {
+        merged
+          .filterNot(_ == p)
+          .find(other => RichPipe.isHashJoinedWithPipe(p, other)) match {
+            case Some(_) =>
+              // special handling for cases where one side of the hashjoin is merged
+              // with the hashjoin result. Cascading no longer allows it,
+              //so we add a checkpoint to the join result as a workaround
+              new Checkpoint(p)
```
rubanm: I think one issue here might be that this can add more Checkpoints compared to the earlier version that uses `reduce`. (If we've already added a checkpoint to `left`, then we do not need another per `isHashJoinedWithPipe`)

rubanm:
@cchepelov Thanks for the patch. Glad that RC3 is working out for you. I'll let @piyushnarang  take a look as well since he is more familiar with the current state of this at Twitter.

cchepelov:
Can this code add more than one Checkpoint per incoming pipe ? This would 
be wrong indeed.

In the "a,b,c,d" example, let's assume for the sake of the example that a = 
z.hashJoin(b) and c=y.hashJoin(d)

The old code would produce
  Merge(Merge(Merge(Checkpoint(a),b), c), d)

Which incurs very hash penalties, at least on Tez and likely on MR as well, 
as the temporaries on the left side of the outer merges are spilled repeatedly

The proposed code would produce
  Merge(Checkpoint(a), b, Checkpoint(c), d)
Indeed one additional Checkpoint.

Do you think it would be better to try to reach something like
  Merge(a, Merge(b, c), d)

(i.e by trying to merge depth-first with the longest chains not requiring 
any checkpoint, checkpointing only if absolutely necessary) ?

The depth-first, longest chain thing would indeed still flatten nicely to a 
single merge in the absence of any prior hash joins.

I'll try that

Envoyé avec AquaMail pour Android
http://www.aqua-mail.com

cchepelov:
Phew. That was more complicated than expected; although it's easier now I noticed that commutativity was already being used (at https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala#L1085).

Added a set of fairly convoluted tests, which seek to verify the proper operation even if multiple level of hash joins have been used in the lineage of the output.

The updated patch now reaches (in yesterday's example) something like

``` scala
     Merge(Checkpoint(Merge(b, c)), d, a)
```

Attempting to remove the Checkpoint results in a planner crash (in the MalformedHashJoin rule), as expected.

The patch also correctly generates

``` scala
    Merge(i1, i2, i3, i4, i5) // assuming i1..i5 have no hash join relationship
```

(which was the case in the older cascading branch of scalding), which helps reducing the inter-node traffic on Tez.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -128,7 +128,9 @@ object RichPipe extends java.io.Serializable {
         // because cascading assumes it knows all the Pipe subtypes
         // and fails to match any others (think of it as a sealed trait)
         // So we handle all special types before checking for the assignName case
```
piyushnarang, line 130: Wondering if we can this this comment block to the top of the function / match statement? It's not really specific to the hashJoin case. 
Would also be nice to add a comment on the hashJoin case on why its being handled differently for folks reading the code later.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,25 +1089,72 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
-          // with the hashjoin result. Cascading no longer allows it,
-          // so we add a checkpoint to the join result as a workaround
-          if (RichPipe.isHashJoinedWithPipe(left, right))
-            new Merge(RichPipe.assignName(new Checkpoint(left)), RichPipe.assignName(right))
-          else if (RichPipe.isHashJoinedWithPipe(right, left))
-            new Merge(RichPipe.assignName(left), RichPipe.assignName(new Checkpoint(right)))
-          else
-            new Merge(RichPipe.assignName(left), RichPipe.assignName(right))
-      }
+      val prepared  = mergeAvoidingHashes(merged)
```
piyushnarang: no need to set to val prepared as we're returning in the next line?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,25 +1089,72 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+
+  /**
+    * Special handling for cases where one side of the hashjoin is merged
+    *  with the hashjoin result. Cascading no longer allows it (as of 3.0),
+    *  so we insert checkpoints and/or intermediate merge stages as appropriate
+    *
+    * @param pipes a list of pipes possibly colliding with each other
+    * @return an updated list of pipes, which can safely be merged
+    */
+  protected def mergeAvoidingHashes(pipes: Vector[Pipe]): Pipe = {
```
piyushnarang: this method seems a bit hard to understand. Trying to think of the best way we can refactor to make it more readable. Could we break it up into smaller methods with more descriptive names?
We have a few nested methods and checks in here, we could refactor those as methods with a descriptive name (cause they all seem to cover a specific scenario)

piyushnarang (commented):
Took a first pass at the code. Still unclear on some parts of it. Shall spend more time on it in the next day / so.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: +1

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,25 +1089,72 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+    require(!pipes.isEmpty)
+
+    // we make use of the fact that the pipe merge operation is not just associative but also commutative
+    val (colliding, uncolliding) = pipes.partition(p => pipes.find(o => (o != p) && RichPipe.isHashJoinedWithPipe(p, o)).isDefined)
```
johnynek: instead of `.find(fn(_).isDefined)` can we use `.exists(fn(_))`?

scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
```diff
@@ -808,6 +808,163 @@ class TypedMergeTest extends WordSpec with Matchers {
   }
 }
 
+class TypedHashAndMergeJob(args: Args) extends Job(args) {
+  import TDsl._
```
johnynek: can we remove this. Generally unneeded.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: so, basically we are building a directed graph: `a -> b` if `a isHashJoinedWith b`. Then pipes with no edges are in the uncolliding set.

Then, we need to deal with the colliding set. For that, we just checkpoint all the colliding nodes, then do one big merge.

Is this correct? If so, why do the recursion? It seems we could just do:

``` scala
new Merge(avoidNameClashes(colliding).map(new Checkpoint(_)) ++ uncolliding: _*)
```

I don't quite see why we need the recursion in the case.

johnynek (changes requested):

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
cchepelov: Hmm. The comment was specific to the case other @(_: Checkpoint ... SubAssembly) case, which used to cover the HashJoin case as well. 
Ideally, the comment applies to both cases (HashJoin and the other Checkpoin...SubAssembly), but not sure now how best to convey the idea

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
cchepelov: a relic from a debugging session; done.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
cchepelov: Yep, I'll try to tidy this up. I was initially fearing I needed recursion, and aimed for tailrec (which I couldn't achieve). 
Now planning to break `val safedColliding = match` as a `resolveCollidingHalf(colliding)` method, but unsure how far down that road I should go

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
cchepelov: Not sure. I initially needed recursion to deal with more complex examples, until I ended up splitting the `case other @ (hj: HashJoin) =>` case in  getChainOfEachs(). 
Now I need to verify whether the recursion still ever does anything (I suspect that now hash joins are transparent to getChainOfEachs(), we no longer really require it)

On your proposed change, I believe this would essentially generate my original code at https://github.com/twitter/scalding/pull/1605/commits/128b09513610f1b155839f3db5809e4a38bed61e, which @rubanm rejected on the grounds it would generate more Checkpoints than necessary. In the example from Monday, with `a, b, c, d` and `a -> b, c -> d`, your code would output

```
new Merge(new Checkpoint(a), b, new Checkpoint(c), d)
```

which is necessary only if we also have `a -> c`; from reading @rubanm, it seemed better to reach 

```
new Merge(new Checkpoint(new Merge(a,c)), b, d)
```

if a and c are hashjoinwise independent.

cchepelov: OK, I tried something. Now outputs `new Merge(new Merge(a,c), b, d)` in the example, still passes the Complex and the Twisted tests

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+  /**
+   * Special handling for cases where one side of the hashjoin is merged
+   *  with the hashjoin result. Cascading no longer allows it (as of 3.0),
+   *  so we insert checkpoints and/or intermediate merge stages as appropriate
+   *
+   * @param pipes a list of pipes possibly colliding with each other
+   * @return an updated list of pipes, which can safely be merged
+   */
+  protected def mergeAvoidingHashes(pipes: Vector[Pipe]): Pipe = {
+    require(!pipes.isEmpty)
```
johnynek: can we instead of the `require` have this method take `head: Pipe, tail: List[Pipe]` then when we call we can do:

``` scala
merged match {
  case Nil => sys.error("we can never merge to create 0 pipes")
  case h :: Nil => h
  case h :: tail => mergeAvoidingHashes(h, tail)
}
```

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+
+    val (innerColliding, innerUncolliding) = colliding.partition(p =>
+      pipes.exists(o => (o != p) && (RichPipe.isHashJoinedWithPipe(p, o) || RichPipe.isHashJoinedWithPipe(o, p))))
+    /* inneruncolliding pipes collide with some pipes in the uncolliding set, but don't collide with one another.
+     It is fine to lump them and merge them together before merging with the 'uncolliding' set.
+
+      innercolliding pipes collide with one another and must each be checkpointed before use in the general merge.
+     */
+    val safedInnerColliding = innerColliding.map(new Checkpoint(_))
+    val safedInnerUncolliding = if (innerUncolliding.isEmpty) Nil else Seq(checkpointOrMergeAvoidingNameClashes(innerUncolliding))
```
johnynek: can we wrap:

``` scala
val safedInnerUncolliding =
  if (innerUncolliding.isEmpty) Nil
  else Seq(checkpointOrMergeAvoidingNameClashes(innerUncolliding))
```

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+    mergeAvoidingNameClashes(safedInnerColliding ++ safedInnerUncolliding ++ uncolliding)
+  }
+
+  /**
+   * Ensure there is either a merge or a checkpoint resulting from the incoming pipes.
+   *
+   * @param pipes
+   * @return
+   */
+  protected def checkpointOrMergeAvoidingNameClashes(pipes: Seq[Pipe]): Pipe = {
```
johnynek: let's take a `p: List[Pipe]` and do a pattern match:

``` scala
def check(head: Pipe, tail: List[Pipe]): Pipe = tail match {
  case Nil => new Checkpoint(head)
  case h :: t => mergeAvoidingNameClashes(head, h, t) // we have at least 2
}
```

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+    }
+  }
+
+  /**
+   * Cascading Merge does not support having multiple incoming pipes with the same name.
+   * Selectively rename pipes to avoid naming conflicts.
+   * @param pipes pipes with potential name clashes
+   * @return a Merge of input pipes with any name clashes removed, or the input pipe if there was only one
+   */
+  protected def mergeAvoidingNameClashes(pipes: Seq[Pipe]): Pipe = {
```
johnynek: same pattern:

``` scala
def merge(h1: Pipe, h2: Pipe, tail: List[Pipe]): Pipe = {
  // we know there are at least 2 pipes, so we are bound to create a merge:
  (h1 :: h2 :: tail).foldLeft(....
}
```

there is no need to consider the case where there is only 1. Also, I'm confused why we don't need checkpoint here, but we have checkpoint above (line 1137).

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+
+  /**
+   * Ensure there is either a merge or a checkpoint resulting from the incoming pipes.
+   *
+   * @param pipes
+   * @return
+   */
+  protected def checkpointOrMergeAvoidingNameClashes(pipes: Seq[Pipe]): Pipe = {
+    if (pipes.size == 1) {
+      new Checkpoint(pipes.head)
```
johnynek: can you comment why we need a checkpoint here, but not in the `else` case?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+   * @param pipes pipes with potential name clashes
+   * @return a Merge of input pipes with any name clashes removed, or the input pipe if there was only one
+   */
+  protected def mergeAvoidingNameClashes(pipes: Seq[Pipe]): Pipe = {
+    require(!pipes.isEmpty)
+
+    if (pipes.size == 1) {
+      pipes.head // avoid generating new Merge(pipes.head)
+    } else {
+      val (result, buf) = pipes.foldLeft((Seq[Pipe](), Set[String]())) {
```
johnynek: Let's use `List` and `::` if order does not matter (since it is fast), where as `:+` is `O(N)` cost on `Seq`.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+        case ((result, names), p) =>
+          if (names.contains(p.getName)) (result :+ RichPipe.assignName(p), names) else (result :+ p, names + p.getName)
```
johnynek: don't we need to add to the `names` set? or are you relying on `assignName` to make a unique name (which I _think_ it does). If so, can we comment?

Also, can we wrap this line?

``` scala
if (names.contains(p.getName)) (RichPipe.assignName(p) :: result, names)
else (p :: result, names + p.getName)
```

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -1089,24 +1089,74 @@ final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) ext
...
+
+  /**
+   * Special handling for cases where one side of the hashjoin is merged
+   *  with the hashjoin result. Cascading no longer allows it (as of 3.0),
+   *  so we insert checkpoints and/or intermediate merge stages as appropriate
+   *
+   * @param pipes a list of pipes possibly colliding with each other
+   * @return an updated list of pipes, which can safely be merged
+   */
+  protected def mergeAvoidingHashes(pipes: Vector[Pipe]): Pipe = {
```
johnynek: can we use private rather than protected? Also, these are all really static methods, right? Can we move them to an object that is `private[scalding]`? I'd rather not add them to `TypedPipe` since they seem to be working just on `Pipe`. Maybe `RichPipe`'s object?

johnynek (changes requested):

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
cchepelov: Uh, since the first operation is to perform a partition over (List(head) :: tail), this is going to be a little ugly, but OK

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
cchepelov: indeed, assignName generates a guaranteed unique name (which it won't ever generate again)

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
cchepelov: Turns out the Checkpoint is always necessary, which dissolves the need for `checkpointOrMergeAvoidingNameClashes()` entirely (I had cases where running without the checkpoint around the inner merge worked, but I couldn't always reliably rule out a violation in the planner's RemoveMalformedHashJoins rule)

cchepelov: Diggigng a little bit, the "flip flop" was due to a typo in mergeAvoidingHashes(), which sometimes (depending on random factors) led to unintentionally outputting (a,b,c,Merge(Checkpoint(d), Checkpoint(e)) 

Now the typo's fixed, we started to reliably generate Merge(a,b,c,Merge(d,e)) (with a=x.hash(d) and b=y.hash(e)), which doesn't satisfy the planner requirements. The correct output is Merge(a,b,c,Checkpoint(Merge(d,e)))

cchepelov:
Just read the release notes for cascading 3.2.0-wip-7 by @cwensel. I wonder if this whole thing isn't moot now. 

Chris, do the new changes (including the c.t.h.DistCacheTap) mean we can now just output a single `Merge( bunch_of_pipes:_*)` and Cascading will take care of guarding against hash joins, or it is better in your opinion to still insert some Checkpoints as appropriate?

cwensel:
The `ReplaceAccumulateTapWithDistCacheTransformer` rule simply decorates any head or intermediate Taps with a `DistCaceTap` on both MR and Tez. MR will see a a larger perf boost.

If you are trying to safeguard against `DualStreamedAccumulatedMergeNodeAssert` from firing the above rule won't make a difference.

But it might make sense for me to dig back into this issue and see if I can plan around the Merge->GroupBy issue (without resorting to a logical optimization).

That said, I strongly recommend, if the assert is the issue at hand, to look at creating an logical optimization rule that collapses a Merge->GroupBy to GroupBy on Tez within a Scalding specific `RuleRegistry` with other possible rules that compensate for Scalding plan artifacts.

johnynek:
@cwensel is there any reason `Merge -> GroupBy` to `GroupBy` and similarly for `CoGroupBy` is not a good rule for all flows (scalding or otherwise)?

cwensel:
Good question. 

My current opinion is that Cascading should not ship (as defaults) with what I would call logical optimizations. The resulting MR/Tez complex can be so complex, I feel it works against the developer to do non-obvious things under the hood. I think what was a good practice for monolithic systems is bad practice at current scales MR/Tez are applied. (anecdotally this made Pig/Hive etc more challenging, it was hard for a developer to build a working mental model of what's going as an app/query become more complex)

Swapping `Merge->GroupBy` with `GroupBy` would fall into that category. On MR Merge has limited effect (MultiInputSplit applied on mapper), on Tez it physically affects the resulting plan topology.

You may also recall in Cascading 2.x an issue where we collapsed 'equivalent' and adjacent taps into one as an optimization. This 'rule' was ported to 3.x (as `CombineAdjacentTapTransformer`) and remains commented out. This is also what I would call a logical optimization ("they are equivalent, so why not re-use one of them"). (I think the nomenclature here might also be re-thought, logical vs physical vs etc)

I think you personally fought this issue a couple years ago as it was not obvious what was happening and led to very weird results and a great deal of frustration.

I do think we could pull down logical optimizations that users can turn on/off via a flag, but that isn't a feature of the planner at this time outside of just swapping out a whole new `RuleRegistry` with useful rules.

As a practical matter, we would have to run the full test suite (6.5 hours) twice with optimizations enabled, along with new tests just to test them -- this leads to an opportunity to re-think how the test suite works to speed things up if possible (we current run the plans and compare results across platforms).

My last point is that the current shipping rules are one-size-fit-all. I recommend higher order frameworks (which naturally work with some space of possible plans. e.g only ever have one sink per Flow, etc) ship with rules that optimize out artifacts (like the above), remove any unnecessary rules that will never apply, and plug in specific optimizations that make sense in the current context (but are problematic otherwise).

johnynek:
Thanks for the quick reply.

I just mean, such a rule might be good for us to add to the cascading code
base rather than in scalding, since others may often want to apply it.

cchepelov:
Thanks for the very detailed reply Chris — the custom rule registry does 
ring a bell ;)

In the short term, it looks to me that wip-7 will still require this PR 
"as is", but might have an interesting performance impact. Will give 
wip-7 + this PR tomorrow.

cwensel:
Without it being applied/verified via existing tests, it will just be code hanging around that Scalding is dependent on. 

I would recommend keeping it in Scalding, or creating a new (sub) project to hold rules and registries people can apply. The later is worthing thinking about if there are enough such rules/registries.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -36,7 +36,23 @@ object RichPipe extends java.io.Serializable {
 
   def getNextName: String = "_pipe_" + nextPipe.incrementAndGet.toString
 
-  def assignName(p: Pipe): Pipe = new Pipe(getNextName, p)
+  private val FormerNameBitLength = 8
+  private val FormerNameBitPattern = "^_pipe_([0-9]+)".r
+  private def getFormerNameBit(p: Pipe): String =
```
johnynek: I like this, but i wonder why not just do:

``` scala
def getFormerNameBit(p: Pipe): Option[Int] = p.getName match {
  case FormerNameBitPattern(pipeNumber) => Some(pipeNumber.toInt)
  case _ => None
}
```

Then use it with: `new Pipe(s"$getNextName-" + getFormerNameBit(p).fold(p.getName)(_.toString))`

johnynek (commented):
I like this. Thanks for the changes. Made one comment to simplify the new naming scheme (and I have some minor anxiety about mixing that with this PR as it seems good, but unrelated).

Maybe we should make a follow up PR to keep it separate in the history?

johnynek:
sounds good.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
cchepelov: meta: indeed (yes, riding on a PR isn't very good form). I'll move that tomorrow morning (UTC+2)
content: indeed too (pattern matching on extractors beyond Some(_) isn't yet on my toolbox' top drawer), will do it that way in the other PR.

Thanks for the feedback!

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -157,6 +175,59 @@ object RichPipe extends java.io.Serializable {
...
+    * @param tail a list of other pipes to be merged within the first
+    * @return a Merge of input pipes with any name clashes removed, or the input pipe if there was only one
+    */
+  private[scalding] def mergeAvoidingNameClashes(head: Pipe, tail: List[Pipe]): Pipe = tail match {
+    case Nil => head // avoid generating new Merge(pipes.head)
+    case _ =>
+      val (result, buf) = tail.foldLeft((List[Pipe](head), Set[String](head.getName))) {
+        case ((result, names), p) =>
+          if (names.contains(p.getName))
+            (assignName(p) :: result, names) // assignName is guaranteed unique and never assigned again
```
piyushnarang: do we need add the assignName(p) values to names as well?

piyushnarang (commented):
Would it make sense to assert things like number of expected steps / details of steps in the PlatformTests? I think we've done it for some of our other tests. Reason I bring it up is that part of what these changes are doing is avoiding checkpoints in hashJoins when they can be avoided. So if we could add verifications for that, it'll help us not regress in the future.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
cchepelov: moved into https://github.com/twitter/scalding/pull/1608

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
cchepelov: we don't, since the result of assignName() is guaranteed unique (so it won't collide with pre-existing items from names) and will never be assigned again (so no pipe yet to be evaluated in the foldLeft will have a name that collides with the freshly-assigned pipe name — unless the developer's being intentionally naming a pipe _pipe_42, but that's a case of "doctor, my eye hurts when I drink coffee while leaving the spoon in the mug" that has always existed). 

I guess I need to pre-pend: "no need to add new name to 'names': " to the comment

johnynek:
:+1: from me

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
piyushnarang: Yeah would be nice to reiterate with a comment to folks reading the code in the future.

piyushnarang:
Changes look good to me. Had a question on the platform tests - would it make sense to assert things like number of expected steps / details of steps in the PlatformTests? I think we've done it for some of our other tests. Reason I bring it up is that part of what these changes are doing is avoiding checkpoints in hashJoins when they can be avoided. So if we could add verifications for that, it'll help us not regress in the future.

johnynek (approved):

johnynek:
I left this open for comments, but it has been > 2 weeks with two +1s. So, I assume we are fine.

Sorry for the delay @cchepelov

PR #1607, opened 2016-10-06
Bump Cascading3 version + add regression test

piyushnarang:
Updating the cascading3 version to include the fix for a user job that was failing due to a planner error. Simplified the user job a bit and added it as a regression test (which fails on the older wip). 
Cascading PR for the issue: https://github.com/cwensel/cascading/pull/57
Shall be running some E2E and other tests and shall report back once they're done.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -305,7 +305,7 @@ lazy val scaldingArgs = module("args")
 lazy val scaldingDate = module("date")
 
 lazy val cascadingVersion =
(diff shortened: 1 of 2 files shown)


piyushnarang:
Verified that our internal E2E tests succeed and the user job that had this error also succeeds.

johnynek:
👍

piyushnarang:
Thanks for taking a look @johnynek :-)

PR #1608, opened 2016-10-07
pipes made with RichPipe.assignName retain a bit of name from parent

cchepelov:
(moved & improved from https://github.com/twitter/scalding/pull/1605#pullrequestreview-3185803 )

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
@@ -36,7 +36,26 @@ object RichPipe extends java.io.Serializable {
 
   def getNextName: String = "_pipe_" + nextPipe.incrementAndGet.toString
 
-  def assignName(p: Pipe) = new Pipe(getNextName, p)
+  private[scalding] val FormerNameBitLength = 12
+  private[scalding] val FormerAssignedPipeNamePattern = "^_pipe_([0-9]+).*$".r
+  private[scalding] val FromUuidPattern = "^.*[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-([0-9a-f]{12}).*$".r
+
+  // grab some bit of the previous pipe name to help walk up the graph across name assignments
+  private def getFormerNameBit(p: Pipe): String = p.getName match {
+    case FormerAssignedPipeNamePattern(pipeNumber) => pipeNumber
+    case FromUuidPattern(lastGroup) => lastGroup /* 12 characters */
+    case s if s.length > FormerNameBitLength => s.substring(s.length - FormerNameBitLength, s.length)
+    case s => s
+  }
+
+  /**
+   * Assign a new, guaranteed-unique name to the pipe.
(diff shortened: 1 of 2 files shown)


johnynek:
:+1:

piyushnarang:
Could we add some unit tests to check / verify the behavior? Also helps folks understand the behavior with a quick glance.

johnynek:
+1 for a test (though I really like the feature. People often get confused by the `_pipe_12` style names. This give some indications.

cchepelov:
+1 on the tests ;) will prepare something shortly. Is it sufficient to 
whitebox-test RichPipe.assignName, or is it necessary to test when/whether 
pipes get renamed as part of CoGroup/Merge (I'm less firm on the best 
policy there TBH)

Envoyé avec AquaMail pour Android
http://www.aqua-mail.com

piyushnarang:
I think whitebox testing the method behavior should be fine. @johnynek feel free to chime in if you'd like something else.

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
```diff
@@ -36,7 +36,26 @@ object RichPipe extends java.io.Serializable {
...
-  def assignName(p: Pipe) = new Pipe(getNextName, p)
+  private val FormerNameBitLength = 12
+  private val FormerAssignedPipeNamePattern = "^_pipe_([0-9]+)".r
+  private val FromGuidPattern = "[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-([a-z0-9]{12})".r
+
+  // grab some bit of the previous pipe name to help walk up the graph across name assignments
+  private def getFormerNameBit(p: Pipe): String = p.getName match {
+    case FormerAssignedPipeNamePattern(pipeNumber) => pipeNumber
+    case FromGuidPattern(lastGroup) => lastGroup /* 12 characters */
+    case s if s.length > FormerNameBitLength => s.substring(s.length - FormerNameBitLength, s.length)
```
ulyssepence, line 47: When would we expect the number to be too long?

ulyssepence:
+1 on adding tests

scalding-core/src/main/scala/com/twitter/scalding/RichPipe.scala
cchepelov: The number would never be too long in the case of already auto-assigned names such as "_pipe_123" (that'd mean way much than Int.MaxValue _pipes_ created within the DAG) 

However, some pipe names will not be named _pipe_123 (e.g. the input heads); those may or may not include a UUID within (most versions of Cascading include a UUID within the name string, but that's a convenience we should not **depend** on; even though when it's here, it's a pretty good signifier)
Those that do, we take the last digit group from the UUID. For those last ones, that are neither named _pipe_123 or include a UUID (presumably, manually-named pipes), then we take the last 12 characters.

cchepelov:
Tests added (didn't find an existing place for RichPipe-specific tests, added one)

piyushnarang:
👍

Issue #1294, opened 2015-05-15
Remove use of reflection to access private field in ExecutionContext.updateFlowStepName


aposwolsky:
blocked on updating to latest version of cascading:

Reflection is used to access "setName" in BaseFlowStep for WithDescription functionality of naming MR steps.  In the latest version of cascading this is a public method.
https://github.com/cwensel/cascading/commit/487a6e9ef#diff-0feab84bc8832b2a39312dbd208e3e69L175

PR #1609, opened 2016-10-11
Remove two obsolete uses of reflection

cchepelov:
(fixes: #1294)

This PR removes two obsolete uses of reflection, which were justified by the need for back-compatibility across versions of Cascading 2.x

On the cascading3 branch, these uses are no longer required.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
@@ -18,6 +18,7 @@ package com.twitter.scalding
 
 import java.lang.reflect.InvocationTargetException
 
+import cascading.util.TraceUtil
 import org.slf4j.{ Logger, LoggerFactory => LogManager }
 
 /**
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Tracing.scala
```diff
@@ -58,7 +55,7 @@ object Tracing {
    * tracing boundary. Normally not needed, but may be useful
    * after a call to unregister()
    */
-  def register(regex: String = defaultRegex) = invokeStaticMethod(traceUtilClassName, "registerApiBoundary", regex)
+  def register(regex: String = defaultRegex) = TraceUtil.registerApiBoundary(regex)
```
cchepelov, line 58: The resulting object can seem quite thin now; however, as it's been part of the public scalding API, I assume it's best to keep it? It has one added value compared to c.u.TraceUtil, which is to know what are the scalding-related API boundaries.

johnynek (commented):
:+1:

PR #1610, opened 2016-10-13
Update Scala and sbt version

zaneli:
- Update Scala version to 2.11.8
- Update sbt version to 0.13.12
- FIx `org.scalamacros#paradise` version.  
  https://travis-ci.org/zaneli/scalding/jobs/167170589#L1780

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
       env: BUILD="base" TEST_TARGET="scalding-args scalding-date maple"
       script: "scripts/run_test.sh"
 
-    - scala: 2.11.7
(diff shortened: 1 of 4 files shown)


piyushnarang:
Changes look good to me. Not sure if we need to continue publishing 2.11.7 artifacts as well. I'd expect that we should be good with 2.11.8 but not entirely sure.

johnynek:
This is great. Thanks. (Piyush: we don't need to publish different minor
scala versions. They are compatible).

piyushnarang:
Ok makes sense. 👍

PR #1613, opened 2016-10-27
Add options for verbose logging in FileSource

isnotinvain:
For a long time we've had issues with getting the data availability checks right in FileSource. It's often hard to debug reports of intermittent errors in FileSource's behavior, because by the time we get to take a look, the HDFS file system may have changed already.

This PR adds a hadoop config variable that can be enabled which adds some very verbose logging, which sort of shows "what the file system looked like" when a job was run and when it was doing its various validations / sanity checks.

It's a little bit messy to have this logging code inside of FileSource but I do think it will help us going forward. Right now we are trying to debug some issues with the new behavior that allows for empty directories that contain success files. It may be doing exactly what it's supposed to, or it may have a bug, but we can't really catch it as it happens. This way we'll at least have the option to know what the FS looked like at run time.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -112,16 +112,15 @@ object AcceptAllPathFilter extends PathFilter {
 object FileSource {
   val LOG = LoggerFactory.getLogger(this.getClass)
 
-  val VerboseFileSourceLoggingKey = "com.twitter.scalding.filesource.verbose.logging"
+  private[this] def verboseLogEnabled(conf: Configuration): Boolean =
+    conf.getBoolean(Config.VerboseFileSourceLoggingKey, false)
 
-  def verboseLogEnabled(conf: Configuration): Boolean = conf.getBoolean(VerboseFileSourceLoggingKey, false)
-
-  def ifVerboseLog(conf: Configuration)(msgFn: => String): Unit = {
+  private[this] def ifVerboseLog(conf: Configuration)(msgFn: => String): Unit = {
     if (verboseLogEnabled(conf)) {
-      val stack = new RuntimeException()
+      val stack = Thread.currentThread
         .getStackTrace
         .iterator
-        .drop(1) // skip this method
+        .drop(2) // skip getStackTrace and ifVerboseLog
         .mkString("\n")
 
       // evaluate call by name param once
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -15,33 +15,24 @@ limitations under the License.
 */
 package com.twitter.scalding
 
-import java.io.{ File, InputStream, OutputStream }
-import java.util.{ UUID, Properties }
+import java.io.{ InputStream, OutputStream }
+import java.util.{ Properties, UUID }
 
 import cascading.scheme.Scheme
-import cascading.scheme.local.{ TextLine => CLTextLine, TextDelimited => CLTextDelimited }
-import cascading.scheme.hadoop.{
-  TextLine => CHTextLine,
-  TextDelimited => CHTextDelimited,
-  SequenceFile => CHSequenceFile
-}
+import cascading.scheme.hadoop.{ SequenceFile => CHSequenceFile, TextDelimited => CHTextDelimited, TextLine => CHTextLine }
+import cascading.scheme.local.{ TextDelimited => CLTextDelimited, TextLine => CLTextLine }
+import cascading.tap.{ MultiSourceTap, SinkMode, Tap }
 import cascading.tap.hadoop.Hfs
-import cascading.tap.MultiSourceTap
-import cascading.tap.SinkMode
-import cascading.tap.Tap
 import cascading.tap.local.FileTap
 import cascading.tuple.Fields
-
 import com.etsy.cascading.tap.local.LocalTap
-import com.twitter.algebird.{ Semigroup, MapAlgebra }
-
+import com.twitter.algebird.{ MapAlgebra, Semigroup }
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.{ FileStatus, PathFilter, Path }
-import org.apache.hadoop.mapred.JobConf
-import org.apache.hadoop.mapred.OutputCollector
-import org.apache.hadoop.mapred.RecordReader
+import org.apache.hadoop.fs.{ FileStatus, Path, PathFilter }
+import org.apache.hadoop.mapred.{ JobConf, OutputCollector, RecordReader }
+import org.slf4j.LoggerFactory
 
-import scala.util.{ Try, Success, Failure }
+import scala.util.{ Failure, Success, Try }
 
 /**
  * A base class for sources that take a scheme trait.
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -119,6 +110,32 @@ object AcceptAllPathFilter extends PathFilter {
 }
 
 object FileSource {
+  val LOG = LoggerFactory.getLogger(this.getClass)
```
johnynek, line 113: have you tested this on the cluster? Seems like serialization woes are possible here.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -119,6 +110,32 @@ object AcceptAllPathFilter extends PathFilter {
...
+
+  val VerboseFileSourceLoggingKey = "com.twitter.scalding.filesource.verbose.logging"
+
+  def verboseLogEnabled(conf: Configuration): Boolean = conf.getBoolean(VerboseFileSourceLoggingKey, false)
```
johnynek: we try to use `scalding.Config` where possible. Can we use that instead?

johnynek: that or make this private please.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -119,6 +110,32 @@ object AcceptAllPathFilter extends PathFilter {
...
+
+  def ifVerboseLog(conf: Configuration)(msgFn: => String): Unit = {
```
johnynek: can this take `shouldLog: Boolean` instead? And maybe make a

``` scala
private def ifVerboseConf(conf: Configuration)(msg: => String): Unit =
  ifVerboreLog(verboseLogEnabled(conf))
```

to minimize exposing `Configuration` on our APIs.

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -7,7 +7,7 @@ import java.net.URI
 import java.nio.ByteBuffer
 
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.mapreduce.filecache.{DistributedCache => HDistributedCache}
+import org.apache.hadoop.mapreduce.filecache.{ DistributedCache => HDistributedCache }
```
johnynek, line 10: can we not commit style changes like this on the side?

scalding-core/src/test/scala/com/twitter/scalding/RichPipeSpecification.scala
```diff
@@ -59,7 +59,6 @@ object RichPipeSpecification extends Properties("RichPipe") {
       (newNum > basePipeNumber) && ap.getName.endsWith(s"-${lastGroup}")
   }
 
-
```
johnynek, line 62: can we remove this from the diff?

johnynek (changes requested):

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -119,6 +110,32 @@ object AcceptAllPathFilter extends PathFilter {
 }
 
 object FileSource {
+  val LOG = LoggerFactory.getLogger(this.getClass)
+
+  val VerboseFileSourceLoggingKey = "com.twitter.scalding.filesource.verbose.logging"
```
piyushnarang: should we declare this in `Config`?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -119,6 +110,32 @@ object AcceptAllPathFilter extends PathFilter {
...
+
+  def verboseLogEnabled(conf: Configuration): Boolean = conf.getBoolean(VerboseFileSourceLoggingKey, false)
+
+  def ifVerboseLog(conf: Configuration)(msgFn: => String): Unit = {
+    if (verboseLogEnabled(conf)) {
+      val stack = new RuntimeException()
```
piyushnarang: could we instead do `Thread.currentThread().getStackTrace()` ?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
piyushnarang: maybe add a short line describing the use of this param? Add if we have a sense of the overhead it causes, that as well?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: I have the same worry. I'm in the middle of testing it out. It's also really hard to get these logs to show up in the tests. Something to do w/ slf4j configuration. So I also want to confirm these logs actually get printed on a real cluster.

I _think_ this'll work, because I see it being done the same way in the reducer estimator and a few other places. Worse case I'll make it lazy transient.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: I can make it private. Is it better to use scalding.Config even if it's private? It seems that what I'll have access to is a Configuration -- does it make sense to convert that to a scalding.Config (probably not?)

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: these can actually all be private, or maybe private[scalding] -- would that be enough?

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
isnotinvain: Ah, so this is what happens when you run the unit tests (scalafmt) -- I wasn't sure whether to keep this. If I don't commit this, next time someone runs the unit tests, it'll do this again. So I thought I'd just keep the changes. I can back them out though and send a separate PR that only does the style fixes.

scalding-core/src/test/scala/com/twitter/scalding/RichPipeSpecification.scala
isnotinvain: same as above, what do you think?

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: yeah good idea, i'll document this

isnotinvain: yes maybe, let me see how that works

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
isnotinvain: Maybe -- I need to check if there's a difference. I think there might be but I'm not 100% sure.

isnotinvain:
OK -- updated to have comments + to use scalding.Config (to some extent) and to make the helper methods private. Let me know what you think about the style changes / if the should be removed.

piyushnarang:
👍 if you were able to verify that this works on the cluster

isnotinvain:
OK I've tested this on the cluster as well. It's pretty noisy -- we could log this to a side file instead maybe.

johnynek (approved):

johnynek:
👍

isnotinvain:
Thanks! Any opinion on log-to-console vs log-to-side-file?

johnynek:
I think we should just use normal terrible java logging. If you want to log
to a file, you set that up in the terrible configuration of slf4j.

I don't see myself running this option, so I am not too worried about it as
long as the default is a no-op.

isnotinvain:
Ya makes sense. I could have done all the turning on/off of this via terrible slf4j configs but I didn't want to do that because it never really works / often gets misconfigured outside of your control. So that's why I used the hadoop conf. I'll leave this as is.

PR #1617, opened 2016-11-04
(🚧🚜 WIP  👷🚧 ) Split fabric-specific code into individual back-ends

cchepelov:
(as being discussed on scalding-dev)

In the things I've left out for now (in addition to the list in the e-mail): scalding-hadoop-test (platform-specific tests on a real minicluster). 
Perhaps it'd make sense to have run / runHadoop (→runFabricLocal ?) / runOnMiniCluster in JobTest, and the relevant Mode implementation assortment?


scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -15,41 +15,53 @@ limitations under the License.
...
 import scala.collection.JavaConverters._
 import scala.collection.mutable.Buffer
 import scala.collection.mutable.{ Map => MMap }
 import scala.collection.mutable.{ Set => MSet }
 import scala.util.{ Failure, Success }
-
 import org.slf4j.LoggerFactory
 
+import scala.annotation.meta.param
+import scala.collection.{ Map, mutable }
```
oscar-stripe, line 50: can we avoid `scala.collection.Map` and explicitly use `Map` for immutable map (standard) and `mutable.Map` when we need a mutable one?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -15,41 +15,53 @@ limitations under the License.
...
 object Mode {
+  private val LOG = LoggerFactory.getLogger(this.getClass)
+
   /**
    * This is a Args and a Mode together. It is used purely as
    * a work-around for the fact that Job only accepts an Args object,
    * but needs a Mode inside.
    */
-  private class ArgsWithMode(argsMap: Map[String, List[String]], val mode: Mode) extends Args(argsMap) {
+  private class ArgsWithMode(argsMap: scala.Predef.Map[String, List[String]], val mode: Mode) extends Args(argsMap) {
```
oscar-stripe, line 64: I'd rather not hide `Map` with an import or use `scala.collection.immutable.Map` here.

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
-      config.set(CascadingFlowConnectorClassKey, DefaultHadoop2Mr1FlowConnector)
-      config.set(CascadingFlowProcessClassKey, DefaultHadoop2Mr1FlowProcess)
-      Hdfs(strictSources, config)
-    } else if (args.boolean("hadoop2-tez")) {
-      config.set(CascadingFlowConnectorClassKey, DefaultHadoop2TezFlowConnector)
-      config.set(CascadingFlowProcessClassKey, DefaultHadoop2TezFlowProcess)
-      Hdfs(strictSources, config)
-    } else
-      throw ArgsException("[ERROR] Mode must be one of --local, --hadoop1, --hadoop2-mr1, --hadoop2-tez or --hdfs, you provided none")
+    lazy val autoMode = if (args.boolean("autoCluster")) {
```
oscar-stripe, line 140: can we prefix new scalding options with `scalding.`?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+    must be loadable whether one of these is missing or not (it is unlikely all can be loaded at the same time)
+     */
+    "local" -> "com.twitter.scalding.Local",
+    "hadoop2-mr1" -> "com.twitter.scalding.Hadoop2Mr1Mode",
+    "hadoop2-tez" -> "com.twitter.scalding.TezMode", "tez" -> "com.twitter.scalding.TezMode",
+    "hadoop1" -> "com.twitter.scalding.LegacyHadoopMode",
+    "hdfs" -> "com.twitter.scalding.LegacyHadoopMode",
+    "flink" -> "com.twitter.scalding.FlinkMode")
+
+  val KnownTestModesMap = Seq(
```
oscar-stripe, line 89: can we make these maps instead?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+    "local-test" -> "com.twitter.scalding.Test",
+    "hadoop1-test" -> "com.twitter.scalding.HadoopTest",
+    "hadoop2-mr1-test" -> "com.twitter.scalding.Hadoop2Mr1TestMode",
+    "tez-test" -> "com.twitter.scalding.TezTestMode",
+    "flink-test" -> "com.twitter.scalding.FlinkTestMode")
+  // TODO: define hadoop2-mr1 (easy), tez and flink (less easy) classes.
+
+  private def getModeConstructor[M <: Mode](clazzName: String, types: Class[_]*) =
```
oscar-stripe, line 97: I wonder if we should just have a contract that a `Mode` needs a constructor that takes exactly one `Args` and exactly one `Configuration`, or even, possibly `Config` which is immutable. Wouldn't that work? I don't want to add too much flexibility if we don't need it.

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+  // access this through the DistributedCacheFile object
+  private[scalding] def addCachedFile(file: UncachedFile): CachedFile
+
+  /** Read the content of a file into a string, UTF-8 encoding is assumed */
+  def readFromFile(filename: String): String
+  /** Read the content of a file into a string, UTF-8 encoding is assumed */
+  def writeToFile(filename: String, text: String): Unit
+}
+
+object StorageMode {
```
oscar-stripe, line 253: I don't think this should be in this file. Can we break `StorageMode` out? Also can we put it in a new namespace/dir? `scalding.fabric` or something?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+  trait TemporarySource[T] {
+    def sink(conf: Config): TypedSink[T]
+    def downstreamPipe(conf: Config): TypedPipe[T]
```
oscar-stripe, line 256: I don't really know what this means. Can you comment?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+  }
+}
+
+trait Mode extends java.io.Serializable {
```
oscar-stripe, line 260: can we add a comment about the requirement of constructors for reflective construction?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+
+  /**
+   * Using a new FlowProcess, which is only suitable for reading outside
+   * of a map/reduce job, open a given tap and return the TupleEntryIterator
+   */
+  final def openForRead(config: Config, tap: Tap[_, _, _]): TupleEntryIterator = executionMode.openForRead(config, tap)
+
+  /** Create a new FlowConnector for this cascading planner */
+  protected // will probably have to end up just "@deprecated"
+  final def newFlowConnector(config: Config): FlowConnector = newFlowConnector(config)
```
oscar-stripe, line 278: this looks like an infinite loop. What am I missing?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+}
 
-  // Returns true if the file exists on the current filesystem.
-  def fileExists(filename: String): Boolean
-  /** Create a new FlowConnector for this cascading planner */
-  def newFlowConnector(props: Config): FlowConnector
+/**
+ * Any Mode that runs using local resources only
+ */
+trait LocalMode extends Mode {
```
oscar-stripe, line 296: no need for `{ }`

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+}
+
+/**
+ * Any Mode that runs over a cluster (as opposed to Local) is a ClusterMode.
+ */
+trait ClusterMode extends Mode {
 }
 
-trait HadoopMode extends Mode {
```
oscar-stripe, line 124: we have user code that assumes this. Can we find a way not to break those folks?

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+ * Any ClusterMode whose main configuration is org.apache.hadoop.conf.Configuration is a HadoopFamilyMode
+ */
+trait HadoopFamilyMode extends ClusterMode {
+  def jobConf: Configuration
+}
+/**
+ * The "HadoopMode" is actually an alias for "a mode running on a fabric that ultimately runs using an execution
+ * engine compatible with some of the Hadoop technology stack (may or may not include Hadoop 1.x, YARN, etc.)
+ */
+trait HadoopExecutionModeBase[ConfigType <: Configuration]
```
oscar-stripe, line 315: maybe we actually want an abstract type here:

``` scala
trait HadoopExecutionModeBase {
  type ConfigType <: Configuration
}
```

since then we can refer to it: `mode.ConfigType`. We might do this to put more of the cascading types and not use `_` so much.

scalding-core/src/main/scala/com/twitter/scalding/Mode.scala
```diff
@@ -63,72 +75,258 @@ object Mode {
...
+  extends ExecutionMode {
   def jobConf: Configuration
```
oscar-stripe, line 317: should this not return `ConfigType`?

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -48,24 +52,76 @@ object StatKey {
...
+    whenever a join between pipes is possible, which would cause significant changes to the API
+
+   */
+  def apply(fp: FlowProcess[_], statKey: StatKey): CounterImpl = {
+    /* TODO: comment this stuff. */
+    val klassName = Option(fp.getStringProperty(CounterImplClass)).getOrElse(sys.error(CounterImplClass + " property is missing"))
+
+    val memo = scala.collection.mutable.Map[String, Constructor[CounterImpl]]()
+    val ctor = memo.synchronized {
+      memo.getOrElse(klassName,
```
oscar-stripe, line 86: let's use `getOrElseUpdate` rather than the nested put.

scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
```diff
@@ -48,24 +52,76 @@ object StatKey {
...
+          val ctor = klass.getConstructor(classOf[FlowProcess[_]], classOf[StatKey])
+          memo.put(klassName, ctor)
+          ctor
+        })
     }
+    ctor.newInstance(fp, statKey)
+  }
+
+  private[scalding] def upcast[T <: FlowProcess[_]](fp: FlowProcess[_])(implicit ev: TypeTag[T]): T = fp match {
+    case hfp: T @unchecked if (ev == typeTag[T]) => hfp // see below
```
oscar-stripe, line 98: this `typeTag[T]` will always match `ev` no? I think so. Why not just explicitly cast in this method:

``` scala
def downCast[T <: U](u: U): T = u.asInstanceOf[T]
```

Also, this is a downcast, no?

oscar-stripe (changes requested):
Great work.

One question: is there any way we can reduce the size of the PR.

For instance, can we keep flink out and just do the existing fabrics? That will make is easier to review. Once we get it in, we can add more.

I like this approach and think it is the most sensible way to get the most out of cascading 3 without breaking users.

PR #1618, opened 2016-11-04
Use correct fields when returning an empty tap in FileSource

isnotinvain:
When using Fields.ALL in IterableSource (when FileSource is returning an empty tap) everything works in the typed api. But in the fields api, we need to actually respect the fields being used and propegate them from the scheme to the tap.

I've added a test that fails in develop and passes with this fix.

The failure looks something like this:

```
cascading.flow.planner.PlannerException: could not build flow from assembly: [could not select fields: [{1}:'keywordNew'], from: [{?}:UNKNOWN]]
```

This does not come up when using something like `mapTo` but it does in `joinWithSmaller` so it seems related to operations that need to pull a field out by name.

### DIFF
--- a/scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
+++ b/scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
@@ -535,7 +535,7 @@ class PlatformTest extends WordSpec with Matchers with HadoopSharedPlatformTest 
           val steps = flow.getFlowSteps.asScala
           steps should have size 1
           val firstStep = steps.headOption.map(_.getConfig.get(Config.StepDescriptions)).getOrElse("")
-          val lines = List(16, 18, 19, 22, 24).map { i =>
+          val lines = List(16, 18, 19, 22, 23).map { i =>
             s"com.twitter.scalding.platform.TypedPipeJoinWithDescriptionJob.<init>(TestJobsWithDescriptions.scala:$i"
           }
           firstStep should include ("leftJoin")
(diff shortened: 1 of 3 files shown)


scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
```diff
@@ -437,11 +435,12 @@ object PlatformTest {
   }
 }
 
-class TestEmptySource extends FileSource with Mappable[(String, Long)] with SuccessFileSource {
+class TestEmptySource(val fields: Fields = new Fields("customNamedOffset", "customNamedLine")) extends FileSource with Mappable[(String, Long)] with SuccessFileSource {
```
piyushnarang: should we create a fields version of this source to decouple the fields and typed source?
e.g.

```
// use this in the existing tests
class TestEmptySource extends FileSource with Mappable[(String, Long)] with SuccessFileSource {
..
} 

// use in the new test
class TestFieldsEmptySource(val fields: Fields = new Fields("customNamedOffset", "customNamedLine")) extends FileSource with Mappable[(String, Long)] with SuccessFileSource {
...}
```

piyushnarang (approved):
Minor decoupling comment, but looks good to me otherwise.

scalding-hadoop-test/src/test/scala/com/twitter/scalding/platform/PlatformTest.scala
isnotinvain: Yeah I can separate them

isnotinvain:
OK some tests that care about line numbers failed. I'll update them.

isnotinvain:
I've pushed the changes to split up the two sources -- waiting for travis to run the tests and tell me the new line numbers needed for the two failing tests :)

isnotinvain:
OK -- I moved the classes whose line numbers we care about to a separate file, that should make this a little less fragile from now on. Letting travis re-run the tests, but I think this should be good to go now.

piyushnarang:
@isnotinvain, thanks for updating! 👍

PR #1621, opened 2016-11-14
Clean up temporary files created by forceToDiskExecution

### LINKED ISSUE #1615: forceToDiskExecution does not clean up data after the execution has completed
We assume that people have tmpdir cleaning set up, but not everyone does.

It would be nice to delete any files created with forceToDiskExecution after the executions have totally completed (when the EvalCache is stopped):

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L151

ajohnson-stripe-zz:
This adds functionality for `WriteExecution` to keep track of temporary files created and then delete them after execution has finished. Currently this is only used by `forceToDiskExecution`, which previously would leave around temporary files every time it was used.

This fixes #1615.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -338,7 +338,7 @@ object Execution {
     type Counters = Map[Long, ExecutionCounters]
     private[this] val cache = new FutureCache[(Config, Execution[Any]), (Any, Counters)]
     private[this] val toWriteCache = new FutureCache[(Config, ToWrite), Counters]
-    protected[EvalCache] val filesToCleanup = mutable.Set[String]()
+    private[this] val filesToCleanup = mutable.Set[String]()
 
     // This method with return a 'clean' cache, that shares
     // the underlying thread and message queue of the parent evalCache
@@ -346,7 +346,7 @@ object Execution {
       val self = this
       new EvalCache {
         override protected[EvalCache] val messageQueue: LinkedBlockingQueue[EvalCache.FlowDefAction] = self.messageQueue
-        override protected[EvalCache] val filesToCleanup: mutable.Set[String] = self.filesToCleanup
+        override def addFilesToCleanup(files: TraversableOnce[String]): Unit = self.addFilesToCleanup(files)
         override def start(): Unit = sys.error("Invalid to start child EvalCache")
         override def finished(mode: Mode): Unit = sys.error("Invalid to finish child EvalCache")
       }
@@ -433,7 +433,9 @@ object Execution {
       res: => Future[(T, Counters)]): Future[(T, Counters)] =
       getOrElseInsertWithFeedback(cfg, ex, res)._2
 
-    def addFilesToCleanup(files: TraversableOnce[String]): Unit = filesToCleanup ++= files
+    def addFilesToCleanup(files: TraversableOnce[String]): Unit = filesToCleanup.synchronized {
+      filesToCleanup ++= files
+    }
   }
 
   private case class FutureConst[T](get: ConcurrentExecutionContext => Future[T]) extends Execution[T] {
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -346,7 +346,7 @@ object Execution {
       val self = this
       new EvalCache {
         override protected[EvalCache] val messageQueue: LinkedBlockingQueue[EvalCache.FlowDefAction] = self.messageQueue
-        override protected[this] val filesToCleanup: mutable.Set[String] = self.filesToCleanup
+        override protected[EvalCache] val filesToCleanup: mutable.Set[String] = self.filesToCleanup
         override def start(): Unit = sys.error("Invalid to start child EvalCache")
         override def finished(mode: Mode): Unit = sys.error("Invalid to finish child EvalCache")
       }
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -338,7 +338,7 @@ object Execution {
     type Counters = Map[Long, ExecutionCounters]
     private[this] val cache = new FutureCache[(Config, Execution[Any]), (Any, Counters)]
     private[this] val toWriteCache = new FutureCache[(Config, ToWrite), Counters]
-    private[this] val filesToCleanup = mutable.Set[String]()
+    protected[EvalCache] val filesToCleanup = mutable.Set[String]()
 
     // This method with return a 'clean' cache, that shares
     // the underlying thread and message queue of the parent evalCache
@@ -346,6 +346,7 @@ object Execution {
       val self = this
       new EvalCache {
         override protected[EvalCache] val messageQueue: LinkedBlockingQueue[EvalCache.FlowDefAction] = self.messageQueue
+        override protected[this] val filesToCleanup: mutable.Set[String] = self.filesToCleanup
         override def start(): Unit = sys.error("Invalid to start child EvalCache")
         override def finished(mode: Mode): Unit = sys.error("Invalid to finish child EvalCache")
       }
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -24,6 +24,10 @@ import scala.concurrent.{ Await, Future, ExecutionContext => ConcurrentExecution
 import scala.util.{ Failure, Success, Try }
 import scala.util.control.NonFatal
 import cascading.flow.{ FlowDef, Flow }
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{ FileSystem, Path }
+import org.slf4j.LoggerFactory
+
 import scala.collection.mutable
 import scala.runtime.ScalaRunTime
 
@@ -148,7 +152,7 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
     // get on Trampoline
     val result = runStats(confWithId, mode, ec)(cec).get.map(_._1)
     // When the final future in complete we stop the submit thread
-    result.onComplete { _ => ec.finished() }
+    result.onComplete { _ => ec.finished(mode) }
     // wait till the end to start the thread in case the above throws
     ec.start()
     result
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -713,11 +742,11 @@ object Execution {
      */
     override def zip[U](that: Execution[U]): Execution[(T, U)] =
       that match {
-        case WriteExecution(h, t, otherFn) =>
+        case WriteExecution(h, t, otherFn, tempFiles) =>
           val newFn = { (conf: Config, mode: Mode) =>
             (fn(conf, mode), otherFn(conf, mode))
           }
-          WriteExecution(head, h :: t ::: tail, newFn)
+          WriteExecution(head, h :: t ::: tail, newFn, tempFiles)
```
oscar-stripe: don't you need to merge the sets from the left and right? Seems like we are losing `this.tempFilesToCleanup`.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
```diff
@@ -69,6 +68,29 @@ class WordCountEc(args: Args) extends ExecutionJob[Unit](args) {
...
+    }
+    val readFn = { (conf: Config, mode: Mode) =>
+      TypedPipe.from(TypedTsv[(String, String)](args("input")))
+    }
+
+    val filesToDeleteFn = { (conf: Config, mode: Mode) =>
+      Set(tempFile)
+    }
+
+    Execution.write(writeFn, readFn, filesToDeleteFn)
```
oscar-stripe, line 72: can you test a `.zip` to be sure we don't lose files there?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -378,7 +385,26 @@ object Execution {
...
+      val fs = mode match {
+        case localMode: CascadingLocal => FileSystem.getLocal(new Configuration)
+        case hdfsMode: HadoopMode =>  FileSystem.get(hdfsMode.jobConf)
+      }
+
+      filesToCleanup foreach { file: String =>
+        try {
+          val path = new Path(file)
+          if (fs.exists(path)) {
+            fs.delete(path, true)
```
oscar-stripe: can you comment what the boolean means at this call site?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -378,7 +385,26 @@ object Execution {
...
      */
-    def finished(): Unit = messageQueue.put(Stop)
+    def finished(mode: Mode): Unit = {
+      messageQueue.put(Stop)
+      val fs = mode match {
+        case localMode: CascadingLocal => FileSystem.getLocal(new Configuration)
+        case hdfsMode: HadoopMode =>  FileSystem.get(hdfsMode.jobConf)
+      }
+
+      filesToCleanup foreach { file: String =>
```
oscar-stripe: can we prefer the `.` style and use `filesToCleanup.foreach {`

oscar-stripe (changes requested):
This is great! Thanks for attacking this problem.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ajohnson-stripe-zz: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ajohnson-stripe-zz: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ajohnson-stripe-zz: Ah good catch, I do need that here.

ajohnson-stripe-zz:
Thanks for the  feedback @oscar-stripe. I incorporated in the changes you suggested.

I also noticed that the REPL expects that the temporary outputs from `forceToDiskExecution` stick around in things like the `snapshot` method on `ShellTypedPipe`. As such I moved the actual deletion to happen in a shutdown hook on JVM exit.

ajohnson-stripe-zz:
Hmm, I'm not sure I see why the CI build is failing. It's complaining about failing when compiling maple: `.[error] (maple/compile:compileIncremental) javac returned nonzero exit code`, but I only see warnings from the compiler, no errors. I can build that fine locally too.

I do see this in the log, maybe related?

```
.tar: Unexpected EOF in archive
tar: rmtlseek not stopped at a record boundary
tar: Error is not recoverable: exiting now
```

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -284,6 +288,34 @@ object Execution {
...
+      filesToCleanup.foreach { file: String =>
+        try {
+          val path = new Path(file)
+          if (fs.exists(path)) {
+            // The "true" parameter here indicates that we should recursively delete everything under the given path
+            fs.delete(path, true)
+          }
+        } catch {
+          // If we fail in deleting a temp file, log the error but don't fail the run
+          case e: Exception => LOG.info(s"Unable to delete temp file $file")
```
piyushnarang: can we pass e to the log statement? that way we get the message and stacktrace? Also should this be a warn?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -284,6 +288,34 @@ object Execution {
   }
 
   /**
+    * This is a Thread used as a shutdown hook to clean up temporary files created by some Execution
+    */
+  case class TempFileCleanup(filesToCleanup: mutable.Set[String], mode: Mode) extends Thread {
```
piyushnarang: since we're just iterating over the filesToCleanup in this case class, maybe we can make this a more generic interface rather than mutable.Set? Iterable / Seq / Set?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -284,6 +288,34 @@ object Execution {
...
+      }
+
+      filesToCleanup.foreach { file: String =>
+        try {
+          val path = new Path(file)
+          if (fs.exists(path)) {
+            // The "true" parameter here indicates that we should recursively delete everything under the given path
+            fs.delete(path, true)
+          }
+        } catch {
```
piyushnarang, line 311: as this is part of a shutdown hook, should we add a catch all clause?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
piyushnarang: maybe also call out that if the job is aborted this will not cleanup stuff (if folks haven't read / know about shutdown hook semantics)

oscar-stripe: +1 to taking an immutable `Iterable` here. Can we make the class private?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
oscar-stripe: Do you mean `catch { case e: Throwable =>` so we catch anything? Is that what you mean?

piyushnarang: @oscar-stripe yeah. If we for whatever reason throw an caught exception in this code that will result in the shutdown code not being run completely as I understand it.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ajohnson-stripe-zz: 👍

ajohnson-stripe-zz:
I've incorporated the next round of feedback. I'm still looking into it, but I'm not able to reproduce the CI build failure.

johnynek:
I restarted. Let's see if it is a CI issue.

piyushnarang:
@ajohnson-stripe changes look good to me. 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -313,7 +347,7 @@ object Execution {
       new EvalCache {
         override protected[EvalCache] val messageQueue: LinkedBlockingQueue[EvalCache.FlowDefAction] = self.messageQueue
         override def start(): Unit = sys.error("Invalid to start child EvalCache")
-        override def finished(): Unit = sys.error("Invalid to finish child EvalCache")
+        override def finished(mode: Mode): Unit = sys.error("Invalid to finish child EvalCache")
```
johnynek, line 351: don't we need to override the `addFilesToCleanup` to forward back to the parent list? I think any files added to the child will not be deleted, right?

ajohnson-stripe-zz: Yeah, that does need to be added

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -305,15 +338,17 @@ object Execution {
     type Counters = Map[Long, ExecutionCounters]
     private[this] val cache = new FutureCache[(Config, Execution[Any]), (Any, Counters)]
     private[this] val toWriteCache = new FutureCache[(Config, ToWrite), Counters]
+    protected[EvalCache] val filesToCleanup = mutable.Set[String]()
```
johnynek: actually, I think we should keep this private and override the add method. The reason is we forgot one more thing: this is not thread-safe, but we have so many threads in play, we should make this thread-safe.

I think we are modifying this value from several threads.

So, if we just forward the method, we can handle the locking a bit more cleanly inside that method.

ajohnson-stripe-zz: Good point, I'll switch that up.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -284,6 +288,35 @@ object Execution {
...
+      filesToCleanup.foreach { file: String =>
+        try {
+          val path = new Path(file)
+          if (fs.exists(path)) {
+            // The "true" parameter here indicates that we should recursively delete everything under the given path
+            fs.delete(path, true)
+          }
+        } catch {
+          // If we fail in deleting a temp file, log the error but don't fail the run
+          case e: Throwable => LOG.warn(s"Unable to delete temp file $file", e)
```
isnotinvain, line 313: style-wise I think `case e => LOG.warn(s"Unable to delete temp file $file", e)` is the same (without the `:Throwable` part)

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -284,6 +288,35 @@ object Execution {
   }
 
   /**
+    * This is a Thread used as a shutdown hook to clean up temporary files created by some Execution
+    *
+    * If the job is aborted the shutdown hook may not run and the temporary files will not get cleaned up
+    */
+  private[scalding] case class TempFileCleanup(filesToCleanup: Iterable[String], mode: Mode) extends Thread {
```
isnotinvain, line 295: should this extend Runnable instead? Does it have to extend Thread?

johnynek:
👍 looks good to me.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ajohnson-stripe-zz: `addShutdownHook` will only take a `Thread` unfortunately

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ajohnson-stripe-zz: It is the same, but leaving out the `: Throwable` generates a compiler warning requesting that you be explicit. I see a mix of both in the codebase now, but I'm happy to go with whatever's the preferred style.

johnynek:
For throwable we should be explicit and avoid the warning in my view.

PR #1626, opened 2016-11-29
Allow supplying better error messages in InvaidSourceTap/Exception

isnotinvain:
Often the constructor of InvalidSourceException / InvalidSourceTap has more info to provide than just the list of paths scanned. They may have an exception to chain, or more details about what's wrong with the source.

Creators of InvalidSourceTap may not even have any 'paths' as it may not be a FileSource that is creating the invalid tap.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Source.scala
@@ -39,7 +39,7 @@ import scala.collection.JavaConverters._
 /**
  * thrown when validateTaps fails
  */
-class InvalidSourceException(message: String) extends RuntimeException(message)
+class InvalidSourceException(message: String, cause: Throwable = null) extends RuntimeException(message, cause)
 
 /**
  * InvalidSourceTap used in createTap method when we want to defer
@@ -51,7 +51,10 @@ class InvalidSourceException(message: String) extends RuntimeException(message)
  *
  * hdfsPaths represents user-supplied list that was detected as not containing any valid paths.
  */
-class InvalidSourceTap(val hdfsPaths: Iterable[String]) extends SourceTap[JobConf, RecordReader[_, _]] {
+class InvalidSourceTap(val e: Throwable) extends SourceTap[JobConf, RecordReader[_, _]] {
+
+  def this(hdfsPaths: Iterable[String]) =
+    this(new InvalidSourceException(s"No good paths in $hdfsPaths"))
 
   private final val randomId = UUID.randomUUID.toString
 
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Source.scala
```diff
@@ -39,7 +39,7 @@ import scala.collection.JavaConverters._
 /**
  * thrown when validateTaps fails
  */
-class InvalidSourceException(message: String) extends RuntimeException(message)
+class InvalidSourceException(message: String, cause: Throwable = null) extends RuntimeException(message, cause)
```
johnynek: can we not have a default argument and instead add a second constructor:
```scala
  this(message: String) = this(message, null)
```

scalding-core/src/main/scala/com/twitter/scalding/Source.scala
```diff
@@ -51,7 +51,10 @@ class InvalidSourceException(message: String) extends RuntimeException(message)
  *
  * hdfsPaths represents user-supplied list that was detected as not containing any valid paths.
  */
-class InvalidSourceTap(val hdfsPaths: Iterable[String]) extends SourceTap[JobConf, RecordReader[_, _]] {
+class InvalidSourceTap(val e: Throwable) extends SourceTap[JobConf, RecordReader[_, _]] {
+
+  def this(hdfsPaths: Iterable[String]) =
+    this(new InvalidSourceException(s"No good paths in $hdfsPaths"))
```
johnynek, line 59: where is `hdfsPaths` coming from now? Why aren't you using `e` as a causedBy exception?

scalding-core/src/main/scala/com/twitter/scalding/Source.scala
isnotinvain: Yeah, wasn't sure if there's a difference.

It turns out that it's hard to extend Exception in scala because of the difference in how multiple constructors work in scala vs. java. The only hack that makes this possible is that passing null as the throwable just happens to be ok w/ the Exception base class. If that wasn't the case it'd be even worse.

scalding-core/src/main/scala/com/twitter/scalding/Source.scala
isnotinvain: this is a 2ndary constructor, so there is no `e` on line 56/57, we are creating `e` and passing it to the primary constructor.

So hdfsPaths comes from the caller, eg:

```
new InvalidSourceTap(somePaths)
```

This is for backwards compatibility as previously InvalidSourceTap's constructor took a list of paths. Does that make sense?

johnynek: ah, yes. Overlooked that somehow.

piyushnarang:
👍

scalding-core/src/main/scala/com/twitter/scalding/Source.scala
isnotinvain: I don't know if it's actually binary compatible though. Should be source compatible.

johnynek (approved):

PR #1627, opened 2016-12-01
TypedText - Added versions of dailyPrefixSuffix for TSV and CSV sources.

cavorite:
TypedText already has a method for reading `OSV` sources with a path built from a prefix and a suffix. This PR adds 2 variants of that method for reading TSV and CSV sources.


johnynek:
Can we have one private method parameterized on the separator and share code across all three?

johnynek:
PS: thanks for the contribution!!

cavorite:
I added a private function for each of the variants of the methods (hourly, daily with prefix and daily with prefix and suffix).

johnynek:
👍 merge when green

cavorite:
I currently don't have write permissions, @johnynek. Could you merge the PR for me?

PR #1628, opened 2016-12-03
add custom JVM opts

sritchie:
The algebird build was 4x faster with these on travis.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,0 +1,12 @@
+# see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
+-Dfile.encoding=UTF8
+-Xms1G
+-Xmx3G
+-XX:MaxPermSize=512M
+-XX:ReservedCodeCacheSize=250M
+-XX:+TieredCompilation
+-XX:-UseGCOverheadLimit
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Tool.scala
```diff
@@ -65,10 +65,10 @@ class Tool extends Configured with HTool {
   def run(args: Array[String]): Int = {
     val (mode, jobArgs) = parseModeArgs(args)
     // Connect mode with job Args
-    run(getJob(Mode.putMode(mode, jobArgs)))
+    runJob(getJob(Mode.putMode(mode, jobArgs)))
   }
 
-  protected def run(job: Job): Int = {
```
johnynek, line 71: What's going on here?

sritchie: oh shit, sorry, this must have been polluting my branch.

sritchie:
Things were not any faster! I wonder if these projects already have `SBT_OPTS` set.

piyushnarang:
@sritchie I've also noticed that travis build have a pretty large variance. So you might want to run a couple of times to make sense of the noise. These settings look reasonable though so we could turn them on anyway.

PR #1632, opened 2017-01-12
Fixes a bug in how we use this stack

ianoc:
So the previous method was using one config for both source and sink, but this isn't safe. Cascading isolates sources but not sinks. So if you apply some fancy settings to one source it can leak if you write to that source in a step.


oscar-stripe:
+1 but this is not binary compatible (as I assume mima will show).

We should probably prefer abstract classes more since they are friendlier for binary compatibility (though, I think maybe this is not the case for 2.12 using java 8).

ianoc:
Is it worth keeping backwards source compatibility with the deprecation here in that case?

oscar-stripe:
I think the source compat is worth it. If we don't try for that, by the time someone goes to update, there is a death by 1000 cuts, where all these small source problems add up.

isnotinvain:
+1

PR #1634, opened 2017-01-17
Bump chill version to 0.8.3

ttim:
In order to have https://github.com/twitter/summingbird working with new storm API we need to depend on kryo 3.0 and therefore one chill 0.8.x. This PR bumps chill dependency to 0.8.3.


johnynek:
This may change the serialization of things that are using kryo for storage (which they shouldn't do, but people don't always do the right thing).

We should definitely do a version bump here.

We should definitely not publish a new version, in my opinion, until we can do 2.12 at the same time.

ttim:
I agree on both. Also I need to investigate travis build failure.

johnynek:
Issue is:
```
should Default serialization should have tokens *** FAILED ***
[info]   java.lang.ClassNotFoundException: void
```
looks like a legit test failure. I guess something changed and void is somehow showing up and breaking that test.

ttim:
@johnynek Since 3.x (actually since some 2.x version) Kryo includes serializer for `void` by default in any `Kryo` instance. This test started to fail because `void`(`Unit` in case of scala) was not in a list of primitive types.

I added `void` to the list, but I'm not 100% sure how legitimate this change is.

If we are going to release it with 2.12 release should we merge this change to develop branch?

johnynek:
seems fine to me.

I think this looks good now. I assume you want to merge so you can do branch publish to get summingbird tested with storm 1.0.2?

ttim:
@johnynek I've done internal twitter release already, so it's better for me to keep it merged but I'm ok either way.

What I really need is - to release scalding with new chill to bump scalding version in summingbird branch to make it mergeable (otherwise two versions of kryo come to summingbird classpath). Also I want to bump scalding dependency in storehaus before storehaus release.

johnynek:
I'm fine to merge it. I'm actually anxious to move many of these
dependencies forward.

johnynek (approved):

Issue #1231, opened 2015-03-24
DistributedCacheFile does not work with Execution


johnynek:
HadoopMode.newFlowConnector ignores (intentionally) the jobConf because we don't want creepy mutation confusing what the Config is.

The problem is, Execution gives no programmatic way to change the Config inside the flow, so DistributedCacheFile, which assumes it can mutate the config, does not work.

There are at least two ways forward:

1) look for the distributed cache keys in the jobconf and mutate those in the Config before running (ugly).

2) Make a thin Execution-based wrapper for distributed cache that encapsulate this exception to the rule, so you have: addToDistributedCache(file): Execution[Unit]. I prefer this route, but it might be more work.

clayrat:
Any chance this could be fixed in the future?

PR #1635, opened 2017-01-25
Added API for Execution/Config to work with DistributedCache

dieu:
one of approaches to integrate DistributedCache with Executions (#1231)

the example of usage:
```
object YourExecJob extends ExecutionApp {
   override def job = 
      Execution.withCached("/path/to/your/file.txt") { cachedFile =>
             doSomething(cachedFile.path)
      }
}
```

multiple files:

```
object YourExecJob extends ExecutionApp {
   override def job = 
       Execution.withCached("/path/to/your/one.txt") { one =>
          Execution.withCached("/path/to/your/second.txt") { second =>
             doSomething(one.path, second.path)
          }
      }
}
```

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -17,10 +17,12 @@ package com.twitter.scalding
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.mapred.JobConf
+import org.apache.hadoop.mapreduce.MRJobConfig
 import org.apache.hadoop.io.serializer.{ Serialization => HSerialization }
 import com.twitter.chill.{ ExternalizerCodec, ExternalizerInjection, Externalizer, KryoInstantiator }
 import com.twitter.chill.config.{ ScalaMapConfig, ConfiguredInstantiator }
 import com.twitter.bijection.{ Base64String, Injection }
+import com.twitter.scalding.filecache.{CachedFile, DistributedCacheFile, HadoopCachedFile}
 
 import cascading.pipe.assembly.AggregateBy
 import cascading.flow.{ FlowListener, FlowStepListener, FlowProps, FlowStepStrategy }
@@ -28,6 +30,7 @@ import cascading.property.AppProps
 import cascading.tuple.collect.SpillableProps
 
 import java.security.MessageDigest
+import java.net.URI
 
 import scala.collection.JavaConverters._
 import scala.util.{ Failure, Success, Try }
@@ -49,6 +52,27 @@ trait Config extends Serializable {
       case (Some(v), r) => (r, this + (k -> v))
       case (None, r) => (r, this - k)
     }
+
+  /**
+   * Add files to be localized to the config. Intended to be used by user code.
+   * @param cachedFiles CachedFiles to be added
+   * @return new Config with cached files
+   */
+  def addDistributedCacheFiles(cachedFiles: CachedFile*): Config =
+    cachedFiles.foldLeft(this) { case (config, file) =>
+        file match {
+          case hadoopFile: HadoopCachedFile =>
+            Config.addDistributedCacheFile(hadoopFile.sourceUri, config)
+          case _ => config
+        }
+    }
+
+  /**
+   * Get cached files from config
+   */
+  def getDistributedCachedFiles: Seq[CachedFile] = {
+    Config.getDistributedCacheFile(this)
+  }
 
   /**
    * This is a name that if present is passed to flow.setName,
@@ -566,6 +590,47 @@ object Config {
     val bytes = fromInputStream(is)
     is.close()
     md5Hex(bytes)
+  }
+
+  /**
+   * Add a file to be localized to the config. Intended to be used by user code.
+   *
+   * @param qualifiedURI The qualified uri of the cache to be localized
+   * @param config Config to add the cache to
+   *
+   * @return new Config with cached files
+   *
+   * @see basic logic from [[org.apache.hadoop.mapreduce.filecache.DistributedCache.addCacheFile]]
+   */
+  private def addDistributedCacheFile(qualifiedURI: URI, config: Config): Config = {
+    val newFile = DistributedCacheFile
+      .symlinkedUriFor(qualifiedURI)
+      .toString
+
+    val newFiles = config
+      .get(MRJobConfig.CACHE_FILES)
+      .map(files => files + "," + newFile)
+      .getOrElse(newFile)
+
+    config + (MRJobConfig.CACHE_FILES -> newFiles)
+  }
+
+  /**
+   * Get distributed cache files from config
+   *
+   * @param config Config with cached files
+   */
+  private def getDistributedCacheFile(config: Config): Seq[CachedFile] = {
+    config
+      .get(MRJobConfig.CACHE_FILES)
+      .toSeq
+      .flatMap(_.split(","))
+      .filter(_.nonEmpty)
+      .map { file =>
+        val symlinkedUri = new URI(file)
+        val qualifiedUri = new URI(symlinkedUri.getScheme, symlinkedUri.getSchemeSpecificPart, null)
+        HadoopCachedFile(qualifiedUri)
+      }
   }
 
   private[this] def buildInj[T: ExternalizerInjection: ExternalizerCodec]: Injection[T, String] =
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -18,6 +18,7 @@ package com.twitter.scalding
 import com.twitter.algebird.monad.{ Reader, Trampoline }
 import com.twitter.algebird.{ Monoid, Monad, Semigroup }
 import com.twitter.scalding.cascading_interop.FlowListenerPromise
+import com.twitter.scalding.filecache.{CachedFile, DistributedCacheFile}
 import com.twitter.scalding.Dsl.flowDefToRichFlowDef
 import java.util.concurrent.LinkedBlockingQueue
 import scala.concurrent.{ Await, Future, ExecutionContext => ConcurrentExecutionContext, Promise }
@@ -256,6 +257,21 @@ object Execution {
 
   def withConfig[T](ex: Execution[T])(c: Config => Config): Execution[T] =
     TransformedConfig(ex, c)
+
+  /**
+   * Distributes the file onto each map/reduce node,
+   * so you can use it for Scalding source creation and TypedPipe, KeyedList, etc. transformations.
+   * Using the [[com.twitter.scalding.filecache.CachedFile]] outside of Execution will probably not work.
+   *
+   * For multiple files you must nested your execution, see docs of [[com.twitter.scalding.filecache.DistributedCacheFile]]
+   */
+  def withCachedFile[T](path: String)(fn: CachedFile => Execution[T]): Execution[T] = {
+    Execution.getMode.flatMap { mode =>
+      val cachedFile = DistributedCacheFile.cachedFile(path, mode)
+
+      withConfig(fn(cachedFile))(_.addDistributedCacheFiles(cachedFile))
+    }
+  }
 
   /**
    * This function allows running the passed execution with its own cache.
--- a/scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
@@ -49,12 +49,34 @@ object URIHasher {
  *
  * {{{
  * class YourJob(args: Args) extends Job(args) {
- *   val theCachedFile = DistributedCacheFile("hdfs://ur-namenode/path/to/your/file.txt")
+ *   val theCachedFile = DistributedCacheFile("/path/to/your/file.txt")
  *
  *   def somethingThatUsesTheCachedFile() {
  *     doSomethingWith(theCachedFile.path) // or theCachedFile.file
  *   }
  * }
+ *
+ * example with Execution:
+ *
+ * {{{
+ * object YourExecJob extends ExecutionApp {
+ *  override def job =
+ *    Execution.withCachedFile("/path/to/your/file.txt") { file =>
+ *      doSomething(theCachedFile.path)
+ *    }
+ * }
+ *
+ * example with Execution and multiple files:
+ *
+ * object YourExecJob extends ExecutionApp {
+ *  override def job =
+ *    Execution.withCachedFile("/path/to/your/one.txt") { one =>
+ *      Execution.withCachedFile("/path/to/your/second.txt") { second =>
+ *        doSomething(one.path, second.path)
+ *      }
+ *    }
+ * }
+ *
  * }}}
  *
  */
@@ -66,11 +88,35 @@ object DistributedCacheFile {
    * @param uri The fully qualified URI that points to the hdfs file to add
    * @return A CachedFile instance
    */
-  def apply(uri: URI)(implicit mode: Mode): CachedFile =
-    UncachedFile(Right(uri)).add()
+  def apply(uri: URI)(implicit mode: Mode): CachedFile = {
+    val cachedFile = UncachedFile(Right(uri)).cached(mode)
 
-  def apply(path: String)(implicit mode: Mode): CachedFile =
-    UncachedFile(Left(path)).add()
+    addCachedFile(cachedFile, mode)
+
+    cachedFile
+  }
+
+  def apply(path: String)(implicit mode: Mode): CachedFile = {
+    val cachedFile = UncachedFile(Left(path)).cached(mode)
+
+    addCachedFile(cachedFile, mode)
+
+    cachedFile
+  }
+
+  private[scalding] def cachedFile(uri: URI, mode: Mode): CachedFile =
+    UncachedFile(Right(uri)).cached(mode)
+
+  private[scalding] def cachedFile(path: String, mode: Mode): CachedFile =
+    UncachedFile(Left(path)).cached(mode)
+
+  private[scalding] def addCachedFile(cachedFile: CachedFile, mode: Mode): Unit = {
+    (cachedFile, mode) match {
+      case (hadoopFile: HadoopCachedFile, hadoopMode: HadoopMode) =>
+        HDistributedCache.addCacheFile(symlinkedUriFor(hadoopFile.sourceUri), hadoopMode.jobConf)
+      case _ =>
+    }
+  }
 
   def symlinkNameFor(uri: URI): String = {
     val hexsum = URIHasher(uri)
@@ -85,9 +131,7 @@ object DistributedCacheFile {
 
 final case class UncachedFile private[scalding] (source: Either[String, URI]) {
 
-  import DistributedCacheFile._
-
-  def add()(implicit mode: Mode): CachedFile =
+  def cached(mode: Mode): CachedFile =
     mode match {
       case Hdfs(_, conf) => addHdfs(conf)
       case HadoopTest(conf, _) => addHdfs(conf)
@@ -123,7 +167,6 @@ final case class UncachedFile private[scalding] (source: Either[String, URI]) {
         case Right(uri) => makeQualifiedURI(uri, conf)
       }
 
-    HDistributedCache.addCacheFile(symlinkedUriFor(sourceUri), conf)
     HadoopCachedFile(sourceUri)
   }
 }
@@ -137,8 +180,8 @@ sealed abstract class CachedFile {
 }
 
 final case class LocallyCachedFile private[scalding] (sourcePath: String) extends CachedFile {
-  def path = file.getCanonicalPath
-  def file = new File(sourcePath).getCanonicalFile
+  def path: String = file.getCanonicalPath
+  def file: File = new File(sourcePath).getCanonicalFile
 }
 
 final case class HadoopCachedFile private[scalding] (sourceUri: URI) extends CachedFile {
--- a/scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
@@ -30,6 +30,7 @@ import scala.util.Random
 import scala.util.{Failure, Success, Try}
 import ExecutionContext._
 import com.twitter.scalding.Execution.TempFileCleanup
+import org.apache.hadoop.conf.Configuration
 
 object ExecutionTestJobs {
   def wordCount(in: String, out: String) =
@@ -261,6 +262,35 @@ class ExecutionTest extends WordSpec with Matchers {
 
       assert(oldCounters != newCounters, "With new configs given the source changed we shouldn't cache so the counters should be different")
 
+    }
+
+    "correctly add cached file into config" in {
+      val execution = Execution.withCachedFile("/path/to/your/file.txt") { cachedFile =>
+        Execution.getConfig.map { config =>
+          config.getDistributedCachedFiles should contain only cachedFile
+        }
+      }
+
+      execution.waitFor(Config.default, Hdfs(strict = true, new Configuration(false))) match {
+        case Success(s) => s
+        case Failure(e) => fail(s"Failed running execution, exception:\n$e")
+      }
+    }
+
+    "correctly add cached files into config" in {
+      val execution =
+        Execution.withCachedFile("/path/to/your/one.txt") { one =>
+          Execution.withCachedFile("/path/to/your/second.txt") { second =>
+            Execution.getConfig.map { config =>
+              config.getDistributedCachedFiles should contain only (one, second)
+            }
+          }
+        }
+
+      execution.waitFor(Config.default, Hdfs(strict = true, new Configuration(false))) match {
+        case Success(s) => s
+        case Failure(e) => fail(s"Failed running execution, exception:\n$e")
+      }
     }
   }
 
--- a/scalding-core/src/test/scala/com/twitter/scalding/ConfigTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/ConfigTest.scala
@@ -15,6 +15,10 @@ limitations under the License.
 */
 package com.twitter.scalding
 
+import com.twitter.scalding.filecache.{ HadoopCachedFile, URIHasher }
+import java.net.URI
+import org.apache.hadoop.mapreduce.MRJobConfig
+
 import org.scalatest.{ WordSpec, Matchers }
 import org.scalacheck.Arbitrary
 import org.scalacheck.Properties
@@ -84,6 +88,41 @@ class ConfigTest extends WordSpec with Matchers {
         prim(c) || Class.forName(c).isArray
       } shouldBe true
     }
+    "addDistributedCacheFile works" in {
+      val (cachedFile, path) = ConfigTest.makeCachedFileAndPath("test.txt")
+
+      Config
+        .empty
+        .addDistributedCacheFiles(cachedFile)
+        .get(MRJobConfig.CACHE_FILES) shouldBe Some(path)
+    }
+    "multiple addDistributedCacheFile work" in {
+      val (cachedFileFirst, pathFirst) = ConfigTest.makeCachedFileAndPath("first.txt")
+      val (cachedFileSecond, pathSecond) = ConfigTest.makeCachedFileAndPath("second.txt")
+
+      Config
+        .empty
+        .addDistributedCacheFiles(cachedFileFirst, cachedFileSecond)
+        .get(MRJobConfig.CACHE_FILES) shouldBe Some(s"$pathFirst,$pathSecond")
+
+      Config
+        .empty
+        .addDistributedCacheFiles(cachedFileFirst)
+        .addDistributedCacheFiles(cachedFileSecond)
+        .get(MRJobConfig.CACHE_FILES) shouldBe Some(s"$pathFirst,$pathSecond")
+    }
+  }
+}
+
+object ConfigTest {
+  def makeCachedFileAndPath(name: String): (HadoopCachedFile, String) = {
+    val uriString = s"hdfs://foo.example:1234/path/to/the/stuff/$name"
+    val uri = new URI(uriString)
+    val hashHex = URIHasher(uri)
+    val hashedFilename = hashHex + s"-$name"
+    val cachedFile = HadoopCachedFile(uri)
+
+    (cachedFile, s"$uriString#$hashedFilename")
   }
 }
 
--- a/scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatform.scala
+++ b/scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatform.scala
@@ -1,0 +1,65 @@
+package com.twitter.scalding.platform
+
+import com.twitter.scalding._
+import com.twitter.scalding.source.TypedText
+
+import java.io.{ BufferedWriter, File, FileWriter }
+
+import org.slf4j.LoggerFactory
+
+trait HadoopPlatform[P, R, T <: HadoopPlatform[P, R, T]] {
+  private val LOG = LoggerFactory.getLogger(getClass)
+
+  val cons: (P) => R
+  val cluster: LocalCluster
+
+  val dataToCreate: Seq[(String, Seq[String])]
+  val sourceWriters: Seq[P => R]
+  val sourceReaders: Seq[Mode => Unit]
+
+  def arg(key: String, value: String): T
+
+  def data(data: (String, Seq[String])): T
+
+  def source[K: TypeDescriptor](location: String, data: Seq[K]): T =
+    source(TypedText.tsv[K](location), data)
+
+  def source[K](out: TypedSink[K], data: Seq[K]): T
+
+  def sink[K: TypeDescriptor](location: String)(toExpect: Seq[K] => Unit): T =
+    sink(TypedText.tsv[K](location))(toExpect)
+
+  def sink[K](in: Mappable[K])(toExpect: Seq[K] => Unit): T
+
+  def run(): Unit
+
+  def init(cons: P => R): R
+
+  def execute(unit: R): Unit
+
+  protected def createSources(): Unit = {
+    dataToCreate foreach {
+      case (location, lines) =>
+        val tmpFile = File.createTempFile("hadoop_platform", "job_test")
+        tmpFile.deleteOnExit()
+        if (lines.nonEmpty) {
+          val os = new BufferedWriter(new FileWriter(tmpFile))
+          os.write(lines.head)
+          lines.tail.foreach { str =>
+            os.newLine()
+            os.write(str)
+          }
+          os.close()
+        }
+        cluster.putFile(tmpFile, location)
+        tmpFile.delete()
+    }
+
+    sourceWriters.foreach { cons => execute(init(cons)) }
+  }
+
+  protected def checkSinks(): Unit = {
+    LOG.debug("Executing sinks")
+    sourceReaders.foreach { _(cluster.mode) }
+  }
+}
--- a/scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformExecutionTest.scala
+++ b/scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatformExecutionTest.scala
@@ -1,0 +1,51 @@
+package com.twitter.scalding.platform
+
+import cascading.flow.Flow
+import com.twitter.scalding._
+import org.apache.hadoop.mapred.JobConf
+import scala.util.{ Failure, Success }
+
+case class HadoopPlatformExecutionTest(
+  cons: (Config) => Execution[_],
+  cluster: LocalCluster,
+  parameters: Map[String, String] = Map.empty,
+  dataToCreate: Seq[(String, Seq[String])] = Vector(),
+  sourceWriters: Seq[Config => Execution[_]] = Vector.empty,
+  sourceReaders: Seq[Mode => Unit] = Vector.empty,
+  flowCheckers: Seq[Flow[JobConf] => Unit] = Vector.empty) extends HadoopPlatform[Config, Execution[_], HadoopPlatformExecutionTest] {
+
+  def config: Config =
+    Config.defaultFrom(cluster.mode) ++ Config.from(parameters)
+
+  override def arg(key: String, value: String): HadoopPlatformExecutionTest =
+    copy(parameters = parameters + (key -> value))
+
+  override def data(data: (String, Seq[String])): HadoopPlatformExecutionTest =
+    copy(dataToCreate = dataToCreate :+ data)
+
+  override def source[K](out: TypedSink[K], data: Seq[K]): HadoopPlatformExecutionTest =
+    copy(sourceWriters = sourceWriters :+ { config: Config =>
+      TypedPipe.from(data).writeExecution(out)
+    })
+
+  override def sink[K](in: Mappable[K])(toExpect: (Seq[K]) => Unit): HadoopPlatformExecutionTest =
+    copy(sourceReaders = sourceReaders :+ { m: Mode => toExpect(in.toIterator(config, m).toSeq) })
+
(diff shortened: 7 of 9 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -17,20 +17,20 @@ package com.twitter.scalding
 
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.mapred.JobConf
-import org.apache.hadoop.io.serializer.{ Serialization => HSerialization }
-import com.twitter.chill.{ ExternalizerCodec, ExternalizerInjection, Externalizer, KryoInstantiator }
-import com.twitter.chill.config.{ ScalaMapConfig, ConfiguredInstantiator }
-import com.twitter.bijection.{ Base64String, Injection }
-
+import org.apache.hadoop.io.serializer.{Serialization => HSerialization}
```
isnotinvain: if you could revert the style changes here that'd be great. I think scalafmt will put them back anyway.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -51,6 +51,20 @@ trait Config extends Serializable {
     }
 
   /**
+   * Add files to be localized to the config. Intended to be used by user code.
+   * @param cachedFiles CachedFiles to be added
+   * @return new Config with cached files
+   */
+  def addDistributedCacheFiles(cachedFiles: CachedFile*): Config =
```
isnotinvain, line 61: `addDistributedCacheFile(uri: URI, config: Config)` below makes sense to me, but this one I don't quite follow. If someone already has a CachedFile, that's only possible if they provided DistributedCacheFile with a mode, and that mode has already been mutated. It seems like we don't really need this method at all? 

I think 2 instance methods (not static ones) on Config should be enough:

`def addDistributedCacheFiles(uris: URII): Config`
`def addDistributedCacheFiles(files: File*): Config`

Execution users will likely not have a CachedFile to begin with, as they won't have a mode to give to DistributedCacheFile right?

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -544,6 +558,27 @@ object Config {
...
       .setSerialization(Right(classOf[serialization.KryoHadoop]))
       .setScaldingVersion
+
+  /**
+   * Add a file to be localized to the config. Intended to be used by user code.
+   * @param uri The uri of the cache to be localized
+   * @param config Config to add the cache to
+   * @return new Config with cached files
+   */
+  def addDistributedCacheFile(uri: URI, config: Config): Config = {
```
isnotinvain: as mentioned above, I think this could just be an instance method?

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -544,6 +558,27 @@ object Config {
...
+    val newFile = DistributedCacheFile
+      .symlinkedUriFor(uri)
```
isnotinvain: there's some logic in DistributedCacheFile for making paths fully qualified -- should we expose those too and and use them here too?

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -544,6 +558,27 @@ object Config {
...
+      .toString
+
+    val newFiles = config
```
isnotinvain: maybe add a comment linking the lines from the hadoop codebase this is taken from so that when we read this we can know where it came from

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -199,6 +199,24 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
   def zip[U](that: Execution[U]): Execution[(T, U)] =
     Zipped(this, that)
 
+  /**
+   * Add an extra config for current execution.
+   * This changes of config are isolated and affect only current execution.
+   * All changes will disappear for any next operation on execution.
+   */
+  def addConfig(extraConfig: Config): Execution[T] =
```
isnotinvain: Do you think we need this method? Since this only "sticks" for just 1 execution, maybe using `Execution.withConfig` is more clear? I'm worried this makes it looks like you could do:

```
myExecuiton
  .addConfig(a)
  .addConfig(b)
  .addConfig(c)
```
 and get a, b, and c all added in the result, when in fact a and b get ignored.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -199,6 +199,24 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
...
+    Execution.withConfig(this) { config =>
+      config ++ extraConfig
+    }
+
+  /**
+   * Replace current the config to a new one.
+   * This changes of config are isolated and affect only current execution.
+   * All changes will disappear for any next operation on execution.
+   */
+  def withConfig(config: Config): Execution[T] =
```
isnotinvain: This one I really don't think we need. It has the same name as `Execution.withConfig` but it's very different. I don't think we ever really want to ignore the old config. Can you think of a use case where this is necessary?

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -57,6 +57,21 @@ object URIHasher {
  * }
  * }}}
  *
+ * example with Execution:
+ * {{{
+ * object YourExecJob extends ExecutionApp {
+ *  override def job = Execution.getConfig.flatMap { config =>
+ *    val theCachedFile = DistributedCacheFile("hdfs://ur-namenode/path/to/your/file.txt")
```
isnotinvain: won't this fail, as there is no implicit mode available for DistributedCacheFile?

isnotinvain (commented):
Thanks for looking into this!

I think we might need to take a slightly different approach, given that DistributedCacheFile.apply() takes in a mode, and also as part of apply() it adds the files to the mode's config. This isn't going to work for executions as there's no mode in scope by default, and if there was, it still wouldn't really make sense to provide it as it's mutations would be ignored.

Maybe we need a new class, DistributedCacheFileExecution, that doesn't take a mode as an argument. Then it could be used something like:

```
override def job = {
  val theCachedFile = DistributedCacheFileExecution("hdfs://ur-namenode/path/to/your/file.txt")

val execution = doSomething(theCachedFile.path)

Execution.withConfig(execution) { c => c.addCachedFiles(theCachedFile) }
```

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: Why does a and b get ignored @isnotinvain? I don't read the code that way.

johnynek:
Alex made some good comments. I'm more positive on the added APIs.

To me I would like to see us split the methods to add Execution APIs into a separate PR that has nothing to do with DistributedCache.

Secondly, since you can get the mode from the Execution (getMode), can we have a method that takes directly does the getMode and Config update in one? Something like: `Execution.withDisributedCached(file: String)(fn: CacheFile => Execution[T]): Execution[T]`

So here you point to the thing to cache and give a function of what to do with the CacheFile (which should not escape in T, and here we would love to use: https://www.cs.purdue.edu/homes/rompf/papers/osvald-oopsla16.pdf )

How does this sound?

Thanks for this. Many have asked for such an API and people will enjoy having this merged.

dieu:
@isnotinvain yes DistributedCacheFileExecution make sense (who will not mutate hadoop config in `mode`), but it still should use `mode` as the parameter to understand where we are, in hadoop mode or local. or you prefer to ignore local case and always modify config? 

PS: Sorry for typos, the correct example should be:
```
object YourExecJob extends ExecutionApp {
 override def job = Execution.getMode.flatMap { mode =>
    val theCachedFile = DistributedCacheFile("hdfs://ur-namenode/path/to/your/file.txt")(mode)

    val newConfig = config
      .addDistributedCacheFiles(theCachedFile)

    doSomethingWithCachedFileAndReturnExecution(theCachedFile.path)
      .withConfig(newConfig)
  }
}
```

Thanks for your feedback!

dieu:
@johnynek 

1. My personal preference to don't mix different entities, I think that `DistributedCache` is the property of `Config` and basically `Execution` don't do anything with it.
2. 
`Execution.withDisributedCached(file: String)(fn: CacheFile => Execution[T]): Execution[T]` 

that kind of API work perfectly but only with one file, in many cases, you want to have multiple files, then you need something like:

` Execution.withDisributedCached(file: String*)(fn: Seq[CacheFile] => Execution[T]): Execution[T]` 

but you can not easily name files by variables, you should provide in API an order of files in `Seq` that for me a little bit fragile API.

Thanks for your feedback!

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: yes, I want to have some API to work with `CachedFile` for backward capability, because I sure that many people already use `DistributedCacheFile` (`DistributedCacheFile` will work fine with execution because `Execution` has method `getMode/getConfigAndMode`). 

but for
```
def addDistributedCacheFiles(uris: URII): Config
def addDistributedCacheFiles(files: File*): Config
```

we need mode, and we can work with it only in Hadoop mode. at this point, I think we should only work with `CachedFile`.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: this method useful because I need to specify which config I want to modify in `foldLeft`. I'll make it as private method.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: we will assume that in this method we work only with qualified uri.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: ok. thanks

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: @isnotinvain no, a,b,c will pass to `myExecution`.

but in case:
```
myExecuiton
  .addConfig(a)
  .addConfig(b)
  .addConfig(c)
  .flatMap(item => doSomething(item))
````
`doSomething` don't get a,b,c.

File: scalding-core/src/main/scala/com/twitter/scalding/Execution.scala (hunk 7 above)
Comment (dieu) [reply scalding-core/src/main/scala/com/twitter/scalding/Execution.scala]: Just as optimization for the case when you use addDistributedCacheFiles on current config, it means that in new config you already have the old one and you don't need ++

File: scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala (hunk 8 above)
Comment (dieu) [reply scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala]: yes, sorry for typos, I'll update example

File: scalding-core/src/main/scala/com/twitter/scalding/Config.scala (hunk 1 above)
Comment (dieu) [reply scalding-core/src/main/scala/com/twitter/scalding/Config.scala]: ok, thanks

File: scalding-core/src/main/scala/com/twitter/scalding/Execution.scala (hunk 7 above)
Comment (ulyssepence) [reply scalding-core/src/main/scala/com/twitter/scalding/Execution.scala]: Agree with @isnotinvain . Breaks expectations users have about the overloaded method. If we were to accept something like this, it should be named like `#replaceConfig` or something.

File: scalding-core/src/main/scala/com/twitter/scalding/Execution.scala (hunk 6 above)
Comment (ulyssepence) [reply scalding-core/src/main/scala/com/twitter/scalding/Execution.scala]: Agree with @isnotinvain that this method is unnecessary (or should be added in a separate PR).

**Comment by ulyssepence** [comment]
I like the approach proposed by @johnynek .

If I understand the spirit of the `Execution` API, users shouldn't be mixing and passing around `Mode`s in general, but should be using it as an indication of the run context.

Thus, I assume the `implicit val mode: Mode` argument in `DistributedCacheFile` is an artifact of `Job` and users shouldn't have to touch it for adding a cached file.

File: scalding-core/src/main/scala/com/twitter/scalding/Config.scala (hunk 2 above)
Comment (isnotinvain) [reply scalding-core/src/main/scala/com/twitter/scalding/Config.scala:61]: No but the DistributedCacheFile class just doesn't make sense in the context of Executions. It has to be called from code that is intended for use only in a Job. Even getting access to the Execution's mode does't help as the mutations are ignored. I really think we need to not conflate these together.

File: scalding-core/src/main/scala/com/twitter/scalding/Config.scala (hunk 4 above)
Comment (isnotinvain) [reply scalding-core/src/main/scala/com/twitter/scalding/Config.scala]: ok maybe that should be in the comments then, as that's not the case in DistributedCacheFile right?

File: scalding-core/src/main/scala/com/twitter/scalding/Execution.scala (hunk 6 above)
Comment (isnotinvain) [reply scalding-core/src/main/scala/com/twitter/scalding/Execution.scala]: Oh, I guess I misunderstood then. In that case addConfig seems fine.

**Comment by isnotinvain** [comment]
I don't think I understand @johnynek's idea that involves a mode.

Ultimately, we need to update a setting in the hadoop conf to get the distributed cache to work. The DistributedCacheFile helper does that by mutating the conf found in the mode. This mutation is ignored by Executions however. 

So all wee really need is helper methods for getting this update in the scalding Config, instead of getting it into a mutate hadoop conf.

File: scalding-core/src/main/scala/com/twitter/scalding/Config.scala (hunk 2 above)
Comment (dieu) [reply scalding-core/src/main/scala/com/twitter/scalding/Config.scala:61]: `CachedFile` itself not depended on `DistributedCacheFile`, it just encapsulates information about qualified URI and has shortcuts for file and path.

File: scalding-core/src/main/scala/com/twitter/scalding/Config.scala (hunk 4 above)
Comment (dieu) [reply scalding-core/src/main/scala/com/twitter/scalding/Config.scala]: I fixed comments and name of variable.

File: scalding-core/src/main/scala/com/twitter/scalding/Execution.scala (hunk 7 above)
Comment (dieu) [reply scalding-core/src/main/scala/com/twitter/scalding/Execution.scala]: @benpence I think, replaceConfig give more misunderstanding, will removed it.

File: scalding-core/src/main/scala/com/twitter/scalding/Execution.scala (hunk 6 above)
Comment (ulyssepence) [reply scalding-core/src/main/scala/com/twitter/scalding/Execution.scala]: I like this style better than the static `.withConfig`, but I think it will only serve to clutter the API. If you want this style, my preference would be to have a method of the same name or perhaps `#transformConfig` on `trait Execution` that makes it trivial to do this:

```scala
val extraConfig: Config = /* ... */

def myExecution(exec: Execution[Unit]): Execution[Unit] = {
  exec
    .transformConfig(_ ++ extraConfig)
    .flatMap { /* ... */ }
    // ...
}
```

and maybe deprecate the `.withConfig` method.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
ulyssepence: Hmm, I'm a little conflicted on this. My preference would be a method on `Execution` like `def addDistributedCacheFile(path: String): Execution[CachedFile]` or something similar. That way, the user doesn't have to worry about the `Mode`. Internally, we can just use the `Mode` in `Execution#run` to decide whether to mutate the Hadoop `Configuration`.

johnynek:
I don't want to have two code paths to use the distributed cache. So even if it is uglier, reusing internally the old code (or making the old API use he new code) would be nice.

Next, the Execution API should be really nice and not bound to 6 year old choices (like mutability of Mode and FlowDef). So something like `addCachedFile(...): Execution[CachedFile]`

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: @benpence in this case, it's huge refactoring to support `extraConfig` in next operations. I think it should be another PR.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: @benpence I have 2 concerns about API like that:
1. DistributedCacheFile is the property of config, not an execution.
2. at now we don't have the ability to pass any changes in config to next operations, so we should create distributed cache before execution, then even if we will have that ability it's will not compatible with older code (you can not use same code for Job API and Execution API), also it's little bit ugly to use for multiple files.

dieu:
@benpence @johnynek not sure, how `addCachedFile(...): Execution[CachedFile]` will works in this cases:

```
TypedPipe
  .flatMap(...)
  .writeExecution(...)
  .addCachedFile(...) // it's dosen't make sense here, right?
```
```
Execution
  .unit
  .addCachedFile(...)
  .addCachedFile(...) // what is type here? Execution[(CachedFile, CachedFile)] or Execution[Seq[CachedFile]]?
```

how to use `addCachedFile(...): Execution[CachedFile]` with `fromFuture`, `fromFn`, etc?

isnotinvain:
@johnynek re: "So even if it is uglier, reusing internally the old code (or making the old API use he new code) would be nice."

This adds the burden on execuiton users to get access to a mode that they don't need, right? 

In this example:

```
object YourExecJob extends ExecutionApp {
 override def job = Execution.getMode.flatMap { mode =>
    val theCachedFile = DistributedCacheFile("hdfs://ur-namenode/path/to/your/file.txt")(mode)

    val newConfig = config
      .addDistributedCacheFiles(theCachedFile)

    doSomethingWithCachedFileAndReturnExecution(theCachedFile.path)
      .withConfig(newConfig)
  }
}
```

The mode is a required argument, it gets mutated, and then that mutation gets ignored, and then the updated `newConfig` is what actually makes it work.

So why not just let users only do the config related part? eg:

```
object YourExecJob extends ExecutionApp {
 override def job = {
   val theCachedFile = DistributedCacheFileExecution("hdfs://ur-namenode/path/to/your/file.txt")
    
   TypedPipe.from(...)
     .filter { x => something(theCachedFile.path, x) }
     .writeExecution
     .withConfig(c => c.addDistributedCacheFiles(theCachedFile))  
  }
}
```

johnynek:
The problem I am concerned with is correctly scoping where the
configuration is in scope. Ideally, anything that compiles will run without
an exception. It seems a bit hard to reach that level here since some map
or reduce function somewhere will assume the existence of a cached file,
but we have not yet proposed an API that proves that existence in the types.

To me, that should be our goal.

As for how to use addCacheFile, you use it with flatMap or for:

```
for {
f1 <- Execution.addCacheFile(p1)
f2 <- Execution.addCacheFile(p2)
res <- useCached(f1, f2)
} yield res
```

isnotinvain:
@johnynek that part makes sense to me, if you're saying we should only support dist cache in executions this way, I'd be in favor of that.

I was concerned with the different choices for how to do this in the config layer -- but if you're saying you'd rather do this only as mentioned above (Execution.addCacheFile) then I'd support that approach.

isnotinvain:
Wait @dieu points out that the Execution.addCacheFile example doesn't actually address how to make that work, in an expression like:

```
e1.addCacheFile(file).flatMap { f => someExecutionThatNeedsF(f) }
```

You'd have to somehow make sure that whatever Execution comes out of someExecutionThatNeedsF has also had its config modified to include f. I don't see where that'd happen. Or would you instead pass `someExecutionThatNeedsF` an `Execution[File]` instead ?

isnotinvain:
ok so unless @johnynek has any hints as to how to make the `Execution[File]` approach work, are we ok with:

```
object YourExecJob extends ExecutionApp {
 override def job = Execution.getMode.flatMap { mode =>
    val theCachedFile = DistributedCacheFile("hdfs://ur-namenode/path/to/your/file.txt")(mode)
    
  doSomethingWithCachedFileAndReturnExecution(theCachedFile.path)
    .withConfig(c => c.addDistributedCacheFile(theCachedFile))
  }
}
```

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
ulyssepence: I mostly agree with you, except for using the same code for Job & Execution API. They seem so different at this point that I don't think that matters.

Anyway, while it's conceptually something in the Config, I think the fact that the current API requires access to the `Mode` couples it to the run environment. So I'm not worried about that either. I agree that the API is not great for multiple files, I think this is acceptable:

```scala
Execution.zip(file1Ex, file2Ex, file3Ex).map { case (file1, file2, file3) =>
  // ...
}
```

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ulyssepence: OK, I think it's fine to drop this for another PR.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: the main problem to have that kind of API is that we can not modify config between executions.

ulyssepence:
Hmm, I thought config changes permeated through `FlatMapped`, but I guess I was wrong: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L459 .

We _could_ make an internal exception for cached files.

@isnotinvain I don't like making the user add something and then retrieve it from the config because then the type system forces you to handle the case where the file is missing from the config, when we know that shouldn't be the case. I'd prefer exposing the file directly somewhere, even if it's something like `def addFile[A](fileToCache: String)(File => Execution[A]): Execution[A]`, although that specific API makes it hard to handle multiple files.

dieu:
@benpence what do you meant "retrieve it from the config"? the user works with `CachedFile` directly.

ulyssepence:
@dieu Oh, you're right. I suppose I should have gotten that from the name `doSomethingWithCachedFileAndReturnExecution`. What @isnotinvain has seems OK for now, assuming he meant to use the existing `Execution.withConfig(...)`

johnynek:
So, I'm actually back to the part where the API I like is:

`Execution.withCached(f: String)(fn: CachedFile => Execution[T]):
Execution[T]`

If you need several files (rare case) nest these calls.

ulyssepence:
@johnynek You're talking about keeping the mutable `Mode` route, then? Otherwise, I think this is impossible without changing the [behavior of `FlatMapped`](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L459) right?

dieu:
@johnynek @benpence @isnotinvain can you look on the recent update, please?

Tried to address all of comments/opinions.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -568,6 +593,47 @@ object Config {
...
+   *
+   * @see basic logic from [[org.apache.hadoop.mapreduce.filecache.DistributedCache.addCacheFile]]
+   */
+  private def addDistributedCacheFile(qualifiedURI: URI, config: Config): Config = {
+    val newFile = DistributedCacheFile
+      .symlinkedUriFor(qualifiedURI)
+      .toString
+
+    val newFiles = config
+      .get(MRJobConfig.CACHE_FILES)
```
johnynek, line 611: should we use both the old `map-red` and `mapreduce` keys (if they are different)? Maybe we don't need to.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -51,6 +54,28 @@ trait Config extends Serializable {
...
+        file match {
+          case hadoopFile: HadoopCachedFile =>
+            Config.addDistributedCacheFile(hadoopFile.sourceUri, config)
+          case _ => config
+        }
+    }
+
+  /**
+   * Get cached files from config
+   * @return seq of cached files
```
ulyssepence: I don't think `@return` adds anything here. I would remove it.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
```diff
@@ -262,6 +263,35 @@ class ExecutionTest extends WordSpec with Matchers {
...
+      val execution =
+        Execution.withCached("/path/to/your/one.txt") { one =>
+          Execution.withCached("/path/to/your/second.txt") { second =>
+            Execution.getConfig.map { config =>
+              config.getDistributedCachedFiles should contain only (one, second)
+            }
+          }
+        }
+
+      execution.waitFor(Config.default, Hdfs(strict = true, new Configuration(false))) match {
```
johnynek, line 290: we don't actually have a test that checks you can get the file. Can we make one?

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
```diff
@@ -568,6 +593,47 @@ object Config {
...
+  private def getDistributedCacheFile(config: Config): Seq[CachedFile] = {
+    config
+      .get(MRJobConfig.CACHE_FILES)
+      .toSeq
+      .flatMap(_.split(","))
+      .filter(_.nonEmpty)
+      .map { file =>
+        val symlinkedUri = new URI(file)
+        val qualifiedUri = new URI(symlinkedUri.getScheme, symlinkedUri.getSchemeSpecificPart, null)
+        HadoopCachedFile(qualifiedUri)
```
ulyssepence, line 632: Hmm, will this be problematic if we try to get these in local mode?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -257,6 +258,17 @@ object Execution {
   def withConfig[T](ex: Execution[T])(c: Config => Config): Execution[T] =
     TransformedConfig(ex, c)
 
+  def withCached[T](path: String)(fn: CachedFile => Execution[T]): Execution[T] = {
```
johnynek: can we add a comment that you cannot return `CachedFile` in `T`. You must use `CachedFile` inside something inside the `Execution[T]` (like a map/reduce function).

ulyssepence: Add a comment here that shows how to use this with multiple files. Consider adding a reminder comment about `CachedFile` expectations i.e. if you're using `Hdfs` mode, the `CachedFile` path will only be valid in the `TypedPipe`/`KeyedList` closures (right?).

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -257,6 +258,17 @@ object Execution {
...
+    Execution.getConfigMode.flatMap { case (config, mode) =>
+      val cachedFile = DistributedCacheFile(path)(mode)
```
johnynek: can we add a comment that this does have the side effect of adding a key to the Configuration, but that should not matter.

On the other hand, maybe we can refactor:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala#L72

to have a `private[scalding]` method that returns `def immutableApply(path: String)(implicit mode: Mode): (() => Mode, CachedFile)` where the funtion in the left is the side-effect of updating the `Configuration`. The current code would then just do:
```scala
def apply(str: String)(implicit m: Mode): CachedFile = {
  val (effect, cf) = immutableApply(str)
  effect()
  cf
}
```

Then our code can just discard the effect. How does that sound?

johnynek (commented):
This looks good to me. A few small comments and suggestions, but I think we are close to merge, as far as I can see.

ulyssepence (commented):
I think this API is close to what we already have and gets the job done. I left a few comments.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: it doesn't have alternative.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
dieu: @johnynek what do you mean "get the file"? tests above test `getDistributedCachedFiles` and `withCached`, they should actually execute `Execution.withCached`.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: done, thanks

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
johnynek: well, in hadoop mode, hadoop has to do some extra steps, so actually using the `CachedFile.file` is a test that step is actually working.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: @johnynek sounds good, but I did a little bit by another way. look, please.

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: done, thank

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
dieu: we don't add files to config in local mode at all. same thing in older code.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: done

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
dieu: @johnynek I found that `scalding-hadoop-test` support only `Job` based tests. Can we do it in other PR(support of run Execution base tests in local cluster)?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -258,6 +259,26 @@ object Execution {
     TransformedConfig(ex, c)
 
   /**
+   * Distribute the file across map/reduce nodes and
+   * give you instance of [[com.twitter.scalding.filecache.CachedFile]].
+   *
+   * The cached file available only inside execution.
+   * You can not return cached file as result of execution and must use inside closure.
+   *
+   * For multiple files you must nested your execution, see docs of [[com.twitter.scalding.filecache.DistributedCacheFile]]
```
piyushnarang: should we add a short snippet indicating how users can do this here?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -258,9 +258,18 @@ object Execution {
...
+   */
   def withCached[T](path: String)(fn: CachedFile => Execution[T]): Execution[T] = {
     Execution.getConfigMode.flatMap { case (config, mode) =>
-      val cachedFile = DistributedCacheFile(path)(mode)
+      val cachedFile = DistributedCacheFile.cachedFile(path, mode)
 
       val newConfig = config
           .addDistributedCacheFiles(cachedFile)
```
johnynek: why isn't this:
```scala
withConfig(fn(cachedFile))(_.addDistributedCacheFiles(cachedFile))
```
it seems the current code is going to ignore any config changes that might happen in `fn` since you are getting the config before you apply `fn`.

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -66,11 +88,37 @@ object DistributedCacheFile {
...
+    val cachedFile = UncachedFile(Left(path)).cached(mode)
 
-  def apply(path: String)(implicit mode: Mode): CachedFile =
-    UncachedFile(Left(path)).add()
+    addCachedFile(cachedFile, mode)
+
+    cachedFile
+  }
+
+  private[scalding] def cachedFile(uri: URI, mode: Mode): CachedFile = {
```
piyushnarang: can probably skip the {} as its a single line function

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -66,11 +88,37 @@ object DistributedCacheFile {
...
+    UncachedFile(Right(uri)).cached(mode)
+  }
+
+  private[scalding] def cachedFile(path: String, mode: Mode): CachedFile = {
+    UncachedFile(Left(path)).cached(mode)
+  }
```
piyushnarang: here too

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -77,11 +88,37 @@ object DistributedCacheFile {
...
+
+  private[scalding] def addCachedFile(cachedFile: CachedFile, mode: Mode) {
```
johnynek: this syntax is deprecated in scala. Can we use: `): Unit = {` rather than `) {`

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -96,9 +133,7 @@ object DistributedCacheFile {
 
 final case class UncachedFile private[scalding] (source: Either[String, URI]) {
 
-  import DistributedCacheFile._
-
-  def add()(implicit mode: Mode): CachedFile =
```
johnynek, line 90: can we not delete this which hurts compatibility? Can you just call the cached method then `addCacheFile`?

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
piyushnarang: Seems like some of the tests here are using Hadoop local mode. If it does support storing and retrieving the cached files, it would be nice.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: snippet available in DistributedCacheFile, and I put the link to it. do you want copy to here also?

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
dieu: ok

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
dieu: ok

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
dieu: @piyushnarang which one? test for a cached file is using `Hdfs` mode. but this test doesn't execute anything on real hadoop, it test only changes in config, for integrated test we should add them to `scalding-hadoop-test` how I understand.

dieu: @piyushnarang which one? test for a cached file is using `Hdfs` mode. but this test doesn't execute anything on real hadoop, it test only changes in config, for integrated test we should add them to `scalding-hadoop-test` how I understand, but for do it, we should add support of execution to test framework.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: ok, but basically `fn` can not change config what use `withConfig`

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
dieu: thanks done

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
dieu: it does not hurt compatibility because we don't return this class anywhere and he has the private constructor. Nobody can't use his directly.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -271,10 +271,7 @@ object Execution {
     Execution.getConfigMode.flatMap { case (config, mode) =>
```
johnynek: can't you just do `Execution.getMode.flatMap { mode =>` now?

dieu: sure, sorry.

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
johnynek: okay, fair enough.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -258,6 +259,23 @@ object Execution {
...
   /**
+   * Distribute the file across map/reduce nodes and
+   * give you instance of [[com.twitter.scalding.filecache.CachedFile]].
+   *
+   * The cached file available only inside execution.
+   * You can not return cached file as result of execution and must use inside closure.
+   *
+   * For multiple files you must nested your execution, see docs of [[com.twitter.scalding.filecache.DistributedCacheFile]]
+   */
+  def withCached[T](path: String)(fn: CachedFile => Execution[T]): Execution[T] = {
```
isnotinvain: how about `withCachedFile` ?

scalding-core/src/main/scala/com/twitter/scalding/filecache/DistributedCacheFile.scala
```diff
@@ -49,12 +49,34 @@ object URIHasher {
...
  *   }
  * }
+ *
+ * example with Execution:
+ *
+ * {{{
+ * object YourExecJob extends ExecutionApp {
+ *  override def job =
+ *    Execution.withCached("/path/to/your/file.txt") { file =>
+ *      doSomething(theCachedFile.path)
```
isnotinvain: should this be `doSomething(file.path)` ? 

do we need to give the user a `CachedFile` and make them call `.path` on it, or should we just give them a `new File(cached.path)` directly?

dieu: @isnotinvain `CachedFile` has `.path` and `.file`

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: done

johnynek:
This looks good to me now. If we could have a test using the hadoop platform somehow to actually demonstrate the caching working in an execution, that would be even better.

dieu:
@johnynek Added tests for this case.

cc: @piyushnarang

johnynek:
👍

scalding-core/src/main/scala/com/twitter/scalding/Config.scala
ulyssepence: Oh good point.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ulyssepence: Minor comment suggestion:
"Distributes the file onto each map/reduce node so you can use it for Scalding source creation and `TypedPipe`, `KeyedList`, etc. transformations. Using the `CachedFile` outside of `Execution` will probably not work.

For multiple..."

ulyssepence (approved):
+1

isnotinvain (approved):

johnynek:
@dieu thanks for getting this done!

Issue #1636, opened 2017-02-03
use Batched in SketchJoin at next binary incompatible release


johnynek:
using batched on the map-side dramatically improves the performance of sketch-join.

PR #1641, opened 2017-02-15
Request for Scalding release 0.17.0

piyushnarang:
Proposing a new Scalding release - 0.17.0. This release includes a major algebird version bump - 0.12.x -> 0.13.0 as well as some binary incompatible changes. If folks are onboard with this, I'll publish artifacts and make it official. 
Here's a link to the release notes: https://github.com/twitter/scalding/releases/tag/v0.17.0


piyushnarang:
cc @isnotinvain / @johnynek  / @pankajroark / @ttim

isnotinvain:
SGTM -- can you post the binary breaking changes (mima should tell you right)?

piyushnarang:
Sure, they're here: https://gist.github.com/piyushnarang/eab87e4ed8f7a4878cb102df5f28bb51

johnynek:
could we possibly hold off just a bit and publish all the projects below (chill with algebird 0.13.0), bijection? Then use the latest of those and get us ready for 2.12?

We use chill-algebird here, but the chill-algebird we use is for the old algebird, so I don't think we want that.

piyushnarang:
@johnynek sure. I think we need to publish of chill with algebird 0.13. Bijection doesn't depend on algebird right? - https://github.com/twitter/bijection/blob/develop/build.sbt. (Unless you're referring to pulling in the bijection version 0.9.5 thats on 2.12 in Scalding prior to this release).

johnynek:
yeah, you are right, the other way around: https://github.com/twitter/algebird/blob/develop/build.sbt#L11

piyushnarang:
@johnynek is someone working on the chill publish?

johnynek:
@piyushnarang not that I know of.

piyushnarang:
Ok, we have the various chill artifacts out. 0.7.7 is the version that we can pick up in Scalding.

ttim:
@piyushnarang Scalding currently depends on 0.8.x branch of chill, do we want it to depend on 0.7.x branch again? This dependency required for SB...

piyushnarang:
@ttim I released chill 0.7.7, 0.8.4 and 0.9.2 (https://github.com/twitter/chill/releases). All 3 of these versions pick up Algebird 0.13.0. In my Scalding PR, I'm picking up 0.8.4 - https://github.com/twitter/scalding/pull/1646.

johnynek:
So, I think we assumed that we would do `0.8` chill for the next scalding because we assumed the next summingbird was going to be on storm 1.0.

Maybe it makes sense to have 1 more `0.7` release of the whole stack with 2.12 support, and then follow up with storm 1.0 in summingbird?

Also note, I think there are no binary incompatible changes actually, so I think a user could use scalding with chill 0.8 now, but they need to realize the serialization may change (but that *should not* be an issue unless the user themselves has persisted some kryo data, which they should never do and if they do they must realize they have to accept responsibility for making it work correctly).

johnynek:
by the way, we really need include #1636 before 0.17. It makes a HUGE difference for us in a forked version and its a bummer for the default to be so slow.

piyushnarang:
Yeah I don't mind either. Currently it seems like we are depending on chill (0.8.x) in Scalding and 0.7.x in Summingbird. Our internal Summingbird seems to be dependent on 0.8.x. @ttim correct me if I'm wrong. 

@johnynek any idea if anyone is working on https://github.com/twitter/scalding/issues/1636 as of now? We can't bump algebird internally cause it breaks a bunch of our other libraries so if this is out in the future would be nice to not block on it.

piyushnarang:
Circling back on this. Now that we have the 2.12 PR merged (https://github.com/twitter/scalding/pull/1646), can we proceed with the 0.17.0 release? The only other PR that seems like an important inclusion is @dieu's - https://github.com/twitter/scalding/pull/1647. Anything else? @johnynek / @dieu / @ttim / @isnotinvain ?

johnynek:
Let's try to get the ones in progress merged. In particular the #1654 to
improve serialization could be a pretty big deal.

piyushnarang:
Ok, it looks like we're pretty much ready to cut this new Scala 2.12 rel. I'll do this on Monday after @johnynek's make private PR is merged: https://github.com/twitter/scalding/pull/1662.

johnynek:
can we not do 2.12? I didn't see it in the change.

piyushnarang:
@johnynek yes we can. I didn't update this after our 2.12 PRs. I've added the 2.12 module list and dropped 2.10.

piyushnarang:
@johnynek I think this is ready to be merged now. I've published the 0.17.0 artifacts (link to the draft release notes: https://github.com/twitter/scalding/releases/tag/v0.17.0). Also set the version to 0.17.1-SNAPSHOT.

johnynek (approved):

PR #1644, opened 2017-02-22
Use JobConf input files list for input size computation used by ReducersEstimators

dieu:
Fix for bug in InputSizeReduceEstimator when he calculates a size of all versions for VersionedKeyValSource instead using only one version.

adopted approach from #1428


johnynek (approved):
:+1:

gerashegalov (commented):
+1

Issue #1510, opened 2016-02-09
RatioBasedEstimator should use MAP_OUTPUT_BYTES 


gerashegalov:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/RatioBasedEstimator.scala#L64

uses FILE_BYTES_READ which has the following issues:
- it's a post-compression counter while the reducer has to decompress in order to sort
- it includes unrelated reads from DistributedCache and other side files.
- less significant since rare, does not deal correctly with in-memory shuffle and merge.

I suggest to use the pre-compression MAP_OUTPUT_BYTES counter instead.

gerashegalov:
This may require a change in hraven. 

Alternatively We may use reduceShuffleBytes to get rid of unrelated local filesystem reads. But we would need to check the map output compression codec and have some heuristic compression values it translates to.

vrushalivc:
Thanks @gerashegalov ! wanted to know what change do you foresee in hRaven, perhaps it is feasible? hRaven respose will have all the counters (whatever params are asked for)

gerashegalov:
@vrushalic MAP_OUTPUT_BYTES is a task-level MR counter. Can we have an O(1) space and runtime mapOutputBytes rolled up for a job.

PR #1645, opened 2017-02-23
Using MAP_OUTPUT_BYTES instead of FILE_BYTES_READ in RatioBasedEstimator

dieu:
fix for #1510


sriramkrishnan:
CC @gerashegalov

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -253,6 +253,7 @@ final case class FlowStepHistory(keys: FlowStepKeys,
   failedReduces: Long,
   mapFileBytesRead: Long,
   mapFileBytesWritten: Long,
+  mapOutputBytes: Long,
   reduceFileBytesRead: Long,
```
piyushnarang, line 257: if this isn't being used anymore, we could drop it?

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -179,7 +181,8 @@ object HRavenHistoryService extends HistoryService {
       } yield toFlowStepHistory(keys, step, tasks)
     }
 
-  private def toFlowStepHistory(keys: FlowStepKeys, step: JobDetails, tasks: Seq[Task]) =
+  private def toFlowStepHistory(keys: FlowStepKeys, step: JobDetails, tasks: Seq[Task]) = {
+    val mapOutputBytes = step.getCounters.getCounter(TaskCounterKey, MapOutputBytesKey).getValue
```
gerashegalov: Too prominent, and inconsistent with the way the other fields are inited. Let us inline.
I think you want getMapCounters here:  
```
step.getMapCounters.getCounter ...
```

piyushnarang: wondering if we should be a bit defensive here in case HRaven returns an empty counters list / the map output bytes counter isn't found. We can't really proceed with the hraven reducer estimator without this value but we could throw a better exception than an NPE.

piyushnarang (commented):
Couple of minor comments. Would be nice to get @gerashegalov to take a look as well as he knows a bit about this too.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -46,6 +46,8 @@ object HRavenHistoryService extends HistoryService {
     "status",
     "startTime",
     "finishTime").asJava
+  private val TaskCounterKey = "org.apache.hadoop.mapreduce.TaskCounter"
```
gerashegalov: Drop this. Let us use depending on Hadoop version:
```
com.twitter.hraven.Constants#TASK_COUNTER
```
or 
```
com.twitter.hraven.Constants#TASK_COUNTER_HADOOP2
```
instead. Ideally hraven should have a generic getWhateverCounter method abstracting Hadoop version from us. cc: @vrushalivc

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -46,6 +46,8 @@ object HRavenHistoryService extends HistoryService {
...
+  private val MapOutputBytesKey = "MAP_OUTPUT_BYTES"
```
gerashegalov: Probably not worth a constant for one time usage

gerashegalov (changes requested):
Thanks for addressing this issue!

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: this value stored in both counters

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: many of them don't being used, I'm not sure that needs to delete.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: I added the check where it used.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: done

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: at now twice :)

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: moved to map counters

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -201,6 +201,20 @@ object HRavenHistoryService extends HistoryService {
       reduceShuffleBytes = step.getReduceShuffleBytes,
       cost = 0,
       tasks = tasks)
+  }
+
+  private def mapOutputBytes(step: JobDetails): Long = {
+    if (step.getHistoryFileType == HistoryFileType.TWO) {
+      getCounterValueAsLong(step.getMapCounters, Constants.TASK_COUNTER_HADOOP2, MapOutputBytesKey)
+    } else {
+      getCounterValueAsLong(step.getMapCounters, Constants.TASK_COUNTER, MapOutputBytesKey)
```
gerashegalov, line 210: Can we move MapOutputBytesKey inside the method or inline it?

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
```diff
@@ -201,6 +201,20 @@ object HRavenHistoryService extends HistoryService {
...
+    }
+  }
+
+  private def getCounterValueAsLong(counters: CounterMap, counterGroupName: String, counterName: String): Long = {
+    val counter = counters.getCounter(counterGroupName, counterName)
+    if (counter != null) counter.getValue else 0L
```
gerashegalov, line 216: nit: maybe more concise to write as

`Option(counter).gerOrElse(0L)`

gerashegalov (approved):
LGTM, 

consider creating an Hraven issue to create public hadoop-version-independent API getCounterValue

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: @gerashegalov why? JVM will inline it.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
dieu: I think this is `Option(counter).map(_.getValue).getOrElse(0L)` less readable, isn't?

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
gerashegalov: I only meant to put it in the scope of where it's actually used, since it's only in this one method.

dieu: My preference to keep it in constant as well as other constant values.

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
gerashegalov: my Option comment is marked nit, so optional to consider :)

scalding-hraven/src/main/scala/com/twitter/scalding/hraven/reducer_estimation/HRavenHistoryService.scala
gerashegalov: noted!

piyushnarang (approved):

PR #1630, opened 2016-12-11
{WIP} Upgrade to 2.12

sritchie:
Before we can finish this out we'll need to

- [ ] publish Algebird 0.13.0 and the latest Chill
- [ ] Use the `sbt-doge` plugin to turn off 2.12 cross-builds for anything that touches Scrooge, which doesn't yet have a 2.12 build

PR #1646, opened 2017-02-25
Scalding updates for Scala 2.12

piyushnarang:
Based on @sritchie's PR (https://github.com/twitter/scalding/pull/1630) so that we can try and get this in and then cut the 0.17.0 release. Putting this up so that we can have travis catch any test failures (2.12 seems to compile ok on my machine, didn't run all tests though).


project/plugins.sbt
```diff
@@ -16,7 +16,6 @@ addSbtPlugin("com.typesafe"       % "sbt-mima-plugin"     % "0.1.12")
 addSbtPlugin("com.typesafe.sbt"   % "sbt-ghpages"         % "0.5.4")
```
johnynek, line 16: mima `0.1.14` is also out with some bug fixes.

johnynek:
Looks like some errors in the test package.

Thanks for pushing on this.

piyushnarang:
@johnynek yeah was checking out some of the failures. Will update the PR with fixes and ping you / others when the build is all green.

piyushnarang:
@johnynek / @isnotinvain / @pankajroark question for you folks. It looks like there's no version of the scrooge-sbt-plugin that supports both 2.10 and 2.12. Support for 2.10 has been dropped on their end a while back. Can we do the same as well? Drop support for 2.10?

johnynek:
if it is not too expensive can we do the same 2.10 strategy we did in chill?

I am not opposed to dropping 2.10, I guess, but honestly, if it is pretty easy to support, maybe not.

That said, if you are on 2.10 you probably aren't updating much, and we are both on 2.11, so maybe 2.11 is fine?

project/plugins.sbt
piyushnarang: Ok, I'll pick it up as well.

piyushnarang:
Yeah if we want to continue supporting 2.10, we'll have to drop the scalding-parquet and scalding-parquet-scrooge projects (as they / their tests depend on scrooge). I'm more in favor of dropping 2.10 support. This change is going to be part of a major release (0.17.0) so we can make some substantial updates (like dropping 2.10). 2.10 seems to have been EOLed around 2 years back (https://www.scala-lang.org/news/2016-schedule.html). It seems like we're going to keep running into issues from time to time if we continue to support it. Seems like it's not worth the support burden when we have 2.11.x and with this change 2.12.x.

johnynek:
sounds good to me. Also testing is slow if we have 3 versions since our tests are already crazy slow.

piyushnarang:
@johnynek yeah I initially thought that my travis build for this PR might have been down yesterday due to this 3 version explosion :-). But it turned out to be s3's fault.

johnynek:
maybe you brought down s3?

;)

project/plugins.sbt
piyushnarang: done

oscar-stripe:
looks like this line fails in 2.12:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/LineNumber.scala#L75

we need to catch the exception for lambdas.

gkk-stripe:
I found this problem with `Box.scala` when trying to upgrade `scalding-core` to Scala 2.12: https://issues.scala-lang.org/browse/SI-10232

The work-around is to remove half of boxes from `Box.scala`. Once I've done it, I got all tests in core to pass.

johnynek:
Note that @gkk-stripe found this issue we are hitting with boxed:

https://issues.scala-lang.org/browse/SI-10232

johnynek:
:) race...

piyushnarang:
Thanks @oscar-stripe. Will take a look in the next couple of days. Got a bit busy with some other stuff.

piyushnarang:
Ah thanks @gkk-stripe that's really cool. I was breaking my head against that one for a while :-). I'll try out your suggestion in the next couple of days (or if you want to push a commit on this PR so that you get credit for it, feel free to do so).

johnynek (commented):
Those may need to be on different objects and we may need to import. Can we make those objects private?

piyushnarang:
@johnynek yeah I can try that. I did try running the scalding-core tests and they worked locally with these changes. Seems to be failing on travis though.

johnynek:
Well, if it works it works.

piyushnarang:
Trying to figure out why the travis build is failing. Tried building a couple of the projects on my machine and they seem to succeed:
`./sbt ++2.12.1 clean scalding-thrift-macros/test` same with scalding-core/test.

johnynek:
I am seeing:
```
  at com.twitter.scalding.LineNumber$.$anonfun$tryNonScaldingCaller$2(LineNumber.scala:75)
[info]   at com.twitter.scalding.LineNumber$.$anonfun$tryNonScaldingCaller$2$adapted(LineNumber.scala:74)
[info]   at com.twitter.scalding.LineNumber$$$Lambda$401/869703951.apply(Unknown Source)
[info]   at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:502)
[info]   at com.twitter.scalding.LineNumber$.headOption$1(LineNumber.scala:68)
[info]   at com.twitter.scalding.LineNumber$.tryNonScaldingCaller(LineNumber```

Which I mentioned as a problem above: 
https://github.com/twitter/scalding/pull/1646#issuecomment-287209798

we need to handle line numbers differently for 2.12 since lambdas might not be able to reflect like that?

**Comment by piyushnarang** [comment]
@johnynek yeah I pinged you on Gitter. So this seems to be an issue only on travis. I’ve tried a bunch of combinations of trying to repro this locally (directly calling ./sbt ++2.12.1 scalding-thrift-macros/test, invoking it via our run_test.sh script, switching to java 8_31 as thats what travis uses). None of those seem to trigger it. Not sure if I’m missing something there, it seems to consistently succeed on my machine (also got Timur to test on his machine).

**Comment by piyushnarang** [comment]
Debugging this locally, I don't see the 'com.twitter.scalding.typed.ReduceStep$$Lambda$275/727580562' showing up in the list of StackTraceElements passed to the method which is why it seems like I don't hit this error on my machine. Not sure why / how they're present on the travis CI build.

**Comment by johnynek** [comment]
Piyush, you are testing locally with 2.12? It would only happen with 2.12.

**Comment by piyushnarang** [comment]
@johnynek while debugging this in Intellij I updated my scalaVersion in the build.sbt file to 2.12.1. While trying to repro before I ran the commands above (and Timur also tried one set) on my command line.

**Comment by johnynek** [comment]
Well, i mean, in any case `Class.forName` might throw. We could put a try/catch and handle that in a decent way. Even if we can't repro, I can see how it may be a bug, I just don't know in what cases it is triggered. The only risk to such an approach is that the bug is elsewhere and we are hiding it with a fix like this. But if we can get the tests green locally and on Travis, that seems okay to me (if we don't have to change the tests to do it).

**Comment by piyushnarang** [comment]
@johnynek yeah I had the changes to catch the CNF handy but I was also worried about the masking of the issue elsewhere which is why I was trying out all those combinations to repro :-). I can push that upstream so that we can see if that makes Travis happy.

**Comment by piyushnarang** [comment]
@johnynek ok the build is green now :-). Do take a look when you get the time. I'll also sanity check this and clean up stuff when I get in tomorrow.

File: build.sbt [hunk 2]
```diff
@@ -15,33 +14,34 @@ def scalaBinaryVersion(scalaVersion: String) = scalaVersion match {
...
-def isScala210x(scalaVersion: String) = scalaBinaryVersion(scalaVersion) == "2.10"
 
 val algebirdVersion = "0.13.0"
 val apacheCommonsVersion = "2.2"
 val avroVersion = "1.7.4"
-val bijectionVersion = "0.9.1"
+val bijectionVersion = "0.9.5"
 val cascadingAvroVersion = "2.1.2"
-val chillVersion = "0.8.3"
+val chillVersion = "0.8.4"
```
johnynek, line 23: are we sure we want to bump to `0.8.x` vs `0.7.7`? Seems summingbird on storm 1.0 has still not landed, so we might want to have one last summingbird with 2.12? (I'm working a lot with summingbird and that may help me).

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -201,8 +201,17 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
     Zipped(this, that)
 
   override val hashCode: Int = ScalaRunTime._hashCode(self)
-  override def equals(other: Any): Boolean =
-    hashCode == other.hashCode && ScalaRunTime._equals(self, other)
+
+  override def equals(other: Any): Boolean = {
+    //removed from Scala 2.12
+    def _equals(x: Product, y: Any): Boolean = y match {
+      case y: Product if x.productArity == y.productArity => x.productIterator sameElements y
```
johnynek: can we wrap this line:
```scala
 case y: Product if x.productArity == y.productArity =>
   x.productIterator sameElements y.productIterator
```

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -41,12 +41,21 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
...
+
     pipe
-      .map { case (k, _) => cms.create(Bytes(serialization(k))) }
-      .sum // sum everything to one value
-      .toTypedPipe
-      .forceToDisk
+      .map { case (k, _) => ((), Batched(cms.create(Bytes(serialize(k))))) }
+      .sumByLocalKeys
+      .map {
+        case (_, batched) => batched.sum
```
johnynek: this looks like some kind of bad merge. I don't think this is part of this. Can you rebase develop and see if this goes away?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
```diff
@@ -106,6 +125,24 @@ class ReducerEstimatorTest extends WordSpec with Matchers with HadoopSharedPlatf
         .run()
     }
 
+    "run with correct number of reducers when we have a glob pattern in path" in {
```
johnynek: this looks like another merge issue.

scalding-repl/src/main/scala/com/twitter/scalding/ScaldingShell.scala
```diff
@@ -124,7 +124,7 @@ trait BaseScaldingShell extends MainGenericRunner {
    * @param args from the command line.
    */
   def main(args: Array[String]): Unit = {
-    val retVal = process(args)
+    val retVal = process(ExpandLibJarsGlobs(args))
```
johnynek: another merge issue.

johnynek (changes requested):

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
piyushnarang: sure, seems cleaner will do.

build.sbt
piyushnarang: @johnynek I can switch this to 0.7.7. If I understand correctly the plan would be:
1) Merge this PR with chill 0.7.7 (might need to revert the test fix here: https://github.com/twitter/scalding/pull/1634/files).
2) Cut a 0.17.0 release 
3) You can use that to cut a SB 2.12 rel on chill 0.7.x
4) Switch back to chill 0.8.4
5) Cut a 0.17.1 rel. 
6) We can use this for our SB 2.12 rel on chill 0.8.x

@ttim - does this sound reasonable to you as well?

ttim: @johnynek summingbird with storm 1.0 support wasn't released because I'm waiting on scalding/storehaus to be released with 0.8.x chill.

In order to release SB with 0.9 storm and 2.12 scala support we also need to release tormenta with the same. Latest tormenta's release uses storm 1.0 (which is not good for above) and supports 2.12.

So I see two possibilities:
1. If we really want to release SB with 0.9 storm and 2.12 scala:
- release old version of tormenta with 2.12 support only, without 1.0 storm
- release scalding 0.17.0 as @piyushnarang said
- release storehaus & SB with 2.12 & 0.9 storm.

And iterate again with scalding/storehaus/SB releases to get SB with 2.12 & 1.0 storm.

2. Or if we are ok without SB for 2.12 and 0.9 storm:
- release scalding with 0.8.x chill & 2.12 support
- release storehaus/SB with 2.12 & 1.0 storm

One more note - we use SB with 1.0 storm at Twitter for last couple of months and it works well. Transition to new api wasn't painful.

scalding-parquet-scrooge-fixtures/src/test/resources/test.thrift
```diff
@@ -86,14 +86,6 @@ struct StructWithReorderedOptionalFields {
   1: optional i32 fieldOne,
 }
 
-struct TestPersonWithRequiredPhone {
```
oscar-stripe, line 89: why did this need to get nuked?

piyushnarang: The scrooge generator was cribbing cause this was being declared twice. The old version seem to not mind.

piyushnarang: In this file: https://github.com/twitter/scalding/blob/7ed0f92a946ad8407645695d3def62324f78ac41/scalding-parquet-scrooge-fixtures/src/test/resources/test.thrift it's on 89 and 107

oscar-stripe: ahh. Cool.

johnynek (approved):

build.sbt
```diff
@@ -15,33 +14,34 @@ def scalaBinaryVersion(scalaVersion: String) = scalaVersion match {
...
+val scalaCheckVersion = "1.13.4"
+val scalaTestVersion = "3.0.1"
+val scroogeVersion = "4.12.0"
 val slf4jVersion = "1.6.6"
 val thriftVersion = "0.5.0"
 val junitVersion = "4.10"
 val macroCompatVersion = "1.1.1"
+val jlineVersion = "2.14.3"
+
+def scrooge(scalaVersion: String) = "com.twitter" %% "scrooge-serializer" % scroogeVersion
```
fwbrasil: What's the purpose of this method? It doesn't use the `scalaVersion` parameter, maybe it should be a `val`?

johnynek:
👍

johnynek:
(when we merge, please squash merge this).

johnynek:
(which github's merge button supports).

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -201,8 +201,17 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
     Zipped(this, that)
 
   override val hashCode: Int = ScalaRunTime._hashCode(self)
-  override def equals(other: Any): Boolean =
-    hashCode == other.hashCode && ScalaRunTime._equals(self, other)
+
+  override def equals(other: Any): Boolean = {
+    //removed from Scala 2.12
+    def _equals(x: Product, y: Any): Boolean = y match {
```
ttim: I prefer to have this method  inlined and without the comment.

ttim (approved):
Minor comment

build.sbt
johnynek: Okay, I think moving forward with 0.8 chill seems fine then. Let's move forward and go ahead to get on 1.0 storm.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
```diff
@@ -560,7 +562,7 @@ object Boxed {
    * using `Any` here instead of parametrizing everything over a type parameter
    * `K`.
    */
-  private[this] val allBoxes = List(
+  val boxes1 = List(
```
fwbrasil: could these lists be `private[serialization]`?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
fwbrasil: +1

build.sbt
johnynek: I think that was old stuff from when we were still supporting scala 2.10 (which scrooge stopped supporting).

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
johnynek: +1 to as private as possible. Good catch.

build.sbt
piyushnarang: yeah I can drop this. Didn't clean it up after a couple of iterations.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Boxed.scala
piyushnarang: ok, let me try that.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
piyushnarang: Cool, will do.

piyushnarang:
@ttim / @fwbrasil - updated with your comments

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -203,14 +203,13 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
   override val hashCode: Int = ScalaRunTime._hashCode(self)
 
   override def equals(other: Any): Boolean = {
-    //removed from Scala 2.12
-    def _equals(x: Product, y: Any): Boolean = y match {
-      case y: Product if x.productArity == y.productArity =>
-        x.productIterator sameElements y.productIterator
+    val productEquals = other match {
+      case p: Product if self.productArity == p.productArity =>
```
ttim: nit: I guess `hashCode` check was made before `_equals` call for perf reasons. Therefore this code has slightly worse performance if it matters.

I recommend  just remove separate `hashCode == other.hashCode` check and put it into `if` part.

piyushnarang: Updated this. Not sure how large those products get. I moved the hashCode into the case body.

fwbrasil:
:+1: LGTM

PR #1648, opened 2017-02-28
Use Batched in Sketch production

johnynek:
I'm not sure why we didn't enable this. Actually at Stripe, we use a basically infinite batch (we assume we can store the serialized versions of all the keys).

Here I batch every 10k keys, since the hashes are 128 bit (or 16 bytes), this is only 160KB of ram. I think we could do even larger, but I'm not sure it adds too much (every 10k additions of copying the CMS).


piyushnarang:
👍

PR #1650, opened 2017-03-02
Fixed reduce estimator for paths with a glob pattern

dieu:


ulyssepence:
+1

I checked on `FileSystem#globStatus(...)`:
https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/fs/FileSystem.html#globStatus(org.apache.hadoop.fs.Path)

Sounds reasonable.

johnynek:
+1

dieu:
@benpence / @johnynek added one test for this case.

piyushnarang:
👍

PR #1652, opened 2017-03-09
Estimate input size only for Hfs and GlobHfs taps

dieu:


scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -63,13 +63,18 @@ object Common {
...
     FileInputFormat
       .getInputPaths(conf)
-      .map { path =>
+      .flatMap { path =>
         val fs = path.getFileSystem(conf)
-        val size = fs.globStatus(path)
-          .map(status => fs.getContentSummary(status.getPath).getLength)
-          .sum
+        val files = fs.globStatus(path)
+        if (files != null) {
```
piyushnarang: add a comment here for when we could end up with null?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -63,13 +63,18 @@ object Common {
 
     FileInputFormat
       .getInputPaths(conf)
-      .map { path =>
+      .flatMap { path =>
```
isnotinvain: an easy way to do this is:

```
FileInputFormat
      .getInputPaths(conf)
      .flatMap { path =>
        val fs = path.getFileSystem(conf)
        Option(fs.globStatus(path)).map { files => 
          val size = files
            .map(status => fs.getContentSummary(status.getPath).getLength)
            .sum
          path -> size
        }
      }
```

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -32,8 +32,8 @@ object InputSizeReducerEstimator {
     Common.inputSizes(info.step) match {
       case Nil =>
         LOG.warn("InputSizeReducerEstimator unable to estimate reducers; " +
-          "cannot compute size of:\n - " +
-          Common.unrollTaps(info.step).filterNot(_.isInstanceOf[Hfs]).mkString("\n - "))
+          "cannot compute size one of:\n - " +
```
isnotinvain: maybe `"cannot compute size of one of:\n"`

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
```diff
@@ -143,6 +180,23 @@ class ReducerEstimatorTest extends WordSpec with Matchers with HadoopSharedPlatf
 
     }
 
+    "run and catch memory source on right way" in {
```
isnotinvain: can you make this more descriptive? Maybe "Don't fail when a memory source is used" or "Ignore memory source in input size estimation" or something like that?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
```diff
@@ -81,6 +81,43 @@ class SimpleGlobJob(args: Args, customConfig: Config) extends Job(args) {
     .write(counts)
 }
 
+class SimpleMemoryJob(args: Args, customConfig: Config) extends Job(args) {
+  import HipJob._
+
+  val inSrc = IterableSource(List(
+    "Direct trade American Apparel squid umami tote bag. Lo-fi XOXO gluten-free meh literally, typewriter readymade wolf salvia whatever drinking vinegar organic. Four loko literally bicycle rights drinking vinegar Cosby sweater hella stumptown. Dreamcatcher iPhone 90's organic chambray cardigan, wolf fixie gluten-free Brooklyn four loko. Mumblecore ennui twee, 8-bit food truck sustainable tote bag Williamsburg mixtape biodiesel. Semiotics Helvetica put a bird on it, roof party fashion axe organic post-ironic readymade Wes Anderson Pinterest keffiyeh. Craft beer meggings sartorial, butcher Marfa kitsch art party mustache Brooklyn vinyl.",
```
isnotinvain, line 90: haha where does this come from?

isnotinvain (changes requested):

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
```diff
@@ -81,6 +81,43 @@ class SimpleGlobJob(args: Args, customConfig: Config) extends Job(args) {
...
+    .flatMap(_.split("[^\\w]+"))
+    .map(_.toLowerCase -> 1)
+    .group
+    // force the number of reducers to two, to test with/without estimation
+    .withReducers(2)
+    .sum
+    .write(counts)
+}
+
+class SimpleFileNotFoundJob(args: Args, customConfig: Config) extends Job(args) {
```
piyushnarang: this isn't being used right? Should we add a test to include this scenario? Source is a Hfs source but the path points to a file that doesn't exist?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: I actually rethinking this and use `Try`, because in many ways it can throw IOException

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
dieu: hipster.txt :)

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
piyushnarang: maybe expand that this can be due to a memory tap / tap with missing files?

scalding-hadoop-test/src/test/scala/com/twitter/scalding/reducer_estimation/ReducerEstimatorTest.scala
dieu: removed

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
dieu: we need change API a little bit to show what we can not estimate, I can do it if you want

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: rewrote it.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
piyushnarang: I'm ok with leaving the API as-is for now. Maybe just update the text to say that we typically can't compute sizes of memory taps / files not found. That will make debugging in the future easier.

dieu: like that `cannot compute size of one of (usually it's memory taps or files not found)`?

piyushnarang: how about `cannot compute size of one of (is it a memory tap / are files not found?): ... ` ?

piyushnarang:
@dieu should we add a test for the file not found scenario? Or are they already covered?

dieu:
@piyushnarang I'm not sure that we can run this tests, because hadoop itself will raise error.

piyushnarang:
Ok, can we test it manually then to ensure it works as expected?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
dieu: done

dieu:
@piyushnarang I added tests for FileNotFound

```
2017-03-09 12:39:41,848 WARN  [pool-47-thread-1] reducer_estimation.InputSizeReducerEstimator$ (InputSizeReducerEstimator.scala:estimateReducersWithoutRounding(34)) - InputSizeReducerEstimator unable to estimate reducers; cannot compute size of one of (usually it's memory taps or files not found):
 - Hfs["TextLine[['offset', 'line']->[ALL]]"]["file.txt"]
2017-03-09 12:39:41,860 INFO  [pool-47-thread-1] flow.FlowStep (BaseFlowStep.java:logInfo(834)) - [com.twitter.scalding.r...] starting step: (1/1) counts.tsv
2017-03-09 12:39:42,402 INFO  [flow com.twitter.scalding.reducer_estimation.SimpleFileNotFoundJob] flow.Flow (BaseFlow.java:logInfo(1378)) - [com.twitter.scalding.r...] stopping all jobs
2017-03-09 12:39:42,403 INFO  [flow com.twitter.scalding.reducer_estimation.SimpleFileNotFoundJob] flow.FlowStep (BaseFlowStep.java:logInfo(834)) - [com.twitter.scalding.r...] stopping: (1/1) counts.tsv
2017-03-09 12:39:42,408 INFO  [flow com.twitter.scalding.reducer_estimation.SimpleFileNotFoundJob] flow.Flow (BaseFlow.java:logInfo(1378)) - [com.twitter.scalding.r...] stopped all jobs
2017-03-09 12:39:42,956 ERROR [ResourceManager Event Processor] resourcemanager.ResourceManager (ResourceManager.java:run(594)) - Returning, interrupted : java.lang.InterruptedException

```

piyushnarang (approved):

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -63,13 +63,15 @@ object Common {
...
-        val fs = path.getFileSystem(conf)
-        val size = fs.globStatus(path)
-          .map(status => fs.getContentSummary(status.getPath).getLength)
-          .sum
-
-        path -> size
+      .flatMap { path =>
+        Try {
+          val fs = path.getFileSystem(conf)
+          val size = fs.globStatus(path)
```
isnotinvain: we should still check for null so that we can have better error messages right? Actually this Try {} block is going to swallow exceptions quietly which doesn't seem great? The explicit check for null via the Option constructor seemed better -- if we do get an IOException here, wouldn't we want it to bubble up?

johnynek: +1. If the NPE is the issue, let's handle it directly:
```scala
val size = fs.globStatus(path) match {
  case null => 0
  case items => items.map(...).sum
}
```

johnynek: That said, I'm not sure you want to be failing the job here if you have an exception. Presumably we want to just log and give warnings that reducer estimation may be off and go ahead assuming they were empty?

dieu:
@isnotinvain rewrote to more safer way.

scalding-commons/src/main/java/com/twitter/scalding/commons/tap/VersionedTap.java
```diff
@@ -107,6 +108,21 @@ public void sinkConfInit(FlowProcess<JobConf> process, JobConf conf) {
   }
 
   @Override
+  public long getSize(JobConf conf) throws IOException {
+    Path path = new Path(getSourcePath(conf));
+    FileSystem fs = getFileSystem(conf);
+    FileStatus[] statuses = fs.globStatus(path);
```
isnotinvain: still need our null checks here probably

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -55,24 +55,30 @@ object Common {
...
-    val conf = step.getConfig
-
-    FileInputFormat
-      .getInputPaths(conf)
-      .map { path =>
-        val fs = path.getFileSystem(conf)
-        val size = fs.globStatus(path)
-          .map(status => fs.getContentSummary(status.getPath).getLength)
+  def size(tap: Hfs, conf: JobConf): Long = {
+    val tapSize = Try { tap.getSize(conf) }.toOption
```
isnotinvain: why the Try / toOption? can we just let errors bubble up? This shouldn't throw in a normal case right? 

At a minimum we should probably log these exceptions

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -55,24 +55,30 @@ object Common {
...
+
+    tapSize
+      .filter(_ > 0)
+      .getOrElse {
+        val fs = tap.getPath.getFileSystem(conf)
+        fs.globStatus(tap.getPath)
```
isnotinvain: check for nulls here too

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -55,24 +55,30 @@ object Common {
...
-  def inputSizes(step: FlowStep[JobConf]): Seq[(Path, Long)] = {
-    val conf = step.getConfig
-
-    FileInputFormat
-      .getInputPaths(conf)
-      .map { path =>
-        val fs = path.getFileSystem(conf)
-        val size = fs.globStatus(path)
-          .map(status => fs.getContentSummary(status.getPath).getLength)
+  def size(tap: Hfs, conf: JobConf): Long = {
```
isnotinvain: lets document how this works:

1) Ask the tap for its size
2) If the reported size is 0, as the tap for it's path and check the size ourself

isnotinvain (changes requested):

scalding-commons/src/main/java/com/twitter/scalding/commons/tap/VersionedTap.java
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: ok

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: because cascading implementation can throw an error even if we can calculate size, first we try to get the size from the tap then fallback to our method to calculate, we don't need to log the error because it will log in next step if it the stable error.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: done

scalding-commons/src/main/java/com/twitter/scalding/commons/tap/VersionedTap.java
isnotinvain: Does this get called after validateTaps? Hopefully otherwise there will be race conditions where this returns  0 but the file shows up later and then we estimate poorly

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -55,24 +55,34 @@ object Common {
...
-      .getInputPaths(conf)
-      .map { path =>
-        val fs = path.getFileSystem(conf)
-        val size = fs.globStatus(path)
-          .map(status => fs.getContentSummary(status.getPath).getLength)
-          .sum
-
-        path -> size
+  def size(tap: Hfs, conf: JobConf): Option[Long] = {
+    val tapSize = Try { tap.getSize(conf) }.toOption
```
isnotinvain: lets just let this fail if it fails?

isnotinvain:
@dieu before I forget, you mentioned this would break all normal HFS instances, so we need to handle that too

dieu:
@isnotinvain no, we handle existing HFS instances, it why I use `Try` to `getSize` on tap, because cascading Hfs doesn't handle glob patterns.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: I know the case when we can calculate by ourself, because generic Hfs getSize doesn't handle paths glob.

dieu:
@piyushnarang / @isnotinvain / @johnynek  please review.

scalding-core/src/main/java/com/twitter/scalding/tap/SizeableHfs.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapred.JobConf;
+import org.apache.hadoop.mapred.OutputCollector;
+import org.apache.hadoop.mapred.RecordReader;
+
+import cascading.scheme.Scheme;
+import cascading.tap.hadoop.Hfs;
+
+public class SizeableHfs extends Hfs {
```
piyushnarang: Lets add a couple of lines of javadoc on why this class is needed. 
Not sure if SizeableHfs is the best name for this as sizeable sounds it you could adjust the size which isn't the case here. Don't have too many great alternatives - GlobSizeHfs or GlobHfs?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -32,8 +32,8 @@ object InputSizeReducerEstimator {
     Common.inputSizes(info.step) match {
       case Nil =>
         LOG.warn("InputSizeReducerEstimator unable to estimate reducers; " +
-          "cannot compute size of:\n - " +
-          Common.unrollTaps(info.step).filterNot(_.isInstanceOf[Hfs]).mkString("\n - "))
+          "cannot compute size of one of (is it a non hfs tap?):\n - " +
```
piyushnarang: this is back to `cannot compute the size of(is it a non Hfs tap): Tap1, Tap2` right? cause we can have a JDBC tap in there with a HDFS based tap and we'd just skip the jdbc one.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -54,23 +54,13 @@ object Common {
...
-      .getInputPaths(conf)
-      .map { path =>
-        val fs = path.getFileSystem(conf)
-        val size = fs.globStatus(path)
-          .map(status => fs.getContentSummary(status.getPath).getLength)
-          .sum
-
-        path -> size
-      }
+    unrollTaps(step) flatMap {
```
piyushnarang: seems like we currently only fail if:
1) There's some exception thrown while computing the size of a tap
2) We can't compute the size of all the taps in a step. In case case we end up with an empty list and log the warning. 

If I understand correctly, this has been the prior behavior. Do we want to log a warning / fail if we can't compute the size of 1 of these n taps? I guess the ones we're explicitly skipping now are just the non-Hfs ones so just logging should be ok?

scalding-core/src/main/java/com/twitter/scalding/tap/SizeableHfs.java
```diff
@@ -0,0 +1,48 @@
...
+    FileSystem fs = path.getFileSystem(conf);
+    FileStatus[] statuses = fs.globStatus(path);
+
+    if (statuses != null) {
+      long size = 0;
+      for (FileStatus status : statuses) {
+        size += fs.getContentSummary(status.getPath()).getLength();
+      }
+      return size;
+    } else {
```
pankajroark: nit: May be do this to avoid indentation to the right:

if (status == null) {
  throw new FileNotFoundException...
}

long size = 0;
...

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
pankajroark: nit: keep '.' i.e. unrollTaps(step).flatMap to be consistent with rest of the file and twitter coding conventions.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
dieu: yes, this case handle when we have only non Hfs taps.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: yes it back to prior behavior, and we don't care about non Hfs taps.

scalding-core/src/main/java/com/twitter/scalding/tap/SizeableHfs.java
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: done

dieu: let's logging non Hfs taps.

scalding-core/src/main/java/com/twitter/scalding/tap/SizeableHfs.java
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -54,23 +56,15 @@ object Common {
   def unrollTaps(step: FlowStep[JobConf]): Seq[Tap[_, _, _]] =
     unrollTaps(step.getSources.asScala.toSeq)
 
-  /**
-   * Get the total size of the input paths, which may contain a glob
-   * pattern in its path, so we must be ready to handle that case.
-   */
-  def inputSizes(step: FlowStep[JobConf]): Seq[(Path, Long)] = {
+  def inputSizes(step: FlowStep[JobConf]): Seq[(String, Long)] = {
```
johnynek, line 59: do we need to change the type of the result here? Could we keep it be `Path`?

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatform.scala
```diff
@@ -33,6 +35,9 @@ trait HadoopPlatform[P, R, T <: HadoopPlatform[P, R, T]] {
 
   def run(): Unit
 
+  def run(fn: Throwable => Unit): Unit =
```
johnynek: can we call this `def runAndHandle` or something?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -1,16 +1,16 @@
...
 
-import cascading.flow.{ FlowStep, Flow, FlowStepStrategy }
+import cascading.flow.{Flow, FlowStep, FlowStepStrategy}
+import cascading.tap.hadoop.Hfs
 import com.twitter.algebird.Monoid
-import com.twitter.scalding.{ StringUtility, Config }
-import cascading.tap.{ Tap, CompositeTap }
-import org.apache.hadoop.fs.Path
-import org.apache.hadoop.mapred.{ FileInputFormat, JobConf }
+import com.twitter.scalding.{Config, StringUtility}
```
piyushnarang: seems like the scalding setting is to include a space after the `{` and before `}`. You might want to update your intellij.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
```diff
@@ -54,23 +56,15 @@ object Common {
...
-          .map(status => fs.getContentSummary(status.getPath).getLength)
-          .sum
-
-        path -> size
-      }
+    unrollTaps(step).flatMap {
+      case tap: GlobHfs => Some(tap.toString -> tap.getSize(conf))
+      case tap: Hfs => Some(tap.toString -> GlobHfs.getSize(tap.getPath, conf))
+      case tap =>
+        LOG.info("InputSizeReducerEstimator unable to calculate size: " + tap)
```
piyushnarang: maybe make this a warn?

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
```diff
@@ -32,7 +32,7 @@ object InputSizeReducerEstimator {
     Common.inputSizes(info.step) match {
       case Nil =>
         LOG.warn("InputSizeReducerEstimator unable to estimate reducers; " +
-          "cannot compute size of:\n - " +
+          "cannot compute size of one of (is it a non hfs tap?):\n - " +
```
piyushnarang: this should be `cannot compute size of (is it a non hfs tap?)` ?

piyushnarang (approved):
Updates look good to me. Couple of minor comments.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: ok. done. I thought about warn, but then thought that is normal path.

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/InputSizeReducerEstimator.scala
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/reducer_estimation/Common.scala
dieu: yes, because we are calculating input size for taps, instead for paths.

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatform.scala
dieu: done

scalding-hadoop-test/src/main/scala/com/twitter/scalding/platform/HadoopPlatform.scala
```diff
@@ -33,6 +35,9 @@ trait HadoopPlatform[P, R, T <: HadoopPlatform[P, R, T]] {
 
   def run(): Unit
 
+  def runAndHandle(fn: Throwable => Unit): Unit =
+    fn(Try { run() }.failed.get)
```
johnynek: wait, is this what you expect? If the job succeeds, `.failed` still gives an exception. Can you comment then intent of this method? Interestingly, here is a case where adding a type parameter helps explain.

If you wrote:
```scala
def runAndHandle[T](fn: Throwable => T): T =
```
we could see in the types that `fn` must be called, so even if the job succeeds, an exception will still be generated and passed to `fn`. But since anyone can create a value of `Unit`, I can't tell if you meant:
```scala
Try(run()) match {
  case Failure(err) => fn(err)
  case Success(_) => ()
}
```

If you really mean what you have, we should probably call it `def runExpectFailure[T]` or something.

dieu: So, this method needs when you expect an exception to check exception path, in this case success unexpected behaviour and user should catch it as exception.

not sure about type parameter, but let's do it.

johnynek:
👍

johnynek (approved):

PR #1655, opened 2017-03-22
Improve Vertica support in scalding-db

alambert:
(AITOOLS-4225, ADI-2692)

I've been attempting to read data from Vertica using scalding-db. I've found that a few of Vertica's type names differ slightly from MySQL's type names, but other than that scalding-db has worked.

This pull request widens the set of mapped type names. It includes unit tests to verify the new type names are accepted. I've tested this with a simple exporter for a Vertica table containing `date`, `varchar`, and `integer` columns.

### DIFF
--- a/scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
+++ b/scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
@@ -212,8 +212,6 @@ object ColumnDefinitionProviderImpl {
           case "BIGINT" =>
             q"""List("INTEGER", "INT", "BIGINT", "INT8", "SMALLINT",
                "TINYINT", "SMALLINT", "MEDIUMINT").contains($typeNameTerm)"""
-          // for Vertica support
-          case "DATE" => q"""List("DATE").contains($typeNameTerm)"""
           case f => q"""$f == $typeNameTerm"""
         }
         val typeAssert = q"""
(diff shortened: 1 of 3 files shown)


scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
```diff
@@ -191,8 +191,10 @@ object ColumnDefinitionProviderImpl {
       case (cf: ColumnFormat[_], pos: Int) =>
         val fieldName = cf.fieldName.toStr
         val typeNameTerm = newTermName(c.fresh(s"colTypeName_$pos"))
+        // MySQL uses names like `DATE`, `INTEGER` and `VARCHAR`;
+        // Vertica uses names like `Date`, `Integer` and `Varchar`
         val typeName = q"""
-        val $typeNameTerm = $rsmdTerm.getColumnTypeName(${pos + 1})
+        val $typeNameTerm = $rsmdTerm.getColumnTypeName(${pos + 1}).toUpperCase(java.util.Locale.US)
```
johnynek, line 197: I'm confused by this. Can you comment how this works for Vertica if it expects non upper cased?

johnynek: Oh. It's matched in the case matching below.

johnynek (commented):
Seems fine to me.

ulyssepence (approved):

PR #1656, opened 2017-03-23
Add andThen to Mappable

lclementi:
Mappable.andThen is defined in TypedSource and when invoked it
returns a TypedSource. In this patch I'm overriding that method
so that when you invoke a Mappable.andThen it returns a Mappable
object. This can be useful when the user of Mappable wants to do
toIeratable after applying a transformation using andThen

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/SourceSpec.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/SourceSpec.scala
@@ -104,3 +104,54 @@ class AddRemoveOneJob(args: Args) extends Job(args) {
 
     .write(RemoveOneTsv("output"))
 }
+
+class MapTypedPipe(args: Args) extends Job(args) {
+  TypedPipe.from(TypedText.tsv[(Int, String)]("input"))
+    .map(MapFunctionAndThenTest.mapFunction)
+    .write(TypedText.tsv[(Int, String, Int)]("output"))
+}
+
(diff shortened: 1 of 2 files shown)


johnynek:
Can we do any test? For instance, start with a mappable, do `.andThen` and see that it is the same as calling `.map` on the TypedPipe.from the original one?

lclementi:
Hey @johnynek ,
does this new version does it?

Thanks for taking a look,
Luca

scalding-core/src/test/scala/com/twitter/scalding/SourceSpec.scala
```diff
@@ -104,3 +104,54 @@ class AddRemoveOneJob(args: Args) extends Job(args) {
...
+  "Mappable.andThen is like TypedPipe.map" should {
+    JobTest(new MapTypedPipe(_))
+      .source(TypedText.tsv[(Int, String)]("input"), input)
+      .typedSink(TypedText.tsv[(Int, String, Int)]("output")){ outputBuffer =>
+        val outMap = outputBuffer.toList
+        "typed pipe return proper results" in {
+          outMap should have size 3
+          outMap shouldBe output
+        }
+      } //(implicitly[TypeDescriptor[(Int, String, Int)]].converter)
```
johnynek: do we need the comment hanging out?

johnynek:
looks good, just one question about a stray comment.

johnynek (approved):

johnynek:
Thanks!

PR #1661, opened 2017-04-04
Simplify match statement and use collection.breakOut

zaneli:


johnynek:
👍 

This is great. Thanks!

PR #1665, opened 2017-04-12
Adds Execution tutorial

ulyssepence:
I've been meaning to write this tutorial for a long time. At Twitter, it seems like there's a paucity of understanding of `Execution`. I don't pretend to be an expert though and would love to hear from other community members about best practices.

Some of the code samples are pretty contrived and could probably use some work.

I'm not too familiar with the cache, so I could definitely use some extra eyes on that.

Also, not sure if how others have differentiated between Scalding "job" and `Job`.

### DIFF
--- a/tutorial/Execution.md
+++ b/tutorial/Execution.md
@@ -1,0 +1,412 @@
+# A Random Walk Down Executions or: You Could Have Invented Executions or: Learn You An Execution For Greater Good
+
+## What is a Scalding Execution?
+
+A Scalding job (lowercase 'j') is the distributed completion of a DAG of Hadoop jobs that read, transform, and write data, usually to the Hadoop Distributed File System. A Scalding `Execution` is conceptually a plan to run zero or more Scalding jobs. The Scalding library has another paradigm, the Scalding `Job` class, that enables users to plan a single Scalding job in the constructor of a class. `Execution` is a newer, more composable, and more functional approach to running Scalding jobs.
+
+## Why would I want to use an Execution?
+
+The Scalding `Job` class uses a more object oriented programming approach to planning Scalding jobs, which often makes it more difficult to reason about and extend functionality. In addition, running multiple Scalding `Job` instances is poorly supported.
+
+Scalding `Execution`, on the other hand, is inspired by functional programming and composes easily. Here are a few things that are much easier with Scalding `Execution`:
+
+* Calling out to a service asynchronously to read or write some state before or after a Scalding job runs
+* Running multiple Scalding jobs or not running any Scalding jobs at all
+* Unit testing
+* Using the output of one Scalding job to decide how to plan future jobs
+
+These situations come up frequently when integrating Scalding jobs into an analytics workflow.
+
+We may want to avoid using Scalding `Execution` because it is more challenging to understand for newcomers to applied category theory (e.g. functors, applicative functors, and monads). This tutorial does not assume any knowledge of these concepts, but familiarity with them will make these explanations much easier to understand. In this respect, Scalding `Job` is simpler because it requires understanding less paradigms.
+
+## Reasoning about how executions work
+
+At its heart, a Scalding `Execution` wraps a function that takes in a Scalding `Config` and a Scalding `Mode` and does *something*. A `Config` contains configuration properties for the running of a Scalding job (although we can access the `Config` properties when creating a Scalding `Execution`, so theoretically they can be used for anything). The `Mode` tells the Scalding job what environment (in-memory, in a local Hadoop cluster, in a remote Hadoop cluster) to run under. The Scalding `Mode` also contains an `Args` object, which represent Scalding-specific arguments passed on the command line.
+
+For now, let's define a simplified `Execution` like so:
+
+```scala
+case class Execution[T](private doSomething: (Config, Mode) => T) {
+  final def run(config: Config, mode: Mode): T = {
+    doSomething(config, mode)
+  }
+}
+
+object Execution {
+  def getConfig: Execution[Config] = {
+    def newDoSomething(config: Config, mode: Mode) = config
+
+    Execution(newDoSomething)
+  }
+
+  def getMode: Execution[Mode] = {
+    def newDoSomething(config: Config, mode: Mode) = mode
+
+    Execution(newDoSomething)
+  }
+}
+```
+
+So, the function `doSomething` above takes in a `Config` and `Mode` and gives us a `T`. It is encapsulated inside of `Execution`, so we cannod access it directly, but only call the `run` method to invoke it.
+
+There's also a companion object `Execution` that enables us get the `Config` or `Mode`. A use of `Execution` might look like:
+
+```scala
+val config: Config = // ...
+val mode: Mode = // ...
+Execution.getConfig.run(config, mode)
+```
+
+Not very useful yet.
+
+### Map
+
+Just like scala's `Seq`, `Option`, `Future`, etc., `Execution` can be *mapped over* with its `map` method. In the context of `Seq`, calling the `map` method will transform every element into a new element via some function. In `Execution`, the `map` method gives us a new `Execution` that will produce the value of the old `Execution` and then transform that value with the function. Remember, `Execution` is a plan to do something; it won't be run until we call the `run` method:
+
+```scala
+case class Execution[T](private doSomething: (Config, Mode) => T) {
+  final def map[U](transform: T => U): Execution[U] = {
+    def newDoSomething(config: Config, mode: Mode): U = {
+      val result = doSomething(config, mode))
+      transform(result)
+    }
+
+    Execution(newDoSomething)
+  }
+
+  // ... other previously defined methods ...
+}
+```
+
+This is pretty powerful. `Execution` can now be used to access anything in the `Config` or `Mode`, or even do arbitrary work:
+
+```scala
+val config: Config = // ...
+val mode: Mode = // ...
+
+val config: ConfigFunProperty = Execution
+  .getConfig
+  .map { config => config.get("fun") }
+  .run(config, mode)
+
+val fivePlusThree = Execution
+  .getMode
+  .map { _ => 5 + 3 }
+```
+
+### From
+
+*Lifting* arbitrary code into `Execution` is used so frequent that we're going to add a helper method, `from`, for it:
+
+```
+object Execution {
+  final def from[T](value: => T): Execution[U] = {
+    def newDoSomething(config: Config, mode: Mode) = value
+    Execution(newDoSomething)
+  }
+
+  // ... other previously defined methods ...
+}
+```
+
+Why is it `value: => T` and not `value: T`? This scala syntax here `=>` is "call by name" and is essentially receiving a function of 0 parameters. Check out the scala documentation for more.
+
+At this point, we can write an `Execution` that reads from a service and uses its result to do something:
+
+```scala
+val config: Config = // ...
+val mode: Mode = // ...
+val service: String => String = // ...
+
+val callOutToService = Execution.from {
+  service("eat galaxies")
+}
+
+val reversedServiceResult = callOutToService.map { serviceResult =>
+  serviceResult.reverse
+}
+
+reversedServiceResult.run(config, mode)
+```
+
+### Running a Scalding job
+
+Writing a single Scalding job with `TypedPipe` and `Grouped` will not be covered in this tutorial; however, in order to run Scalding jobs with `Execution`, we need to be familiar with with 2 methods on `TypedPipe`: `writeExecution` and `toIterableExecution`:
+
+```scala
+val config: Config = // ...
+val mode: Mode = // ...
+
+val source: TypedSource[Int] = // ...
+val transform: Int => String = // ...
+val sink: TypedSink[String] = // ...
+
+val scaldingJob: Execution[Unit] = TypedPipe
+  .from(source)
+  .map(transform)
+  .writeExecution(sink)
+
+scaldingJob.run(config, mode)
+
+val scaldingJobResults: Execution[Iterable[String]] = TypedPipe
+  .from(source)
+  .map(transform)
+  .toIterableExecution
+
+scaldingJobResults
+  .map { results => results.foreach(println) }
+  .run(config, mode)
+```
+
+The `TypedPipe` method `writeExecution` parallels the normal `write` method on `TypedPipe`, but instead of planning the result inside of a Scalding `Job` class, `writeExecution` yields a plan to run that Scalding job as an `Execution`. The type parameter is `Unit` because the purpose of the `Execution` created from the `TypedPipe` is to have side-effects by writing to the `TypedSink` argument.
+
+The `TypedPipe` method `toIterableExecution` creates an `Execution` plan to expose the output of the `TypedPipe` in the local environment. This is useful if we need to use the output of a Scalding job to decide what to do next. Note that there's no guarantee on the order of the data and for large datasets, the length of the `Iterable` could be significant.
+
+### Flatmap
+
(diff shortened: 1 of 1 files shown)


tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+  .map { _ => 5 + 3 }
+```
+
+### From
+
+*Lifting* arbitrary code into `Execution` is used so frequent that we're going to add a helper method, `from`, for it:
+
+```
+object Execution {
+  final def from[T](value: => T): Execution[U] = {
```
johnynek: `Execution[T]` should be.

Can we use `tut` here: https://github.com/tpolecat/tut to catch these kinds of compilation errors?

johnynek:
This is really great and I'm happy we have better documentation here.

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
+# A Random Walk Down Executions or: You Could Have Invented Executions or: Learn You An Execution For Greater Good
```
isnotinvain, line 1: I like the title :)

It gets rendered a little strange though, maybe add some new lines?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+## Why would I want to use an Execution?
+
+The Scalding `Job` class uses a more object oriented programming approach to planning Scalding jobs, which often makes it more difficult to reason about and extend functionality. In addition, running multiple Scalding `Job` instances is poorly supported.
+
+Scalding `Execution`, on the other hand, is inspired by functional programming and composes easily. Here are a few things that are much easier with Scalding `Execution`:
+
+* Calling out to a service asynchronously to read or write some state before or after a Scalding job runs
+* Running multiple Scalding jobs or not running any Scalding jobs at all
+* Unit testing
+* Using the output of one Scalding job to decide how to plan future jobs
```
isnotinvain: maybe add loops here (line 16 implies loops, but not obvious to someone new)

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+These situations come up frequently when integrating Scalding jobs into an analytics workflow.
+
+We may want to avoid using Scalding `Execution` because it is more challenging to understand for newcomers to applied category theory (e.g. functors, applicative functors, and monads). This tutorial does not assume any knowledge of these concepts, but familiarity with them will make these explanations much easier to understand. In this respect, Scalding `Job` is simpler because it requires understanding less paradigms.
```
isnotinvain: maybe mention that Executions are like Futures for anyone familiar w/ futures

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Map
+
+Just like scala's `Seq`, `Option`, `Future`, etc., `Execution` can be *mapped over* with its `map` method. In the context of `Seq`, calling the `map` method will transform every element into a new element via some function. In `Execution`, the `map` method gives us a new `Execution` that will produce the value of the old `Execution` and then transform that value with the function. Remember, `Execution` is a plan to do something; it won't be run until we call the `run` method:
+
+```scala
+case class Execution[T](private doSomething: (Config, Mode) => T) {
+  final def map[U](transform: T => U): Execution[U] = {
+    def newDoSomething(config: Config, mode: Mode): U = {
+      val result = doSomething(config, mode))
```
isnotinvain: would `this.doSomething` make this clearer?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+  .run(config, mode)
+
+val fivePlusThree = Execution
+  .getMode
+  .map { _ => 5 + 3 }
+```
+
+### From
+
+*Lifting* arbitrary code into `Execution` is used so frequent that we're going to add a helper method, `from`, for it:
```
isnotinvain: frequently

isnotinvain: maybe explain what lifting means?

tutorial/Execution.md
isnotinvain: +1

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+val callOutToService = Execution.from {
+  service("eat galaxies")
+}
+
+val reversedServiceResult = callOutToService.map { serviceResult =>
+  serviceResult.reverse
+}
+
+reversedServiceResult.run(config, mode)
```
isnotinvain: here, and everywhere else you call .run() can you show what the returned result is? I think tut will do that for you

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Running a Scalding job
+
+Writing a single Scalding job with `TypedPipe` and `Grouped` will not be covered in this tutorial; however, in order to run Scalding jobs with `Execution`, we need to be familiar with with 2 methods on `TypedPipe`: `writeExecution` and `toIterableExecution`:
+
+```scala
+val config: Config = // ...
+val mode: Mode = // ...
+
+val source: TypedSource[Int] = // ...
```
isnotinvain: I think (here and elsewhere) examples are easier to follow when you don't use primitive types. Eg, make this:

`TypedPipe[User]` and `val getAddress = u: User => u.address` etc. (and use `Address` instead of String for address). No need to actually define the User and Address types tho.

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+val transform: Int => String = // ...
+val sink: TypedSink[String] = // ...
+
+val scaldingJob: Execution[Unit] = TypedPipe
+  .from(source)
+  .map(transform)
```
isnotinvain: maybe inline transform here?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+  .writeExecution(sink)
+
+scaldingJob.run(config, mode)
+
+val scaldingJobResults: Execution[Iterable[String]] = TypedPipe
```
isnotinvain: maybe a separator between these two examples explain the difference ?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+scaldingJobResults
+  .map { results => results.foreach(println) }
+  .run(config, mode)
+```
+
+The `TypedPipe` method `writeExecution` parallels the normal `write` method on `TypedPipe`, but instead of planning the result inside of a Scalding `Job` class, `writeExecution` yields a plan to run that Scalding job as an `Execution`. The type parameter is `Unit` because the purpose of the `Execution` created from the `TypedPipe` is to have side-effects by writing to the `TypedSink` argument.
+
+The `TypedPipe` method `toIterableExecution` creates an `Execution` plan to expose the output of the `TypedPipe` in the local environment. This is useful if we need to use the output of a Scalding job to decide what to do next. Note that there's no guarantee on the order of the data and for large datasets, the length of the `Iterable` could be significant.
+
```
isnotinvain: somewhere we should mention that "orphaned" or ignored executions never execute

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+### Flatmap
+
+So far, we've seen how to examine the `Config`, `Mode`, lift a value into the Execution environment, and run a Scalding job. We've alluded to the idea that we can use the output of one `Execution` to plan another, but haven't talked about how to do that. One may be familiar with the method `flatMap` on various collections in scala, just like `map`. `flatMap` acts differently depending on the type of collection:
+
+```scala
+val toThree        = Seq.range(0, 3)                    // Seq(0, 1, 2)
```
isnotinvain: Seq.flatMap is kind of confusing compared to Option.flatMap, which works more like Execution/Future .flatMap. Maybe stick with Option?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+val mode: Mode = // ...
+
+val scaldingJob: Execution[Unit] = // ...
+implicit val executor: ExecutionContext = // ...
+
+Future { /* ... asynchronous task ... */ }
+  .flatMap{ _ => scaldingJob.run(config, mode)(executor) }
+```
+
+But in general, we recommend that one extends `ExecutionApp` (below) and plans asynchronous actions from within that environment instead of integrating `Execution`s into existing asynchronous environments.
```
isnotinvain, line 281: I think practical notes about "you should never, ever, call .run or .waitFor. Let ExecutionApp do that for you. If you're calling .run or .waitFor, you're probably doing it wrong"

so many users reach for .waitFor when they need .flatMap because types-wise, it does what they need (unboxes the Execution)

isnotinvain (commented):
This is awesome, thanks for posting!

isnotinvain:
When you're done, you can throw it on here: https://wiki.haskell.org/Monad_tutorials_timeline haha

tutorial/Execution.md
oscar-stripe: not crazy about this at all.

This is all also true of `List`. Should we also be worried about using `List`? I think we should, as Alex suggests, instead describe `Execution` as a `Future` for big data jobs.

tutorial/Execution.md
piyushnarang: yeah would be nice to mention iterative ML style workflows often use execution.

tutorial/Execution.md
piyushnarang: yeah I think the Future analogy is really useful. Helps folks with a more imperative background conceptualize this as well.

piyushnarang: So on the one may want to avoid Executions point, should we say that if you don't have a usecase that warrants it (e.g. calling a service before / after or iterative jobs etc) you could stick to using a job? Or do we want to nudge users towards Executions?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+These situations come up frequently when integrating Scalding jobs into an analytics workflow.
+
+We may want to avoid using Scalding `Execution` because it is more challenging to understand for newcomers to applied category theory (e.g. functors, applicative functors, and monads). This tutorial does not assume any knowledge of these concepts, but familiarity with them will make these explanations much easier to understand. In this respect, Scalding `Job` is simpler because it requires understanding less paradigms.
+
+## Reasoning about how executions work
+
+At its heart, a Scalding `Execution` wraps a function that takes in a Scalding `Config` and a Scalding `Mode` and does *something*. A `Config` contains configuration properties for the running of a Scalding job (although we can access the `Config` properties when creating a Scalding `Execution`, so theoretically they can be used for anything). The `Mode` tells the Scalding job what environment (in-memory, in a local Hadoop cluster, in a remote Hadoop cluster) to run under. The Scalding `Mode` also contains an `Args` object, which represent Scalding-specific arguments passed on the command line.
```
piyushnarang: I'm not sure I understand the "although we can access the `Config` properties when creating a Scalding `Execution`, so theoretically they can be used for anything" bit.

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+For now, let's define a simplified `Execution` like so:
```
piyushnarang: I'm wondering if this portion (building a simple execution) should be in a separate section. It seems great to have if users want to understand executions in depth, but it doesn't seem essential to know from a usage standpoint. I'd imagine there'd be a bunch of users who'd come to this page to understand conceptually what an execution is and how to use it. They might want to skip this portion. What do you think?

tutorial/Execution.md
piyushnarang: yeah that's an important one. Lots of users have been tripped by that.

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+## The Real Execution
+
+Thus far, we have only discussed the simplified version of `Execution` to make it easier to understand. Here, we will talk about the real `Execution` in Scalding and how to make best use of it.
+
+### Executions are actually asynchronous
+
+Our `case class Execution` above does all work synchronously--only one thing at a time. In reality, `Execution` uses scala's `scala.concurrent.ExecutionContext` to schedule tasks asynchronously. So, if we want to turn non-Scalding asynchronous tasks into planned `Execution`s for use with Scalding, we can use the `Execution` companion object's method `def fromFuture[T](fn: ExecutionContext => Future[T]): Execution[T]` to schedule a scala `Future` on the same `ExecutionContext` that the `Execution` is using:
```
piyushnarang, line 260: maybe link to the scala doc of `fromFuture`?

tutorial/Execution.md
```diff
@@ -0,0 +1,412 @@
...
+
+### Handling failures
+
+Much like scala's `Try` and `Future` types, `Execution`s can either succeed with some result or fail with a `Throwable`. There are methods on `Execution` to handle failures and recover from them:
+
+```scala
+def withRetries(retries: Int)(execution: Execution[Unit]): Execution[Unit] = {
+  execution.recoverWith {
+    case t: Throwable if retries > 0 =>
+      logger.waring(t, s"Failed to run execution. Retrying $retries more time(s).")
```
piyushnarang: s/waring/warning

piyushnarang (commented):
Thanks a ton for taking the time to write this! Future generations of scalding users will thank you :-)

tutorial/Execution.md
ulyssepence: What do you mean by "strange"?

tutorial/Execution.md
ulyssepence: I meant that sometimes we use the `Config` properties outside of Scalding to make decisions when planning `Execution`s. I'll remove this statement.

tutorial/Execution.md
ulyssepence: I think building a mental model is extremely important and diving into the code or skipping to looking at uses of `Execution` won't necessarily do that for most users. Although I can be convinced otherwise.

tutorial/Execution.md
ulyssepence: Added.

tutorial/Execution.md
ulyssepence: OK sounds cool. I can make the undefined stuff just use `???`; however, a lot of the examples have `// ... other previously defined methods ...`. Should I expand this by copy-pasting these methods in?

tutorial/Execution.md
ulyssepence: Added.

tutorial/Execution.md
ulyssepence: Done.

tutorial/Execution.md
ulyssepence: I added a link to the `Execution` documentation at the top.

tutorial/Execution.md
ulyssepence: @johnynek, what are your feelings on this? We use `ExecutionApp` pretty much exclusively at Twitter, so I don't want to misrepresent best practices across the community. E.g. are there teams that make use of services like `A => Future[(Config, Mode)]` to schedule `Execution`s alongside other async code?

tutorial/Execution.md
ulyssepence: Done.

athuras:
godspeed

oscar-stripe:
👍

ulyssepence:
@oscar-stripe so I didn't add the `tut` stuff in this PR. I looked at the project and I couldn't figure out a way of showing just part of the code and keeping it compilable. Did you have any suggestions?

johnynek:
will, you can use `???` liberally, but sure. If you are showing code
sketches, we can punt.

tutorial/Execution.md
ulyssepence: I'm going to skip `tut` for now. I will try to make it as obvious as possible what the return types are.

tutorial/Execution.md
ulyssepence: Removed.

tutorial/Execution.md
ulyssepence: We could look into changing this in a followup PR. For now, I think I'm going to keep the primitives, but I like the idea.

tutorial/Execution.md
ulyssepence: Done.

tutorial/Execution.md
ulyssepence: Done.

PR #1666, opened 2017-04-14
Refactor the cascading code in the typed API into a self contained backend

johnynek:
As expected, this was actually easier than we might have suspected for a while.

This is not ready to ship. We will need to test it a fair amount. Also we need to probably add optimizations (the graph optimizer from summingbird should be great here).

But the code is much cleaner.

It is not totally factored, as I didn't move all the Grouped and CoGrouped code into the CascadingBackend, but that is easy, I just wanted to keep the diff size down a bit.

With this code, we could almost plan to any backend (when we are well factored) and just error if someone starts with a raw cascading pipe.  The backend would need a map of `TypedSource[T]` to the platform's source type. We will also need to update `Execution[_]` to deal with pluggable backends, but that should be in a separate PR in my view.

cc @piyushnarang @ianoc @isnotinvain @fwbrasil @dieu 

todo:
- [x] ~benchmark the approach we have to function composition vs the previous stack consuming approach (stacks should not be deep, so just do the absolute fastest)~ just avoid the fancy stack avoiding stuff and do it like we used to.
- [x] fix the fact that we are regressing on some line numbers currently (previously TypedPipeFactory put a line number on almost everything). We keep line numbers on grouping and writing, which is probably enough for many cases, but we should not regress (although the tests are clearly weak on this since we are only failing 1 and only on 2.11.8).


johnynek:
Also, this still blends some optimizations in the types (join composition is still handled by CoGrouped). Ideally we want to separate and do no optimizations as we build the AST, and then when we plan apply the general rules (those that apply to all graphs, e.g. `(a ++ b).map(f) == (a.map(f) ++ b.map(f))` vs those that only apply to Cascading (a GroupBy node can accept any number of input pipes).

johnynek:
This still does not pass all the tests.

The two issues are:

1. line numbers in many cases were lost since they were handled as a side effect of how TypedPipeFactory worked.
2. optimizations around zip of Execution seems a bit fragile, so we seem to have blocked some of them here. I think it has to do with some of the pipes looking different, and I am not sure the test is exactly a real failure in that I fear there may have been other fragile cases in the past. We should fix the issue and maybe make it more robust.

In a follow up PR, we should move more of the Execution related code into `CascadingBackend`.

I'm fairly optimistic this approach could allow us to:
a. both make the planning more clear and understandable.
b. enable new, whole-graph optimizations which we can't do well now.
c. allow us to run on other platforms (spark, flink, concurrent local) with minimal changes (error if the user converts directly from a cascading pipe but plans to another platform, so we can keep source compatibility with the existing API).

johnynek:
another optimization that occurs to me is that we could now pretty easily avoid lifting every map function to `A => TraversableOnce[B]`. Instead if we have a `A => B` we could avoid the allocation of `Iterator.single`. Since we do that on each row, that could actually be a measurable improvement for such jobs.

piyushnarang:
This is pretty cool, thanks for putting this up. Will take a look in the next day / two.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+          loop(cp, rest)
+      }
+      go(r)
+    }
+
+    import Dsl._ // Ensure we hook into all pipes coming out of the typed API to apply the FlowState's properties on their pipes
+    val pipe =
+      loop(p, FlatMappedFn.identity[U])
+        .applyFlowConfigProperties(flowDef)
+    RichPipe.setPipeDescriptionFrom(pipe, LineNumber.tryNonScaldingCaller)
```
fwbrasil: would it be possible to move the line numbers to the `TypedPipe` AST instances to keep the information precise?

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+import scala.collection.mutable.{ Map => MMap }
+import java.util.WeakHashMap
+
+object CascadingBackend {
+  import TypedPipe._
+
+  /**
+   * we want to cache renderings of some TypedPipe to Pipe so cascading
+   * will see them as the same. Without this, it is very easy to have
+   * a lot of recomputation. Ideally we would plan an entire graph
```
fwbrasil: I'm not sure I understand the flow here. Isn't `TypedPipe.toPipe` called only once given a job?

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+   * at once, and not need a static cache here, but currently we still
+   * plan one TypedPipe at a time.
+   */
+  private[this] val pipeCache = new WeakHashMap[TypedPipe[Any], Map[Mode, CascadingPipe[Any]]]()
```
fwbrasil: a global static cache for mutable values is pretty scary :) Is there a way that we could reduce its scope? For instance, if the flow def creation calls `TypedPipe.toPipe` multiple times but using the same thread, we could use a `ThreadLocal`-bound cache that can be cleaned after the planning.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+      val emptyFD = new FlowDef
+      val res = p(emptyFD)
+      pipeCache.put(t, mmc + (m -> res.asInstanceOf[CascadingPipe[Any]]))
+      res
+    }
+
+    pipeCache.synchronized {
+      pipeCache.get(t) match {
+        case null => add(Map.empty)
+        case somemap if somemap.contains(m) => somemap(m).asInstanceOf[CascadingPipe[T]]
```
fwbrasil: It seems that some of the AST nodes have values that use the identity hash code (e.g. `CascadingPipe.flowDef`), so even though nodes may represent the same computation they may not match the same hashmap entry. I'm not sure that's a problem, but I thought it's worth pointing out.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
fwbrasil: Wdyt about moving the caching logic to a separate class?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -119,6 +113,53 @@ object TypedPipe extends Serializable {
...
+
+   // TODO: CascadingPipe is still tightly bound to cascading
+   case class CascadingPipe[T](pipe: Pipe,
+    fields: Fields,
+    @transient localFlowDef: FlowDef, // not serializable.
+    @transient mode: Mode,
+    converter: TupleConverter[T]) extends TypedPipe[T]
+
+   case class CrossPipe[T, U](left: TypedPipe[T], right: TypedPipe[U]) extends TypedPipe[(T, U)] {
+     def viaHashJoin: TypedPipe[(T, U)] =
```
fwbrasil, line 125: This transformation seems much simpler than the previous implementations of `cross`. Is it a safe change? Would it be worth keeping the old behavior to reduce the scope of the change?

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+        }
+        go(cp)
+
+      case cp@CrossPipe(_, _) => loop(cp.viaHashJoin, rest)
+
+      case cv@CrossValue(_, _) => loop(cv.viaHashJoin, rest)
+
+      case DebugPipe(p) =>
+        // There is really little that can be done here but println
+        loop(p.map { t => println(t); t }, rest)
```
fwbrasil: How often is `.debug` used? Is it useful? Reading the code, it seems that the new implementation will output way less information.

fwbrasil (commented):
Initial review. I'm still wrapping my head around the code and changes, but it looks promising! I wonder if it'd be better to reduce the change set even further and avoid behavior changes.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: what values are mutable here?

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: No. It is called on each `.write` and some jobs write many outputs. It is probably also called somewhere inside Twitter by someone doing some mixing of the fields and typed API.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: I kind of want to keep it as local as possible and hopefully kill it eventually. If you prefer, I can make a class that encapsulates this WeakHashMap and lookup.

Note this previously existed in some form (and was needed to get tests to pass) in the previous version as well (inside TypedPipeFactory)

fwbrasil: aren't the cascading values mutable? (fields, flowDef)

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
fwbrasil: I see, thanks for the clarification.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: I believe that in fact we will never put a `CascadingPipe` directly into this cache. We could assert that to make it clear. It could come in as a grouping from someone that uses `from(pipe: Pipe, f: Field)...`. That said, in that case, I think we want identity hashCode and equals. I think we can add a discussion of this in the comment.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
fwbrasil: Fair enough, I agree that it's better to keep it limited to this scope 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: can you explain? The old implementation just called cascading Debug pipe on things that were already scheduled, which does this:
https://github.com/Cascading/cascading/blob/3.1/cascading-core/src/main/java/cascading/operation/Debug.java#L234

which, since we have it print every item, I think is just different in that it prints it with cascading fields, which are meaningless to the user of scalding's typed API (and actually confusing).

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: I think we should. I would like to add another AST node to annotate with line numbers rather than add it to each node. I wanted to keep this diff as small as possible though so people can read it.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: you mean here:
https://github.com/twitter/scalding/pull/1666/files/a76865d69948a2fef0953b8a061a95da9c070bf0#diff-fbe95168deecb5a4b5e9ace30182593eL981

I'm happy to add more optimization rules here, but I think all of those (other than the Merged change) are small potatoes. doing a cross is super rare anyway (because it multiplies the number of outputs) and the optimization for using iterables via closure capture is was not motivated by benchmarks.

The real reason I kept it simple is before we launch this we will want to actually run a lot of more principled and tested optimization rules rather than ad-hoc ones and the ones on cross seemed both ad-hoc (there was even a TODO to remove them) and also rarely exercised (since cross is so rare).

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: while they are technically mutable, we are not mutating any of them as we store them in here. This is kind of the whole game with this API. We are putting immutable fascades over mutable types, but we are being sure to make sure they are not mutated.

Note: all the items added to the cache we create in this file, and we never return them back to the user, so I think it is totally safe, but the type system it not helping us prove that.

fwbrasil: Cool! I thought they'd be mutated by the cascading calls. Maybe we could encapsulate the mutable values and expose only safe methods? I worry if someone mutates the values by mistake since they are accessible as `CascadingPipe` fields.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
fwbrasil: I'm probably missing the big picture, but it seems fragile to rely on the identity hash code of values to determine that two AST nodes represent the same computation. Maybe this discussion doesn't make sense, though. What's the impact if eventually a hash code mismatch happens? Is it just performance of the flow planning?

johnynek: if we have to semantically equal but not identical flowDefs it means we may wind up recomputing something (if cascading does not dedup at its level). We could implement a `FlowDef` wrapper that implements a hashCode and equality, but I don't think it is high priority actually.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
fwbrasil: I see. I saw https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/PipeDebug.scala and https://github.com/twitter/scalding/pull/559/files, and thought that we could be losing relevant debug information.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: it would be interesting to think about that, but it is not totally simple. Since we have to get them back out periodically to pass to cascading. We could copy at that point since the performance of planning is not really critical (it is very fast), the runtime is totally dominated by job running time and a little bit of IO at scheduling time.

I'd say this is something to add to the roadmap for the next version (if we adopt this PR), but not a blocker for this PR, if you ask me.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
fwbrasil: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
fwbrasil: sounds good, thanks for the clarification

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
piyushnarang: Actually I was thinking a inner class would be nice too. I don't mind keeping it in this class (rather than breaking to a separate file)

scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala
```diff
@@ -20,38 +20,73 @@ import java.io.Serializable
...
+    Filter[A, A](fn, implicitly)
+
+  def filterKeys[K, V](fn: K => Boolean): FlatMapping[(K, V), (K, V)] =
+    filter { kv => fn(kv._1) }
 
-/* This is the initial way we get a FlatMapFn */
-case class Converter[R](conv: TupleConverter[R]) extends FlatMapFn[R] {
-  // make sure not to start with an Iterator to keep everything lazy
-  def apply(te: TupleEntry) = Iterator(conv(te))
+  case class Identity[A, B](ev: A =:= B) extends FlatMapping[A, B]
```
piyushnarang, line 38: should we move these up? or you wanted to keep them within this object?

scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala
```diff
@@ -20,38 +20,73 @@ import java.io.Serializable
...
+  case class Filter[A, B](fn: A => Boolean, ev: A =:= B) extends FlatMapping[A, B]
+  case class Map[A, B](fn: A => B) extends FlatMapping[A, B]
+  case class FlatM[A, B](fn: A => TraversableOnce[B]) extends FlatMapping[A, B]
 }
 
-/* This is the mzero of this Monad */
-case object Empty extends FlatMapFn[Nothing] {
-  def apply(te: TupleEntry) = Iterator.empty
+/**
+ * This is a composition of one or more FlattMappings
```
piyushnarang: s/FlattMappings/FlatMappings ?

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+
+    def singlePipe[T](t: TypedPipe[T], force: Boolean = false): CascadingPipe[T] =
+      cacheGet(t, mode) { localFD =>
+        val pipe = t.toPipe(f0)(localFD, mode, singleSetter)
+        val p = if (force) RichPipe(pipe).forceToDisk else pipe
+        CascadingPipe[T](p, f0, localFD, mode, singleConverter)
+      }
+
+    def loop[T](t: TypedPipe[T], rest: FlatMappedFn[T, U]): Pipe = t match {
+      case cp@CascadingPipe(_, fields, localFlowDef, m, _) =>
```
piyushnarang: I'm wondering if we could split this up. This method is a super big case match which is a bit hard to read. Would it make sense to move the loop method to each of these specific subclasses? So we'd have overrides for each type of pipe?

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+import cascading.pipe.{ Each, Pipe }
+import cascading.tuple.{ Fields, TupleEntry }
+import com.twitter.scalding.TupleConverter.{ singleConverter, tuple2Converter }
+import com.twitter.scalding.TupleSetter.{ singleSetter, tup2Setter }
+import com.twitter.scalding._
+import com.twitter.scalding.serialization.{ CascadingBinaryComparator, OrderedSerialization, Boxed }
+import scala.collection.mutable.{ Map => MMap }
+import java.util.WeakHashMap
+
+object CascadingBackend {
```
piyushnarang: should we move this to a separate package (`cascading`?)

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
```diff
@@ -0,0 +1,268 @@
...
+        // There is really little that can be done here but println
+        loop(p.map { t => println(t); t }, rest)
+
+      case EmptyTypedPipe =>
+        // just use an empty iterable pipe.
+        // Note, rest is irrelevant
+        IterableSource(Iterable.empty, fieldNames)(setter, singleConverter[U]).read(flowDef, mode)
+
+      case fk@FilterKeys(_, _) =>
+        def go[K, V](node: FilterKeys[K, V]): Pipe = node match {
```
piyushnarang: maybe give this method a nicer name? doFilterKeys? / filterKeys?

piyushnarang: same for the others below?

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -45,20 +45,23 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
...
-    // Note, the Ordering must have that compare(x,y)== 0 being consistent with hashCode and .equals to
-    // otherwise, there may be funky issues with cascading
-    TypedPipeFactory({ (fd, mode) =>
-      val newPipe = new HashJoin(
-        RichPipe.assignName(mapside.toPipe(('key, 'value))(fd, mode, tup2Setter)),
-        Field.singleOrdered("key")(keyOrdering),
-        getForceToDiskPipeIfNecessary(fd, mode),
-        Field.singleOrdered("key1")(keyOrdering),
-        WrappedJoiner(new HashJoiner(joinFunction, joiner)))
+  //def hashCogroupOn[V1, R](mapside: TypedPipe[(K, V1)])(joiner: (K, V1, Iterable[V]) => Iterator[R]): TypedPipe[(K, R)] =
```
piyushnarang: delete?

scalding-core/src/main/scala/com/twitter/scalding/typed/HashJoinable.scala
```diff
@@ -109,11 +112,11 @@ trait HashJoinable[K, +V] extends CoGroupable[K, V] with KeyedPipe[K] {
     eachOperation match {
       case f: FlatMapFunction[_, _] =>
         f.getFunction match {
-          case _: Converter[_] => true
-          case _: FilteredFn[_] => false //we'd like to forceToDisk after a filter
-          case _: MapFn[_, _] => false
-          case _: FlatMappedFn[_, _] => false
-          case _ => false // treat empty fn as a Filter all so forceToDisk
+          case fmp: FlatMappedFn[_, _] if (FlatMappedFn.asId(fmp).isDefined) =>
```
piyushnarang, line 106: let's also update the comment at the top of this method. It refers to the old converter / emptyFn etc.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: I would really rather not do that since that puts the Pipe on the subclasses of TypedPipe which we are trying to avoid.

The case match is big, but each is a pretty clear block and we have a check that we have a total match.

scalding-core/src/main/scala/com/twitter/scalding/typed/CascadingBackend.scala
johnynek: I don't want to do that because I think it actually obscures things. If scala were smarter with type inference this would not even be needed (note we just define a method to call it). I think keeping a standard pattern here actually serves to tell the reader this is doing the logic for the match, except we can name the types.

scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala
johnynek: I like keeping subclasses of a sealed trait in the companion object to avoid polluting the outer namespace with things that should be implementation details.

scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala
johnynek: 👍

johnynek:
okay, other than benchmarking the new function composition, I think this addresses most comments.

I think after the benchmark, this should be mergeable. Can you look again @fwbrasil and @piyushnarang ?

johnynek:
turns out of benchmarks don't compile, which is a bummer. Rather than shave that yak right now, I'm just making the function composition use `andThen` to compose except be more efficient than before (avoid allocating a Iterator except at the very end for map sequences, be smarter about filter as well).

I think this is mergeable. What do you folks say?

scalding-core/src/main/scala/com/twitter/scalding/typed/FlatMappedFn.scala
piyushnarang: ok, sounds good.

fwbrasil:
LGTM! The build is failing, though.

piyushnarang:
👍 This looks good to me as well.

PR #1670, opened 2017-04-22
Refactor the cascading planning code out of Grouped

johnynek:
This is a follow on to #1666 to bring almost all of the cascading planning out of Grouped related to reduce steps. There is still a small bit which overlaps with CoGrouped, which will be moved after we tackle CoGrouped.

r? @fwbrasil @piyushnarang @ianoc

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
@@ -193,7 +193,7 @@ sealed trait ReduceStep[K, V1, V2] extends KeyedPipe[K] {
    */
   def mapped: TypedPipe[(K, V1)]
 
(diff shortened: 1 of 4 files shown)


ianoc:
all seems pretty lgtm, like all the sealed traits. Code looks nice and readable too 👍 

(Overall these plots actually I think are making the planning code overall more readable so thats pretty awesome too)

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -113,6 +106,7 @@ object Grouped {
     (boxfn, boxordSer)
   }
 
+  // TODO move this to CascadingBackend when we refactor joins
```
piyushnarang, line 109: maybe cite the github issue so that we dont lose track of it?

piyushnarang (approved):

PR #1673, opened 2017-04-24
Be more paranoid about Kryo registration order

johnynek:
We had an error that seemed like it *might* be related to kryo tokens being registered incorrectly with Execution. This is a stab in the dark, but it can't hurt.

Related to #1654

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
@@ -92,9 +92,14 @@ class KryoHadoop(@transient config: Config) extends KryoInstantiator {
     /**
      * Register any cascading tokenized classes not already registered
      */
-    val tokenizedClasses = CascadingTokenUpdater.parseTokens(config.get(ScaldingConfig.CascadingSerializationTokens)).values
+    val tokenizedClasses =
+      CascadingTokenUpdater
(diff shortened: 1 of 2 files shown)


ianoc-stripe:
lgtm

scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
```diff
@@ -92,9 +92,14 @@ class KryoHadoop(@transient config: Config) extends KryoInstantiator {
     /**
      * Register any cascading tokenized classes not already registered
      */
-    val tokenizedClasses = CascadingTokenUpdater.parseTokens(config.get(ScaldingConfig.CascadingSerializationTokens)).values
+    val tokenizedClasses =
+      CascadingTokenUpdater
+        .parseTokens(config.get(ScaldingConfig.CascadingSerializationTokens))
+        .toList
+        .sorted // Go through this list in order the tokens were allocated
```
ulyssepence, line 99: Can you add a reference to an issue or Kryo API or something?

ulyssepence (approved):

johnynek:
What do you mean? Maps are not ordered. I'm just trying to make them
ordered. Otherwise no change. Why should this code add extra Kryo API links
when we use it in many places related to serialization?

piyushnarang (approved):

PR #1678, opened 2017-05-01
Register all Boxed classes in Kryo

### LINKED ISSUE #1524: should register all boxed classes
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala#L57

if we don't register, they are not assigned tokens, which slows serialization down.

jcdavis:
They should never be kryo serialized (hence the ThrowingSerializer),
but this means they will be given cascading tokens.

Fixes #1524


johnynek (commented):
:+1:

PR #1679, opened 2017-05-02
Remove cascading dependency from Grouped/CoGrouped, seal the traits

johnynek:
With this change, scalding grouped/cogrouped no longer has any dependency on cascading. The cascading backend owns all the code to convert to joins/reduces and grouped/cogrouped knows nothing about cascading fields, tuples, etc...

The next step is to make Execution know nothing about cascading and be able to punt, via the Mode, into the correct backend.


johnynek:
cc @fwbrasil @piyushnarang @pankajroark

johnynek:
note I had to move CoGrouped/CoGroupable into Grouped in order to make them sealed classes, which is something we want (so when writing backends we can write matches that are total).

ulyssepence:
Looks good to me

PR #1681, opened 2017-05-03
Pull cascading flowDef thread into cascading_backend

johnynek:
This is a first step towards factoring Execution so that it does not know about the backend. This is a good cleanup regardless, because we separate the concern of caching from producing Futures from FlowDefs.

But I want to drive towards the `Mode` providing something to take a sequence of `TypedPipe[T], TypedSink[T]` pairs and returning a future `Unit` (and maybe counters) when they have all been written.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -15,22 +15,16 @@ limitations under the License.
  */
 package com.twitter.scalding
 
-import com.twitter.algebird.monad.{ Reader, Trampoline }
+import cascading.flow.{ FlowDef, Flow }
+import com.twitter.algebird.monad.Trampoline
 import com.twitter.algebird.{ Monoid, Monad, Semigroup }
 import com.twitter.scalding.cascading_interop.FlowListenerPromise
-import com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner
 import com.twitter.scalding.filecache.{CachedFile, DistributedCacheFile}
-import com.twitter.scalding.Dsl.flowDefToRichFlowDef
-import scala.concurrent.{ Await, Future, ExecutionContext => ConcurrentExecutionContext, Promise }
-import scala.util.{ Failure, Success, Try }
-import scala.util.control.NonFatal
-import cascading.flow.{ FlowDef, Flow }
-import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.{ FileSystem, Path }
-import org.slf4j.LoggerFactory
-
+import com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner
 import scala.collection.mutable
+import scala.concurrent.{ Await, Future, ExecutionContext => ConcurrentExecutionContext, Promise }
 import scala.runtime.ScalaRunTime
+import scala.util.{ Failure, Success, Try }
 
 /**
  * Execution[T] represents and computation that can be run and
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -551,20 +551,23 @@ object Execution {
    * but with proof that pipe matches sink
    */
 
-  private trait ToWrite {
-    def write(config: Config, flowDef: FlowDef, mode: Mode): Unit
-  }
-  private case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite {
-    def write(config: Config, flowDef: FlowDef, mode: Mode): Unit = {
-      // This has the side effect of mutating flowDef
-      pipe.write(sink)(flowDef, mode)
-      ()
-    }
+  sealed trait ToWrite
+  object ToWrite {
+    case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite
+    case class PreparedWrite[T](fn: (Config, Mode) => SimpleWrite[T]) extends ToWrite
   }
 
-  private case class PreparedWrite[T](fn: (Config, Mode) => SimpleWrite[T]) extends ToWrite {
-    def write(config: Config, flowDef: FlowDef, mode: Mode): Unit =
-      fn(config, mode).write(config, flowDef, mode)
+  /**
+   * Something that can handle a batch of writes that may be optimized
+   * before running. Return a unique Long for each run and Counters
+   */
+  trait Writer {
+    def execute(
+      conf: Config,
+      mode: Mode,
+      head: ToWrite,
+      rest: List[ToWrite])(implicit cec: ConcurrentExecutionContext): Future[Map[Long, ExecutionCounters]]
+
   }
 
   /**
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -0,0 +1,201 @@
...
+  }
+  /*
+   * This is called after we are done submitting all jobs
+   */
+  def finished(mode: Mode): Unit = {
+    messageQueue.put(Stop)
+    // get an immutable copy
+    val cleanUp = filesToCleanup.synchronized { filesToCleanup.toList }
+    if (cleanUp.nonEmpty) {
+      Runtime.getRuntime.addShutdownHook(TempFileCleanup(cleanUp, mode))
```
ianoc, line 150: Since i'm reading this, any thoughts on how we should handle this instead. Not for doing in this PR obviously. I'm guessing we don't run the cleanup right now because we might have done a toIterableExecution and are using it post the run?

ianoc:
👍  lgtm

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: It's a good question. I'm currently thinking of making some methods like:
```scala
object Execution {
  def forceToDisk[T](t: TypedPipe[T]): Execution[TypedPipe[T]] = ...
  def toIterable[T](t: TypedPipe[T]): Execution[Iterable[T]] = ...
}
```
and just have those two be new subclasses of the Execution trait. Then each Writer can have a functions like:
```scala

trait Writer {
  def forceToDisk[T](c: Config, m: Mode, t: TypedPipe[T]): Future[TypedPipe[T]]
  def toIterable[T](c: Config, m: Mode, t: TypedPipe[T]): Future[Iterable[T]]
}
```
which execution will just call through to. What do you think of that?

When we get to that stage, basically Execution really is just a memo-izing State/Reader monad. In cats-like description `Execution[T] = StateT[Future, (Config, Mode, ExecutionContext), T]` with some memoization.

johnynek: I wasn't that clear: if we do the above, the logic about cleanup files can happen when you do forceToDisk, which Writer is driving, so it can keep the list of clean-up files it needs.

ianoc: Yeah i like this ^^ makes it more pluggable, and ultimately separating the 2 notions inside the execution feels like a good place to handle it. The question of maybe when to do the delete is still rough but not sure what to do there. I don't know if we can hook into some sort of hadoop fs shutdown thing. But separating this logic out from typed pipe to the platforms is 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -675,20 +545,23 @@ object Execution {
...
-  private case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite {
-    def write(config: Config, flowDef: FlowDef, mode: Mode): Unit = {
-      // This has the side effect of mutating flowDef
-      pipe.write(sink)(flowDef, mode)
-      ()
-    }
+  sealed trait ToWrite
+  object ToWrite {
+    case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite
+    case class PreparedWrite[T](fn: (Config, Mode) => SimpleWrite[T]) extends ToWrite
```
ulyssepence, line 551: Do we need to change the visibility of these?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -0,0 +1,201 @@
...
+          // If we fail in deleting a temp file, log the error but don't fail the run
+          case e: Throwable => LOG.warn(s"Unable to delete temp file $file", e)
+        }
+      }
+    }
+  }
+}
+
+/**
+ * This holds an internal thread to submit run
```
ulyssepence, line 71: "submit run"

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -0,0 +1,201 @@
...
+      Runtime.getRuntime.addShutdownHook(TempFileCleanup(cleanUp, mode))
+    }
+  }
+
+  def addFilesToCleanup(files: TraversableOnce[String]): Unit =
+    filesToCleanup.synchronized {
+      filesToCleanup ++= files
+    }
+
+  private def toFuture[T](t: Try[T]): Future[T] =
```
ulyssepence, line 159: This is already defined elsewhere. Can we just put it in some central place or use bijection's or inline it?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -0,0 +1,201 @@
...
+    thread.setDaemon(true)
+    thread.start()
+  }
+  /*
+   * This is called after we are done submitting all jobs
+   */
+  def finished(mode: Mode): Unit = {
+    messageQueue.put(Stop)
+    // get an immutable copy
+    val cleanUp = filesToCleanup.synchronized { filesToCleanup.toList }
```
ulyssepence, line 148: Why the addition?

ulyssepence (commented):
Looks good. A few minor comments.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: Yes, the idea is that we want to be able to write new backends. To do that, you will need to be able to see these types (so, in this case AsyncFlowDefRunner needs to be able to pattern match on them).

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: The previous stuff was not thread safe.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: It's so simple, I hear you, but I don't want to make a public API which we can't delete, not use add a bunch of syntax to call bijection here. Can we let this minor duplication slide?

johnynek:
merging now, and can continue fixes in the next in the series.

PR #1682, opened 2017-05-06
Move toIterableExecution and forceToDiskExecution into Execution

johnynek:
This is a follow up to #1681 

Here we have done almost all of the work to make Execution delegate to a `Writer` that handles all the side effects.

With this we can basically do 100% of having an in-memory or spark backend as long as users don't use certain cascading only features (toPipe, TypedPipe from a Pipe, or Execution from a FlowDef).

We are getting very close to where I want to be for the next published version. I want to make an example in-memory backend just to exercise everything to see if we really can run non-trivial jobs. We can leave a spark or flink backend to a future work.

I would like to remove the Execution from a FlowDef, I doubt it was ever used. I'd like to move toPipe to an implicit enrichment that users opt into if they really need it (which I doubt they would).


johnynek:
@ianoc any time for this one?

johnynek:
ping @benpence

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -531,10 +531,16 @@ object Execution {
...
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
-      lazy val future = cache
-        .runner
-        .validateAndRun(conf, mode)(result)
-        .map { m => ((), m) }
+      lazy val future = {
+        cache.writer match {
+          case ar: AsyncFlowDefRunner =>
+            ar.validateAndRun(conf, mode)(result).map { m => ((), Map(m)) }
+          case other =>
```
ulyssepence, line 538: Why not also make this part of the interface like in `#start()` instead of matching? Do we anticipate handling writers differently?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -622,11 +668,12 @@ object Execution {
                   Future.successful(Map.empty)
               }
 
-            failFastZip(otherResult, localFlowDefCountersFuture).map {
-              case (lCounters, fdCounters) =>
-                val summedCounters = (fdCounters :: lCounters).reduce(_ ++ _)
-                (fn(conf, mode), summedCounters)
-            }
+            val bothFutures = failFastZip(otherResult, localFlowDefCountersFuture)
```
ulyssepence, line 671: So this `localFlowDefCountersFuture` is a `Future` for the stats from finishing the cascading flow?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -510,54 +509,8 @@ sealed trait TypedPipe[+T] extends Serializable {
...
-          Set[String]()
-        case _: HadoopMode =>
-          Set(temporaryPath(conf, cachedRandomUUID))
-      }
-    }
-
-    Execution.write(writeFn, readFn, filesToDeleteFn)
-  }
+  def forceToDiskExecution: Execution[TypedPipe[T]] =
+    Execution.forceToDisk(this)
```
ulyssepence, line 513: This is much easier to read :)

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -68,14 +73,51 @@ object AsyncFlowDefRunner {
 }
 
 /**
+ * This holds an internal thread to run
  * This holds an internal thread to submit run
```
ulyssepence, line 77: Duplicate comment

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -68,14 +73,51 @@ object AsyncFlowDefRunner {
...
+  private[this] var state: State = State(Map.empty, Map.empty)
+
+  private def updateState[S](fn: State => (State, S)): S =
+    mutex.synchronized {
+      val (st1, s) = fn(state)
+      state = st1
+      s
+    }
+  private def getState: State =
+    updateState { s => (s, s) }
```
ulyssepence, line 119: I find this a little hard to read because of the use of `S` and `s` and I anticipate people unfamiliar with this construct will find it confusing. Can we use `R` and `result` instead or some more verbose in the naming?

ulyssepence (commented):
I think I understand what's going on in each localized scope, but am less familiar with the whole picture here.

Overall, it looks fine. Left a few questions/comments.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ianoc: I'm not sure we can easily do this differently with the current interface -- in that FlowDefExecution is tied to cascading. this match would eventually go away with a more cross platform interface I would think. Since we would want to eventually get references to FlowDef out of this file

johnynek: what Ian said. The goal here is to make hopefully all of the typed API not depend on cascading and just make the cascading backend one implementation. I think no one uses this function anyway (creating an Execution from a FlowDef) and I want to just remove it, but in the mean time, I want to not dig us in any deeper.

If no one at Twitter is using it, I would suggest we remove the function, but keeping it around for 1 version deprecated may work okay too.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: :)

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: that's a merge issue I think... :/

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: can you be concrete about how you want to change this? These is only one variable here, the state and you are copying that to be the result. Can you write the code you want me to replace it with because I don't follow your comment.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: First note, I didn't change that, I just captured the future into a name so I could use a for comprehension and have it read nicer, in my opinion.

About how this works, I hope the code is somewhat readable, but it is not simple. This is where a lot of the hard work of Execution is happening and it got more complicated over time due to finding and fixing issues with its use at Twitter. The thing we are working very hard to do here is not ever re-run a ToWrite aside from also not re-running ANY execution. The problem is, if things are zipped differently, the same ToWrite might end up in two different Executions. So we look through for all the ToWrites we didn't yet run, we run those in one call to the Writer, and we merge the counter results from the other previous writes (otherwise the counter results are wrong).

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -68,14 +73,51 @@ object AsyncFlowDefRunner {
...
+ * This holds an internal thread to run
  * This holds an internal thread to submit run
  * a Config, Mode, FlowDef and return a Future holding the
  * JobStats
  */
 class AsyncFlowDefRunner extends Writer { self =>
   import AsyncFlowDefRunner._
 
-  private[this] val filesToCleanup = mutable.Set[String]()
+  private[this] val mutex = new AnyRef
```
ianoc, line 84: do we need the mutex rather than self?

johnynek: personal OCD. I don't like to lock things others can see. I always like to lock a private val.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
ulyssepence: Sure something along the lines of

```scala
private def updateState[R](modify: State => (State, R)): R =
  mutex.synchronized {
    val (newState, result) = modify(state)
    state = newState
    result 
  }
private def getState: State =
  // Expose current state as the result value
  updateState { state => (state, state) }
```

although as is it's probably OK.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ulyssepence: Ya I was just a little confused by what was going on here in the old code. I think it makes sense now. I think the concept of "we" and "someoneElse" was not clear to me before.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
ulyssepence: We do have a few `#fromFn` uses in our code, specifically in a couple places where we have `FlowDef => Unit` functions to accommodate both `Job` & `Execution`. Deprecation is definitely a good idea though.

ulyssepence (approved):

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
benpence: Something along these lines:

```scala

def planScaldingJob(flowDef: FlowDef): Unit = ???

Execution.fromFn {
  val flowDef = new FlowDef
  planScalding(flowDef)
  flowDef
}

class MyJob extends Job {
  override def buildFlow = {
    planScaldingJob(flowDef)
    super.buildFlow
  }
}
```

ianoc:
lgtm

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: I'll sweep this change into the next one if that's okay.

ulyssepence: Sure.

PR #1683, opened 2017-05-07
Remove CascadingPipe node from TypedPipe ADT

johnynek:
It occurred to me that we could have used a TypedSource already to implement the `TypedPipe.from[T](p: Pipe, f: Fields)(implicit fd: FlowDef, m: Mode, c: TupleConverter[T]): TypedPipe[T]` method.

I'm sorry I didn't notice this earlier. With this we can remove the `CascadingPipe` subclass of TypedPipe, which was a TODO in the code.

The type is still useful so I moved it into CascadingBackend.

Along the way, I did some more unification of the code so there is less redundancy.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -44,7 +44,20 @@ object TypedPipe extends Serializable {
    */
   def from[T](pipe: Pipe, fields: Fields)(implicit flowDef: FlowDef, mode: Mode, conv: TupleConverter[T]): TypedPipe[T] = {
     val localFlow = flowDef.onlyUpstreamFrom(pipe)
-    CascadingPipe[T](pipe, fields, localFlow, mode, conv).withLine
+    val pipeSrc = new TypedSource[T] {
+      def converter[U >: T]: TupleConverter[U] =
+        TupleConverter.asSuperConverter[T, U](conv)
+      def read(implicit fd: FlowDef, m: Mode): Pipe = {
+        // This check is not likely to fail unless someone does something really strange.
+        // for historical reasons, it is not checked by the typed system
+        require(m == mode,
+          s"Cannot switch Mode between TypedPipe.from and toPipe calls. Pipe: $pipe, pipe mode: $m, outer mode: $mode")
+        fd.mergeFrom(localFlow)
+        pipe
+      }
+      override def sourceFields: Fields = fields
+    }
+    from(pipeSrc)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


johnynek:
cc @fwbrasil @benpence @piyushnarang

ianoc:
FYI this looks to be failing CI

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -44,7 +44,20 @@ object TypedPipe extends Serializable {
    */
   def from[T](pipe: Pipe, fields: Fields)(implicit flowDef: FlowDef, mode: Mode, conv: TupleConverter[T]): TypedPipe[T] = {
```
ianoc, line 44: Do you think we can pull this method out into an implicit on the typed pipe type in a cascading package/some other package possibly and delurk them from here entirely? its not binary compatible, but if we split scalding-core into some sort of scalding-typed and scalding-core where scalding-core has the cascading dep, we might be able to make it source compatible I think?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -149,21 +154,18 @@ object CascadingBackend {
...
-        require(m == mode,
-          s"Cannot switch Mode between TypedSource.read and toPipe calls. Pipe: $p, pipe mode: $m, outer mode: $mode")
-        flowDef.mergeFrom(localFlowDef)
-
-        def go[T1 <: T](cp: CascadingPipe[T1]): Pipe = {
-          val withRest = RichPipe(cp.pipe).flatMapTo[T1, U](fields -> fieldNames)(rest)(cp.converter, setter)
-          applyDescriptions(withRest, descriptions)
-        }
-        go(cp)
+    def finish[T](cp: CascadingPipe[T],
```
ianoc: might be worth a quick comment on this and loop below just to say what they do.

This i think means we merge our new flow def from the cascading pipe with an historical one, flatMap with the function we kept inside the pipe, finally applying the descriptions on the cascading pipe.

ianoc:
Change looks great to me, looks very much on track for the separation!

johnynek:
@ianoc thanks for the review.

Strangely, this caught an issue that a `TypedPipe` was in a closure just to get a function out. We actually don't serialize TypedPipe (which we probably could now, but really shouldn't need to). I fixed another issue that looked similar. They seemed to be the only two cases of a TypedPipe being accessed inside a closure.

I consolidated code here, but I'm not sure why the optimization paths were different. In any case, I'm glad we caught it.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: I do think so, the only question is when to do that. I think we can basically pull all the cascading specific stuff into an enrichment now and stay source compatible. I think your plan is exactly what I want to drive towards: scalding-typed is a dependency of scalding-core and does not depend on cascading at all.

fwbrasil: :+1: to postponing the decision on how we're going deal with the cascading dependency. I think it's a separate discussion.

fwbrasil:
👍  LGTM

PR #1685, opened 2017-05-12
Fix serialization of `KryoHadoop`

ttim:
In process of Summingbird migration I found that serialization of `KryoHadoop` class was broken recently, this PR adds test for that and fixes it.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/KryoHadoop.scala
@@ -22,10 +22,17 @@ import com.twitter.chill.algebird._
 import com.twitter.chill.config.Config
 import com.twitter.chill.{ IKryoRegistrar, KryoInstantiator, ScalaKryoInstantiator, SingletonSerializer }
 
-class KryoHadoop(@transient config: Config) extends KryoInstantiator {
+class KryoHadoop(
   // keeping track of references is costly for memory, and often triggers OOM on Hadoop
-  val useRefs = config.getBoolean("scalding.kryo.setreferences", false)
-  val cascadingSerializationTokens = config.get(ScaldingConfig.CascadingSerializationTokens)
+  useRefs: Boolean,
+  cascadingSerializationTokens: String
+) extends KryoInstantiator {
+
+  def this(config: Config) =
+    this(
+      config.getBoolean("scalding.kryo.setreferences", false),
+      config.get(ScaldingConfig.CascadingSerializationTokens)
+    )
 
   /**
    * TODO!!!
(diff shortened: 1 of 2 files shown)


dieu (approved):

johnynek:
Lgtm. We can also add a serializer for Config by just converting to and from a Map[String, String] I think.

ttim:
@johnynek Do you mean to keep `Config` serializable instead and therefore there will be no need to have parts of it as `val`s?

I guess it was more or less this way before, but after @ianoc changed behavior to what we see now https://github.com/twitter/scalding/commit/2c72bac6717cffb1db005bcb370981844d77cff7.

johnynek:
I think we should make your change AND make Config Kryo serializable. Belt and suspenders.

ttim:
Now I see. It's not simple (at least in this PR) for now because `Config` doesn't expose a way to iterate over keys, so it's impossible to convert it into `Map[String, String]`.

dieu:
@ttim look on trait `Config` in scalding, it has method`toMap`

ttim:
@dieu small clarification - what we pass there isn't scalding config but chill's one, which doesn't contain this method.

johnynek:
ahh yes. We should open an issue on Chill to make all of the Config Serializable and Kryo serializable.

ttim:
@johnynek I'm wondering how was it working before @ianoc change (https://github.com/twitter/scalding/commit/2c72bac6717cffb1db005bcb370981844d77cff7)? For me it looks like `Config` *is* serializable already (or maybe `HadoopConfig`, I don't know)?

dieu:
@ttim everything is serializable in `chill.Config`, but only by kryo I think, because `chill.Config` is not marked as serializable.

ttim:
@johnynek So I guess we don't need to create an issue in chill because `Config` is already kryo serializable.

Also I did a small refactoring to make logic clearer (and avoid transient fields), what do you think?

johnynek (commented):
I'm not crazy about this at all. Multiple constructors is non-idiomatic in scala and many folks subclass this as it is. I'd rather do the first fix personally.

johnynek (commented):
To put it another way: this type, and hose like them, are created with reflection and expected to have a constructor that accepts a Config. This is part of chill-hadoop.

Also, if we add another option we would be adding another parameter to the constructor vs just reading more from the Config.

ttim:
@johnynek my main motivation was to make object's state more clear (and I guess `transient` plays against it) and do separation between internal state and how do we create `KryoHadoop` clearer. 
At the same time I don't like two constructors either and missed a point of creation by reflection, so I'm happy to revert this back.

pankajroark:
@johnynek Do you think this PR is ready to go? We need this fix for upgrading summingbird to 2.12 https://github.com/twitter/summingbird/pull/721. Thanks

johnynek:
👍

PR #1687, opened 2017-05-22
Fix stack overflow in typedPipeMonoid.zero

pankajroark:
typedPipeMonoid.zero calls Monoid.empty which calls typedPipeMonoid.zero leading to stack overfflow, fix that.

@ttim We'll need this fix in the release for scala 2.12.


scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala
```diff
@@ -121,6 +122,15 @@ class TypedSumByKeyTest extends WordSpec with Matchers {
   }
 }
 
+class TypedPipeMonoidTest extends WordSpec with Matchers {
+  "typedPipeMonoid.zero" should {
+    "be equal to typedPipeMonoid.empty" in {
+      val mon = implicitly[Monoid[TypedPipe[Int]]]
+      assert(mon.zero == mon.empty)
```
ttim: `TypedPipe.empty` makes more sense to me.

ttim (changes requested):
Can you make a PR against 0.17.x branch as well after this one will be merged?

I guess `.DS_Store` should be excluded from this PR.

ttim (approved):

johnynek:
this is due to `Monoid` now extending `cats.kernel.Monoid`

PR #1689, opened 2017-05-22
Update the sbt runner script

### LINKED ISSUE #1680: Sbt launch script is out of date
It doesn't have 2.12 support, and uses the wrong sbt version.

Can I just copy/paste the latest https://github.com/paulp/sbt-extras/blob/master/sbt ?

I might also be able to modify the existing one to support 2.12 if that is easier

jcdavis:
This uses the updated sbt version (0.13.17), and adds support for 2.12, removing support for 2.10.
This is copy/paste from the master version, minus changing scala versions to only support 2.11.8 and 2.12.1

This is a fix for #1680

### DIFF
--- a/sbt
+++ b/sbt
@@ -3,20 +3,89 @@
 # A more capable sbt runner, coincidentally also called sbt.
 # Author: Paul Phillips <paulp@improving.org>
 
-# todo - make this dynamic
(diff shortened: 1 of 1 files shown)


johnynek:
Thanks!

PR #1693, opened 2017-05-26
Lookup classses using the mirror instead of by name

jcdavis:
This fixes finding of inner classes and should be more robust overall.

@dieu : I know you pointed this case out earlier, any other situations I might want to look into?


johnynek:
👍

PR #1697, opened 2017-06-03
First draft of pure-scalding memory backend

johnynek:
This follows up the thread of work leading to #1682 

This gives an in-memory backend without using cascading (which for the basic tests is MUCH faster).

This is not a production quality backend yet:
1. no support for joins (hashJoin yes cogroup no).
2. parallelism has not been carefully tuned, so we only get very naive parallelism at the moment.

The main point of this is to exercise using the execution API without cascading in the loop. I think this proof of concept shows that a spark backend would not be very hard at this point and the memory backend should be a guide for someone looking to do that.

I think we should merge this despite it not being complete because the PR is already dense enough. I'd like to improve the quality of the test coverage and support all the cases in later PRs.

r? @fwbrasil @piyushnarang

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
@@ -244,9 +244,12 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
         case cp@CrossPipe(_, _) =>
           plan(m, cp.viaHashJoin)
 
-        case cv@CrossValue(_, _) =>
-          plan(m, cv.viaHashJoin)
-
+        case CrossValue(left, EmptyValue) => (m, Op.empty)
+        case CrossValue(left, LiteralValue(v)) =>
+          val (m1, op) = plan(m, left)
+          (m1, op.flatMap { a => Iterator.single((a, v)) })
+        case CrossValue(left, ComputedValue(right)) =>
+          plan(m, CrossPipe(left, right))
         case DebugPipe(p) =>
           // There is really little that can be done here but println
           plan(m, p.map { t => println(t); t })
@@ -393,7 +396,27 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
       }
     }
 
-  def planHashJoinable[K, V](m: Memo, hk: HashJoinable[K, V]): (Memo, Op[(K, V)]) = ???
+  def planHashJoinable[K, V](m: Memo, hk: HashJoinable[K, V]): (Memo, Op[(K, V)]) = hk match {
+    case IdentityReduce(_, pipe, _, _) => plan(m, pipe)
+    case UnsortedIdentityReduce(_, pipe, _, _) => plan(m, pipe)
+    case imr@IteratorMappedReduce(_, _, _, _, _) =>
+      def go[K, U, V](imr: IteratorMappedReduce[K, U, V]) = {
+        val IteratorMappedReduce(_, pipe, fn, _, _) = imr
+        val (m1, op) = plan(m, pipe)
+        (m1, op.mapAll[(K, U), (K, V)] { kvs =>
+          val m = kvs.groupBy(_._1).iterator
+          val res = ArrayBuffer[(K, V)]()
+          m.foreach { case (k, kus) =>
+            val us = kus.iterator.map { case (k, u) => u }
+            fn(k, us).foreach { v =>
+              res += ((k, v))
+            }
+          }
+          res
+        })
+      }
+      go(imr)
+  }
 
   case class State(
     id: Long,
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
@@ -115,13 +115,12 @@ object MemoryPlanner {
     }
 
     case class Concat[O](left: Op[O], right: Op[O]) extends Op[O] {
-      def result(implicit cec: ConcurrentExecutionContext) =
-        for {
-          l <- left.result
-          r <- right.result
-        } yield {
-          ArrayBuffer.concat(l, r)
-        }
+      def result(implicit cec: ConcurrentExecutionContext) = {
+        // start both futures in parallel
+        val f1 = left.result
+        val f2 = right.result
+        f1.zip(f2).map { case (l, r) => ArrayBuffer.concat(l, r) }
+      }
     }
 
     case class Map[I, O](input: Op[I], fn: I => TraversableOnce[O]) extends Op[O] {
@@ -178,11 +177,12 @@ object MemoryPlanner {
       opB: Op[B],
       fn: (IndexedSeq[A], IndexedSeq[B]) => ArrayBuffer[C]) extends Op[C] {
 
-      def result(implicit cec: ConcurrentExecutionContext) =
-        for {
-          as <- opA.result
-          bs <- opB.result
-        } yield fn(as, bs)
+      def result(implicit cec: ConcurrentExecutionContext) = {
+        // start both futures in parallel
+        val f1 = opA.result
+        val f2 = opB.result
+        f1.zip(f2).map { case (a, b) => fn(a, b) }
+      }
     }
   }
 
@@ -375,7 +375,18 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
           }
           go(hcg)
 
-        case cgp@CoGroupedPipe(_) => ???
+        case CoGroupedPipe(cg) =>
+          def go[K, V](cg: CoGrouped[K, V]) = {
+            val inputs = cg.inputs
+            val joinf = cg.joinFunction
+            /**
+             * we need to expand the Op type to deal
+             * with multi-way or we need to keep
+             * the decomposed series of joins
+             */
+            ???
+          }
+          go(cg)
 
         case ReduceStepPipe(IdentityReduce(_, pipe, _, descriptions)) =>
           plan(m, pipe)
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
@@ -1,0 +1,493 @@
+package com.twitter.scalding.typed.memory_backend
+
+import cascading.flow.{ FlowDef, FlowConnector }
+import cascading.pipe.Pipe
+import cascading.tap.Tap
+import cascading.tuple.{ Fields, TupleEntryIterator }
+import com.twitter.scalding.typed._
+import com.twitter.scalding.{ Config, Execution, ExecutionCounters, Mode }
+import java.util.concurrent.atomic.AtomicReference
+import java.util.{ ArrayList, Collections, UUID }
+import scala.concurrent.{ Future, ExecutionContext => ConcurrentExecutionContext, Promise }
+import scala.collection.mutable.{ ArrayBuffer, Map => MMap }
+import scala.collection.JavaConverters._
+
+import Execution.{ ToWrite, Writer }
+
+class AtomicBox[T <: AnyRef](init: T) {
+  private[this] val ref = new AtomicReference[T](init)
+
+  def lazySet(t: T): Unit =
+    ref.lazySet(t)
+
+  /**
+   * use a pure function to update the state.
+   * fn may be called more than once
+   */
+  def update[R](fn: T => (T, R)): R = {
+
+    @annotation.tailrec
+    def loop(): R = {
+      val init = ref.get
+      val (next, res) = fn(init)
+      if (ref.compareAndSet(init, next)) res
+      else loop()
+    }
+
+    loop()
+  }
+
+  def get(): T = ref.get
+}
+
+final class MemoryMode private (srcs: Map[TypedSource[_], Iterable[_]], sinks: Map[TypedSink[_], AtomicBox[Option[Iterable[_]]]]) extends Mode {
+
+  def newWriter(): Writer =
+    new MemoryWriter(this)
+
+  def openForRead(config: Config, tap: Tap[_, _, _]): TupleEntryIterator = ???
+  def fileExists(filename: String): Boolean = ???
+  def newFlowConnector(props: Config): FlowConnector = ???
+
+  def addSource[T](src: TypedSource[T], ts: Iterable[T]): MemoryMode =
+    new MemoryMode(srcs + (src -> ts), sinks)
+
+  def addSink[T](sink: TypedSink[T]): MemoryMode =
+    new MemoryMode(srcs, sinks + (sink -> new AtomicBox[Option[Iterable[_]]](None)))
+
+  /**
+   * This has a side effect of mutating this MemoryMode
+   */
+  def writeSink[T](t: TypedSink[T], iter: Iterable[T]): Unit =
+    sinks(t).lazySet(Some(iter))
+
+  def readSink[T](t: TypedSink[T]): Option[Iterable[T]] =
+    sinks.get(t).flatMap(_.get).asInstanceOf[Option[Iterable[T]]]
+
+  def readSource[T](t: TypedSource[T]): Option[Iterable[T]] =
+    srcs.get(t).asInstanceOf[Option[Iterable[T]]]
+}
+
+object MemoryMode {
+  def empty: MemoryMode = new MemoryMode(Map.empty, Map.empty)
+}
+
+object MemoryPlanner {
+
+  sealed trait Op[+O] {
+    def result(implicit cec: ConcurrentExecutionContext): Future[ArrayBuffer[_ <: O]]
+
+    def flatMap[O1](fn: O => TraversableOnce[O1]): Op[O1] =
+      Op.Map(this, fn)
+
+    def mapAll[O1 >: O, O2](fn: IndexedSeq[O1] => ArrayBuffer[O2]): Op[O2] =
+      Op.MapAll[O1, O2](this, fn)
+  }
+  object Op {
+    def source[I](i: Iterable[I]): Op[I] = Source(_ => Future.successful(i))
+    def empty[I]: Op[I] = source(Nil)
+
+    case class Source[I](input: ConcurrentExecutionContext => Future[Iterable[I]]) extends Op[I] {
+      private[this] val promise: Promise[ArrayBuffer[I]] = Promise()
+
+      def result(implicit cec: ConcurrentExecutionContext): Future[ArrayBuffer[I]] = {
+        if (!promise.isCompleted) {
+          promise.tryCompleteWith {
+            val iter = input(cec)
+            iter.map { i => ArrayBuffer.concat(i) }
+          }
+        }
+
+        promise.future
+      }
+    }
+
+    case class Materialize[O](op: Op[O]) extends Op[O] {
+      private[this] val promise: Promise[ArrayBuffer[_ <: O]] = Promise()
+
+      def result(implicit cec: ConcurrentExecutionContext) = {
+        if (!promise.isCompleted) {
+          promise.tryCompleteWith(op.result)
+        }
+
+        promise.future
+      }
+    }
+
+    case class Concat[O](left: Op[O], right: Op[O]) extends Op[O] {
+      def result(implicit cec: ConcurrentExecutionContext) =
+        for {
+          l <- left.result
+          r <- right.result
+        } yield {
+          ArrayBuffer.concat(l, r)
+        }
+    }
+
+    case class Map[I, O](input: Op[I], fn: I => TraversableOnce[O]) extends Op[O] {
+      def result(implicit cec: ConcurrentExecutionContext): Future[ArrayBuffer[O]] =
+        input.result.map { array =>
+          val res = ArrayBuffer[O]()
+          val it = array.iterator
+          while(it.hasNext) {
+            val i = it.next
+            fn(i).foreach(res += _)
+          }
+          res
+        }
+    }
+
+    case class MapAll[I, O](input: Op[I], fn: IndexedSeq[I] => ArrayBuffer[O]) extends Op[O] {
+      def result(implicit cec: ConcurrentExecutionContext) =
+        input.result.map(fn)
+    }
+
+    case class Reduce[K, V1, V2](
+      input: Op[(K, V1)],
+      fn: (K, Iterator[V1]) => Iterator[V2],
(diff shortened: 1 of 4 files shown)


johnynek:
cc @ianoc

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+
+  def get(): T = ref.get
+}
+
+final class MemoryMode private (srcs: Map[TypedSource[_], Iterable[_]], sinks: Map[TypedSink[_], AtomicBox[Option[Iterable[_]]]]) extends Mode {
+
+  def newWriter(): Writer =
+    new MemoryWriter(this)
+
+  def openForRead(config: Config, tap: Tap[_, _, _]): TupleEntryIterator = ???
```
piyushnarang: throw an exception instead? this will evaluate to a NotImplementedError which might lead the caller / user to think it's a implementation bug?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+   * use a pure function to update the state.
+   * fn may be called more than once
+   */
+  def update[R](fn: T => (T, R)): R = {
+
+    @annotation.tailrec
+    def loop(): R = {
+      val init = ref.get
+      val (next, res) = fn(init)
+      if (ref.compareAndSet(init, next)) res
```
ulyssepence, line 33: Just to clarify: This is like STM where you're only making the change if the initial conditions when calling `fn` are met? Could this loop forever if there is another thread that consistently gets scheduled between the read and write?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+}
+
+final class MemoryMode private (srcs: Map[TypedSource[_], Iterable[_]], sinks: Map[TypedSink[_], AtomicBox[Option[Iterable[_]]]]) extends Mode {
+
+  def newWriter(): Writer =
+    new MemoryWriter(this)
+
+  def openForRead(config: Config, tap: Tap[_, _, _]): TupleEntryIterator = ???
+  def fileExists(filename: String): Boolean = ???
+  def newFlowConnector(props: Config): FlowConnector = ???
```
ulyssepence: These are just not implemented yet? Is there an exception we could throw instead of using undefined? Or is this the standard for scala?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+      else loop()
+    }
+
+    loop()
+  }
+
+  def get(): T = ref.get
+}
+
+final class MemoryMode private (srcs: Map[TypedSource[_], Iterable[_]], sinks: Map[TypedSink[_], AtomicBox[Option[Iterable[_]]]]) extends Mode {
```
ulyssepence, line 43: Will this `Map` work if we read from and write to the same path at different points during the flow?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+
+    case class Reduce[K, V1, V2](
+      input: Op[(K, V1)],
+      fn: (K, Iterator[V1]) => Iterator[V2],
+      ord: Option[Ordering[_ >: V1]]
+      ) extends Op[(K, V2)] {
+
+      def result(implicit cec: ConcurrentExecutionContext): Future[ArrayBuffer[(K, V2)]] =
+        input.result.map { kvs =>
+          val m = MMap[K, ArrayList[V1]]()
```
ulyssepence: Can we rename for clarity to like `valuesGroupedByKey` or something?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: yes. But I don't think that can happen because we only mutate it when jobs finish and there are only a finite number of jobs running at a time, so I think it is exponentially unlikely that they do spin forever.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: yes, because the sinks and sources are separate. You can't make a source EXCEPT something that is already a list on this platform.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: These cannot be implemented because this platform does not know about cascading. Ultimately I'd like to remove these methods from `Mode` and move them down to a class like `CascadingBackedMode`, but I want to do that with your help since you likely have some code somewhere that assumes `Mode` has these methods.

fwbrasil: @johnynek could you add a comment or a specific exception with this explanation?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+}
+
+object MemoryMode {
+  def empty: MemoryMode = new MemoryMode(Map.empty, Map.empty)
+}
+
+object MemoryPlanner {
+
+  sealed trait Op[+O] {
+    def result(implicit cec: ConcurrentExecutionContext): Future[ArrayBuffer[_ <: O]]
```
fwbrasil, line 86: Would this new mode be used only for testing? I think a simpler synchronous implementation without parallellism would suffice for that.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+          promise.tryCompleteWith(op.result)
+        }
+
+        promise.future
+      }
+    }
+
+    case class Concat[O](left: Op[O], right: Op[O]) extends Op[O] {
+      def result(implicit cec: ConcurrentExecutionContext) = {
+        // start both futures in parallel
```
fwbrasil: Why not inline these? They will run in parallel regardless.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+object MemoryMode {
+  def empty: MemoryMode = new MemoryMode(Map.empty, Map.empty)
+}
+
+object MemoryPlanner {
+
+  sealed trait Op[+O] {
+    def result(implicit cec: ConcurrentExecutionContext): Future[ArrayBuffer[_ <: O]]
+
+    def flatMap[O1](fn: O => TraversableOnce[O1]): Op[O1] =
```
fwbrasil: I find the naming here confusing. `flatMap` indicates a monad, but function doesn't return a monad instance. Also, it's strange the implementation creates a `Map` for a `flatMap`. Maybe the method could be called `mapValues` and the implementation be on top of `mapAll`?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+      Op.Map(this, fn)
+
+    def mapAll[O1 >: O, O2](fn: IndexedSeq[O1] => ArrayBuffer[O2]): Op[O2] =
```
fwbrasil: wdyt about `transform` instead of `mapAll`?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+      }
+    }
+
+    case class Join[A, B, C](
+      opA: Op[A],
+      opB: Op[B],
+      fn: (IndexedSeq[A], IndexedSeq[B]) => ArrayBuffer[C]) extends Op[C] {
+
+      def result(implicit cec: ConcurrentExecutionContext) = {
+        // start both futures in parallel
```
fwbrasil, line 226: inline as well?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+
+}
+
+/**
+ * These are just used as type markers which are connected
+ * to inputs via the MemoryMode
+ */
+case class SourceT[T](ident: String) extends TypedSource[T] {
+  def converter[U >: T] = ???
+  def read(implicit flowDef: FlowDef, mode: Mode): Pipe = ???
```
fwbrasil: add a comment or a specific exception with more context

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+}
+
+/**
+ * These are just used as type markers which are connected
+ * to outputs via the MemoryMode
+ */
+case class SinkT[T](indent: String) extends TypedSink[T] {
+  def setter[U <: T] = ???
+  def writeFrom(pipe: Pipe)(implicit flowDef: FlowDef, mode: Mode): Pipe = ???
```
fwbrasil: ditto

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+          def go[T](f: Filter[T]): (Memo, Op[T]) = {
+            val Filter(p, fn) = f
+            val (m1, op) = plan(m, f)
+            (m1, op.flatMap { t => if (fn(t)) t :: Nil else Nil })
+          }
+          go(f)
+
+        case f@FlatMapValues(_, _) =>
+          def go[K, V, U](node: FlatMapValues[K, V, U]) = {
+            // don't capture node, which is a TypedPipe, which we avoid serializing
```
fwbrasil: unnecessary?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+        case Fork(pipe) =>
+          val (m1, op) = plan(m, pipe)
+          (m1, Op.Materialize(op))
+
+        case IterablePipe(iterable) =>
+          (m, Op.source(iterable))
+
+        case f@MapValues(_, _) =>
+          def go[K, V, U](node: MapValues[K, V, U]) = {
+            // don't capture node, which is a TypedPipe, which we avoid serializing
```
fwbrasil: unnecessary?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+            val mvfn = node.fn
+            val (m1, op) = plan(m, node.input)
+            (m1, op.flatMap { case (k, v) => Iterator.single((k, mvfn(v))) })
+          }
+
+          go(f)
+
+        case Mapped(input, fn) =>
+          val (m1, op) = plan(m, input)
+          (m1, op.flatMap { t => fn(t) :: Nil })
```
fwbrasil: it might be worth adding a `map` method to avoid creating one list for each element.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+        case MergedTypedPipe(left, right) =>
+          val (m1, op1) = plan(m, left)
+          val (m2, op2) = plan(m1, right)
+          (m2, Op.Concat(op1, op2))
+
+        case SourcePipe(src) =>
+          (m, Op.Source({ cec =>
+            mem.readSource(src) match {
+              case Some(iter) => Future.successful(iter)
+              case None => Future.failed(new Exception(s"Source: $src not wired"))
```
fwbrasil: wdyt about explaining how the user can fix the error in the exception message?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+              res += ((k, v))
+            }
+          }
+          res
+        })
+      }
+      go(imr)
+  }
+
+  case class State(
```
fwbrasil: could it be `private[this]`?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+    id: Long,
+    memo: MemoryPlanner.Memo,
+    forced: Map[TypedPipe[_], Future[TypedPipe[_]]]
+    )
+
+  val state = new AtomicBox[State](State(0, MemoryPlanner.Memo.empty, Map.empty))
```
fwbrasil: could it be `private[this]`?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: I want to actually make this production grade as we move forward. I don't think it will be *that* hard. So I don't want to back out the design of using Futures since a better system will have more explicit parallelism.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: I hear you on `flatMap`. I can rename concatMap, but the fact is we used `flatMap` for scalding to make it a bit easier for newbies to scala (since it is like flatMap on List in that List accepts a similarly broad return type). `TypedPipe[T]` is definitely not a monad (it is Applicative!). But I can change this name here.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: sure.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: bad comment from when I did `for { }` which is not parallel unless we start the futures first. Now, it doesn't matter (using zip).

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: yeah, left over copypasta.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: +1.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: yes.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: yes.

johnynek:
will send an update addressing these comments. Thank you for taking the time to look.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
fwbrasil: Fair enough. I think `Future` might not be the best tool to achieve parallelism given its high execution overhead, but I think it's fine to use `Future` for now and revisit the decision if necessary.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
fwbrasil: thanks!

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+  /**
+   * use a pure function to update the state.
+   * fn may be called more than once
+   */
+  def update[R](fn: T => (T, R)): R = {
+
+    @annotation.tailrec
+    def loop(): R = {
+      val init = ref.get
+      val (next, res) = fn(init)
```
piyushnarang, line 32: do we need to worry about fn being computationally expensive? We could call and cache the value in the enclosing method if that's the case

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
piyushnarang: could we name the generic type something else rather than O1?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
piyushnarang: something apart from O1, O2?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+                }
+              }
+              val res = new ArrayBuffer[(K, V)](map.size)
+              map.foreach { res += _ }
+              res
+            })
+          }
+          sum(slk)
+
+        case tp@TrappedPipe(_, _, _) => ???
```
piyushnarang, line 411: throw not yet implemented exception?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+          // this can be interpretted as catching any exception
+          // on the map-phase until the next partition, so it can
+          // be made to work, but skipping for now
+
+        case WithDescriptionTypedPipe(pipe, description, dedup) =>
+          plan(m, pipe)
+
+        case WithOnComplete(pipe, fn) => ???
```
piyushnarang: here too?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -0,0 +1,527 @@
...
+        case CoGroupedPipe(cg) =>
+          def go[K, V](cg: CoGrouped[K, V]) = {
+            val inputs = cg.inputs
+            val joinf = cg.joinFunction
+            /**
+             * we need to expand the Op type to deal
+             * with multi-way or we need to keep
+             * the decomposed series of joins
+             */
+            ???
```
piyushnarang: should this be a todo?

piyushnarang (commented):
Really cool stuff. Would be nice to add more scenarios in the memory test but we can follow up with that in a future pr.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: `???` means not implemented.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
johnynek: same.

johnynek:
okay.

Sorry for the delay. Can you all take another look. As you know, this is a long line of changes and I am trying to keep each one somewhat digestible (shooting for ~400 lines of diff). This is slightly longer, so I'm hoping we can address any outstanding issues in a follow up.

I'd love to get the optimizations in place so we could think about releasing scalding 0.18 with this change to typed pipe, and in fact this memory platform is just a proof of concept that you can run without cascading. We can polish it more and make it as nice as we like, but the main purpose is to have a realistic example to prove that the API basically works without getting into the weeds of spark or flink.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
fwbrasil: If I understood correctly the purpose of this class, caching doesn't seem possible since `fn` depends on the current value and needs to be computed on each retry (the value will be different on each retry). @johnynek maybe renaming `loop` to `retry` would be more clear?

piyushnarang:
Looks good to me. Seems like the CI build has been hitting the 50 min timeout on the hadoop tests (noticed that on: https://github.com/twitter/scalding/pull/1700 as well). We'll need to either bump the timeout / maybe breakout the tests in that suite.

fwbrasil (approved):
LGTM

PR #1698, opened 2017-06-04
Make an ADT for CoGrouped

johnynek:
CoGrouped was the last scalding typed trait to be a black box that we cannot look inside at planning.

This change continues to expose the old methods so you can use the direct composition of CoGrouped, or you can take them apart and re-assemble in a backend as may be more appropriate with global information or for particular compute substrates.

After this is merged, we can pretty easily add join support to #1697 without having to directly support N-way joins on the memory backend.

This will also be useful for spark, which does not support N-way joins in the RDD API.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
@@ -60,41 +60,8 @@ sealed trait CoGroupable[K, +R] extends HasReducers with HasDescription with jav
    * fewer values per key on the right. If both sides are similar, no need to worry.
    * If one side is a one-to-one mapping, that should be the "smaller" side.
    */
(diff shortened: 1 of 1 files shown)


johnynek:
cc @fwbrasil

ulyssepence (approved):
Looks reasonable. 

"This change continues to expose the old methods so you can use the direct composition of CoGrouped, or you can take them apart and re-assemble in a backend as may be more appropriate with global information or for particular compute substrates."

Am I understanding correctly that you expect users to use the case classes directly? Or do you mean for porting Scalding to different platforms?

johnynek:
A normal user of scalding would never use this for anything that comes to mind now.

Someone writing some specific optimizer of scalding TypedPipes would use it, or someone writing a new scalding backend would use it.

Pretty much everyone else should never notice anything.

PR #1704, opened 2017-07-12
Revert memory estimator changes on 0.17.x branch

piyushnarang:
Reverted the commit that brought in the memory estimator changes into 0.17.x (https://github.com/twitter/scalding/pull/1700). This will allow us to cut a 0.17.2 release that is binary & source compatible.


ttim (approved):

dieu (approved):

.travis.yml
```diff
@@ -60,14 +60,6 @@ matrix:
       script: "scripts/run_test.sh"
 
     - scala: 2.11.8
-      env: BUILD="base" TEST_TARGET="scalding-estimators-test"
-      script: "scripts/run_test.sh"
-
-    - scala: 2.12.1
```
johnynek, line 66: wait, didn't we lose `2.12` CI here?

johnynek: oh nope. sorry.

build.sbt
```diff
@@ -25,7 +25,7 @@ val elephantbirdVersion = "4.15"
 val hadoopLzoVersion = "0.4.19"
```
johnynek, line 25: would be nice to turn mima back on in this PR or I guess the next and make sure we are compatibile with `0.17.0`.

piyushnarang: Yeah I was going to follow up with a PR to set mima against 0.17.0 and also enable it in the `run_test.sh` script

johnynek:
we really need a migration guide here.

People configure reducer estimators with reflection (class names). I'm afraid this is a major pain now since compilation failures won't catch it.

Also, it is not clear what ReducerEstimator should be replaced with. Estimator? Can you make a short set of comments @dieu ?

dieu:
@johnynek sorry for the delay, I will on this today and give you more detail about this.

PR #1705, opened 2017-07-13
Try turning on mima

piyushnarang:
Updated the version to 0.17.0 and also seeing if we can trigger this from the run_test script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - rvm use 2.2.6 --install --fuzzy
   - gem update --system
   - gem install sass
-  - gem install jekyll -v 3.5.0
+  - gem install jekyll -f -v 3.2.1
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - rvm use 2.2.6 --install --fuzzy
   - gem update --system
   - gem install sass
-  - gem install jekyll -v 3.2.1
+  - gem install jekyll -v 3.5.0
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


johnynek:
looks like all the builds timed out.

ttim:
I think this PR should be against `0.17.x` branch and for `develop` branch `youngestForwardCompatible` should return null.

Meanwhile I checked locally and mima check passed on current 0.17.x branch.

piyushnarang:
Seems to be stuck on a weird issue:
```
Fetching: listen-3.0.8.gem (100%)
listen's executable "listen" conflicts with sass-listen
Overwrite the executable? [yN]  
```

piyushnarang:
Seems like we're installing this for our microsite setup: https://github.com/twitter/scalding/blob/develop/.travis.yml#L10
```
- rvm use 2.2.3 --install --fuzzy
--
  | - gem update --system
  | - gem install sass
  | - gem install jekyll -v 3.2.1
```

piyushnarang:
@johnynek what do you think of @ttim 's suggestion of only enabling this check in 0.17.x (as we are planning to cut 0.18.0 based off develop anyway and we have a bunch of mima failures here already due to the cascading backend rewrites + memory estimator updates). Flip side of not having this on here is that new PRs that are potentially not compatible and could be easily updated to be compatible won't be caught.

johnynek:
I think develop is already broken with respect to mima, so we really can't turn it on, so I think that's right.

piyushnarang:
Ok, closing this review for now. We can set this up again when we have a 0.18.0 release for develop to compare against.

PR #1711, opened 2017-08-03
Add summingbird graph library as a zero-dep module

johnynek:
This moves the summingbird graph code into a zero dependency module in scalding.

This will allow us to later delete it from summingbird and depend on it here if we choose. Also it allows us to use the same graph code to do transformations and optimizations of the TypedPipe graphs.

After this is merged, my next step is to use this to make the new cascading backend better (look for fan-out nodes that the user did not use `fork` on, so that we avoid recomputation. We can also use this code to apply some optimization rules (e.g. pushing up `filterKeys` on TypedPipe into the previous step, which we can't currently do).

### DIFF
--- a/scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
+++ b/scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
@@ -1,0 +1,58 @@
+/*
+ Copyright 2013 Twitter, Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package com.twitter.scalding.graph
+
+/**
+ * Given Dag and a List of immutable nodes, and a function to get
+ * dependencies, compute the dependants (reverse the graph)
+ */
+abstract class DependantGraph[T] {
+  def nodes: List[T]
+  def dependenciesOf(t: T): Iterable[T]
+
+  lazy val allTails: List[T] = nodes.filter { t =>
+    fanOut(t) match {
+      case Some(n) => n == 0
+      case None => false
+    }
+  }
+  private lazy val nodeSet: Set[T] = nodes.toSet
+
+  /**
+   * This is the dependants graph. Each node knows who it depends on
+   * but not who depends on it without doing this computation
+   */
+  private lazy val graph: NeighborFn[T] = reversed(nodes)(dependenciesOf(_))
+
+  private lazy val depths: Map[T, Int] = dagDepth(nodes)(dependenciesOf(_))
+
+  /**
+   * The max of zero and 1 + depth of all parents if the node is the graph
+   */
+  def isNode(p: T): Boolean = nodeSet.contains(p)
+  def depth(p: T): Option[Int] = depths.get(p)
+
+  def dependantsOf(p: T): Option[List[T]] =
+    if (isNode(p)) Some(graph(p).toList) else None
+
+  def fanOut(p: T): Option[Int] = dependantsOf(p).map { _.size }
+  /**
+   * Return all dependendants of a given node.
+   * Does not include itself
+   */
+  def transitiveDependantsOf(p: T): List[T] = depthFirstOf(p)(graph)
+}
--- a/scalding-graph/src/main/scala/com/twitter/scalding/graph/Expr.scala
+++ b/scalding-graph/src/main/scala/com/twitter/scalding/graph/Expr.scala
@@ -1,0 +1,84 @@
+/*
+ Copyright 2014 Twitter, Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package com.twitter.scalding.graph
+
+/**
+ * The Expressions are assigned Ids. Each Id is associated with
+ * an expression of inner type T.
+ *
+ * This is done to put an indirection in the ExpressionDag that
+ * allows us to rewrite nodes by simply replacing the expressions
+ * associated with given Ids.
+ *
+ * T is a phantom type used by the type system
+ */
+final case class Id[T](id: Int)
+
+/**
+ * Expr[T, N] is an expression of a graph of container nodes N[_] with
+ * result type N[T]. These expressions are like the Literal[T, N] graphs
+ * except that functions always operate with an indirection of a Id[T]
+ * where N[T] is the type of the input node.
+ *
+ * Nodes can be deleted from the graph by replacing an Expr at Id = idA
+ * with Var(idB) pointing to some upstream node.
+ *
+ * To add nodes to the graph, add depth to the final node returned in
+ * a Unary or Binary expression.
+ *
+ * TODO: see the approach here: https://gist.github.com/pchiusano/1369239
+ * Which seems to show a way to do currying, so we can handle general
+ * arity
+ */
+sealed trait Expr[T, N[_]] {
+  def evaluate(idToExp: HMap[Id, ({ type E[t] = Expr[t, N] })#E]): N[T] =
+    Expr.evaluate(idToExp, this)
+}
+case class Const[T, N[_]](value: N[T]) extends Expr[T, N] {
+  override def evaluate(idToExp: HMap[Id, ({ type E[t] = Expr[t, N] })#E]): N[T] = value
+}
+case class Var[T, N[_]](name: Id[T]) extends Expr[T, N]
+case class Unary[T1, T2, N[_]](arg: Id[T1], fn: N[T1] => N[T2]) extends Expr[T2, N]
+case class Binary[T1, T2, T3, N[_]](arg1: Id[T1],
+  arg2: Id[T2],
+  fn: (N[T1], N[T2]) => N[T3]) extends Expr[T3, N]
+
+object Expr {
+  def evaluate[T, N[_]](idToExp: HMap[Id, ({ type E[t] = Expr[t, N] })#E], expr: Expr[T, N]): N[T] =
+    evaluate(idToExp, HMap.empty[({ type E[t] = Expr[t, N] })#E, N], expr)._2
+
+  private def evaluate[T, N[_]](idToExp: HMap[Id, ({ type E[t] = Expr[t, N] })#E],
+    cache: HMap[({ type E[t] = Expr[t, N] })#E, N],
+    expr: Expr[T, N]): (HMap[({ type E[t] = Expr[t, N] })#E, N], N[T]) = cache.get(expr) match {
+    case Some(node) => (cache, node)
+    case None => expr match {
+      case Const(n) => (cache + (expr -> n), n)
+      case Var(id) =>
+        val (c1, n) = evaluate(idToExp, cache, idToExp(id))
+        (c1 + (expr -> n), n)
+      case Unary(id, fn) =>
+        val (c1, n1) = evaluate(idToExp, cache, idToExp(id))
+        val n2 = fn(n1)
+        (c1 + (expr -> n2), n2)
+      case Binary(id1, id2, fn) =>
+        val (c1, n1) = evaluate(idToExp, cache, idToExp(id1))
+        val (c2, n2) = evaluate(idToExp, c1, idToExp(id2))
+        val n3 = fn(n1, n2)
+        (c2 + (expr -> n3), n3)
+    }
+  }
+}
--- a/scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
+++ b/scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
@@ -1,0 +1,372 @@
+/*
+ Copyright 2014 Twitter, Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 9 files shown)


scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
```diff
@@ -0,0 +1,58 @@
...
+ limitations under the License.
+ */
+
+package com.twitter.scalding.graph
+
+/**
+ * Given Dag and a List of immutable nodes, and a function to get
+ * dependencies, compute the dependants (reverse the graph)
+ */
+abstract class DependantGraph[T] {
```
ttim, line 23: Do we really want to keep it inheritance based here? Why don't do class instead (with `nodes` and `dependenciesOf` (or even just `Map[T, List[T]]`) as an input)?

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
```diff
@@ -0,0 +1,58 @@
...
+      case None => false
+    }
+  }
+  private lazy val nodeSet: Set[T] = nodes.toSet
+
+  /**
+   * This is the dependants graph. Each node knows who it depends on
+   * but not who depends on it without doing this computation
+   */
+  private lazy val graph: NeighborFn[T] = reversed(nodes)(dependenciesOf(_))
```
ttim, line 39: Why `graph` made this way while other things (ie `transitiveDependantsOf`) made other way?

piyushnarang:
👍 looks good to me.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The max of zero and 1 + depth of all parents if the node is the graph
+   */
+  def isNode(p: T): Boolean = nodeSet.contains(p)
+  def depth(p: T): Option[Int] = depths.get(p)
+
+  def dependantsOf(p: T): Option[List[T]] =
+    if (isNode(p)) Some(graph(p).toList) else None
+
+  def fanOut(p: T): Option[Int] = dependantsOf(p).map { _.size }
```
ttim, line 52: I don't think it should be `Option[Int]`

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  private lazy val depths: Map[T, Int] = dagDepth(nodes)(dependenciesOf(_))
+
+  /**
+   * The max of zero and 1 + depth of all parents if the node is the graph
+   */
+  def isNode(p: T): Boolean = nodeSet.contains(p)
+  def depth(p: T): Option[Int] = depths.get(p)
+
+  def dependantsOf(p: T): Option[List[T]] =
```
ttim, line 49: Just `List[T]`?

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
```diff
@@ -0,0 +1,58 @@
...
+   */
+  private lazy val graph: NeighborFn[T] = reversed(nodes)(dependenciesOf(_))
+
+  private lazy val depths: Map[T, Int] = dagDepth(nodes)(dependenciesOf(_))
+
+  /**
+   * The max of zero and 1 + depth of all parents if the node is the graph
+   */
+  def isNode(p: T): Boolean = nodeSet.contains(p)
+  def depth(p: T): Option[Int] = depths.get(p)
```
ttim, line 47: Just `Int`?

scalding-graph/src/main/scala/com/twitter/scalding/graph/Expr.scala
```diff
@@ -0,0 +1,84 @@
...
+ * TODO: see the approach here: https://gist.github.com/pchiusano/1369239
+ * Which seems to show a way to do currying, so we can handle general
+ * arity
+ */
+sealed trait Expr[T, N[_]] {
+  def evaluate(idToExp: HMap[Id, ({ type E[t] = Expr[t, N] })#E]): N[T] =
+    Expr.evaluate(idToExp, this)
+}
+case class Const[T, N[_]](value: N[T]) extends Expr[T, N] {
+  override def evaluate(idToExp: HMap[Id, ({ type E[t] = Expr[t, N] })#E]): N[T] = value
```
ttim, line 52: We don't really need to override `evaluate` here, right?

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
```diff
@@ -0,0 +1,372 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package com.twitter.scalding.graph
+
+/////////////////////
+// There is no logical reason for Literal[T, N] to be here,
```
ttim, line 20: Should be addressed I guess?

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
```diff
@@ -0,0 +1,372 @@
...
+// java.lang.Error: typeConstructor inapplicable for <none>
+//   at scala.tools.nsc.symtab.SymbolTable.abort(SymbolTable.scala:34)
+//   at scala.tools.nsc.symtab.Symbols$Symbol.typeConstructor(Symbols.scala:880)
+////////////////////
+
+/**
+ * This represents literal expressions (no variable redirection)
+ * of container nodes of type N[T]
+ */
+sealed trait Literal[T, N[_]] {
```
ttim, line 32: This represents something similar `Expr` but without `Id`s, right? Do we want to carry both of them?

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
ttim: This class in general is something unrelated to other classes, I don't think we want to copy it here.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
```diff
@@ -0,0 +1,372 @@
...
+          case BinaryLit(in1, in2, fn) =>
+            val (h1, p1) = evaluate(hm, in1)
+            val (h2, p2) = evaluate(h1, in2)
+            val p3 = fn(p1, p2)
+            (h2 + (lit -> p3), p3)
+        }
+    }
+}
+
+sealed trait ExpressionDag[N[_]] { self =>
```
ttim, line 72: I don't really understand why it's `trait`.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
```diff
@@ -0,0 +1,372 @@
...
+
+  /**
+   * These have package visibility to test
+   * the law that for all Expr, the node they
+   * evaluate to is unique
+   */
+  protected[graph] def idToExp: HMap[Id, E]
+  protected def nodeToLiteral: GenFunction[N, Lit]
+  protected def roots: Set[Id[_]]
+  protected def nextId: Int
```
ttim, line 85: This class is expected to be immutable, right? nit: Then this `nextId` is misleading. I guess something like `maxId` is better, but not sure.

scalding-graph/src/main/scala/com/twitter/scalding/graph/package.scala
```diff
@@ -0,0 +1,92 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package com.twitter.scalding
+
+import scala.collection.mutable.{ Map => MMap }
+
+/** Collection of graph algorithms */
+package object graph {
```
ttim, line 22: Same as `DependantGraph`

ttim (commented):
I think having some general dag optimizing library on the level of scalding is great. 

I guess `DependantGraph` is a class we don't move to scalding.

Also I have one high level concern - I'm not sure I have a good mental model for `ExpressionDag`. 
To be precise - on the level of `Rule` `N[_]` is given and returned. But when you optimise your DAG you can change `UnaryLit` & `BinaryLit` in your graph, right? Why `N[_]` is exposed then and not `Literal[T, N[_]]` there?
But if we expose `Literal[T, N[_]]` it would be nice to pattern match against concrete `BinaryLit` & `UnaryLit` which currently represented by passed functions there. For me simpler abstraction will be to have `BinaryLit` & `UnaryLit` as a base algebra for Dag + two interpreters: one to convert `TypedPipe` into `ExpressionDag` of your base algebra and another to actually evaluate optimized `ExpressionDag` back to `TypedPipe` dag (I guess this is an `evaluate` part right now). I guess I can miss a way how `Rule`s work thou.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
johnynek: I think we do want it. It allows you to reverse the dependency graph to see what depends on a given node. That is often useful when doing graph walks.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
johnynek: Option because the node may not be in the graph.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
johnynek: same comment about Option.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
johnynek: same comment... Node may not be in the graph.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
johnynek: this is expensive, so it was made lazy in case you don't actually need it. You may only want depths, for instance.

We could reconsider that, but I'm reluctant to refactor while copying the code since it makes the history harder to follow. An exact copy is easier to trace in my opinion.

The def below is lazy since all defs are lazy, but some of them trigger the computation of graph or depths.

scalding-graph/src/main/scala/com/twitter/scalding/graph/Expr.scala
johnynek: I guess not. But I'd rather do refactoring separately from the copy.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
johnynek: I think so, I can fix in a follow up PR.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
johnynek: Yeah, I think we could not have both of them if we really want to remove them. The difference is `Expr` has variables and can point with `Var`. Literal does not have that and everything has been fully evaluated.

Again, we can try to rewrite this code after we do the copy in my opinion.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
johnynek: well, this is a common scala issue. `sealed abstract class` is fine too. I'm happy to make that change in a following PR. In scala 2.12, I don't think there is a big cost. Honestly, this trait/abstract class distinction is a real issue in learning scala. If not for the jvm crap I would say there should be no abstract class.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
johnynek: this is immutable, yes. This will be the id of the next `Id` node: https://github.com/twitter/scalding/pull/1711/files/f4b77ce5a70c76568117b9b24f534bea8b124ac5#diff-8fde3d3d5fd7e43cad322bdbb784c5eaR213 which I think is an accurate name.

johnynek (commented):
unless you strongly object @ttim I'd rather merge an exact copy from summingbird and then do the changes in later smaller PRs.

johnynek:
about the question @ttim posed: why do we have `Rule` on `N` rather than `Literal`. The reason is that it is much easier to write rules in terms of the underlying nodes.

See the summingbird rules:
https://github.com/twitter/summingbird/blob/develop/summingbird-core/src/main/scala/com/twitter/summingbird/planner/DagOptimizer.scala#L354

There are pretty easy to write and read on the original nodes, but on `Literal` they would be almost impossible since you need to match on the functions that are embedded in the `Unary` or `Binary` operations.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
ttim: Yes, it's useful, sure, just feels completely separate. I'm not against of copying it actually. Let's do that.

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
ttim: I see the point of "just copy", but I'm afraid we will not change it then forever =)

scalding-graph/src/main/scala/com/twitter/scalding/graph/DependantGraph.scala
ttim: Hm, I see. Then I prefer to have a comment about this.

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
ttim: My question was more - why it should be abstract at all?

scalding-graph/src/main/scala/com/twitter/scalding/graph/ExpressionDag.scala
ttim: Yes, I understand how it used, but I think it's better to describe what is it, not how it used because former contains mutability semantic.

But I'm ok to keep it as is.

ttim:
@johnynek I'm not strongly against, I see a point to merge it as is and do some follow-ups.

About `Rule`s on `N[_]` - I'm totally agree it's simpler, what I don't understand (and that's where my mental model breaks) - if you expressed change of dag structure on the level of `N[_]` how this can be reflected on dag itself?

PR #1713, opened 2017-08-09
Check fan-out before composing functions

johnynek:
This makes the cascading plan look closer to what the user writes. If
the user consumes a TypedPipe in two branches, we will materialize that
branch for cascading (which may or may not materialize before the fork).

This does not change behavior of explicit forks or forceToDisks

Note without Execution, we currently don't see all the writes at once,
so this will not perfectly work with `.write` vs `.writeExecution`. In
the worst case, it should behave as it did previously, but plans may
be better in some cases (where there are diamonds) or with Executions.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -137,6 +137,52 @@ object TypedPipe extends Serializable {
     }
   }
 
+  /**
+   * Which TypedPipes does the given TypedPipe directly depend on
+   * this must form a DAG by construction. This is useful for
+   * backend implementations.
+   */
+  def dependencies(of: TypedPipe[Any]): List[TypedPipe[Any]] =
+    of match {
+      case CrossPipe(a, b) => List(a, b)
+      case CrossValue(p, EmptyValue) => List(p)
+      case CrossValue(p, LiteralValue(_)) => List(p)
+      case CrossValue(p, ComputedValue(vpipe)) => List(p, vpipe)
+      case DebugPipe(p) => List(p)
+      case FilterKeys(p, _) => List(p)
+      case Filter(p, _) => List(p)
+      case Fork(p) => List(p)
+      case FlatMapValues(p, _) => List(p)
+      case FlatMapped(p, _) => List(p)
+      case ForceToDisk(p) => List(p)
+      case IterablePipe(_) => Nil
+      case MapValues(p, _) => List(p)
+      case Mapped(p, _) => List(p)
+      case MergedTypedPipe(p1, p2) => List(p1, p2)
+      case SourcePipe(_) => Nil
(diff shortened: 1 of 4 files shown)


johnynek:
@ttim @piyushnarang any thoughts?

piyushnarang:
Sorry missed this. Shall take a look.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -137,6 +137,52 @@ object TypedPipe extends Serializable {
     }
   }
 
+  /**
+   * Which TypedPipes does the given TypedPipe directly depend on
+   * this must form a DAG by construction. This is useful for
+   * backend implementations.
+   */
+  def dependencies(of: TypedPipe[Any]): List[TypedPipe[Any]] =
```
piyushnarang, line 145: to confirm, you're using Any here cause we don't know of the types of the dependency TypedPipes right?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -222,9 +223,24 @@ class AsyncFlowDefRunner extends Writer { self =>
 
     val done = Promise[Unit]()
 
+    val writtenPipes = writes.map {
+      case Force(pipe) => pipe
+      case ToIterable(pipe) => pipe
+      case SimpleWrite(pipe, _) => pipe
+    }
+
+    val depGraph = new DependantGraph[TypedPipe[Any]] {
```
piyushnarang, line 232: should we extract this to a helper method in TypedPipe? it's identical to the code in CascadingBackend for `dependants` right?

piyushnarang (approved):
Do we need any tests that verify the updates to the behavior for pipes we're now treating as implicit forks?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: yes (and TypedPipe is covariant, so all of them can be considered as Any).

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: it's not. This one has a number of written pipes, the CascadingBackend works for a single pipe.

johnynek: actually, you are right. It's the same (we just call it with a single pipe). I will combine.

johnynek:
I'm not convinced this is the way to go actually... I'm closing in favor of a different direction.

PR #1717, opened 2017-09-22
Revert "Add summingbird graph library as a zero-dep module"

johnynek:
This reverts commit f4b77ce5a70c76568117b9b24f534bea8b124ac5.

We factored the graph library out into https://github.com/stripe/dagon let's just use that if we need it.


johnynek:
@ttim review?

piyushnarang:
👍

PR #1720, opened 2017-09-23
remove a few warnings

johnynek:
I'm trying to chip away at the giant wall of warnings we have.

review @ttim @ianoc ?


scalding-core/src/main/scala/com/twitter/scalding/HfsConfPropertySetter.scala
```diff
@@ -54,7 +54,7 @@ private[scalding] class ConfPropertiesHfsTap(
  * Changes here however will not show up in the hadoop UI
  */
 trait HfsConfPropertySetter extends HfsTapProvider {
-  @deprecated("Tap config is deprecated, use sourceConfig or sinkConfig directly. In cascading configs applied to sinks can leak to sources in the step writing to the sink.")
+  @deprecated("Tap config is deprecated, use sourceConfig or sinkConfig directly. In cascading configs applied to sinks can leak to sources in the step writing to the sink.", "2017-01-12")
```
piyushnarang: might have been nice if we could tie this to a scalding version. E.g. deprecated since 0.17.1 or so..

piyushnarang (approved):

scalding-core/src/main/scala/com/twitter/scalding/HfsConfPropertySetter.scala
johnynek: I can fix that.

Issue #1721, opened 2017-09-23
clean up macro related warnings


johnynek:
We no longer support scala 2.10, so we can port the macros over to 2.11/2.12 style. Most of our warnings are related to this.

relates to #1603 

cc @ianoc any interest in killing some time on this one? ;)

PR #1725, opened 2017-09-25
Adds more tests around the macros for trait edge cases and others

ianoc-stripe:
This should fail in CI, will follow up with code changes to remove warnings and fix these tests.

### DIFF
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/BinaryOrdering.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/BinaryOrdering.scala
@@ -8,4 +8,4 @@ trait BinaryOrdering {
   implicit def ordSer[T]: OrderedSerialization[T] = macro com.twitter.scalding.serialization.macros.impl.OrderedSerializationProviderImpl[T]
 }
 
-object BinaryOrdering extends BinaryOrdering {}
+object BinaryOrdering extends BinaryOrdering
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/BinaryOrdering.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/BinaryOrdering.scala
@@ -4,10 +4,8 @@ import com.twitter.scalding.serialization.OrderedSerialization
 
 import scala.language.experimental.macros
 
-/**
- * @author Mansur Ashraf.
- */
-object BinaryOrdering {
-
+trait BinaryOrdering {
   implicit def ordSer[T]: OrderedSerialization[T] = macro com.twitter.scalding.serialization.macros.impl.OrderedSerializationProviderImpl[T]
 }
+
+object BinaryOrdering extends BinaryOrdering {}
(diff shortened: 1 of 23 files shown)


scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/BinaryOrdering.scala
```diff
@@ -4,10 +4,8 @@ import com.twitter.scalding.serialization.OrderedSerialization
...
-/**
- * @author Mansur Ashraf.
- */
-object BinaryOrdering {
-
+trait BinaryOrdering {
   implicit def ordSer[T]: OrderedSerialization[T] = macro com.twitter.scalding.serialization.macros.impl.OrderedSerializationProviderImpl[T]
 }
+
+object BinaryOrdering extends BinaryOrdering {}
```
johnynek: I don't *think* you need `{}`

ianoc:
Does a bunch of damage on #1721 but there's more warnings, some of which the linter fires in generated code, not too sure how to stop that even....

@johnynek As you requested :) 


(Also included is failing test + bug fix for some trait handling for classes with lots of fields failing to deserialize. And some better escape hatch handling so users can supply their own OrderedSerializations when the macros fail more often).

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ProductOrderedBuf.scala
```diff
@@ -42,54 +48,182 @@ object ProductOrderedBuf {
...
+    def freshT(id: String) = TermName(c.freshName(id))
 
     val dispatcher = buildDispatcher
     val elementData: List[(c.universe.Type, TermName, TreeOrderedBuf[c.type])] =
-      outerType
-        .declarations
+      outerType.declarations
         .collect { case m: MethodSymbol => m }
-        .filter(m => m.name.toString.startsWith("_"))
+        .filter(m => m.name.toTermName.toString.startsWith("_"))
```
johnynek: I had to remove a bunch of these since they were calling `toTermName` on things that were already `TermName`, but I guess with `blackbox` this changed?

johnynek (approved):
Thanks for removing the curlies!

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ProductOrderedBuf.scala
ianoc-stripe: I suspect i've glossed over some of those warnings migrating in bulk the code with the fixes in it. I'll do a clean compile or two to make sure i've not added new warnings.

ianoc-stripe:
Ok i think that should be better minus those few fixes.

There are a bunch of warnings about unreachable code in match statements, though they seem spurrious. I've added a sys.error into one of them and it fires...

johnynek:
I'm merging as is and we can deal with the blackbox issue in a follow up (I would love to get the warnings cleaned up in scalding).

PR #1727, opened 2017-09-26
Use a Leibniz-like type to remove casts in FlatMappedFn

johnynek:
We can remove casts with zero job-time (but some plan-time) costs by using an improved version of `=:=` which works on `F[A] => F[B]`.

I have done this here and also made all the inlined functions case classes so they have good identities (since we sometimes put these things into hashmaps).


non:
LGTM. 👍 

Leibniz strikes again! 😎

non:
In case my comment is too cryptic, here's a blog post that explains the basic idea behind this:

https://typelevel.org/blog/2014/07/02/type_equality_to_leibniz.html

PR #1728, opened 2017-09-26
[WIP] PullByteBufferOut as a default ordering

ianoc-stripe:
First extraction.

Right now i don't think we need the macros and can probably drop them again.

Naming/code organization and how this should all get imported open to opinions

We could have the defaults show up in the package object ordered serialization itself? 

thoughts ? @johnynek @travisbrown


scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Exported.scala
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright 2015 Twitter, Inc.
```
johnynek: can we skip the bogus copyright?

1) year is wrong
2) stripe is maybe paying you to write this code.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Exported.scala
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package com.twitter.scalding.serialization
+
+case class Exported[T](instance: T) extends AnyVal
```
johnynek: I don't understand this pattern. Can you document it or link to an explanation? I know @travisbrown likes it.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/ExportMacros.scala
```diff
@@ -0,0 +1,24 @@
+package com.twitter.scalding.serialization.macros
+
+import com.twitter.scalding.serialization.Exported
+import com.twitter.scalding.serialization.{OrderedSerialization, DefaultOrderedSerialization}
+import scala.reflect.macros.whitebox
+
+class ExportMacros(val c: whitebox.Context) {
```
johnynek: are we actually calling this anywhere?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Exported.scala
ianoc: The general gist is just that by supplying instances of `Exported` you can import an object/package supplying them and inject them at a lower priority than a normal import.

So here if you import the default object it would just supply low priority implicits so won't override a user supplied one.

https://github.com/milessabin/export-hook

(i can add a link in the code too)

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/ExportMacros.scala
ianoc: Nope, i think its killable/not sure if adds something. Was discussing that with @travisbrown towards EOD here.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Exported.scala
ianoc: Yeah i was going to ask you if you thought we need to put the licence and/or copyright in these at all

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/ComplexHelper.scala
```diff
@@ -0,0 +1,105 @@
...
+      case scala.util.control.NonFatal(e) =>
+        scala.util.Failure(e)
+    }
+
+  final def read(in: InputStream): Try[T] =
+    try {
+      if (staticSize.isEmpty)
+        in.readPosVarInt
+
+      _root_.scala.util.Success(unsafeRead(in))
```
travisbrown: Is the `_root_` just an artifact of a previous macro context?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/ExportMacros.scala
travisbrown: Right now the only exported instance has a concrete type, but once you have generically derived ones the macro here will be necessary (although it turns out it doesn't need to be whitebox, I believe). I've got some notes on this from when I was prepping [this talk](https://meta.plasm.us/slides/scalaworld/)—I'll try to find them.

scalding-serialization/src/test/scala/com/twitter/scalding/serialization/macros/MacroOrderingProperties.scala
```diff
@@ -510,7 +512,7 @@ class MacroOrderingProperties
   }
 
   test("Test out ByteBuffer") {
-    BinaryOrdering.ordSer[ByteBuffer]
+    implicitly[OrderedSerialization[ByteBuffer]]
```
travisbrown: It'd be nice to have a test confirming that an explicit instance doesn't get overridden by the `DefaultOrderedSerialization` import.

ianoc-stripe: Done, did a test with a companion object defined ordered serialization not being overridden

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/ComplexHelper.scala
ianoc-stripe: Yep, i've cleaned these up i think so it should be more sane now

ianoc:
Ok I think this is looking a bit more cleaned up and may be worth looking at again design wise. I'd be slow to merge this as-is if we think develop should always be ready to released since i think for a release we would want to get everything from the macro out and under provided split up. It would be binary breaking since today a single method can be used to do everything. (Though we could possibly inject a different/new whitebox macro to smooth over most of this... but i'm not sure its worth that vs just moving onto a nicer split up world).

There is also a pre-release question here if the old school macros should be under the same package/import as say our ByteBuffer implementations. This would make it harder to shop around and use the ByteBuffer implementation, but then also use Shapeless.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/ComplexHelper.scala
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * This is the worst case: we have to serialize in a side buffer
+   * and then see how large it actually is. This happens for cases, like
+   * string, where the cost to see the serialized size is not cheaper than
+   * directly serializing.
+   */
+  private[this] def noLengthWrite(element: T, outerOutputStream: OutputStream): Unit = {
+    // Start with pretty big buffers because reallocation will be expensive
+    val baos = new ByteArrayOutputStream(512)
+    unsafeWrite(baos, element)
```
johnynek: this implies that `unsafeWrite` means no size. Can we add that to the comments below?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/HasUnsafeCompareBinary.scala
```diff
@@ -0,0 +1,59 @@
...
+*/
+package com.twitter.scalding.serialization
+
+import com.twitter.scalding.serialization.JavaStreamEnrichments._
+import java.io.InputStream
+import java.io.OutputStream
+
+trait HasUnsafeCompareBinary[T] extends OrderedSerialization[T] {
+  def unsafeCompareBinary(inputStreamA: InputStream, inputStreamB: InputStream): Int
+  def unsafeWrite(out: java.io.OutputStream, t: T): Unit
```
johnynek: can we drop `java.io`? Also, I think this has to be the unsized output if there is ever a size header added, right? It seems confusing above since it is used that way, but it is not clear.

It may or may not be sized? Can you add some laws about how to reason about these things?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/HasUnsafeCompareBinary.scala
```diff
@@ -0,0 +1,59 @@
...
+  def unsafeRead(in: java.io.InputStream): T
+  def unsafeSize(t: T): Option[Int]
```
johnynek: what is the contract here? Again similar concerns as above.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/HasUnsafeCompareBinary.scala
```diff
@@ -0,0 +1,59 @@
...
+        // Members declared in scala.math.Ordering
+        def compare(x: T, y: T): Int = o.compare(x, y)
+
+        def dynamicSize(e: T) = o.dynamicSize(e)
+        def unsafeSize(t: T): Option[Int] = o.dynamicSize(t)
+
+        // Members declared in com.twitter.scalding.serialization.Serialization
+        def read(in: java.io.InputStream): scala.util.Try[T] = o.read(in)
+        def staticSize: Option[Int] = o.staticSize
+        def unsafeWrite(out: java.io.OutputStream, t: T): Unit = o.write(out, t).get
```
johnynek: here, `unsafeWrite` could have a size if the original did, but then I guess you could add two sizes, couldn't you (since above we might call `noLengthWrite`)?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/provided/OrderedSerializationByteBuffer.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  def unsafeWrite(outputStream: java.io.OutputStream, element: ByteBuffer): Unit = {
+    outputStream.writePosVarInt(element.remaining)
+    outputStream
+      .writeBytes(element.array, element.arrayOffset + element.position, element.remaining)
+  }
+
+  def unsafeRead(inputStream: java.io.InputStream): ByteBuffer = {
+    val lenA = inputStream.readPosVarInt
```
johnynek, line 59: do we write an additional length header on this thing currently?

ianoc-stripe: Yep we always write a length header for ByteBuffers in the existing master
https://github.com/twitter/scalding/blob/develop/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ByteBufferOrderedBuf.scala#L82

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/ComplexHelper.scala
ianoc-stripe: It implies no outer size, it can be a bit wasteful. I'm adding a comment to the definition of this method that I hope will help here a little:
```
  // This will write out the interior data as a blob with no prepended length
  // This means binary compare cannot skip on this data.
  // However the contract remains that one should be able to _read_ the data
  // back out again.
  def unsafeWrite(out: java.io.OutputStream, t: T): Unit
```

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/Serialization.scala
```diff
@@ -53,6 +54,11 @@ trait Serialization[T] extends Equiv[T] with Hashing[T] with Serializable {
    * otherwise the caller should just serialize into an ByteArrayOutputStream
    */
   def dynamicSize(t: T): Option[Int]
+
+  // Override this to provide more efficient
+  def skip(in: InputStream): Try[Unit] = {
```
johnynek, line 59: I think we may want `def skip(count: Int, in: InputStream): Try[Unit]`  so in say `List[T]` we can skip the rest of the collection. If `count <= 0` do nothing, and otherwise in the worst case just read and throw them away as you do below.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala
```diff
@@ -32,9 +32,17 @@ trait OrderedSerialization[T] extends Ordering[T] with Serialization[T] {
    * the InputStreams is mutated to be the end of the record.
    */
   def compareBinary(a: InputStream, b: InputStream): OrderedSerialization.Result
+
+  /**
+   * This compares two InputStreams. After this call, the position in
+   * the InputStreams may or may not be at the end of the record.
+   */
+  def compareBinaryNoConsume(a: InputStream, b: InputStream): OrderedSerialization.Result = {
```
johnynek, line 40: I get really worried about how to compose methods like this that lack a strong contract. Also, I don't see that we ever call this.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/provided/UnsafeCompareBinary.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  final def read(in: InputStream): Try[T] =
+    try {
+      Success(unsafeRead(in))
+    } catch {
+      case NonFatal(e) =>
+        Failure(e)
+    }
+
+  override def compareBinaryNoConsume(inputStreamA: InputStream, inputStreamB: InputStream): OrderedSerialization.Result =
```
johnynek, line 74: should this be final?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/provided/UnsafeCompareBinary.scala
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      val r = unsafeSwitchingCompareBinaryNoConsume(inputStreamA, inputStreamB, false)
+      OrderedSerialization.resultFrom(r)
+    } catch {
+      case NonFatal(e) =>
+        OrderedSerialization.CompareFailure(e)
+    }
+
+  override def compareBinary(inputStreamA: InputStream, inputStreamB: InputStream): OrderedSerialization.Result =
```
johnynek, line 83: can this be final?

johnynek (changes requested):
What about this: can you update the PR to include 3 hand-written combinators:

`Tuple2OrderedSerialization` `EitherOrderedSerialization` and `ListOrderedSerialization`. I think if we can do those three (static-sized product, static-sized sum/union, dynamic-sized product) I think we will see what methods we want to have to enable.

You can tell something is wrong with OrderedSerialization because in our current tuple2 we have no way to avoid deserializing the second part.

I think if we exercise your code in the same PR that does those three we will be able to see more clearly if we have the API improved or not, or if we are still missing something.

PR #1729, opened 2017-09-27
Use a null check rather than foreach

johnynek:
counters are often hit in a tight loop. We can sacrifice the compiler check on safety here for the big win we could get from inlining a static null check.


ianoc (approved):

ianoc:
merge when green

PR #1730, opened 2017-09-28
Cherrypick for 0.17.3

johnynek:
what do you think @ttim @ianoc @pankajroark ?

If you give the +1, I'll publish this as 0.17.3

I don't think anything else can go in as it is not binary compatible.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Scalding #
 
+## Version 0.17.3 ###
+Cherry picks of develop PRs to fix an issue with null counter reporters
+* https://github.com/twitter/scalding/pull/1729
+* https://github.com/twitter/scalding/pull/1726
+
 ### Version 0.17.2 ###
 This version is basically the same as 0.17.1 but backward compatible with 0.17.0.
 * Revert memory estimator changes on 0.17.x branch: #1704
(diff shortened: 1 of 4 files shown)


johnynek:
cc @tdyas

tdyas:
thanks!

ianoc:
👍

johnynek:
@tdyas 0.17.3 should be making its way to maven central now. Let us know if you learn any more about these null counters. I really don't like silently dropping the counters. Maybe in the future we will add a configuration flag like "optional counters" which allows it to silently ignore counters and otherwise fails if we get a null reporter.

tdyas:
Thanks. This will let me get off our local fork of Scalding.

As for tracking down the cause, we did track it to a specific changeset although the "why" still eludes me. The "weird" part of the changeset in question was that the job had a `Option[TypedPipe]` as a `val` at the toplevel of the job which was later converted to either a transformed version of the wrapped `TypedPipe` getOrElse `TypedPipe.empty`. (The developer was trying to optionally include prior data into their computation.) That could be a total red herring though, I haven't gotten far enough into the issue to know any more.

I agree regarding a config flag. Obviously, I need to work around this issue, but from a correctness standpoint, the need to do that raises a red flag.

PR #1738, opened 2017-10-09
Make TypedPipe Optimizations configurable

### LINKED ISSUE #1735: Make optimizations in 0.18 configurable
For testing optimizations, we want to be able to run the job without any, then run with a set and verify the result is the same.

We could have a test-only method that does this, but it would be nice to give users the ability to control more easily the optimizations applied.

Of course, they can apply them themselves before sending them over, so that may not be as needed for adding optimizations, but they may want to disable some optimizations (which may on some unlikely inputs potentially perform worse that doing nothing, though I don't know of any examples).

johnynek:
closes #1735 

This makes it possible to control what optimizations are applied. This allows us to improve the tests since we can optimize only parts of the graph.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -16,6 +16,7 @@ limitations under the License.
 package com.twitter.scalding
 
 import cascading.flow.{ FlowDef, Flow }
+import com.stripe.dagon.{ Dag, Id, Rule, HMap }
 import com.twitter.algebird.monad.Trampoline
 import com.twitter.algebird.{ Monoid, Monad, Semigroup }
 import com.twitter.scalding.cascading_interop.FlowListenerPromise
@@ -556,6 +557,32 @@ object Execution {
     final case class Force[T](pipe: TypedPipe[T]) extends ToWrite
     final case class ToIterable[T](pipe: TypedPipe[T]) extends ToWrite
     final case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite
+
+    /**
+     * Optimize these writes into new writes and provide a mapping from
+     * the original TypedPipe to the new TypedPipe
+     */
+    def optimizeWriteBatch(writes: List[ToWrite], rules: Seq[Rule[TypedPipe]]): HMap[TypedPipe, TypedPipe] = {
+      val dag = Dag.empty(typed.OptimizationRules.toLiteral)
+      val (d1, ws) = writes.foldLeft((dag, List.empty[Id[_]])) {
+        case ((dag, ws), Force(p)) =>
+          val (d1, id) = dag.addRoot(p)
+          (d1, id :: ws)
+        case ((dag, ws), ToIterable(p)) =>
+          val (d1, id) = dag.addRoot(p)
+          (d1, id :: ws)
+        case ((dag, ws), SimpleWrite(p, sink)) =>
+          val (d1, id) = dag.addRoot(p)
+          (d1, id :: ws)
+      }
+      // now we optimize the graph
+      val d2 = d1.applySeq(rules)
+      // convert back to TypedPipe:
+      ws.foldLeft(HMap.empty[TypedPipe, TypedPipe]) {
+        case (cache, id) =>
+          cache + (d1.evaluate(id) -> d2.evaluate(id))
+      }
+    }
   }
 
   /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -590,10 +590,9 @@ sealed trait TypedPipe[+T] extends Serializable {
    * @return a pipe equivalent to the current pipe.
    */
   def write(dest: TypedSink[T])(implicit flowDef: FlowDef, mode: Mode): TypedPipe[T] = {
-    // Make sure that we don't render the whole pipeline twice:
-    val res = fork
-    dest.writeFrom(res.toPipe[T](dest.sinkFields)(flowDef, mode, dest.setter))
-    res
+    dest.writeFrom(toPipe[T](dest.sinkFields)(flowDef, mode, dest.setter))
+    // We want to fork after this point
+    fork
   }
 
   /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
@@ -88,6 +88,13 @@ class AsyncFlowDefRunner extends Writer { self =>
   type StateKey[T] = (Config, Mode, TypedPipe[T])
   type WorkVal[T] = Future[TypedPipe[T]]
 
+  /**
+   * @param filesToCleanup temporary files created by forceToDiskExecution
+   * @param initToOpt this is the mapping between user's TypedPipes and their optimized versions
+   * which are actually run.
+   * @param forcedPipes these are all the side effecting forcing of TypedPipes into simple
+   * SourcePipes or IterablePipes. These are for both toIterableExecution and forceToDiskExecution
+   */
   private case class State(
     filesToCleanup: Map[Mode, Set[String]],
     initToOpt: HMap[TypedPipe, TypedPipe],
@@ -102,7 +109,8 @@ class AsyncFlowDefRunner extends Writer { self =>
       }
 
     /**
-     * Returns true if we actually add
+     * Returns true if we actually add this optimized pipe. We do this
+     * because we don't want to take the side effect twice.
      */
     def addForce[T](c: Config,
       m: Mode,
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
@@ -17,7 +17,9 @@ import com.twitter.scalding.{
   Mode,
   TypedPipe
 }
+import com.twitter.scalding.typed.TypedSink
 import com.twitter.scalding.cascading_interop.FlowListenerPromise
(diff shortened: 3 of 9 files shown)


ianoc:
Is there any examples using this with sets of optimization phases? (didn't see one other than empty?)

But i guess  one API question here is should we point to an optimization phase as a single class like this or should it be a list of rules? (Thus requiring rules or having rule builders that are reflectively instantiated -- but ultimately this would let us for different jobs on the command line possibly flip rules on and off? )

I also presume in the seq here each rule is itself a combined set of all rules for that phase? -- might be too verbose to do what i suggest above is my own devils advocate here. In that for full configurability we would want to do a spec something like 
`scalding.rules=[[com.twit...,com.twitt],[com.twit,...]]`
or similar to express the rules in each phase? Though that might be pretty powerful for customization -- but possibly overkill.

(I've no super strong preference here, but this change is more API impacting that a lot of the previous ones since adding to Config)

johnynek:
I saw two comments:

1. are we exercising the optimization enough: do we know it is being respected when configured? I think we don't know for sure. I'll add some more tests to convince ourselves.

2. could we make it easier to configure rules? I assume it will be somewhat rare to want to do this, and doing it by making a class with no constructor args and returning what you want would only be a few lines of code. I'd actually like to punt on making it easier to configure for now. I think we could come back if we thought it was too hard to configure, and we could do that even without breaking APIs I think.

How does it sound if I add more tests to address 1. above?

ianoc:
if your confident there isn't an API breakage concern/back into a corner i think it really addresses most of my (1) and (2). Since (1) really was trying to get at exercising the new API.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -556,6 +557,32 @@ object Execution {
...
     final case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite
+
+    /**
+     * Optimize these writes into new writes and provide a mapping from
+     * the original TypedPipe to the new TypedPipe
+     */
+    def optimizeWriteBatch(writes: List[ToWrite], rules: Seq[Rule[TypedPipe]]): HMap[TypedPipe, TypedPipe] = {
+      val dag = Dag.empty(typed.OptimizationRules.toLiteral)
+      val (d1, ws) = writes.foldLeft((dag, List.empty[Id[_]])) {
+        case ((dag, ws), Force(p)) =>
```
ianoc, line 568: is this force a Force from a force to disk or a force to disk execution?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -590,10 +590,9 @@ sealed trait TypedPipe[+T] extends Serializable {
    * @return a pipe equivalent to the current pipe.
    */
   def write(dest: TypedSink[T])(implicit flowDef: FlowDef, mode: Mode): TypedPipe[T] = {
-    // Make sure that we don't render the whole pipeline twice:
-    val res = fork
-    dest.writeFrom(res.toPipe[T](dest.sinkFields)(flowDef, mode, dest.setter))
-    res
+    dest.writeFrom(toPipe[T](dest.sinkFields)(flowDef, mode, dest.setter))
+    // We want to fork after this point
+    fork
```
ianoc, line 595: why did we make this change? I guess it should be ok now with the better optimizer? i'm guessing before we needed to do it the first way to avoid re-computation?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -83,9 +85,13 @@ class AsyncFlowDefRunner extends Writer { self =>
 
   private[this] val mutex = new AnyRef
 
+  type StateKey[T] = (Config, Mode, TypedPipe[T])
+  type WorkVal[T] = Future[TypedPipe[T]]
+
   private case class State(
     filesToCleanup: Map[Mode, Set[String]],
-    forcedPipes: Map[(Config, Mode, TypedPipe[Any]), Future[TypedPipe[Any]]]) {
+    initToOpt: HMap[TypedPipe, TypedPipe],
```
ianoc: could we comment on this field

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -95,19 +101,37 @@ class AsyncFlowDefRunner extends Writer { self =>
         case None => this
       }
 
-    def addPipe[T](c: Config,
+    /**
+     * Returns true if we actually add
```
ianoc: could you expand on this comment

Returns true if we will add to our forced pipes, false if cache hit and noop. I think?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -83,9 +85,13 @@ class AsyncFlowDefRunner extends Writer { self =>
...
   private[this] val mutex = new AnyRef
 
+  type StateKey[T] = (Config, Mode, TypedPipe[T])
+  type WorkVal[T] = Future[TypedPipe[T]]
+
   private case class State(
     filesToCleanup: Map[Mode, Set[String]],
-    forcedPipes: Map[(Config, Mode, TypedPipe[Any]), Future[TypedPipe[Any]]]) {
+    initToOpt: HMap[TypedPipe, TypedPipe],
+    forcedPipes: HMap[StateKey, WorkVal]) {
```
ianoc, line 101: probably worth ensuring we have good comments on all the state fields here (i know this is old, but helps review/reading). i.e. does force mean the same as in terms of the TypedPipe force to disk. Is it something we need to materialize as an output. And i.e. these are our results from the execution run ultimately?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: ForceToDiskExecution.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: I'm not sure why it was ever the other way. Many `.writes` are not consumed. But by putting the fork before, you are always going to do some extra boxing in and out of a single tuple.

I can revert it, I don't think tests pass or fail either way. We really need a better set of performance vs just correctness tests.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: yes. The thing is, writes are side-effects. We only want to take that side effect once. So, if you return true, we should actually do the side effect. I'll update the comment.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -303,61 +303,61 @@ object CascadingBackend {
...
+        val force =
+          if (config.getHashJoinAutoForceRight) OptimizationRules.ForceToDiskBeforeHashJoin
+          else Rule.empty[TypedPipe]
+        std(force)
+    }
+  }
+
+  final def toPipe[U](p: TypedPipe[U], fieldNames: Fields)(implicit flowDef: FlowDef, mode: Mode, setter: TupleSetter[U]): Pipe = {
 
+    val phases = defaultOptimizationRules(
```
ianoc, line 337: If we are optimizing on toPipe does the optimization rules in the Execution have any effect?

ianoc:
lgtm modulo few comments and a final question

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
johnynek: we are careful to avoid calling `toPipe` in Execution. We call `toPipeUnoptimized`. This is because we want to do the optimization on the whole graph (multiple sinks) that Execution can see. This is likely to be better and allow more aggressive optimization than the per-pipe approach of `.write`.

johnynek:
Thanks for the review! I'll add the comments you wanted and if those look good, will merge (and on to the next item!).

I'm getting excited about scalding 0.18. I hope we don't hit any problems shipping it to twitter (I think we have tried to keep the API changes small).

Of course, we still do have the reducer estimation API change that twitter sent that we reverted in 0.17. Was that a big pain to change code? Can we do anything to make that easier for people to adopt?

ianoc:
@johnynek Nah its not that bad, i suspect most people don't have custom reducer estimators so its not likely to bleed too far from that POV. I had a branch to adapt some of Stripe's code for it. I think we've only 2 reducer estimators internally. So should be ok....

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
ianoc: Do we not hit toPipe going in and out of groupBy's?

johnynek:
not anymore. groupBy is just building an ADT, then we recursively apply the planning, and we never call toPipe in that recursion (you can double check).

ianoc:
mmm if we are breaking the API here should we mark the toPipe and from in TypedPipe as deprecated? (to be removed in scalding 0.19.0), and re-point at some cascading specific version of them?

johnynek:
Added #1745 so we can add configuration options without binary breaks.

About toPipe being deprecated, I definitely want to get there before 0.18 but I want to see how tractable it is to factor scalding-typed into a separate project, and I will try to do it there.

johnynek:
@ianoc how does this look? I think I addressed your concerns about comments. Also I added tests about the configuration of optimization rules to make sure they are triggered as expected in both the toPipe and Execution cases.

johnynek:
got verbal signoff from @ianoc

Issue #1734, opened 2017-10-06
remove use of anonymous function literals in TypedPipe/Grouped


johnynek:
TypedPipe is immutable and so it would be nice to treat it fully as a value. But our heavy use of anonymous literals in TypedPipe and Grouped.scala means if a user does something like:
```scala
TypedPipe.from(List(1, 2, 3)).groupAll == TypedPipe.from(List(1, 2, 3)).groupAll`
```
they get false. Which is probably okay for many cases, but means TypedPipes are not very useful in caches where we might cache that we have already computed something.

We could fix this by making case classes that extends `Function1` to use internally, so if the inputs of a function are the same, the typed pipe is the same.

This also helps people building systems on top of scalding that want this property.

See the TODOs in this PR: #1731

PR #1746, opened 2017-10-15
Replace function literals with case classes

johnynek:
this addresses #1734 

There are two reasons to use case classes for internal functions:

1. we preserve equality on case classes that wrap them.
2. we don't accidentally create a closure over something we don't want to serialize

@benpence @ttim @pankajroark @ianoc


ianoc:
lgtm

PR #1748, opened 2017-10-17
Add DateRange.prepend (opposite direction of extend)

FlavSF:


johnynek:
Looks good to me, but we are trying to increase test coverage. Can we add a test?

isnotinvain:
@FlavSF looks like https://github.com/twitter/scalding/blob/develop/scalding-date/src/test/scala/com/twitter/scalding/DateTest.scala is probably the right place to add the test

FlavSF:
Thanks both. I started working on unit tests but I'm facing some laptop issues. Give me a few days and I'll get it done.

johnynek (approved):
Thank you sir!

PR #1752, opened 2017-10-31
Extend TextLine with TypedSink

snoble:
https://github.com/twitter/scalding/issues/1751

attn @johnynek

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -523,13 +523,12 @@ object TextLine {
     new TextLine(p, sm, textEncoding)
 }
 
-class TextLine(p: String, override val sinkMode: SinkMode, override val textEncoding: String)(
-    implicit val tset: TupleSetter[String]
-  ) extends FixedPathSource(p) with TextLineScheme with TypedSink[String] {
+class TextLine(p: String, override val sinkMode: SinkMode, override val textEncoding: String) extends FixedPathSource(p) with TextLineScheme with TypedSink[String] {
   // For some Java interop
 
   def this(p: String) = this(p, TextLine.defaultSinkMode, TextLine.defaultTextEncoding)
-  override def setter[U <: String] = TupleSetter.asSubSetter[String, U](tset)
+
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -523,9 +524,12 @@ object TextLine {
     new TextLine(p, sm, textEncoding)
 }
 
-class TextLine(p: String, override val sinkMode: SinkMode, override val textEncoding: String) extends FixedPathSource(p) with TextLineScheme {
+class TextLine(p: String, override val sinkMode: SinkMode, override val textEncoding: String) extends FixedPathSource(p) with TextLineScheme with TypedSink[String] {
   // For some Java interop
+  implicit val tset: TupleSetter[String]
```
johnynek: does this need to be implicit? And can it be `private[this] val...`

johnynek:
can we add a test to https://github.com/twitter/scalding/blob/develop/scalding-core/src/test/scala/com/twitter/scalding/TypedPipeTest.scala. We should use `.runHadoop` I think to test it out.

snoble:
@johnynek test added

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -523,9 +523,13 @@ object TextLine {
     new TextLine(p, sm, textEncoding)
 }
 
-class TextLine(p: String, override val sinkMode: SinkMode, override val textEncoding: String) extends FixedPathSource(p) with TextLineScheme {
+class TextLine(p: String, override val sinkMode: SinkMode, override val textEncoding: String)(
+    implicit val tset: TupleSetter[String]
```
johnynek: can we remove this and instead do:
`override def setter[U <: String] = TupleSetter.asSubSetter[String, U](TupleSetter.of[String])` taking the implicit in the constructor changes the binary compatibility without a need that I see.

snoble: oh cool! kk

snoble:
I'll figure out why this test is failing tomorrow

johnynek:
Nice work @snoble ! Thanks for fixing this broken window!

PR #1755, opened 2017-12-11
introduce scalding-quotation sub-project

fwbrasil:
See https://github.com/twitter/scalding/pull/1754 for more details

### DIFF
--- a/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Liftables.scala
+++ b/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Liftables.scala
@@ -1,0 +1,41 @@
+package com.twitter.scalding.quotation
+
+import scala.reflect.macros.blackbox.Context
+
+trait Liftables {
+  val c: Context
+  import c.universe.{ TypeName => _, _ }
+
+  protected implicit val sourceLiftable: Liftable[Source] = Liftable {
+    case Source(path, line) => q"_root_.com.twitter.scalding.quotation.Source($path, $line)"
+  }
+
+  protected implicit val projectionsLiftable: Liftable[Projections] = Liftable {
+    case p => q"_root_.com.twitter.scalding.quotation.Projections(${p.set})"
+  }
+
+  protected implicit val typeNameLiftable: Liftable[TypeName] = Liftable {
+    case TypeName(name) => q"_root_.com.twitter.scalding.quotation.TypeName($name)"
+  }
+  
+  protected implicit val accessorLiftable: Liftable[Accessor] = Liftable {
+    case Accessor(name) => q"_root_.com.twitter.scalding.quotation.Accessor($name)"
+  }
+  
+  protected implicit val quotedLiftable: Liftable[Quoted] = Liftable {
+    case Quoted(source, call, fa) => q"_root_.com.twitter.scalding.quotation.Quoted($source, $call, $fa)"
+  }
+
+  protected implicit val projectionLiftable: Liftable[Projection] = Liftable {
+    case p: Property => q"$p"
+    case p: TypeReference => q"$p"
+  }
+
+  protected implicit val propertyLiftable: Liftable[Property] = Liftable {
+    case Property(path, accessor, tpe) => q"_root_.com.twitter.scalding.quotation.Property($path, $accessor, $tpe)"
+  }
+
+  protected implicit val typeReferenceLiftable: Liftable[TypeReference] = Liftable {
+    case TypeReference(name) => q"_root_.com.twitter.scalding.quotation.TypeReference($name)"
+  }
+}
--- a/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
+++ b/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
@@ -1,0 +1,146 @@
+package com.twitter.scalding.quotation
+
+import scala.annotation.tailrec
+
+case class Accessor(asString: String) extends AnyVal
+case class TypeName(asString: String) extends AnyVal
+
+sealed trait Projection {
+  def andThen(accessor: Accessor, typeName: TypeName): Projection =
+    Property(this, accessor, typeName)
+}
+
+/**
+ * A reference of a type. If not nested within a `Property`, it means that all fields are used.
+ */
+final case class TypeReference(typeName: TypeName) extends Projection {
+  override def toString = typeName.asString.split('.').last
+}
+
+/**
+ * A projection property (e.g. `Person.name`)
+ */
+final case class Property(path: Projection, accessor: Accessor, typeName: TypeName) extends Projection {
+  override def toString = s"$path.${accessor.asString}"
+}
+
+/**
+ * Utility class to deal with a collection of projections.
+ */
+final class Projections private (val set: Set[Projection]) extends Serializable {
+
+  /**
+   * Returns the projections that are based on `tpe` and limits projections
+   * to only properties that extend from `superClass`.
+   */
+  def of(typeName: TypeName, superClass: Class[_]): Projections = {
+
+    def byType(p: Projection) = {
+      @tailrec def loop(p: Projection): Boolean =
+        p match {
+          case TypeReference(`typeName`) => true
+          case TypeReference(_) => false
+          case Property(p, _, _) => loop(p)
+        }
+      loop(p)
+    }
+
+    def bySuperClass(p: Projection): Option[Projection] = {
+
+      def isSubclass(c: TypeName) =
+        try
+          superClass.isAssignableFrom(Class.forName(c.asString))
+        catch {
+          case _: ClassNotFoundException =>
+            false
+        }
+
+      def loop(p: Projection): Either[Projection, Option[Projection]] =
+        p match {
+          case TypeReference(tpe) =>
+            Either.cond(!isSubclass(tpe), None, p)
+          case p @ Property(path, name, tpe) =>
+            loop(path) match {
+              case Left(_) =>
+                Either.cond(!isSubclass(tpe), Some(p), p)
+              case Right(path) =>
+                Right(path)
+            }
+        }
+
+      loop(p) match {
+        case Left(path) => Some(path)
+        case Right(opt) => opt
+      }
+    }
+
+    Projections(set.filter(byType).flatMap(bySuperClass))
+  }
+
+  /**
+   * Given a set of base projections, returns the projections based on them.
+   *
+   * For instance, given a quoted function
+   * 	 `val contact = Quoted.function { (c: Contact) => c.contact }`
+   * and a call
+   *   `(p: Person) => contact(p.name)`
+   * returns the projection
+   *   `Person.name.contact`
+   */
+  def basedOn(base: Set[Projection]): Projections = {
+    def loop(base: Projection, p: Projection): Option[Projection] =
+      p match {
+        case TypeReference(tpe) =>
+          base match {
+            case TypeReference(`tpe`) => Some(base)
+            case Property(_, _, `tpe`) => Some(base)
+            case other => None
+          }
+        case Property(path, name, tpe) =>
+          loop(base, path).map(Property(_, name, tpe))
+      }
+    Projections {
+      set.flatMap { p =>
+        base.flatMap(loop(_, p))
+      }
+    }
+  }
+
+  def ++(p: Projections) =
+    Projections(set ++ p.set)
+
+  override def toString =
+    s"Projections(${set.mkString(", ")})"
+
+  override def equals(other: Any) =
+    other match {
+      case other: Projections => set == other.set
+      case other => false
+    }
+
+  override def hashCode =
+    31 * set.hashCode
+}
+
+object Projections {
+  val empty = apply(Set.empty)
+
+  /**
+   * Creates a normalized projections collection. For instance,
+   * given two projections `Person.contact` and `Person.contact.phone`,
+   * creates a collection with only `Person.contact`.
+   */
+  def apply(set: Set[Projection]) = {
+    @tailrec def isNested(p: Projection): Boolean =
+      p match {
+        case Property(path, acessor, property) =>
+          set.contains(path) || isNested(path)
+        case _ => 
+          false
+      }
+    new Projections(set.filter(!isNested(_)))
+  }
+
+  def flatten(list: Iterable[Projections]): Projections =
+    list.foldLeft(empty)(_ ++ _)
+}
--- a/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
+++ b/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
@@ -1,0 +1,106 @@
+package com.twitter.scalding.quotation
+
+import scala.reflect.macros.blackbox.Context
+
+trait ProjectionMacro extends TreeOps with Liftables {
+  val c: Context
+  import c.universe.{ TypeName => _, _ }
+
+  def projections(params: List[Tree]): Tree = {
+
+    def typeName(t: Tree) =
+      TypeName(t.symbol.typeSignature.typeSymbol.fullName)
+
+    def accessor(m: TermName) =
+      Accessor(m.decodedName.toString)
+
+    def typeReference(tpe: Type) =
+      TypeReference(TypeName(tpe.typeSymbol.fullName))
+
+    def isFunction(t: Tree) =
+      Option(t.symbol).map {
+        _.typeSignature
+          .erasure
+          .typeSymbol
+          .fullName
+          .contains("scala.Function")
+      }.getOrElse(false)
+
+    val nestedList =
+      params.flatMap {
+        case param @ q"(..$inputs) => $body" =>
+
+          val inputSymbols = inputs.map(_.symbol).toSet
+
+          object ProjectionExtractor {
+            def unapply(t: Tree): Option[Tree] =
+              t match {
+
+                case q"$v.$m(..$params)" => unapply(v)
+
+                case q"$v.$m" if t.symbol.isMethod =>
+
+                  if (inputSymbols.contains(v.symbol)) {
+                    val p =
+                      TypeReference(typeName(v))
+                        .andThen(accessor(m), typeName(t))
+                    Some(q"$p")
+                  } else
+                    unapply(v).map { n =>
+                      q"$n.andThen(${accessor(m)}, ${typeName(t)})"
+                    }
+
+                case t if inputSymbols.contains(t.symbol) =>
+                  Some(q"${TypeReference(typeName(t))}")
+
+                case _ => None
+              }
+          }
+
+          def functionCall(func: Tree, params: List[Tree]) = {
+            val paramProjections = params.flatMap(ProjectionExtractor.unapply)
+            q"""
+              $func match {
+                case f: _root_.com.twitter.scalding.quotation.QuotedFunction =>
+                  f.quoted.projections.basedOn($paramProjections.toSet)
+                case _ =>
+                  _root_.com.twitter.scalding.quotation.Projections(Set(..$paramProjections))
+              }
+            """
+          }
+
+          collect(body) {
+            case q"$func.apply[..$t](..$params)" =>
+              functionCall(func, params)
+            case q"$func(..$params)" if isFunction(func) =>
+              functionCall(func, params)
+            case t @ ProjectionExtractor(p) =>
+              q"_root_.com.twitter.scalding.quotation.Projections(Set($p))"
+          }
+
+        case func if isFunction(func) =>
+          val paramProjections =
+            func.symbol.typeSignature.typeArgs.dropRight(1)
+              .map(typeReference)
+          q"""
+            $func match {
+              case f: _root_.com.twitter.scalding.quotation.QuotedFunction =>
+                f.quoted.projections
+              case _ =>
+                _root_.com.twitter.scalding.quotation.Projections(Set(..$paramProjections))                
+            }
+          """ :: Nil
+
+        case method if method.symbol != null && method.symbol.isMethod =>
+          val paramRefs =
+            method.symbol.asMethod.paramLists.flatten
+              .map(param => typeReference(param.typeSignature))
+          q"${Projections(paramRefs.toSet)}" :: Nil
+
+        case other =>
+          Nil
+      }
+
+    q"_root_.com.twitter.scalding.quotation.Projections.flatten($nestedList)"
+  }
+}
--- a/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/QuotedMacro.scala
+++ b/scalding-quotation/src/main/scala/com/twitter/scalding/quotation/QuotedMacro.scala
@@ -1,0 +1,111 @@
+package com.twitter.scalding.quotation
+
+import language.experimental.macros
+import scala.reflect.macros.blackbox.Context
+import scala.reflect.internal.util.RangePosition
+import scala.reflect.internal.util.OffsetPosition
+import scala.reflect.macros.runtime.{ Context => ReflectContext }
+import java.io.File
+
+class QuotedMacro(val c: Context)
+  extends TreeOps
+  with TextMacro
+  with ProjectionMacro
+  with Liftables {
+  import c.universe._
+
+  def internal: Tree = quoted
+
+  def method: Tree = {
+    rejectScaldingSources
+    quoted
+  }
+
+  private def quoted: Tree =
+    quoted(
+      c.asInstanceOf[ReflectContext]
+        .callsiteTyper
+        .context
+        .tree
+        .asInstanceOf[Tree])
+
+  val QuotedCompanion = q"_root_.com.twitter.scalding.quotation.Quoted"
+
+  private def quoted(tree: Tree): Tree = {
+    val source = Source(tree.pos.source.path, tree.pos.line)
+
+    find(tree) { t =>
+      t.pos != NoPosition && t.pos.start <= c.enclosingPosition.start
+    }.flatMap { t =>
+      collect(t) {
+
+        // the start position of vals is wrong, so we workaround
+        case q"val $name = $body" => quoted(body)
+
+        case q"$m.method" if m.symbol.fullName == classOf[Quoted].getName =>
+          c.abort(
+            c.enclosingPosition,
+            "Quoted.method can be invoked only as an implicit parameter")
+
+        case tree @ q"$instance.$method[..$t]" =>
+          q"${Quoted(source, Some(callText(method, t)), Projections.empty)}"
+
+        case tree @ q"$instance.$method[..$t](...$params)" =>
+          q"""
+            $QuotedCompanion(
+              $source, 
+              Some(${callText(method, t ++ params.flatten)}), 
+              ${projections(params.flatten)})
+          """
+
+      }.headOption
+    }.getOrElse {
+      q"${Quoted(source, None, Projections.empty)}"
+    }
+  }
+
+  def function(f: Tree): Tree = {
+    val source = Source(f.pos.source.path, f.pos.line)
+    val text = paramsText(TermName("function"), f)
+    f match {
+      case q"(..$params) => $body" =>
+        c.untypecheck {
+          q"""
+            new ${f.tpe.finalResultType} with ${c.symbolOf[QuotedFunction]} {
+              override def apply(..$params) = $body
+              override def quoted = 
+                $QuotedCompanion(
+                  $source, 
+                  Some($text), 
+                  ${projections(f :: Nil)}
+                )
+            }
+          """
+        }
+      case _ =>
+        c.abort(f.pos, "Expected a function")
+    }
+  }
+
+  private def rejectScaldingSources = {
+
+    def whitelist =
+      Set("test", "example", "tutorial")
+        .exists(c.enclosingPosition.source.path.contains)
+
+    def isScalding(sym: Symbol): Boolean =
+      sym.fullName.startsWith("com.twitter.scalding") || {
+        sym.owner match {
+          case NoSymbol => false
+          case owner => isScalding(owner)
+        }
+      }
+
+    if (!whitelist && isScalding(c.internal.enclosingOwner))
+      c.abort(
+        c.enclosingPosition,
+        "The quotation must happen at the level of the user-facing API. Add an `implicit q: Quoted` to the enclosing method. " +
+          "If that's not possible and the transformation doesn't introduce projections, use Quoted.internal.")
+  }
+}
+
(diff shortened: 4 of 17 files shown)


scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
+package com.twitter.scalding.quotation
+
+sealed trait Projection {
+  def andThen(prop: String, tpe: String): Projection
```
johnynek: can we move the default implementation here? Looks like both cases have the same implementation.

johnynek: I also wonder if we should use some `AnyVal` wrappers here instead of `String`. Something like:
```scala
case class Accessor(asString: String) extends AnyVal
case class TypeName(asString: String) extends AnyVal
```

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionMacroTest.scala
```diff
@@ -0,0 +1,105 @@
...
+  val test = new TestClass
+
+  "no projection" in {
+    test.noProjection(42)._1.projections.set mustEqual Set.empty
+  }
+
+  "method with params isn't considered as projection" in {
+    test
+      .function[Person, String](_.name.substring(1))._1
+      .projections.set mustEqual Set(name)
```
johnynek: I'm confused what `name` is here? How is this compiling?

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/package.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import org.scalatest.MustMatchers
+import org.scalatest.FreeSpec
+
+package object quotation {
+  def cls[T](implicit ct: reflect.ClassTag[T]) = ct.runtimeClass.getName
+  trait Test extends FreeSpec with MustMatchers
+
+  val person = TypeReference(cls[Person])
+  val name = person.andThen("name", cls[String])
```
johnynek: ahh, here it is. Can we not do this and instead add an explicit import? It makes it harder to read the tests without a clear benefit if you ask me.

johnynek (commented):
still reading, but wanted to share some initial comments.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+final class Projections private (val set: Set[Projection]) extends Serializable {
+
+  /**
+   * Returns the projections that are based on `tpe` and limits projections
+   * to only properties that extend from `superClass`.
+   */
+  def of(tpe: String, superClass: Class[_]): Projections = {
+
+    def byType(p: Projection) = {
+      def loop(p: Projection): Boolean =
```
johnynek: can we add the `@annotation.tailrec` here?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+
+      def isSubclass(c: String) =
+        try
+          superClass.isAssignableFrom(Class.forName(c))
+        catch {
+          case _: ClassNotFoundException =>
+            false
+        }
+
+      def loop(p: Projection): Either[Projection, Option[Projection]] =
```
johnynek, line 58: can we add `@annotation.tailrec`?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+        p match {
+          case TypeReference(tpe) =>
+            Either.cond(!isSubclass(tpe), None, p)
+          case p @ Property(path, name, tpe) =>
+            loop(path) match {
+              case Left(path) =>
```
johnynek: can we use `Left(_) => ` here since we are not using `path`?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+      def loop(p: Projection): Boolean =
+        p match {
+          case TypeReference(`tpe`) => true
+          case TypeReference(_) => false
+          case Property(p, _, _) => loop(p)
+        }
+      loop(p)
+    }
+
+    def bySuperClass(p: Projection) = {
```
johnynek: can we put the return type? I think `Option[Projection]`?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+   *   `(p: Person) => contact(p.name)`
+   * returns the projection
+   *   `Person.name.contact`
+   */
+  def basedOn(base: Set[Projection]): Projections = {
+    def loop(base: Projection, p: Projection): Option[Projection] =
+      p match {
+        case TypeReference(tpe) =>
+          base match {
+            case TypeReference(`tpe`) => Some(p)
```
johnynek: `Some(p) == Some(base)` in this branch right? Can we write it that way so we can see in both branches we are returning `Some(base)` or `None`. In fact, maybe we should return Boolean instead so it is easier to follow and use filter below?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Creates a normalized projections collection. For instance,
+   * given two projections `Person.contact` and `Person.contact.phone`,
+   * creates a collection with only `Person.contact`.
+   */
+  def apply(set: Set[Projection]) = {
+    new Projections(
+      set.filter {
+        case Property(path, name, tpe) => !set.contains(path)
```
johnynek: not clear to me this algorithm works. It seems the order you add things in matters here. It seems you need to add shortest paths first and just going through the set may cause you to add a path that later you will not want in there.

Do you follow my point?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,143 @@
...
+        case _ => true
+      })
+  }
+
+  def flatten(list: List[Projections]): Projections =
```
johnynek: why not `Iterable` here since `++` is commutative right? We don't care about an particular order do we?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
johnynek: Also, the property you want to maintain, how are we maintaining it through `++`? Can't you have two different projections in two different sets that need to have one filtered after a `++`? Can you spell out the invariant above the class in the comments?

johnynek (commented):
some more comments.

PS: in my view, the idea PR is about 400 lines. This is better since it is smaller, but still quite large.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: this method is not tail recursive

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: `loop` here does more than just filter the projection. It reconstructs the projections on top of the base projections using the type as the matching criteria. Take a look at line 99, where it uses option.map to recreate the property on top of the new base.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: I'm not sure what you mean by ordering since it filters an unordered set based on the same complete unordered set, but indeed this algorithm has a bug. It doesn't consider nested intersections. For instance, `person.contact` and `person.contact.phone.number` would return both projections, which is wrong. I've just fixed it.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/package.scala
fwbrasil: I've moved the projections to the `Person` companion object

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: The invariant holds for `++` since it also uses the `apply` method. I'm not sure the invariant needs to be in the class comments since it's already in the apply method, but I don't feel strongly about it.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
johnynek: duh.  yep.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
johnynek: right. I missed that branch. But line 95 could be `Some(base)`  right (since we are checking equality)?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
```diff
@@ -0,0 +1,106 @@
...
+
+                case _ => None
+              }
+          }
+
+          def functionCall(func: Tree, params: List[Tree]) = {
+            val paramsProjecttions = params.flatMap(Projection.unapply)
+            q"""
+              $func match {
+                case f: com.twitter.scalding.quotation.QuotedFunction =>
```
johnynek: do we need to use `_root_.com` here? I thought we needed to.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
```diff
@@ -0,0 +1,106 @@
...
+          .contains("scala.Function")
+      }.getOrElse(false)
+
+    val nestedList =
+      params.flatMap {
+        case param @ q"(..$inputs) => $body" =>
+
+          val inputSymbols = inputs.map(_.symbol).toSet
+
+          object Projection {
```
johnynek: why is this an object rather than just an `def unapply(t: Tree): Option[Tree]` method? Can you comment?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
```diff
@@ -0,0 +1,106 @@
...
+
+                case t if inputSymbols.contains(t.symbol) =>
+                  Some(q"${TypeReference(typeName(t))}")
+
+                case _ => None
+              }
+          }
+
+          def functionCall(func: Tree, params: List[Tree]) = {
+            val paramsProjecttions = params.flatMap(Projection.unapply)
```
johnynek: I think `Projections` is spelled wrong.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
```diff
@@ -0,0 +1,106 @@
...
+              }
+            """
+          }
+
+          collect(body) {
+            case q"$func.apply[..$t](..$params)" =>
+              functionCall(func, params)
+            case q"$func(..$params)" if isFunction(func) =>
+              functionCall(func, params)
+            case t @ Projection(p) =>
```
johnynek: okay, I guess you are using unapply here. Can we just comment on the object.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/QuotedMacro.scala
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  private def rejectScaldingSources = {
+
+    def whitelist =
+      Set("test", "example", "tutorial")
+        .exists(c.enclosingPosition.source.path.contains)
+
+    def isScalding(sym: Symbol): Boolean =
+      sym.fullName.contains("com.twitter.scalding") || {
```
johnynek: don't we want `startsWith` here?

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionMacroTest.scala
```diff
@@ -0,0 +1,105 @@
...
+          .projections.set mustEqual Set(Person.nameProjection)
+      }
+    }
+  }
+
+  "function instance" - {
+    "non-quoted" in {
+      val function = (p: Person) => p.name
+      test.function[Person, String](function)._1
+        .projections.set mustEqual Set(Person.typeReference)
```
johnynek, line 82: in this case, does `typeReference` mean we are keeping the whole thing (when we see that, we can't do any projections)?

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionTest.scala
```diff
@@ -0,0 +1,168 @@
...
+        Projections.flatten(list).set mustEqual Set(t1)
+      }
+    }
+
+    "++" - {
+      "simple" in {
+        val p = Projections(Set(p1)) ++ Projections(Set(p2))
+        p.set mustEqual Set(p1, p2)
+      }
+      "with merge" in {
```
johnynek, line 92: seems like we could do a scalacheck test here. We could test that for all `List[Projections]` then if we `++` all of them, we have a final result where there are not Projections which are "suffixes" of another.

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionTest.scala
```diff
@@ -0,0 +1,168 @@
...
+        Projections.flatten(list).set mustEqual Set(t1)
+      }
+    }
+
+    "toString" - {
+      "empty" in {
+        Projections.empty.toString mustEqual "Projections()"
+      }
+      "non-empty" in {
+        Projections(Set(p1, p2)).toString mustEqual "Projections(T1.p1, T2.p2)"
```
johnynek, line 105: Another scalacheck might be that `Projections(someSet).toSet.size <= someSet.size`

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: Yes, it could. It doesn't seem to matter, though?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
fwbrasil: It's good to have it for hygiene. If the user has a `com` object it'll fail, for instance. Note that even `_root_` doesn't guarantee hygiene because that's a value that the user could define, but that should never happen :)

I'll add `_root_` to all trees

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
fwbrasil: I've renamed it to `ProjectionExtractor`. Is it clear enough to avoid the comment?

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionMacroTest.scala
fwbrasil: Yes, it must assume that all properties are used since the function isn't introspectable

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionTest.scala
fwbrasil: I'm not a big fan of property-based testing. Is there something that would have better coverage if I use it?

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionTest.scala
fwbrasil: ditto

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/QuotedMacro.scala
fwbrasil: good catch. Fixed

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
fwbrasil: fixed

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionTest.scala
johnynek: WHAT!?!? Who is not a big fan of property based checks!? Why don't you like it? I love it. It gives you an ability to clearly state properties and a decent approach at checking if what you claim is true actually is. Of course, with poor generators, it often doesn't hit tricky cases, but I have found it to be hugely helpful.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
johnynek: I think it is easier to read that both branches are returning the same thing. I feel that is a value to someone understanding the algorithm.

fwbrasil: Makes sense. I've changed it

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/ProjectionTest.scala
fwbrasil: If the code is refactored into simple functions that don't require complex structures, unit tests are easier to implement and reason about. Property-based tests take longer to implement, are much harder to debug, and don't bring many benefits if compared to simple unit tests.

johnynek: I guess the reason I brought it up is that I thought I noticed a bug which I think a scalacheck would have caught.

fwbrasil: Sure, if I had thought about that scenario and checked the property :) The same is valid for unit tests. I can implement the property-based test if you feel strongly about it, though.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/TreeOps.scala
```diff
@@ -0,0 +1,46 @@
...
+    t.traverse(tree)
+    res
+  }
+
+  /**
+   * Similar to tree.collect but it doesn't collect the children of a
+   * collected tree.
+   */
+  def collect[T](tree: Tree)(f: PartialFunction[Tree, T]): List[T] = {
+    var res = List[T]()
```
johnynek: what about using a `List.newBuilder[T]` and using `+=` on it which I think gives O(1) append (because they cheat and mutate the list). Currently I think this code is `O(N^2)` since it is using append on a list.

scalding-quotation/src/test/scala/com/twitter/scalding/quotation/LimitationsTest.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  class TestClass {
+    def function[T, U](f: T => U)(implicit q: Quoted) = (q, f)
+  }
+
+  val test = new TestClass
+
+  "nested transitive projection" in pendingUntilFixed {
+    test.function[Person, Option[String]](_.alternativeContact.map(_.phone))._1.projections.set mustEqual
+      Set(Person.typeReference.andThen(Accessor("alternativeContact"), typeName[Option[Contact]]).andThen(Accessor("phone"), typeName[String]))
```
johnynek, line 13: this confuses me. How do we distinguish an `andThen` projection inside or outside an `Option`? Seems like we need to record the input type we are dealing with also. We have `Option[Contact]` as the output of the previous Projection, but how do we jump across the `.map` on the Option?

fwbrasil: Note that this is a failing test with `pendingUntilFixed`. The projections in this scenario are `Person.alternativeContact` and `Contact.phone`, which are unrelated and thus don't get filtered out by `Projections`. Ideally, the macro should apply a transformation similar to beta reduction to produce only the projection `Person.alternativeContact.phone`.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/TreeOps.scala
fwbrasil: fixed

johnynek (approved):

fwbrasil:
@johnynek @benpence @dieu @ttim Could someone merge this? I don't have access. I'd like to move forward and submit the second PR.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Liftables.scala
```diff
@@ -0,0 +1,41 @@
+package com.twitter.scalding.quotation
+
+import scala.reflect.macros.blackbox.Context
+
+trait Liftables {
```
ttim, line 5: Can we add comment on what `Liftable` means?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,146 @@
...
+ * A reference of a type. If not nested within a `Property`, it means that all fields are used.
+ */
+final case class TypeReference(typeName: TypeName) extends Projection {
+  override def toString = typeName.asString.split('.').last
+}
+
+/**
+ * A projection property (e.g. `Person.name`)
+ */
+final case class Property(path: Projection, accessor: Accessor, typeName: TypeName) extends Projection {
```
ttim, line 23: `typeName` is type of this `Projection`? It sound like something which belongs to any `Projection`, does it make sense to put it on the level of `Projection` then?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
ttim: `typeName` in comment instead of `tpe`?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Utility class to deal with a collection of projections.
+ */
+final class Projections private (val set: Set[Projection]) extends Serializable {
+
+  /**
+   * Returns the projections that are based on `tpe` and limits projections
+   * to only properties that extend from `superClass`.
+   */
+  def of(typeName: TypeName, superClass: Class[_]): Projections = {
```
ttim, line 36: `of` sounds like method on companion object for me and I was reading this method this way. Maybe `filterBySuperClass`?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,146 @@
...
+
+    def byType(p: Projection) = {
```
ttim, line 38: Can we do `rootProjection(Projection): TypeReference` and match on it? I think logic will be clearer then.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+      loop(p) match {
+        case Left(path) => Some(path)
+        case Right(opt) => opt
+      }
+    }
+
+    Projections(set.filter(byType).flatMap(bySuperClass))
```
ttim, line 77: I read this couple of times and have issues with understanding what `bySuperClass` does. Can we add a comment on it?

ttim: And/or rename it and put on `Projection` itself?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
```diff
@@ -0,0 +1,146 @@
...
+   * Given a set of base projections, returns the projections based on them.
+   *
+   * For instance, given a quoted function
+   * 	 `val contact = Quoted.function { (c: Contact) => c.contact }`
+   * and a call
+   *   `(p: Person) => contact(p.name)`
+   * returns the projection
+   *   `Person.name.contact`
+   */
+  def basedOn(base: Set[Projection]): Projections = {
```
ttim, line 90: Clarifying for myself: we loose correspondence between function arguments and their `Projection`s and because of that we do this match between all `base` and all `set` projections?

Can we rename `loop(base, p)` and put it to `Projection` class?

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
```diff
@@ -0,0 +1,106 @@
...
+        _.typeSignature
+          .erasure
+          .typeSymbol
+          .fullName
+          .contains("scala.Function")
+      }.getOrElse(false)
+
+    val nestedList =
+      params.flatMap {
+        case param @ q"(..$inputs) => $body" =>
```
ttim, line 31: This is kinda hard to read because you can't easily see alternatives together, can we extract bodies into separate functions?

ttim (commented):
I guess the only real concern from me - `Projections` sounds more like `Map[ParamName, Set[Projections]]` and not like `Set[Projection]`. 

Feel free to address comments in subsequent reviews.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Liftables.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: I don't think so, there isn't immediate benefit in abstracting it and the `typeName` of a `TypeReference` doesn't really have the same meaning as the `typeName` of a `Property`.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: `filterBySuperClass` doesn't represent well what this method does. Would you have another suggestion? I find `of` a good name to be honest.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
fwbrasil: done

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/ProjectionMacro.scala
fwbrasil: done

fwbrasil:
@ttim I addressed your feedback on https://github.com/twitter/scalding/pull/1761

> I guess the only real concern from me - Projections sounds more like Map[ParamName, Set[Projections]] and not like Set[Projection].

We could use beta reded uction to deal with projections based on the actual parameters. It doesn't seem necessary, though. It would make the code much more complex without clear benefit since the projections are aggregated at the end and it doesn't really matter from which parameter they came from.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
ttim: I like `of`, but I perceive it as something on companion objection like `Projections.of(clazz)`.

scalding-quotation/src/main/scala/com/twitter/scalding/quotation/Projection.scala
ttim: My point more about `typeName` being part of any `Projection` in some way (I guess it's type of `Projection`). I definitely don't want to introduce it as an abstraction, more like a property which we have for any `Projection`. If you can see it as something non consistent between implementations I'm ok with it.

ttim:
> We could use beta reded uction to deal with projections based on the actual parameters. It doesn't seem necessary, though. It would make the code much more complex without clear benefit since the projections are aggregated at the end and it doesn't really matter from which parameter they came from.

I don't think it complicates a logic anyhow significantly, but it makes part where you combine projections between function call and function body more straightforward and correct.

fwbrasil:
> I don't think it complicates a logic anyhow significantly, but it makes part where you combine projections between function call and function body more straightforward and correct.

The projection logic is based on type references right now. In order to do detect which projection is from which parameter, we'd need to introduce a logic similar to beta reduction. I don't see how it'd be more correct or precise than the current implementation.

PR #1756, opened 2017-12-14
Make Config an abstract class

### LINKED ISSUE #1745: make Config an abstract class for better binary compatibility
Traits are very sensitive to binary incompatibility.

alexeygorobets:
closes [#1745](https://github.com/twitter/scalding/issues/1745)


johnynek:
👍

PR #1757, opened 2017-12-14
 introduce modes for required ordered serialization

fwbrasil:
## Problem

We're planning to roll out `OrderedSerialization` to more jobs. The initial plan was to add the `RequiredBinaryComparators` traits to our execution and job classes, but jobs could fail at runtime if `TypedPipe` transformations are done outside the job/execution body since the implicit macro wouldn't be available. For instance, we recommend people to extract transformations in companion objects for testability.

## Solution

Introduce a mechanism that allows us to add the traits without failing at runtime if a transformation doesn't have the ordered serialization. By overriding `requireOrderedSerializationMode`, we can set the mode to `Log`, which will produce an error message at runtime.

## Notes

- I've enhanced the error message so users know how to fix the problem without having to ask us or look at the FAQ
- I"ve kept backward compatibility, but I'm not sure if it's necessary

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
@@ -23,6 +23,7 @@ import com.twitter.scalding.ExecutionContext.getDesc
 import java.io.InputStream
 import java.util.Comparator
 import scala.util.{ Failure, Success, Try }
+import org.slf4j.LoggerFactory
 
 /**
  * This is the type that should be fed to cascading to enable binary comparators
@@ -40,11 +41,13 @@ class CascadingBinaryComparator[T](ob: OrderedSerialization[T]) extends Comparat
 
 object CascadingBinaryComparator {
 
+  private val LOG = LoggerFactory.getLogger(this.getClass)
+
   /**
    * This method will walk the flowDef and make sure all the
    * groupBy/cogroups are using a CascadingBinaryComparator
    */
-  private[scalding] def checkForOrderedSerialization[T](flow: Flow[T]): Try[Unit] = {
+  private[scalding] def checkForOrderedSerialization[T](flow: Flow[T], mode: RequireOrderedSerializationMode): Try[Unit] = {
     import collection.JavaConverters._
     import cascading.pipe._
     import com.twitter.scalding.RichPipe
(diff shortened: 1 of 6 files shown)


fwbrasil:
cc/ @johnynek

scalding-core/src/test/scala/com/twitter/scalding/typed/RequireOrderedSerializationTest.scala
```diff
@@ -18,24 +18,25 @@ package com.twitter.scalding
 import com.twitter.scalding.serialization.CascadingBinaryComparator
 import com.twitter.scalding.serialization.OrderedSerialization
 import com.twitter.scalding.serialization.StringOrderedSerialization
+import com.twitter.scalding.serialization.RequireOrderedSerializationMode
 
 import org.scalatest.{ Matchers, WordSpec }
 
-class NoOrderdSerJob(args: Args) extends Job(args) {
+class NoOrderdSerJob(args: Args, mode: String) extends Job(args) {
```
johnynek: mode is confusing here since it is almost always `scalding.Mode` named that. Can we rename?

scalding-core/src/main/scala/com/twitter/scalding/serialization/CascadingBinaryComparator.scala
```diff
@@ -53,8 +56,17 @@ object CascadingBinaryComparator {
...
-    def failure(s: String): Try[Unit] =
-      Failure(new RuntimeException("Cannot verify OrderedSerialization: " + s))
+    def failure(s: String): Try[Unit] = {
+      val message = 
+        s"Cannot verify OrderedSerialization: $s. Add `import com.twitter.scalding.serialization.RequiredBinaryComparators._`"
+      mode match {
+        case RequireOrderedSerializationMode.Fail =>
+          Failure(new RuntimeException(message))
+        case RequireOrderedSerializationMode.Log =>
+          LOG.error(message)
```
johnynek: should this be an error or warning? I kind of feel warning.

johnynek (approved):
some minor nits. Take or leave them.

fwbrasil:
@johnynek I've just addressed your comments. Thank you for the quick review!

scalding-core/src/main/scala/com/twitter/scalding/serialization/RequiredBinaryComparatorsConfig.scala
```diff
@@ -2,6 +2,13 @@ package com.twitter.scalding.serialization
...
+sealed trait RequireOrderedSerializationMode
+object RequireOrderedSerializationMode {
+  case object Fail extends RequireOrderedSerializationMode
+  case object Log extends RequireOrderedSerializationMode
+}
+
 trait RequiredBinaryComparatorsConfig extends Job {
-  override def config = super.config + (Config.ScaldingRequireOrderedSerialization -> "true")
+  def requireOrderedSerializationMode: RequireOrderedSerializationMode = RequireOrderedSerializationMode.Fail
+  override def config = super.config + (Config.ScaldingRequireOrderedSerialization -> requireOrderedSerializationMode.toString)
```
ulyssepence, line 13: What's the `.toString` supposed to output here? Aren't we matching on lowercase in `getRequireOrderedSerializationMode ` above?

johnynek:
👍

PR #1758, opened 2017-12-16
Check for success files based on configuration flag

moulimukherjee:
Adding a configuration flag, based on which the `pathisGood` method would check for `_SUCCESS` file.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
@@ -229,7 +229,13 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
    * TODO: consider writing a more in-depth version of this method in [[TimePathedSource]] that looks for
    * TODO: missing days / hours etc.
    */
-  protected def pathIsGood(globPattern: String, conf: Configuration) = FileSource.globHasNonHiddenPaths(globPattern, conf)
+  protected def pathIsGood(globPattern: String, conf: Configuration) = {
+    if (conf.getBoolean("scalding.require_success_file", false)) {
+      FileSource.allGlobFilesWithSuccess(globPattern, conf, true)
+    } else {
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
```diff
@@ -229,7 +229,13 @@ abstract class FileSource extends SchemedSource with LocalSourceOverride with Hf
    * TODO: consider writing a more in-depth version of this method in [[TimePathedSource]] that looks for
    * TODO: missing days / hours etc.
    */
-  protected def pathIsGood(globPattern: String, conf: Configuration) = FileSource.globHasNonHiddenPaths(globPattern, conf)
+  protected def pathIsGood(globPattern: String, conf: Configuration) = {
+    if (conf.getBoolean("scalding.require_success_file", false)) {
```
oscar-stripe, line 233: can we add this setting to `Config.scala` as well? `.setRequireSuccessFile(value: Boolean): Config` or something to make it easy also for people to programmatically set the option?

oscar-stripe: actually, not. Let's leave it as it is so we don't cause any binary incompatibility by cherry picking this for scalding 0.17.4

moulimukherjee: Ack 👍. Adding it.

oscar-stripe:
👍 
Looks good.

If this is green I'll publish a 0.17.4 with this change so we can roll it out with minimal work at Stripe.

scalding-core/src/main/scala/com/twitter/scalding/FileSource.scala
moulimukherjee: Okay.

johnynek:
@ianoc @pankajroark can I get +1 to make a 0.17.4 that has this PR?

PR #1759, opened 2017-12-19
Add a setting to skip null counters

### LINKED ISSUE #1732: Add an option to make jobs not fail is counters are null
We merged a fix to ignore when a counter reporter is null

#1726 

But this means counters are lossy in principle. When making the counter, we can see the jobconf so we can see a key like `ignoreNullCounterReporter` or something. So users who hit the error and want to hack around can set to ignore.

By default, we should probably not silently ignore these errors but we did publish a version to quickly get someone unblocked.

johnynek:
We have observed NPEs for null counters at least two companies
using scalding. We have not root-caused this issue or found
a good fix, but previously set to ignore all null counters.

Since some people rely on counters this is not a great plan,
so instead this patch makes ignoring an opt-in behavior.

closes #1732 

related to #1716 and #1726 

cc @tdyas @rstewart this will change the default to error on null again. When you upgrade you will need to opt in to silently dropping counters using `scalding.counters.skipnull=true`

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Config.scala
@@ -432,6 +432,19 @@ abstract class Config extends Serializable {
   def setVerboseFileSourceLogging(b: Boolean): Config =
     this + (VerboseFileSourceLoggingKey -> b.toString)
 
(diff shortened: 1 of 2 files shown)


ianoc:
lgtm

PR #1763, opened 2018-01-08
add quotation to external APIs

fwbrasil:
This PR is the second part of https://github.com/twitter/scalding/pull/1754. It adds the `Quoted` implicit param to the user-facing APIs. After this change, the last PR will use the projection information to do automatic pushdown.


fwbrasil:
@ttim @benpence @dieu @johnynek this is ready for review

ttim (commented):
In general this looks good but I have a suggestion about how `Quoted` works in general.

I think right now the logic with passing quoted explicitly / implicitly (and `Quoted.internal`) is a bit fuzzy. Can we do instead something similar to `cause` in exceptions? 

Let's make `Quoted` nested and basically putting `quoted` on method declaration means you want to pass information about this call site. And then if you have quoted parameter in the scope & apply quoted macro - put quoted from the context as `.cause` call site information.

I don't think we really need `Quoted.internal` in this case and overall logic seems more clear to me (quoted parameter - want to save information about call site).

Also `Quoted` become more or less call site information  in this case (which maybe is a better name?).

As cons for that I can see is - you need to somehow encode how different properties (such as used field projection) propagate through `cause`. But I think it actually represents what we need.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -501,7 +505,7 @@ final case class ValueSortedReduce[K, V1, V2](
    * After sorting, then reducing, there is no chance
    * to operate in the mappers. Just call take.
    */
-  override def bufferedTake(n: Int) = take(n)
+  override def bufferedTake(n: Int)(implicit q: Quoted) = take(n)
```
johnynek, line 781: why do we need a `Quoted` here. Not clear to me. Just to get position information of the caller?

scalding-core/src/test/scala/com/twitter/scalding/typed/NoStackLineNumberTest.scala
```diff
@@ -1,51 +0,0 @@
...
-import scala.concurrent.{ ExecutionContext => SExecutionContext, _ }
-import SExecutionContext.Implicits.global
-import scala.concurrent.duration.{ Duration => SDuration }
-
-import cascading.flow.FlowDef
-import org.apache.hadoop.conf.Configuration
-
-class NoStackLineNumberTest extends WordSpec {
-
-  "No Stack Shouldn't block getting line number info" should {
```
johnynek, line 30: can we replace this with an equivalent test that we can see the caller correctly?

johnynek (commented):
This is both exciting, but a bit scary. This touches virtually every public API and make everything binary incompatible.

In exchange for that, it is not yet clear what we are getting. I'd like to see maybe a quoted branch where we can use the final product with say Parquet to do some filter pushdown or projection pushdown.

If we don't have that as a killer app, I'm nervous about merging and getting to it later.

What do you think about making a quoted branch to make this PR into and when we have filter/projection pushdown working, we merge into develop?

fwbrasil:
> This is both exciting, but a bit scary. This touches virtually every public API and make everything binary incompatible.

Yeah, it's an intrusive change. I thought it was clear that it'd be necessary when we discussed it, though.



> In exchange for that, it is not yet clear what we are getting. I'd like to see maybe a quoted branch where we can use the final product with say Parquet to do some filter pushdown or projection pushdown.
> 
> If we don't have that as a killer app, I'm nervous about merging and getting to it later.
> 
> What do you think about making a quoted branch to make this PR into and when we have filter/projection pushdown working, we merge into develop?

That's https://github.com/twitter/scalding/pull/1754. These pull requests are just a break down of that initial one. Note that only projection pushdown is in the scope of this change.

fwbrasil:
> I think right now the logic with passing quoted explicitly / implicitly (and Quoted.internal) is a bit fuzzy. Can we do instead something similar to cause in exceptions?
> 
> Let's make Quoted nested and basically putting quoted on method declaration means you want to pass information about this call site. And then if you have quoted parameter in the scope & apply quoted macro - put quoted from the context as .cause call site information.

@ttim I can't see the benefit of this proposal. The quotation propagation is just a way to make the initial information available in the final `TypedPipe` composition. Could you give more details?

> I don't think we really need Quoted.internal in this case and overall logic seems more clear to me (quoted parameter - want to save information about call site).

`Quoted.internal` is just for safety since people might forget to propagate the quotation when making chanegs to Scalding. See the "Quoted propagation" section on https://github.com/twitter/scalding/pull/1754

scalding-core/src/main/scala/com/twitter/scalding/typed/GeneratedFlattenGroup.scala
```diff
@@ -17,7 +19,7 @@ object FlattenGroup {
   }
 
   class FlattenLeftJoin3[KEY, KLL[KLL_K, +KLL_V] <: KeyedListLike[KLL_K, KLL_V, KLL], A, B, C](nested: KLL[KEY, ((A, B), C)]) {
-    def flattenValueTuple: KLL[KEY, (A, B, C)] = nested.mapValues { tup => FlattenGroup.flattenNestedTuple(tup) }
+    def flattenValueTuple(implicit q: Quoted): KLL[KEY, (A, B, C)] = nested.mapValues { tup => FlattenGroup.flattenNestedTuple(tup) }
```
johnynek, line 22: I'm a bit confused why we need the implicits here. Can you comment? Can't we statically see anything we need for the purposes of projection? Is this just to collect line numbers?

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -195,14 +199,14 @@ sealed trait CoGrouped[K, +R] extends KeyedListLike[K, R, CoGrouped]
    * but it is not clear how to generalize that for general cogrouping functions.
    * For now, just do a normal take.
    */
-  override def bufferedTake(n: Int): CoGrouped[K, R] =
+  override def bufferedTake(n: Int)(implicit q: Quoted): CoGrouped[K, R] =
```
johnynek, line 278: I'd rather not take an implicit here if it can't contribute to row/column filtering, which I can't see how it can in the case of take.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -272,16 +276,16 @@ object Grouped {
  * of each key in memory on the reducer.
  */
 sealed trait Sortable[+T, +Sorted[+_]] {
-  def withSortOrdering[U >: T](so: Ordering[U]): Sorted[T]
+  def withSortOrdering[U >: T](so: Ordering[U])(implicit q: Quoted): Sorted[T]
```
johnynek, line 378: we can't project/filter here right since we know sorting is 1 in 1 out.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -406,20 +410,20 @@ final case class UnsortedIdentityReduce[K, V1](
...
   }
 
   // It would be nice to return IdentityReduce here, but
   // the type constraints prevent it currently
-  override def mapValues[V2](fn: V1 => V2) =
+  override def mapValues[V2](fn: V1 => V2)(implicit q: Quoted) =
     UnsortedIdentityReduce(keyOrdering, mapped.mapValues(fn), reducers, descriptions)
 
-  override def sum[U >: V1](implicit sg: Semigroup[U]) = {
+  override def sum[U >: V1](implicit sg: Semigroup[U], q: Quoted) = {
```
johnynek: since semigroups emit the whole object, do we really need the quoted here? Seems like we can assume there can be no projection/filtering here.

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -134,50 +135,50 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
...
-  def forceToReducers: This[K, T] =
+  def forceToReducers(implicit q: Quoted): This[K, T] =
     mapValueStream(Identity())
 
   /**
    * Use this to get the first value encountered.
    * prefer this to take(1).
    */
-  def head: This[K, T] = sum(HeadSemigroup[T]())
+  def head(implicit q: Quoted): This[K, T] = sum(HeadSemigroup[T](), q)
```
johnynek, line 161: we can't really filter here, can we? why to we need quoted?

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -134,50 +135,50 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
...
-  def flattenValues[U](implicit ev: T <:< TraversableOnce[U]): This[K, U] =
+  def flattenValues[U](implicit ev: T <:< TraversableOnce[U], q: Quoted): This[K, U] =
     flatMapValues(Widen(SubTypes.fromEv(ev)))
 
   /**
    * This is just short hand for mapValueStream(identity), it makes sure the
    * planner sees that you want to force a shuffle. For expert tuning
    */
-  def forceToReducers: This[K, T] =
+  def forceToReducers(implicit q: Quoted): This[K, T] =
```
johnynek, line 154: can't we avoid quoted here?

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -123,9 +124,9 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
   /**
    * Use Algebird Aggregator to do the reduction
    */
-  def aggregate[B, C](agg: Aggregator[T, B, C]): This[K, C] =
+  def aggregate[B, C](agg: Aggregator[T, B, C])(implicit q: Quoted): This[K, C] =
```
johnynek, line 129: I don't think we can look inside the aggregrator and filter, can we? Can we really use the quoted meaningfully?

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -189,102 +190,102 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
...
   /** For each key, check to see if a predicate is true for all Values*/
-  def forall(fn: T => Boolean): This[K, Boolean] =
+  def forall(fn: T => Boolean)(implicit q: Quoted): This[K, Boolean] =
     mapValues(fn).product
 
   /**
    * For each key, selects all elements except first n ones.
    */
-  def drop(n: Int): This[K, T] =
+  def drop(n: Int)(implicit q: Quoted): This[K, T] =
```
johnynek, line 246: we can't filter here, can we?

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -189,102 +190,102 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
...
+  def takeWhile(p: T => Boolean)(implicit q: Quoted): This[K, T] =
     mapValueStream(TakeWhile(p))
 
   /**
    * Folds are composable aggregations that make one pass over the data.
    * If you need to do several custom folds over the same data, use Fold.join
    * and this method
    */
-  def fold[V](f: Fold[T, V]): This[K, V] =
+  def fold[V](f: Fold[T, V])(implicit q: Quoted): This[K, V] =
```
johnynek, line 274: can we look inside the Fold?

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
```diff
@@ -293,31 +294,31 @@ trait KeyedListLike[K, +T, +This[K, +T] <: KeyedListLike[K, T, This]]
...
-  def distinctSize: This[K, Long] =
+  def distinctSize(implicit q: Quoted): This[K, Long] =
     toSet[T].mapValues(SizeOfSet())
 
   /**
    * For each key, remove duplicate values. WARNING: May OOM.
    * This assumes the values for each key can fit in memory.
    */
-  def distinctValues: This[K, T] = toSet[T].flattenValues
+  def distinctValues(implicit q: Quoted): This[K, T] = toSet[T].flattenValues
```
johnynek, line 322: we can't filter here, can we?

johnynek (commented):
Maybe I'm missing the point, but I'd love to minimize the callsites that have to change.

To me, that could be done by only accepting Quoted where we take a lambda and only where we expect we could in principle do some filter pushdown or projection pushdown. Does that make sense?

Why else do we want the quoted?

fwbrasil:
> To me, that could be done by only accepting Quoted where we take a lambda and only where we expect we could in principle do some filter pushdown or projection pushdown. Does that make sense?
> 
> Why else do we want the quoted?

@johnynek take a look at the description on https://github.com/twitter/scalding/pull/1754. We'd also like to use `Quoted` to provide a better description of the job to users, which requires adding `Quoted` to all user-facing APIs.

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -45,7 +46,7 @@ case class Sketched[K, V](pipe: TypedPipe[(K, V)],
   lazy val sketch: TypedPipe[CMS[Bytes]] = {
     // every 10k items, compact into a CMS to prevent very slow mappers
     lazy implicit val batchedSG: com.twitter.algebird.Semigroup[Batched[CMS[Bytes]]] = Batched.compactingSemigroup[CMS[Bytes]](10000)
-
+    implicit val q: Quoted = Quoted.internal
```
johnynek: should we take a `Quoted` in the constructor actually? That should be created where we call `.sketch` which I think could improve reporting on internal methods.

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
```diff
@@ -84,6 +85,8 @@ case class SketchJoined[K: Ordering, V, V2, R](left: Sketched[K, V],
 
   def reducers = Some(numReducers)
 
+  private implicit val q: Quoted = Quoted.internal
```
johnynek: same comment: should we take `implicit q: Quoted` on the constructor?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -251,21 +249,21 @@ sealed trait TypedPipe[+T] extends Serializable {
...
   @annotation.implicitNotFound(msg = "For asKeys method to work, the type in TypedPipe must have an Ordering.")
-  def asKeys[U >: T](implicit ord: Ordering[U]): Grouped[U, Unit] =
+  def asKeys[U >: T](implicit ord: Ordering[U], m: Quoted): Grouped[U, Unit] =
     map(WithConstant(())).group
 
   /**
    * If T <:< U, then this is safe to treat as TypedPipe[U] due to covariance
    */
-  protected def raiseTo[U](implicit ev: T <:< U): TypedPipe[U] =
+  protected def raiseTo[U](implicit ev: T <:< U, m: Quoted): TypedPipe[U] =
```
johnynek: I don't think we are using `m` here are we? This is really just a typesafe cast.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -274,22 +272,22 @@ sealed trait TypedPipe[+T] extends Serializable {
...
+    TypedPipe.CrossValue(this, p).withQuoted
 
   /** prints the current pipe to stdout */
-  def debug: TypedPipe[T] =
-    TypedPipe.DebugPipe(this).withLine
+  def debug(implicit m: Quoted): TypedPipe[T] =
+    TypedPipe.DebugPipe(this).withQuoted
 
   /** adds a description to the pipe */
   def withDescription(description: String): TypedPipe[T] =
```
johnynek: why don't we want `Quoted` here. Can you add the comment why? Is it the assumption that users get full control of the note, and there is no need to add the line/file info that Quoted can give?

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -721,7 +719,8 @@ sealed trait TypedPipe[+T] extends Serializable {
     delta: Double = 0.01, //5 rows (= 5 hashes)
     seed: Int = 12345)(implicit ev: TypedPipe[T] <:< TypedPipe[(K, V)],
       serialization: K => Array[Byte],
-      ordering: Ordering[K]): Sketched[K, V] =
+      ordering: Ordering[K],
+      m: Quoted): Sketched[K, V] =
     Sketched(ev(this), reducers, delta, eps, seed)
```
johnynek: Can we pass `m` explicitly to `Sketched` so we can see that it is not discarded?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -25,9 +25,13 @@ import com.twitter.scalding.serialization.{
 }
 import java.util.WeakHashMap
 import scala.collection.mutable.{ Map => MMap }
+import org.apache.hadoop.mapred.JobConf
+import com.twitter.scalding.quotation.Quoted
 
 object CascadingBackend {
   import TypedPipe._
+  
+  private implicit val q: Quoted = Quoted.internal
```
johnynek: I'm nervous about this hanging out implicitly. Can we not use it implicitly and instead explicitly call it: `interalQuoted` where needed? I fear this will actually destroy our `Quoted` information when we are doing composing operations, and defeat the whole purpose.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -262,12 +266,12 @@ object CascadingBackend {
...
             CascadingPipe[u](pipe, sink.sinkFields, fd, conv)
-          case (WithDescriptionTypedPipe(input, descr, dedup), rec) =>
+          case (WithDescriptionTypedPipe(input, descr, dedup, quoted), rec) =>
 
             @annotation.tailrec
             def loop[A](t: TypedPipe[A], acc: List[(String, Boolean)]): (TypedPipe[A], List[(String, Boolean)]) =
               t match {
-                case WithDescriptionTypedPipe(i, desc, ded) =>
+                case WithDescriptionTypedPipe(i, desc, ded, quoted) =>
                   loop(i, (desc, ded) :: acc)
```
johnynek: don't we need to accumulate the `quoted`? Seems like we need a `Semigroup` on quoted (maybe a `Monoid`).

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
```diff
@@ -301,6 +302,7 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
 
   def plan[T](m: Memo, tp: TypedPipe[T]): (Memo, Op[T]) =
     m.plan(tp) {
+      implicit val q: Quoted = Quoted.internal
```
johnynek: similar to above, can we not keep dummy implicit Quoted around. I fear we won't know where we are actually sending them. I prefer to be explicit with the dummies (and add a comment to that effect).

johnynek (commented):
Okay.... It is scary, touch every API, but I guess we do it...

Have you tried to build twitter's Source against this change? i want to know what we are signing up for. Is there 100 places we need to change, or 5000?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
fwbrasil: @johnynek `desc` already has the description information from `quoted` (see `TypedPipe.withQuoted`). I think we don't need to accumulate `quoted` at this point since the purpose is to set the description only (no projections).

scalding-core/src/main/scala/com/twitter/scalding/typed/Sketched.scala
fwbrasil: fixed

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
fwbrasil: removed

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
fwbrasil: just updated the scaladoc

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
fwbrasil: fixed

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
fwbrasil: fixed

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
fwbrasil: fixed

fwbrasil:
@johnynek I'm still working on updating source to develop without this change. I'll test these changes after it

PR #1769, opened 2018-01-25
Make TypedPipe abstract class for better binary compatibility

johnynek:
in 2.11, every new method on a trait breaks binary compatibility. Before we release 0.18, we should make this change.


johnynek:
@piyushnarang what do you think?

or @fwbrasil @ianoc

ianoc:
Can't think of a good reason off hand this isn't a strictly better setup

lgtm

piyushnarang:
Yeah I think it sounds good to me as well. 👍

PR #1773, opened 2018-01-28
Move all TypedPipe keyed functions to Keyed

### LINKED ISSUE #1733: Proposal move keyed TypedPipe operations into an implicit class
Right now, we implement a number of keyed operations on typed pipes by accepting an `implicit ev: T <:< (K, V)`. This is hard for many users to understand and also complicates type inference.

A mostly-source (but not binary) compatible change would be to add to the TypedPipe companion:
```scala
implicit class KeyedTypedPipe[K, V](val kvpipe: TypedPipe[(K, V)]) extends AnyVal {
  def filterKeys(fn: K => Boolean): TypedPipe[(K, V)]
  def mapValues[V2](fn: V => V2): TypedPipe[(K, V2)] 
  def group: Grouped[K, V]
  def keys: TypedPipe[K]
  def values: TypedPipe[V]
  def sumByKey(implicit sg: Semigroup[V]): UnsortedGrouped[K, V]
  // etc..
}
```
Notice most of those methods no longer need type parameters. This breaks the type inference in half so scala only needs to prove there is a `(K, V)` in the `TypedPipe` to call these methods, then the types have been more constrained. Also, if we add this inside `object TypedPipe` users don't need to import anything, and intellij understands this (as far as I know).

If you have used scalding, you may know that `mapValues` is notorious for this. You usually have to specify currently three parameters or type annotate the function:
```scala
val kv: TypedPipe[(K, V)] = ...
val next = kv.mapValues[K, V, V1] { x => fn(x) }
val next1 = kv.mapValues { x: V1 => fn(x) }
```
if we make this change, users who specifically wrote the type parameters code won't compile since they will have the wrong number of parameters.

But I think it would be a nice thing from a usability standpoint overall.

What I propose is that we make a branch of scalding 0.17.3 that has this change as a test and Twitter and Stripe report how many call sites get broken by the change. If it looks manageable, and the teams don't mind fixing that in a single PR when they bump, then I propose we do this.

Thoughts?

@pankajroark @ttim @ianoc ?

johnynek:
closes #1733 

I'm very happy with this. Note that in the tests only 1 place did we have to change some type parameters. However, with this change type-inference on mapValues, flatMapValues and eitherValues should actually work correctly. That's a nice win.

It also removes a lot of junky `T <:< (K, V)` and finally it centralizes all the keyed methods in one place. I am more and more happy about this change and I think it is a real win.

cc @fwbrasil @avibryant @ianoc

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/KeyedList.scala
@@ -28,6 +28,9 @@ import com.twitter.scalding.typed.functions._
 object KeyedListLike {
   /** KeyedListLike items are implicitly convertable to TypedPipe */
   implicit def toTypedPipe[K, V, S[K, +V] <: KeyedListLike[K, V, S]](keyed: KeyedListLike[K, V, S]): TypedPipe[(K, V)] = keyed.toTypedPipe
+
+  implicit def toTypedPipeKeyed[K, V, S[K, +V] <: KeyedListLike[K, V, S]](keyed: KeyedListLike[K, V, S]): TypedPipe.Keyed[K, V] =
+    new TypedPipe.Keyed(keyed.toTypedPipe)
 }
 
 /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TDsl.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TDsl.scala
@@ -38,6 +38,12 @@ object TDsl extends Serializable with GeneratedTupleAdders {
 
   implicit def sourceToTypedPipe[T](src: TypedSource[T]): TypedPipe[T] =
     TypedPipe.from(src)
+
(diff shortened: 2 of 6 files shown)


ianoc:
This looks good to me -- should we have any concern about intellij users? i know usage of implicits can make them sad sometimes (those random `toTypedPipe` i see in codebases...).

Spark has this pattern for its keyed operations so might be able to ignore the ^^ given that its already pretty common?

I suspect somewhere in one of the main code bases something will get confused -- and as you had it'll require some changes if anyone has explicitly provided the implicit args to the group methods. But ¯\_(ツ)_/¯ don't think thats a huge deal.

I'm personally 👍  and bullish. No changes needed I can see. Not sure if twitter folks have any fud tho since I expect they will have manual changes in a few places

non:
👍  from me too

johnynek:
I've confirmed with @gkk-stripe our local intellij expert that this PR should not hamper intellij users.

Given that this will improve the type inference (since it forces scala to first fix `K, V`), I think this will be a usability win.

Issue #1736, opened 2017-10-06
Add more optimizations


johnynek:
We have not turned on filterKeys push-down yet.

we could introduce a `Sum` node into the Grouped/CoGrouped AST which allows us to see that the `k,v` has a single item. This could allow us to be more efficient in joins with a summed value. Users could then use `.head` to signal that something has a single item, or something like `.singleValueAssert` which is like `.head` but throws if there are two items, to communicate which side of a join has a single key.

We can push `TypedPipe.mapValues/flatMapValues` up into a Grouped/CoGrouped to avoid Tuple boxing/unboxing.

If we have a pipe that fans out after a flatMap, it might be better convert it to a filter + flatMap (isEmpty  + nonEmpty parts) then apply the flatMap on each of the readers rather than materializing a potentially large explosion and the reading it twice. This is probably not an "always on" type optimization and one users might use to transform their own graphs.

`a ++ a == a.flatMap { x => List(x, x) }` and more generally we can do this for an entire tree of merges. The most general version of this might be called tree-de-diamonding where we find the minimal distinct inputs to a merge DAG and then execute the diamond portions in mappers rather than forking and merging.

In cascading (and many platforms) you can do a free merge before a groupBy operation. We could look in our planner if we have a merge just before the groupBy and not plan the merge, but get all the pipes and directly pass them to cascading. This is a small effect in cascading 2 (mostly skipping some Tuple boxing) but in cascading 3 on tez is a big deal.

Issue #1669, opened 2017-04-20
optimizations we could do with a full AST at plan time


johnynek:
If we merge #1666 and continue that with putting Grouped, CoGrouped, and HashJoinable in the AST, we could do a number of optimizations, fairly easily if we steal the summingbird graph optimizer code that does not actually depend on summingbird. Some examples:

- [x] Push up mapValues/flatMapValues after a toTypedPipe on a group/cogroup to avoid cascading tuple boxing
- [x] Push up filterKeys after a toTypedPipe on a group to before the shuffle.
- [ ] Push a flatMapValues into the joiner on a hashJoin on the right so we don't expand it before we broadcast.
- [x] Make sure to force a hashJoin to disk at a good spot (after a filter, but before a flatMap). We do this now, but in a pretty brittle way.
- [ ] in map-reduce, serialization is expensive, not so much computation to a good approximation, so we can convert a flatMap to a filter + flatMap where the filter is checking if the flatMap has 1 or more outputs. We can use this to insert filters before communication barriers to hopefully reduce the number of items communicated.
- [x] Push up filterKeys to both sides of a hashJoin/cogroup/group (may be done now in all cases, but need to make sure we don't regress)
- [ ] Remove cascading Merge nodes before GroupBy (helps with tez, simplifies the cascading graph).
- [ ] distribute joins: `a.join(b ++ c) == a.join(b) ++ a.join(c)` and `(a ++ b).join(c) == (a.join(c) ++ b.join(c))`, so we should be able to do that in 1 map-reduce step.

johnynek:
pretty much a dup of #1736

johnynek:
Another interesting rule is the following
```scala
val p1 = p.filter(fn1)
val p2 = p.filter(fn2)
```
if `p1` and `p2` have no other children and don't merge back into a single mapper (via de-diamonding), then we might want to go from `p -> p.filter { x => fn1(x) || fn(x) }` since function application is cheap, then we can make sure we don't checkpoint a giant data set just to filter it out downstream.

johnynek:
Another rule: `a.join(b).toTypedPipe.join(c)` could be `a.join(b).join(c)` and make sure we stay in 1 map-reduce job.

also: `a.join(b.join(c).toTypedPipe) == a.join(b.join(c))`

This may be a big usability win so extra `.toTypedPipe` calls don't change the efficiency of jobs.

johnynek:
[...]

PR #1774, opened 2018-01-28
Add back some optimizations we had in 0.17

johnynek:
This is related to #1736 and #1669 

This turns back on two optimizations we had before:

1. do filters on Iterable pipes before sending to cascading.
2. defer merges so we can combine more operations into single cascading nodes.

We are still lacking the `a ++ a == a.flatMap { v => List(v, v) }` which can avoid some cases of allocating cascading Merge nodes at all. It is not clear how useful that optimization actually is in practice however.

cc @fwbrasil maybe this fixes some of the issues you saw with different behavior in 0.18.

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
@@ -8,7 +8,7 @@ import org.apache.hadoop.conf.Configuration
 import com.twitter.scalding.{ Config, ExecutionContext, Local, Hdfs, FlowState, FlowStateMap, IterableSource }
 import com.twitter.scalding.typed.cascading_backend.CascadingBackend
 import org.scalatest.FunSuite
-import org.scalatest.prop.PropertyChecks.forAll
+import org.scalatest.prop.PropertyChecks
 import org.scalatest.prop.GeneratorDrivenPropertyChecks.PropertyCheckConfiguration
 import org.scalacheck.{ Arbitrary, Gen }
 import scala.util.{ Failure, Success, Try }
@@ -21,7 +21,6 @@ object TypedPipeGen {
   }
 
   def mapped(srcGen: Gen[TypedPipe[Int]]): Gen[TypedPipe[Int]] = {
-    val mappedRec = Gen.lzy(mapped(srcGen))
     val next1: Gen[TypedPipe[Int] => TypedPipe[Int]] =
       Gen.oneOf(
         tpGen(srcGen).map { p: TypedPipe[Int] =>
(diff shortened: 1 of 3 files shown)


scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -30,7 +29,7 @@ object TypedPipeGen {
         tpGen(srcGen).map { p: TypedPipe[Int] =>
           { x: TypedPipe[Int] => x.cross(ValuePipe(2)).values }
         },
-        Gen.const({ t: TypedPipe[Int] => t.debug }),
+        //Gen.const({ t: TypedPipe[Int] => t.debug }), debug spews a lot to the terminal
```
ianoc, line 32: should we leave this commented out/enabled ? (just don't know if we have any concerns over debug's correctness, we won't test it like this. But that might be fine)

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -234,13 +240,13 @@ class OptimizationRulesTest extends FunSuite {
 
   test("all optimization rules don't change results") {
     import TypedPipeGen.{ genWithIterableSources, genRule }
-    implicit val generatorDrivenConfig: PropertyCheckConfiguration = PropertyCheckConfiguration(minSuccessful = 100000)
+    implicit val generatorDrivenConfig: PropertyCheckConfiguration = PropertyCheckConfiguration(minSuccessful = 500)
```
ianoc, line 243: did you mean to drop these to 500 in the commit/merge ?

ianoc:
this looks good to me once CI is happy. 2 small questions just about commented/test changes that look like might have been for faster turn around locally?

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
johnynek: I pretty much hate scalatest. It was ignoring this parameter until I made this extend PropertyChecks. After I did this, 500 checks takes a while (200 seconds). I figured that would be enough for now.

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
johnynek: We almost never use debug in real life and it was making test logs unreadable.

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
ianoc: Yeah its a great exemplar of something with 2 many ways to do anything/everything and then its just too hard to do things right.

👍

PR #1778, opened 2018-01-31
Improve HashJoin related rules

johnynek:
These are rules I am using to try to get cascading 3 to work. Might as well have them in develop.

Also reduced the number of trials of our tests since they have been timing out on travis.

cc @erik-stripe @non


ianoc-stripe:
lgtm

PR #1780, opened 2018-02-01
Add a regression test (currently failing on cascading 3)

johnynek:
I want to make sure any time we find a graph that fails on cascading 3 we add a test for it in develop. There is some chance it could be a graph that also fails on cascading 2.

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/RegressionTests.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/RegressionTests.scala
@@ -1,0 +1,24 @@
+package com.twitter.scalding
+
+import org.scalatest.FunSuite
+
+class RegressionTests extends FunSuite {
+  test("hashJoins + merges that fail in cascading 3") {
+    val p1 =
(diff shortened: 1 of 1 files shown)


piyushnarang:
👍  Ruban, Cyrille and I added a few of these based on specific Cascading3 failures. I can try and consolidate them all here in a followup PR too.

johnynek:
Interestingly, this actually passes on the cascading3 branch with the standard scalding optimization rules turned on.

So, I assume it is the DeferMerge rule + existing other checks that make it work. Perhaps we *have* to run the DeferMerge for our current tricks to work?

johnynek:
I checked:
```
class Cascading3Bugfix extends OptimizationPhases {
  def phases = List(OptimizationRules.DeferMerge.orElse(OptimizationRules.DescribeLater),
    OptimizationRules.ForceToDiskBeforeHashJoin)
}
```
But that wasn't enough to fix the issue. So, it is some other of the full set of rewrites we do that it is saving it.

PR #1781, opened 2018-02-02
Give good equality to TupleConversion classes

### LINKED ISSUE #1739: Remove identity TypedPipe converter + identity + setter operations.
We can do this if we give single and double arity converters and setters good equality. This can avoid some boxing.

johnynek:
closes #1739 

This changes the generated tuple conversion code to be case classes. This gives us nice toString, but more importantly it makes it safe to skip some `.mapTo` operations in the CascadingBackend that are only reformatting the tuple.

There may be slightly more room for optimization, but this should be an improvement all around.


johnynek:
ptal @piyushnarang

ianoc:
looks great to me, even just to have better equality and toString are good reasons for the change. Perf optimization is nice too ofc.

lgtm

piyushnarang:
Not super familiar with this part of the code but from what I understand, looks good to me 👍

johnynek:
Thanks for the review folks!

PR #1784, opened 2018-02-04
Consolidate a scalding-safe Ordering cast

johnynek:
This removes 4 totally general casts, and replaces them with one highly constrained casting function which should be safe for all of scalding uses.

This removes the need to audit each of these casts and consolidates the explanation as to the safety in one place.


johnynek:
@non @erik-stripe another in a similar vein of cleaning up types. Can you review this?

johnynek:
The test failure here seems unrelated, but real. Filed #1786 to run it down. Yay for property checks! I wish I had made these TypedPipe generators sooner!

non:
LGTM 👍

Issue #1743, opened 2017-10-09
Consider adding a Sum node to the grouped and cogrouped AST


johnynek:
Sums are very common in scalding, but we hide them inside generic functions. By keeping them in the AST, we can write optimization rules and smarter translations, especially when we join a summed values.

PR #1785, opened 2018-02-04
Optimize joins when the right side is summed

johnynek:
Here we give a potentially very large performance optimization when joining a left side with many values against something that has 0 or 1 value. Here we can avoid recomputing the reduction on the right side, which in some cases will be pretty expensive when multiplied by the cardinality of the left.

This shows the value of our case class functions to allow us to take apart what we have done to make changes at plan time. This reduces, but not eliminates, the need to do #1743 since we should be able to see with extra work now where sums are occurring (as long as we are diligent in always making case class functions).


ianoc:
lgtm

PR #1789, opened 2018-02-07
Cache some hashCodes to prevent exponential complexity

johnynek:
When you have a hashCode on a type like `MergedTypePipe` it can create exponential cost hashCode.

Consider the Fibonacci typed pipe that merges the two previous pipes to get the next. For small numbers of recursion, this can totally blow up (80 seconds to compute hashCode when you are 45 deep). We actually have graphs that big at Stripe, so this is not academic.


non:
LGTM 👍

PR #1790, opened 2018-02-08
update to dagon 0.2.4

johnynek:
there was a bug in 0.2.3 that we need to not use.

cc @non @erik-stripe


johnynek:
this is kind of a no-brainer. So I'm merging.

PR #1791, opened 2018-02-08
Globally Optimize Job-based TypedPipe.write calls

johnynek:
Before this, in 0.18-develop we were calling the optimizer after each write. This not only makes a sub-optimal optimization, but also for long jobs with many writes at different stages makes optimization an O(N^2) operation (times how much it costs, which is significant, to optimize each set of TypedPipes).

This pattern comes up in summingbird, so it is a practical issue for us when trying to use scalding 0.18 with summingbird at Stripe.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -15,11 +15,13 @@ limitations under the License.
 */
 package com.twitter.scalding
 
-import com.twitter.algebird.Semigroup
 import cascading.flow.{ Flow, FlowDef, FlowListener, FlowStep, FlowStepListener, FlowSkipStrategy, FlowStepStrategy }
 import cascading.pipe.Pipe
 import cascading.property.AppProps
 import cascading.stats.CascadingStats
+
+import com.twitter.algebird.Semigroup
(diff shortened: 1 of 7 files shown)


ianoc-stripe:
looks like a pretty straightforward win

lgtm

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -251,6 +253,7 @@ class Job(val args: Args) extends FieldConversions with java.io.Serializable {
   // called before run
   // only override if you do not use flowDef
   def validate(): Unit = {
```
johnynek, line 255: this won't get called in the case where we write DOTs in Tool. Need to fix.

ianoc: :( i guess we lack a test for that path

johnynek: actually, I was wrong, we call `buildFlow`, which will make sure this stuff gets built, but we don't have a test for DOT generation, I don't think.

ianoc: Shame no test, but its not a critical path

PR #1792, opened 2018-02-09
Add ability to convert Job to Execution

### LINKED ISSUE #1744: Make a function to convert a Job to an Execution
I think we could have a function like:
```scala
def toExecution(c: Class[_ <: Job]): Execution[Unit]
```
which would work like a Job. If we had this, it could aid it converting more code to the execution API.

johnynek:
closes #1744 

This allows us to use existing jobs as Executions. It should work for the vast majority of Jobs using only the TypedPipe.

A more sophisticated version could also work for the Fields API work by making some changes to the Execution API (or using Execution.fromFn, which I want to remove, so I am reluctant to make the function when I assume the number of fields API jobs is decreasing).

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Job.scala
@@ -46,6 +46,86 @@ object Job {
       .newInstance(args)
       .asInstanceOf[Job]
   }
+
+  /**
+   * Make a job reflectively from the given class
+   * and the Args contained in the Config.
+   */
+  def makeJob[J <: Job](cls: Class[J]): Execution[J] =
+    Execution.getConfigMode.flatMap { case (conf, mode) =>
+      // Now we need to allocate the job
+      Execution.from {
+        val argsWithMode = Mode.putMode(mode, conf.getArgs)
+        cls.getConstructor(classOf[Args])
+          .newInstance(argsWithMode)
+      }
+    }
+
+  /**
+   * Create a job reflectively from a class, which handles threading
+   * through the Args and Mode correctly in the way Job subclasses expect
+   */
+  def toExecutionFromClass[J <: Job](cls: Class[J], onEmpty: Execution[Unit]): Execution[Unit] =
+    makeJob(cls).flatMap(toExecution(_, onEmpty))
+
+  /**
+   * Convert Jobs that only use the TypedPipe API to an Execution
+   *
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -46,6 +46,86 @@ object Job {
...
+              }
+            case Some(fs) =>
+              // we can't handle if there have been anything other than TypedPipe.write on the
+              // TypedPipe
+              Execution.failed(new Exception(s"expected empty FlowState other than TypedWrites, found: $fs"))
+            case None => onEmpty
+          }
+        }
+        else {
+          // We only support
```
avibryant, line 111: Nit: this comment seems to have gotten truncated (though it's clear from context)

scalding-core/src/main/scala/com/twitter/scalding/Job.scala
```diff
@@ -46,6 +46,86 @@ object Job {
...
+          // be updated
+          FlowStateMap.get(fd) match { // Note, we want this to be a pure function so we don't mutate the FlowStateMap
+            case Some(FlowState(srcs, confs, writes)) if srcs.isEmpty && confs.isEmpty =>
+              writes match {
+                case Nil => onEmpty
+                case nonEmpty =>
+                  def write[A](f: FlowStateMap.TypedWrite[A]): Execution[Unit] =
+                    f.pipe.writeExecution(f.sink)
+
+                  Execution.sequence(nonEmpty.map(write(_))).unit
```
avibryant, line 101: This is where I get a little bit lost - the rest of this looks great. I'm assuming this doesn't handle a case that would need to be a `writeThrough` in `Execution`? If so, I'm missing how, and if not, it's not clear to me what the failure would look like; maybe there could be a test for that?

johnynek: That is already handled by Executions:
```scala
pipe1.writeExecution(sink1)
  .zip(pipe1.map(fn).writeExecution(sink2))
```
will work. Internally the optimizer sees the fork and that is exposed to cascading (they share everything above the fork).

Really, this is just a trick to get things written into Execution form. You question would still apply to Execution if it were an issue.

avibryant (approved):

ianoc-stripe:
neet

lgtm

PR #1782, opened 2018-02-03
Make scalding native MemoryBackend usable

johnynek:
This finishes the scalding native memory platform into what appears to be a usable mode.

* add a test that our Memory mode agrees with cascading local mode
* remove some methods from Mode to CascadingMode to prepare for more non-cascading backends (e.g. spark)

PR #1793, opened 2018-02-09
Add an AnyVal adapter for removed Mode methods

johnynek:
add an enrichment to return some methods removed #1782 

@fwbrasil can you review?


fwbrasil:
Thank you! LGTM

Issue #1795, opened 2018-02-09
Not serializable function in `MapGroup`


fwbrasil:
I'm investigating this error in one of our e2e tests. The exception happens when running with the latest develop

```
Caused by: cascading.flow.planner.PlannerException: could not build flow from assembly: [Neither Java nor Kyro works for class: class com.twitter.scalding.typed.CoGrouped$MapGroup$$anonfun$joinFunction$2 instance: <function3>
export CHILL_EXTERNALIZER_DEBUG=true to see both stack traces]
	at cascading.flow.planner.FlowPlanner.handleExceptionDuringPlanning(FlowPlanner.java:578)
	at cascading.flow.hadoop.planner.HadoopPlanner.buildFlow(HadoopPlanner.java:286)
	at cascading.flow.hadoop.planner.HadoopPlanner.buildFlow(HadoopPlanner.java:80)
	at cascading.flow.FlowConnector.connect(FlowConnector.java:459)
	at com.twitter.scalding_internal.job.FileSourceWarningFlowConnector.connect(FileSourceWarningFlowConnector.scala:42)
	at com.twitter.scalding.ExecutionContext$class.buildFlow(ExecutionContext.scala:75)
	at com.twitter.scalding.ExecutionContext$$anon$1.buildFlow(ExecutionContext.scala:168)
	at com.twitter.scalding.Job$$anonfun$buildFlow$1.apply(Job.scala:231)
	at com.twitter.scalding.Job$$anonfun$buildFlow$1.apply(Job.scala:231)
	at scala.util.Success.flatMap(Try.scala:231)
	at com.twitter.scalding.Job.buildFlow(Job.scala:231)
	at com.twitter.scalding_internal.job.TwitterJob.buildFlow(Common.scala:84)
	at com.twitter.scalding.Job.run(Job.scala:301)
	at com.twitter.scalding.Tool.start$1(Tool.scala:124)
	at com.twitter.scalding.Tool.run(Tool.scala:141)
	at com.twitter.scalding.Tool.run(Tool.scala:68)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:70)
	at com.twitter.scalding.Tool$.main(Tool.scala:149)
	... 7 more
```

johnynek:
This looks like a regression on using Externalizer on functions, or at least I hope that is it.

Serialization problems are the worst.

johnynek:
ahh, I see this issue...

[...]

PR #1796, opened 2018-02-09
Avoid capturing in a closure in MapGroup

johnynek:
@fwbrasil this should fix (I hope) #1795 

It would be good to reproduce the error in a test we can check in, but it is the correct behavior anyway.


ianoc:
lgtm

PR #1797, opened 2018-02-09
Make TypedPipe.WithDescriptionPipe more composable

johnynek:
When planning giant graphs, as many as 1/2 or 1/3 of the pipes are description pipes.

We can't reduce that count and the tend to be more and more of the nodes as we optimize. With this change we can compose several descriptions onto a single node, which allows us to shink the graph.

This shrinking somewhat improves performance.


non:
Seems legit to me. 👍

PR #1798, opened 2018-02-12
Actually fix exponential TypedPipe equality

johnynek:
We were actually not solving the exponential complexity problem on TypedPipe equality. With this, we have a test that it was exponentially slow, then fix it.

Without this, applying rules is exponentially slow, which is obviously no good.


non:
This looks good to me. 👍 

In the long run it would be nice for the timing/algorithmic complexity testing code to be pulled out in a more generic place so we could use it other places.

PR #1799, opened 2018-02-13
Slight optimization for TypedPipe equality

johnynek:
bail out early in TypedPipe equality if hashCodes don't match, which they must in the case of equality.


gerashegalov:
👍

PR #1800, opened 2018-02-14
Add WritePartitioner to break large flows up

johnynek:
Here is some code that we want to use internally inside giant summingbird flows to break up a bunch of writes into a smaller set.

It probably isn't totally optimal yet, but it seems to do the trick of breaking things up.

Since we don't have any fancy recursion schemes going on, there is a lot of code duplication from a very similar, but not super trivial to share, function called toLiteral in the OptimizationRules.

@ianoc and @erik-stripe @non can you both give feedback?


scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
```diff
@@ -153,7 +153,7 @@ object OptimizationRules {
             Unary[TypedPipe, (K, V1), (K, V)](handleCoGrouped(wd.on, recurse), { (tp: TypedPipe[(K, V1)]) =>
               tp match {
                 case ReduceStepPipe(rs) =>
-                  withDescription(rs, desc)
+                  ReduceStepPipe(ReduceStep.withDescription(rs, desc))
```
ianoc, line 156: we were loosing some information before here i imagine?

ianoc: Oh nm i see, this logic got moved out of this file

scalding-core/src/main/scala/com/twitter/scalding/typed/WritePartitioner.scala
```diff
@@ -0,0 +1,297 @@
...
+import scala.language.higherKinds
+
+object WritePartitioner {
+  private[this] val logger = LoggerFactory.getLogger(getClass)
+
+  type PairK[F[_], G[_], T] = (F[T], G[T])
+
+  /**
+   * This breaks a job at all the places it explicitly fans out,
+   * (and currently after each reduce/join).
```
ianoc, line 15: Do we have enough optimization rules to pull `.map` functions into the reducer ? or does it make more sense to explicitly not do that and let user code do mapValues/etc.. to stay in the reducer?

johnynek: This partitioning is not yet optimal. It may wind up in more steps that cascading would have taken. It would be nice to get it so that is never the case, and we could just turn it on all the time.

Since the optimizer works, for now, on the same nodes as TypedPipe, we can't push map into a reduce because it can change the keys. We can push mapValues/flatMapValues/filterKeys. We are actually not *yet* doing mapValues and flatMapValues. I just didn't get around to it. Since adding rules is not binary incompatible, I've been punting on adding a few of the last rules.

johnynek:
@ianoc I added a test that we only have 1 more step than cascading by doing this (the trailing map that we don't yet pull up). If we nailed that, we could turn this on by default, or add a config flat to automatically do this inside of executions.... for now I'll let us get more experience with it.

non:
It's great to see this `Materializer` type class being used to write those tests!

Also great use of `EqTypes` (i.e. *Leibniz*).

👍

ianoc:
👍 

lgtm not in the critical path so any iterations to perf whatever can be done on develop i think. This looks to have good test coverage/very useful features. 

(side interesting benefit is we could actually do some absolute crazyness and if the ordered serializers have been registered for a type use those at our materialization points now... since the signatures don't include them it would need to be some sort of i guess cache.... but be an interesting experiment)

PR #1803, opened 2018-02-16
Improve rules, rule testing, and bump dagon

johnynek:
This seems to fix an issue with an infinite loop I was seeing that we were interpreting as it taking a long time.

I think when randomly making rules: one would change the graph in one way, the next would change it back, creating an infinite loop.

I seem to fix it by removing a rule from random selection. I also check that the standard rules don't loop infinitely.


scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
```diff
@@ -545,12 +527,28 @@ object OptimizationRules {
...
   }
 
+  /**
+   * This rule is important in that it allows us to reduce
+   * the number of nodes in the graph, which is helpful to speed up rule application
+   */
   object ComposeDescriptions extends PartialRule[TypedPipe] {
+    def combine(descs1: List[(String, Boolean)], descs2: List[(String, Boolean)]): List[(String, Boolean)] = {
+      val combined = descs1 ::: descs2
+      val unordered = combined.collect { case (d, true) => d }.toSet
```
ianoc-stripe: this is unused?

ianoc-stripe:
lgtm

scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
johnynek: Nope. I somehow forgot it and implemented without it below. Good catch.

non:
lgtm 👍

johnynek:
the only test that failed is the one that is fixed by #1805.

Merging since that one should fix it up.

PR #1806, opened 2018-02-16
Fix exponential equals problem with Execution

johnynek:
This bug becomes acute with #1800 because after using that code you have as many Executions as you had TypedPipes. Almost none of them are hadoop jobs, but you have a ton of map/zip nodes. The zip nodes blow up equals run times.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -23,11 +23,13 @@ import com.twitter.scalding.cascading_interop.FlowListenerPromise
 import com.twitter.scalding.filecache.{CachedFile, DistributedCacheFile}
 import com.twitter.scalding.typed.functions.{ ConsList, ReverseList }
 import com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner
+import com.stripe.dagon.{Memoize, RefPair}
 import java.util.UUID
 import scala.collection.mutable
 import scala.concurrent.{ Await, Future, ExecutionContext => ConcurrentExecutionContext, Promise }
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -198,12 +200,59 @@ sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
...
+            case (RefPair(UniqueIdExecution(fn0), UniqueIdExecution(fn1)), _) =>
+              fn0 == fn1
+            case (RefPair(WithNewCache(ex0), WithNewCache(ex1)), rec) =>
+              rec(RefPair(ex0, ex1))
+            case (RefPair(WriteExecution(h0, t0, f0), WriteExecution(h1, t1, f1)), _) =>
+              (f0 == f1) && ((h0 :: t0) == (h1 :: t1))
+            case (RefPair(Zipped(a0, b0), Zipped(a1, b1)), rec) =>
+              rec(RefPair(a0, a1)) && rec(RefPair(b0, b1))
+            case (rp, _) =>
+              require(rp._1.getClass != rp._2.getClass)
```
johnynek, line 251: this was an added safety check to make sure we didn't overlook type of Execution (we didn't, but could in the future I guess). Since this would only get hit if we have correct code in the case that two items have collided, but are really different, it is rare enough that the extra check seems a nice safety measure.

ianoc:
find it a little gnarly and concerning that we have to break out so much in the equality for it to not be expensive. But at the same time, this lgtm

johnynek:
@ianoc I think you can make a generic version of this that works using productIterator. Even better if that was the default scala implementation.

PR #1809, opened 2018-02-17
Improve Config hashCode/equals

johnynek:
yet another hashCode/equals fix....

Running partitioning creates a ton of executions, but virtually all have the same config. They were taking forever linearly traversing the config over and over.


non:
👍

PR #1810, opened 2018-02-17
Add an option to cleanup temp Execution files on finish

### LINKED ISSUE #1808: add a config option to immediately clean up execution temp files
We register in execution a thread to run on shutdown to clean up.

This doesn't always (ever?) actually work because shutdown hooks seem unreliable here.

Alternatively, in many cases we know that when the execution is done, any ForceToDisk or ToIterable have been consumed, so we can clean up immediately. If we had a config option for this, we could be safer, and most users could turn this on by default in their ExecutionApp subclass that they use, or by passing in a config option.

johnynek:
closes #1808 

This adds a new option to remove temp files in cascading-based Executions, when the outer-most execution completes. This is safe to do when the outer-most execution is an Execution[Unit], which it usually is for an application. So in Job.toExecution or ExecutionApp, this can be the default setting. Otherwise, we default to false.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
@@ -48,7 +48,7 @@ object AsyncFlowDefRunner {
    *
    * If the job is aborted the shutdown hook may not run and the temporary files will not get cleaned up
    */
-  case class TempFileCleanup(filesToCleanup: List[String], mode: Mode) extends Thread {
+  final case class TempFileCleanup(filesToCleanup: List[String], mode: Mode) extends Thread {
 
     val LOG = LoggerFactory.getLogger(this.getClass)
 
@@ -88,6 +88,14 @@ class AsyncFlowDefRunner extends Writer { self =>
   type StateKey[T] = (Config, Mode, TypedPipe[T])
   type WorkVal[T] = Future[TypedPipe[T]]
 
+  private case class FilesToCleanUp(onFinish: Set[String], onShutdown: Set[String]) {
+    def addFile(conf: Config, m: Mode, s: String): FilesToCleanUp =
+      if (conf.getExecutionCleanupOnFinish) copy(onFinish = onFinish + s)
+      else copy(onShutdown = onShutdown + s)
+  }
+  private object FilesToCleanUp {
+    def empty: FilesToCleanUp = FilesToCleanUp(Set.empty, Set.empty)
+  }
   /**
    * @param filesToCleanup temporary files created by forceToDiskExecution
    * @param initToOpt this is the mapping between user's TypedPipes and their optimized versions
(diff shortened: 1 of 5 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -227,7 +236,14 @@ class AsyncFlowDefRunner extends Writer { self =>
...
     cleanUp.foreach {
       case (mode, filesToRm) =>
-        Runtime.getRuntime.addShutdownHook(TempFileCleanup(filesToRm.toList, mode))
+        if (filesToRm.onShutdown.nonEmpty) {
+          Runtime.getRuntime.addShutdownHook(TempFileCleanup(filesToRm.onShutdown.toList, mode))
+        }
+        if (filesToRm.onFinish.nonEmpty) {
+          val cleanUpThread = TempFileCleanup(filesToRm.onFinish.toList, mode)
+          // run it that the outer most execution is complete
+          cleanUpThread.start()
```
ianoc, line 245: we probably should block on this thread competing here?  -- or i'd be more inclined that this didn't use a thread at all but we could swallow the errors. (Threads have the unfortunate problem that either its a foreground thread and we'll have some unexpected strange hanging behavior, or if a background thread we might just not wind up doing the work before we exit.)

ianoc:
modulo using a thread to actually do the cleanup -- lgtm

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
johnynek: sure. Sounds good. Only note: this is triggered in a thread anyway: it is called by `.onComplete` of a Future, so already it is on the threadpool independent of blocking the result on the Future.

ianoc: From some light googling it feels like that will eventually be its own sort of bug around this -- its not a huge one in the sense cleaning up is an optimistic operation.

We could do a try {} around the cleanup block and put the Stop into the message queue in a finally block ?

johnynek: I don't think I understand. We are already doing a try {} call inside the state cleanup. And the STOP message has already been sent (and that thread is not a daemon thread anyway).

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala#L62

ianoc: Concretely what i mean is we do want to block the final future returning for this i believe, which means we want to move the stop from 
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala#L225

after
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala#L231

We can skip the try finally if the inner method has it. But I think this would avoid any races with the application exiting normally before doing the deletes.

johnynek: That won’t block the final future from returning. The final future returning *calls* finished. We register an onComplete of the final future to call finish.

We could convert onComplete to a transform and synchronously call this, but I’m not really sure it is worth it since anyone is going to have to be prepared for crashes or users not setting this option anyway. So you will need to have cron cleanup your temp dir in any case.

ianoc: Sorry your quite right, i hadn't quite tweaked that the STOP is really just occuring in the on complete itself. Now in practice i believe moving it below, or indeed I think undoing my first comment on this would make the issue go away -- since both threads are foreground threads and would stop the application exiting until they complete. 

So i'm afraid the state is basically i think my comment has put us in a worse position, and 3 options not to make this probably not very effective (deletes could take several sections and i imagine we normally exit fast):

1) Undo my requested change and do the work on the created thread, this i believe is a foreground thread so will block the exiting
2) Move the STOP to after the synchronous operation, since this is a foreground thread it should have the same effect as (1)
3) Make it a transform rather than an onComplete to execute the side effect

One side effect pro of (3) if there was an exception in sending the stop message i think the app would hang. Though its pretty hard for that to actually occur and in practice i've never seen it.

So i guess in summary any one of those options at this point seems fine its a background cleanup after all. Simplest might be to just undo my requested change?

johnynek: Okay. I'll revert the fix keeping in mind this is easy to change in a point release since it does not change the API at all.

Thanks for thinking through some of these subtler issues.

ianoc (approved):

johnynek:
Thank you very much for your time on the review!

PR #1811, opened 2018-02-17
Add the Execution step id to the name

### LINKED ISSUE #1807: Emit the execution step id in the name of the cascading job
Currently, Executions have 1 name for every step in the execution.

This is not ideal. We should add to the name the id, which the cascading submit thread uses to distinguish stats objects.

johnynek:
close #1807 

This makes it possible to distinguish one step from another with executions.

Unfortunately, this is not a reproducible number since steps can race (due to zip running in parallel), but it seems better than nothing.

It seems somewhat non-trivial to give a deterministic numbering to all the steps in the presence of flatMap on Execution. We may revisit it later.


ianoc:
lgtm

cascading step numbers aren't stable anyway... so not any worse

PR #1813, opened 2018-02-18
Add a few serializable annotations

johnynek:
@fwbrasil maybe these objects were being captured as outer refs from the inner classes they contain, and maybe they were not serializable?

Just a shot in the dark on #1795 but it is a good idea anyway.


fwbrasil:
LGTM

johnynek:
Thank you very much for the review, @fwbrasil !

Issue #1819, opened 2018-02-20
Nested mutation bug


fwbrasil:
I'm debugging a test that fails on the latest `develop` revision. `FlowStateMap.mutate` calls are nested and mutate the same `FlowDef`. The outer mutation overrides the mutations done by the nested call. Stack trace from the debugging session:

<img width="370" alt="screen shot 2018-02-20 at 1 37 37 pm" src="https://user-images.githubusercontent.com/831175/36454125-84f417d4-164f-11e8-8f2c-8dcc639f00c2.png">

Note the two `mutate` calls. 

Isolated example:

```scala
object NestedMutationBug extends App {
  implicit val fd = new FlowDef
  implicit val m = Local(false)
  val t = TypedPipe.from(TypedText.tsv[String]("")).write(NullSink)
  CascadingBackend.planTypedWrites(fd, m)
  val state = FlowStateMap.get(fd).get
  require(state.sourceMap.nonEmpty) // fails
}
```

johnynek:
working on this now.

PR #1822, opened 2018-02-21
Fix the nested mutation issue 1819

johnynek:
The fix is pretty simple, just re-read the state after doing the reads.

This passes the test and seems correct.

Please review @fwbrasil

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -410,10 +410,15 @@ object CascadingBackend {
         val optPipe = optDag.evaluate(pair._1)
         val dest = pair._2
         val cpipe = toPipeUnoptimized[A](optPipe, dest.sinkFields)(fd, mode, dest.setter)
+        // Note this changes the FlowState, so we need to re-read it at the bottom
         dest.writeFrom(cpipe)(fd, mode)
       }
       todos.foreach(doWrite(_))
       (st.copy(pendingTypedWrites = Nil), ())
+      val finalFS = FlowStateMap(fd)
+      require(finalFS.pendingTypedWrites == st.pendingTypedWrites,
+        s"pending writes changed from ${st.pendingTypedWrites} to ${finalFS.pendingTypedWrites}")
+      (finalFS.copy(pendingTypedWrites = Nil), ())
     }
 
   /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -406,7 +406,6 @@ object CascadingBackend {
       val optPipe = optDag.evaluate(pair._1)
       val dest = pair._2
       val cpipe = toPipeUnoptimized[A](optPipe, dest.sinkFields)(fd, mode, dest.setter)
-      // Note this changes the FlowState, so we need to re-read it at the bottom
       dest.writeFrom(cpipe)(fd, mode)
     }
     todos.foreach(doWrite(_))
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -414,7 +414,7 @@ object CascadingBackend {
         dest.writeFrom(cpipe)(fd, mode)
       }
       todos.foreach(doWrite(_))
-      (st.copy(pendingTypedWrites = Nil), ())
+
       val finalFS = FlowStateMap(fd)
       require(finalFS.pendingTypedWrites == st.pendingTypedWrites,
         s"pending writes changed from ${st.pendingTypedWrites} to ${finalFS.pendingTypedWrites}")
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -8,7 +8,7 @@ import com.stripe.dagon.{ FunctionK, HCache, Id, Memoize, Rule, Dag }
 import com.twitter.scalding.TupleConverter.{ singleConverter, tuple2Converter }
 import com.twitter.scalding.TupleSetter.{ singleSetter, tup2Setter }
 import com.twitter.scalding.{
-  CleanupIdentityFunction, Config, Dsl, Field, FlatMapFunction, FlowStateMap, GroupBuilder,
+  CleanupIdentityFunction, Config, Dsl, Field, FlatMapFunction, FlowState, FlowStateMap, GroupBuilder,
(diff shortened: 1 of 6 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -410,10 +410,15 @@ object CascadingBackend {
         val optPipe = optDag.evaluate(pair._1)
         val dest = pair._2
         val cpipe = toPipeUnoptimized[A](optPipe, dest.sinkFields)(fd, mode, dest.setter)
+        // Note this changes the FlowState, so we need to re-read it at the bottom
         dest.writeFrom(cpipe)(fd, mode)
       }
       todos.foreach(doWrite(_))
-      (st.copy(pendingTypedWrites = Nil), ())
+
+      val finalFS = FlowStateMap(fd)
```
fwbrasil: I find `FlowStateMap.mutate` too fragile, even after the fix. If someone changes the implementation of any of the other method calls to also mutate the global state, we'd have another potential bug.

Maybe `FlowStateMap.mutate` could take functions that return deltas that are collected in a `ThreadLocal` and merged/applied by the initial `mutate` call. The merge could detect conflicts and throw if necessary. It seems involved, but having an unprotected global mutable state is pretty scary.

johnynek: I don’t think thread-local is safe because Execution is multi-threaded (via Future/ExectionContext). I don’t think your proposed fix will be correct.

I agree I don’t like FlowStateMap. Since you are at Twitter you could change the use of FlowDef to a GraphState that we could remove the need for FlowStateMap. I’m reluctant to send that change without being able to see how much breakage it causes and how large the diff would be to make Source green.

Frankly we might have been too conservative about Source changes at Twitter but there were previous periods of strong pushback against any source changes and a lot of fear since jobs being under tested.

fwbrasil: > I don’t think thread-local is safe because Execution is multi-threaded (via Future/ExectionContext). I don’t think your proposed fix will be correct.

This bug doesn't involve concurrency. It occurs when the same thread acquires the lock and does nested mutations, so a `ThreadLocal` should work. 

Concurrency is potentially another issue: if a nested mutation happens in another thread, it could produce a deadlock. That's a separate issue, though.

> I agree I don’t like FlowStateMap. Since you are at Twitter you could change the use of FlowDef to a GraphState that we could remove the need for FlowStateMap. I’m reluctant to send that change without being able to see how much breakage it causes and how large the diff would be to make Source green.

We should definitely plan to get rid of `FlowStateMap` in the future. We now have a good infrastructure to run scalafix rewrites, so a migration might not be that hard.

> Frankly we might have been too conservative about Source changes at Twitter but there were previous periods of strong pushback against any source changes and a lot of fear since jobs being under tested.

We're also planning to create a tool to verify correctness of scalding jobs. Something similar to Diffy but for batch processing. That'll make this kind of change easier for us.

johnynek: I know this bug does not involve concurrency, but I don't see how we can use a ThreadLocal since planning is done in multiple threads. The FlowState cannot be thread-local state, it has to be accessible across threads.

fwbrasil: My proposal was to hold the deltas in a `ThreadLocal` and merge them if the `mutate` call is not nested. `FlowStateMap` would continue to use the weak hashmap.

johnynek: PS: about Diffy note TypedPipeDiff which was created when we attempted something like this before. The E2E tests were a basic version of what you describe, but they are a selected set of jobs.

ianoc: To me these changes seem like a clear improvement in the state of the world. Cleaner interfaces , more comments. 

If someone wants to figure out multi-threaded interactions , or removing this entirely those feel like separate PR's?

Anyone -1 on just merging this as is?

fwbrasil: > If someone wants to figure out multi-threaded interactions , or removing this entirely those feel like separate PR's?

It's not really a concurrency/multi-thread issue. If we keep this global mutable state unprotected, it's easy to introduce other bugs like this one.

> Anyone -1 on just merging this as is?

I think it's important to address the underlying design issue to avoid future bugs. I can provide another PR with a more restrictive solution if you like.

johnynek: Since we have made the unsafe mutate method private, I think we are in a good shape.

We don't seem to be communicating very well here. It seems like you comment implies that we haven't made the dangerous method private, also it seems like you are not concerned with the fact that we need to update FlowState from multiple threads potentially (even though this particular bug is not a thread-safety bug).

Lastly, we don't exactly have global mutable state: you need to have your hands on the mutable FlowDef to mutate the FlowState. It is not like anyone can reach in and do anything: if you have the mutable FlowDef, you can mutate the FlowState. I don't see how this is fundamentally more dangerous than having a mutable FlowDef. I don't like any mutability, but Twitter last I knew had not adopted Execution very much (which I wrote almost 4 years ago), and without more adoption of that, we need something mutable to record where writes have taken place.

Seems like we think we should merge and you are of course welcome to send a new PR to improve it even more.

ianoc: @fwbrasil I think you should press ahead and try build another PR and we can debate the pro's and con's of that approach on that PR. And just merge this since it fixes some handling/and improves things here. Incremental improvements should really be accepted when offered imo.

johnynek:
closes #1819

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
fwbrasil: > We don't seem to be communicating very well here.

I've just replied on gitter. I think "protected" wasn't a good word. I'd like to restrict the implementation so developers can't introduce bugs like this one.

> It seems like you comment implies that we haven't made the dangerous method private, also it seems like you are not concerned with the fact that we need to update FlowState from multiple threads potentially (even though this particular bug is not a thread-safety bug).

I am concerned with thread safety, but it's not related to this bug. The example I gave of a potential deadlock with nested mutations running on different threads could happen with the current implementation. 

> It is not like anyone can reach in and do anything: if you have the mutable FlowDef, you can mutate the FlowState. I don't see how this is fundamentally more dangerous than having a mutable FlowDef.

I'd say that `FlowStateMap` is more fragile than a regular mutable `FlowDef`. For instance, this bug probably wouldn't happen if we were just mutating an instance.

> @fwbrasil I think you should press ahead and try build another PR and we can debate the pro's and con's of that approach on that PR. And just merge this since it fixes some handling/and improves things here. Incremental improvements should really be accepted when offered imo.

Sounds good, I can work on it as a separate PR. I think it'll be easier to understand my proposal if we have the code.

PR #1824, opened 2018-02-22
Improve the WritePartition API so we can use it without FlowStateMap externally

johnynek:
in #1822 we made FlowStateMap private to prevent external users from depending on that mutable state.

Unfortunately, that left no way for an external user to convert a FlowDef to an Execution, which we need internally at Stripe since we are using the write partitioning to allow us to plan fast.


ianoc:
lgtm

PR #1825, opened 2018-02-22
bump algebird to the lastest (bugfix)

johnynek:
There was a bug in 0.13.2 and 0.13.3, which we are not yet using, but others may be. By bumping this we should push them to a later (and compatible) version.


fwbrasil:
lgmt

PR #1826, opened 2018-02-23
Add ComposeReduceStep optimization

johnynek:
related to #1669 

`a.join(b.sumByKey.toTypedPipe)` should be converted to 1 map-reduce job with this optimization.

We really need more tests to show not only that the optimizations don't change the results (which is what we have now) but that they actually work in cases we think they do (and that they aren't thwarted somehow by aspects of the graph we don't notice now.

A related optimization (but not this one) allows `a.join(b).toTypedPipe.join(c)` to be a single map-reduce job. That is a follow up to the current optimization.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
@@ -1,7 +1,7 @@
 package com.twitter.scalding.typed
 
 import com.stripe.dagon.{ FunctionK, Memoize, Rule, PartialRule, Dag, Literal }
-import com.twitter.scalding.typed.functions.{ FlatMapping, FlatMappedFn, FilterKeysToFilter, FilterGroup, Fill, MapGroupMapValues, MapGroupFlatMapValues }
+import com.twitter.scalding.typed.functions.{ FlatMapping, FlatMappedFn, FilterKeysToFilter, FilterGroup, Fill, MapGroupMapValues, MapGroupFlatMapValues, SumAll, MapValueStream }
 import com.twitter.scalding.typed.functions.ComposedFunctions.{ ComposedMapFn, ComposedFilterFn, ComposedOnComplete }
 
 object OptimizationRules {
@@ -105,19 +105,23 @@ object OptimizationRules {
     private def handleCoGrouped[K, V](cg: CoGroupable[K, V], recurse: FunctionK[TypedPipe, LiteralPipe]): LiteralPipe[(K, V)] = {
       import CoGrouped._
 
-      def pipeToCG[V1](t: TypedPipe[(K, V1)]): CoGroupable[K, V1] =
+      def pipeToCG[V1](t: TypedPipe[(K, V1)]): (CoGroupable[K, V1], List[(String, Boolean)]) =
         t match {
           case ReduceStepPipe(cg: CoGroupable[K @unchecked, V1 @unchecked]) =>
             // we are relying on the fact that we use Ordering[K]
             // as a contravariant type, despite it not being defined
             // that way.
-            cg
+            (cg, Nil)
           case CoGroupedPipe(cg) =>
             // we are relying on the fact that we use Ordering[K]
             // as a contravariant type, despite it not being defined
             // that way.
-            cg.asInstanceOf[CoGroupable[K, V1]]
-          case kvPipe => IdentityReduce[K, V1, V1](cg.keyOrdering, kvPipe, None, Nil, implicitly)
+            (cg.asInstanceOf[CoGroupable[K, V1]], Nil)
+          case WithDescriptionTypedPipe(pipe, descs) =>
+            val (cg, d1) = pipeToCG(pipe)
+            (cg, ComposeDescriptions.combine(d1, descs))
+          case kvPipe =>
+            (IdentityReduce[K, V1, V1](cg.keyOrdering, kvPipe, None, Nil, implicitly), Nil)
         }
 
       cg match {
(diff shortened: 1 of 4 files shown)


johnynek:
Here is a new rule, not yet turned on by default, but I think we should definitely consider it.

This is one of the more powerful optimizations we can do because it reduces the number of map/reduce steps and serialization boundaries. This optimization is kind of the holy-grail of the whole optimization work. This is exactly the kind of optimization we couldn't do before, so users had to micro manage when they wrote .toTypedPipe or not.

With this change, if enabled by default, we will probably see many jobs reduce the number of MR-steps, and also the ergonomics are very nice since users can concern themselves more with the logic and less with details of the map-reduce model to get good performance.

please take a look @ianoc @fwbrasil

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -509,4 +520,34 @@ class OptimizationRulesTest extends FunSuite with PropertyChecks {
...
+    }
+  }
+
+  test("needless toTypedPipe is removed") {
+
+    def kv(s: String) =
+      TypedPipe.from(TypedText.tsv[(Int, Int)](s))
+
+    optimizedSteps(OptimizationRules.standardMapReduceRules ::: List(OptimizationRules.ComposeReduceSteps), 1) {
+      kv("a").sumByKey.toTypedPipe.group.max
```
johnynek, line 539: this is 2 steps without ComposeReduceSteps

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -509,4 +520,34 @@ class OptimizationRulesTest extends FunSuite with PropertyChecks {
...
+    }
+
+    optimizedSteps(OptimizationRules.standardMapReduceRules ::: List(OptimizationRules.ComposeReduceSteps), 1) {
+      kv("a").join(kv("b")).toTypedPipe.group.max
```
johnynek, line 543: this is 2 steps without ComposeReduceSteps

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -509,4 +520,34 @@ class OptimizationRulesTest extends FunSuite with PropertyChecks {
...
+    }
+
+    optimizedSteps(OptimizationRules.standardMapReduceRules ::: List(OptimizationRules.ComposeReduceSteps), 1) {
+      kv("a").sumByKey.toTypedPipe.join(kv("b")).toTypedPipe.group.max
```
johnynek, line 547: this is 3 steps without ComposeReduceSteps

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -509,4 +520,34 @@ class OptimizationRulesTest extends FunSuite with PropertyChecks {
...
+    }
+    optimizedSteps(OptimizationRules.standardMapReduceRules ::: List(OptimizationRules.ComposeReduceSteps), 1) {
+      kv("a").join(kv("b").sumByKey.toTypedPipe).toTypedPipe.group.max
```
johnynek: this is 3 steps without ComposeReduceSteps

johnynek:
By the way: keep in mind, I don't think users should write `.toTypedPipe` explicitly ever, but sometimes methods return `TypedPipe[(K, V)]` when it is really a joined or grouped thing. With the current optimization subsequent `.group.sum` could be done without an additional map-reduce step.

ianoc:
I wonder has the main usage i've seen of `toTypedPipe` gone away with the changes you've made around the pair implicit class... i think it was all about making intellij happy with the type. 

That said this is great, users returning complex grouping types from functions so that they didn't hit extra steps if they combined them was rather unfortunate leak of internal scalding types. With this i think users in those cases can just always return the typed pipe. if things can stay in a single step they will... which is super neet for cleaner/more performant code

(lgtm)

johnynek:
Thanks for the review @ianoc, merging with develop to make sure the tests are still green then will merge.

fwbrasil:
I'm reviewing right now

johnynek:
We are green.

Note we have not seen the MapGroup serialization failure of #1795 since the fix of #1827 -- so, that is looking more and more good with each passing CI.

scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
```diff
@@ -451,6 +462,32 @@ object OptimizationRules {
...
+        rs2.mapped match {
+          case ReduceStepPipe(rs1) =>
+            ReduceStep.maybeCompose(rs1, rs2).map(ReduceStepPipe(_))
+          case WithDescriptionTypedPipe(ReduceStepPipe(rs1), descs) =>
+            ReduceStep.maybeCompose(rs1, rs2).map { rs3 =>
+              WithDescriptionTypedPipe(ReduceStepPipe(rs3), descs)
+            }
+          case CoGroupedPipe(cg1) =>
+            CoGrouped.maybeCompose(cg1, rs2).map(CoGroupedPipe(_))
+          case WithDescriptionTypedPipe(CoGroupedPipe(cg1), descs) =>
```
fwbrasil, line 481: Is there a guarantee that there aren't nested descriptions here?

fwbrasil:
okay, I'm stopping the review

scalding-core/src/main/scala/com/twitter/scalding/typed/OptimizationRules.scala
johnynek: There is not, but there is another optimization rule that collapses those, so this rule should run after that rule.

johnynek:
Sorry I missed that you were reviewing @fwbrasil if you have any comments or changes, let me know, I can send a follow up PR.

PR #1827, opened 2018-02-24
Add MultiJoinFunction to remove more anonymous functions

### LINKED ISSUE #1795: Not serializable function in `MapGroup`

johnynek:
This removes the use of anonymous functions for joining and replaces with a sealed class. This is a worthwhile improvement in any case but it may also fix #1795 since we use externalizer at each element in the ADT.

This hints at an approach we could use elsewhere: in the Composed functions, we could use Externalizer, so that each function can be serialized separately.

We also fix a few issues:
1. we were incorrectly closing over `Pair` which transitively would pull in the entire graph reachable from the join.
2. our tracking of 1-1-ness or many -> 1-ness was incomplete. This improves it, so that we potentially improve the performance of our joins since we can see that the right-side does not need to be recomputed for each value on the left without blowing up memory.

This *may* be the problem with #1795 but since we don't yet have a repro test, we can't be sure.

Please take a look @fwbrasil @ianoc 

While this may not fix the issue, it is certainly *a* bug, so we should merge it in my view.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/MultiJoinFunction.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/MultiJoinFunction.scala
@@ -1,0 +1,110 @@
+package com.twitter.scalding.typed
+
+import java.io.Serializable
+import com.twitter.scalding.serialization.Externalizer
+
+/**
+ * This is a weakly typed multi-way join function. By construction,
+ * it should be kept in sync with the types in a Seq[TypedPipe[(K, Any)]]
+ *
+ * a more sophisticated typing could use an HList of TypedPipe
+ * and another more advanced coding here to prove the types line up.
+ * However, this is somewhat easy to test and only exposed to
+ * those writing backends, so we are currently satisfied with the
+ * weak typing in this case
+ *
+ * We use Externalizer internally to independently serialize each function
+ * in the composition. This, in principle, should allow Externalizer
+ * to work better since different functions may be serializable with
+ * Kryo or Java, but currently Externalizer has to use java or kryo
+ * for the entire object.
+ */
+sealed abstract class MultiJoinFunction[A, +B] extends Serializable {
+  def apply(key: A, leftMost: Iterator[Any], rightStreams: Seq[Iterable[Any]]): Iterator[B]
+}
+
+object MultiJoinFunction extends Serializable {
+  final case class Casting[A, B]() extends MultiJoinFunction[A, B] {
+    def apply(k: A, iter: Iterator[Any], empties: Seq[Iterable[Any]]) = {
+      assert(empties.isEmpty, "this join function should never be called with non-empty right-most")
+      iter.asInstanceOf[Iterator[B]]
+    }
+  }
+
+  final case class PairCachedRight[K, A, B, C](
+    leftSeqCount: Int,
+    left: MultiJoinFunction[K, A],
+    right: MultiJoinFunction[K, B],
+    @transient fn: (K, Iterator[A], Iterable[B]) => Iterator[C]) extends MultiJoinFunction[K, C] {
+
+    private[this] val fnEx = Externalizer(fn)
+
+    def apply(key: K, leftMost: Iterator[Any], rightStreams: Seq[Iterable[Any]]): Iterator[C] = {
+      val (leftSeq, rightSeq) = rightStreams.splitAt(leftSeqCount)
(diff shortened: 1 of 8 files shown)


johnynek:
Well, it didn't fix it. We tripped the same error. It is still worth merging, but not the problem.

It's a puzzle how MapGroup is involved here.

johnynek:
maybe we got lucky, but this went green on the first try (which has been pretty rare).

johnynek:
If we got lucky, we got lucky twice... Not that common. So, this seems positive to me.

scalding-core/src/main/scala/com/twitter/scalding/typed/MultiJoinFunction.scala
```diff
@@ -0,0 +1,122 @@
...
+    right: MultiJoinFunction[K, B],
+    @transient fn: (K, Iterator[A], Iterable[B]) => Iterator[C]) extends MultiJoinFunction[K, C] {
+
+    private[this] val fnEx = Externalizer(fn)
+
+    val inputSize = left.inputSize + right.inputSize
+    private[this] val leftSeqCount = left.inputSize - 1
+
+    def apply(key: K, leftMost: Iterator[Any], rightStreams: Seq[Iterable[Any]]): Iterator[C] = {
+      require(rightStreams.size == inputSize - 1, s"expected ${inputSize} inputSize, found ${rightStreams.size + 1}")
```
ianoc: probably good to add a comment describing some of the logic here around the input sizes

This is that the RHS is composed of N streams and we only cache the far right of those streams in this case?

ianoc: Ohhh its how many of the rightStreams are already covered in prior joins by the left hand multijoin function? so we need to feed them in there?

johnynek:
Three times green. I think this is a fix.

scalding-core/src/main/scala/com/twitter/scalding/typed/MultiJoinFunction.scala
johnynek: It is just an extra check (which should never possibly fail unless we have a programming bug) that the number of streams we are joining matches the the total joining operation we have.

Since we have one stream in leftMost, the others should be in rightStreams. I'll add this as a comment.

ianoc:
lgtm

(With the comment i was really referring really how the layered multi-joins worked not the require itself, but all good)

PR #1828, opened 2018-02-26
Add requireSingleValuePerKey

johnynek:
This has come up as a frequently useful function at Stripe for debugging.

In addition to debugging, the new optimizer can make use of this information and potentially reduce the cost of joins.


johnynek:
@ianoc can you take a look?

ianoc:
lgtm

PR #1829, opened 2018-02-26
Avoid map-side cache for toList

johnynek:
This is kind of weird how we never changed this, but it makes no sense to use the monoid for list on the map-side (especially since list concatenation is O(N^2) without being careful).

This just moves that operation to always just call toList on the iterator, which will be very close to optimal.

@ianoc can you review?


ianoc:
lgtm

fwbrasil:
lgtm

PR #1830, opened 2018-02-27
Prevent scala to generate toString function for Password class

dieu:
We can leak password in logs by using standard `toString` function of case classes. To prevent this behavior, we can define our own `toString` function.


johnynek (approved):

johnynek:
do you want to add a test that `Password(x).toString` does not contain `x.toString` for some non-trivial `x`?

dieu:
@johnynek added property testing.

johnynek (approved):

PR #1831, opened 2018-03-02
Make arguments to TypedPipes transient

johnynek:
This should help in some cases when a TypedPipe gets swept into a closure. If we are serializing a TypedPipe, it is never needed, since the code that runs on the mappers/reducers has already been planned, and TypedPipes are for planning.

We are running this PR at Stripe, so it seems safe.


johnynek:
@ianoc @fwbrasil ptal?

fwbrasil (approved):

PR #1832, opened 2018-03-02
Spark backend

johnynek:
This is a very basic beginning to a spark backend.

It is not complete, but does support map-only operations.

There is one big question: can we really just lie to spark and say we have AnyRef everywhere? I think it may just make serialization worse (kryo writing the classnames), but maybe we can circumvent that later since scalding allows configs to have registered classes named, maybe we can pass that information to spark somehow.

cc @ianoc

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
@@ -1,0 +1,161 @@
+package com.twitter.scalding.spark_backend
+
+import com.stripe.dagon.{ FunctionK, Memoize }
+import com.twitter.scalding.typed._
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.apache.spark.storage.StorageLevel
+import scala.concurrent.ExecutionContext
+import scala.reflect.ClassTag
+
+object SparkPlanner {
+
+  private def widen[A, B >: A](rdd: RDD[A]): RDD[B] = rdd.asInstanceOf[RDD[B]]
+
+  private def cast[A](rdd: RDD[_]): RDD[A] = rdd.asInstanceOf[RDD[A]]
+
+  // TODO, this may be just inefficient, or it may be wrong
+  implicit private def fakeClassTag[A]: ClassTag[A] = ClassTag(classOf[AnyRef]).asInstanceOf[ClassTag[A]]
+
+  /**
+   * Convert a TypedPipe to an RDD
+   */
+  def plan(ctx: SparkContext)(implicit ec: ExecutionContext): FunctionK[TypedPipe, RDD] =
+    Memoize.functionK(new Memoize.RecursiveK[TypedPipe, RDD] {
+      import TypedPipe._
+
+      def toFunction[A] = {
+        case (cp @ CounterPipe(_), rec) =>
+          // TODO: counters not yet supported
+          def go[A](p: CounterPipe[A]): RDD[A] = {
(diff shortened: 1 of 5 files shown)


johnynek:
@ianoc can you take a look?

This is not finished, but what we have is testable and in the interest of keeping the PRs small, I'd like to merge this and then follow up with more parts:

1. (this) basic framework, map-only operation support.
2. reduce operation support.
3. join support
4. full Mode/Execution support

johnynek:
build fails because spark is not there for 2.12. Will remove spark 2.12 from the CI.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.apache.spark.storage.StorageLevel
+import scala.concurrent.ExecutionContext
+import scala.reflect.ClassTag
+
+object SparkPlanner {
+
+  // TODO, this may be just inefficient, or it may be wrong
+  implicit private def fakeClassTag[A]: ClassTag[A] = ClassTag(classOf[AnyRef]).asInstanceOf[ClassTag[A]]
```
ianoc, line 16: i think they just register these to kryo so i imagine this should just hit the inefficient paths. For a normal execution app though this i guess does drop some sort of performance.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+          }
+
+          go(f)
+        case (FlatMapped(prev, fn), rec) =>
+          val op = rec(prev) // linter:disable:UndesirableTypeInference
+          op.concatMap(fn)
+        case (ForceToDisk(pipe), rec) =>
+          rec(pipe).persist(StorageLevel.DISK_ONLY)
+        case (Fork(pipe), rec) =>
+          rec(pipe).persist(StorageLevel.MEMORY_ONLY)
```
ianoc, line 140: don't need to bother updating this here, but i'd leave these to DISK_ONLY until we can figure out how/when we can upgrade the state. (Spark also has the output of shuffles persisted, so we might be able to get the planner to realize when we should have shuffle data and not do the persisting then).

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -0,0 +1,226 @@
...
+          def go[A](m: MergedTypedPipe[A]): R[A] = {
+            R.merge(rec(m.left), rec(m.right))
+          }
+          go(m)
+        case (SourcePipe(src), rec) =>
+          // TODO
+          ???
+        case (slk @ SumByLocalKeys(_, _), rec) =>
+          def sum[K, V](sblk: SumByLocalKeys[K, V]): R[(K, V)] = {
+            // we can use Algebird's SummingCache https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala#L36
```
ianoc, line 160: since these partitions are ~usually ondisk or can fit in memory we might just want MapAlgebra.sumByKey possibly (or sort and fold i guess..).

ianoc:
Some comments but they are more of an ongoing discussion than anything else. LGTM

johnynek:
thanks for the comments. these seem like good points to keep in mind as we optimize. All of your comments are doable for sure. Will address in the follow ups.

PR #1833, opened 2018-03-02
Add unit tests for kryo serialization of toIterableExecution

tchow-stripe:
an older version of kryo (2.21) fails to serialize results of toIterableExecution. Issue: https://github.com/EsotericSoftware/kryo/issues/15

```
Caused by: com.esotericsoftware.kryo.KryoException: java.lang.IllegalArgumentException: Can not set final com.twitter.scalding.typed.TypedPipe field com.twitter.scalding.Execution$$anonfun$toIterable$1.t$8 to com.twitter.scalding.Execution$ToWrite$ToIterable
Serialization trace:
t$8 (com.twitter.scalding.Execution$$anonfun$toIterable$1)
result (com.twitter.scalding.Execution$WriteExecution)
$outer (com.twitter.scalding.Execution$WriteExecution$$anonfun$map$1)
result (com.twitter.scalding.Execution$WriteExecution)
prev (com.twitter.scalding.Execution$FlatMapped)
usedSketch (com.stripe.zoolander.util.features.PrecomputedSketched$SketchInputs)
si$1 (com.stripe.zoolander.util.features.PrecomputedSketched$SketchedPipe$$anonfun$1)
fn$1 (com.stripe.zoolander.util.features.PrecomputedSketched$SketchedPipe$$anonfun$flatMapWithReplicas$1$1)
$outer (com.stripe.zoolander.util.features.PrecomputedSketched$SketchedPipe$$anonfun$flatMapWithReplicas$1$1$$anonfun$apply$2)
next (com.twitter.scalding.typed.functions.FlatMapFunctions$FromMapCompose)
next (com.twitter.scalding.typed.functions.FlatMapFunctions$FromFlatMapCompose)
com$twitter$scalding$typed$functions$FlatMappedFn$$toFn (com.twitter.scalding.typed.functions.FlatMappedFn$Series)
	at com.esotericsoftware.kryo.serializers.FieldSerializer$ObjectField.read(FieldSerializer.java:626)
	at com.esotericsoftware.kryo.serializers.FieldSerializer.read(FieldSerializer.java:221)
	at com.esotericsoftware.kryo.Kryo.readObject(Kryo.java:648)
```
- added tests around the externalizer and KryoHadoop to assert on serialization. This test fails running against kryo 2.21 and succeeds in the development branch.
- make TypedPipe transient inside of ToWrite
- add default kryo serialization strategy for Execution, TypedPipe, Execution.ToWrite

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -24,6 +24,7 @@ import com.twitter.scalding.filecache.{CachedFile, DistributedCacheFile}
 import com.twitter.scalding.typed.functions.{ ConsList, ReverseList }
 import com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner
 import com.stripe.dagon.{Memoize, RefPair}
+import java.io.Serializable
 import java.util.UUID
 import scala.collection.mutable
 import scala.concurrent.{ Await, Future, ExecutionContext => ConcurrentExecutionContext, Promise }
@@ -50,7 +51,7 @@ import scala.util.hashing.MurmurHash3
  * in some libraries) composes two Executions is parallel. Prefer
  * zip to flatMap if you want to run two Executions in parallel.
  */
-sealed trait Execution[+T] extends java.io.Serializable { self: Product =>
+sealed trait Execution[+T] extends Serializable { self: Product =>
   import Execution.{ EvalCache, FlatMapped, GetCounters, ResetCounters, Mapped, OnComplete, RecoverWith, Zipped }
 
   /**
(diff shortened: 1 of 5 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -603,11 +604,11 @@ object Execution {
    * but with proof that pipe matches sink
    */
 
-  sealed trait ToWrite
-  object ToWrite {
-    final case class Force[T](pipe: TypedPipe[T]) extends ToWrite
-    final case class ToIterable[T](pipe: TypedPipe[T]) extends ToWrite
-    final case class SimpleWrite[T](pipe: TypedPipe[T], sink: TypedSink[T]) extends ToWrite
+  sealed trait ToWrite extends Serializable
+  object ToWrite extends Serializable {
```
non, line 611: It makes sense to have the `ToWrite` companion object be serializable if we are going to capture/serialize it, but it's not obvious to me how that would happen. Is that something we've observed?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -603,11 +604,11 @@ object Execution {
...
+    final case class Force[T](@transient pipe: TypedPipe[T]) extends ToWrite
+    final case class ToIterable[T](@transient pipe: TypedPipe[T]) extends ToWrite
+    final case class SimpleWrite[T](@transient pipe: TypedPipe[T], @transient sink: TypedSink[T]) extends ToWrite
```
non, line 614: It might be worth documenting (here or in `KryoHadoop.scala`) that our strategy for serializing captured pipes/executions is to skip these fields (so they're initialized to `null`) and to explicitly serialize them as `null` elsewhere. At least, I think that's the strategy we're using, right?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: they were captured inside the ToIterable Execution.

non: OK, makes sense.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: yeah, that's a good idea, since they are only used in planning, it should be safe, as the note in KryoHadoop says, but copying here is probably good too.

PR #1835, opened 2018-03-03
Add Resolver type to clean up non-cascading backends

johnynek:
To break the pattern of making Sources and Sinks know about all the platforms, we need some way to look up the backend specific way to read/write. Much of that code will be the same, so I introduced a Resolver type to give us some basic combinators to work with (orElse, andThen, empty, pair being the most common).

This will also be used for the spark backend which is developing in #1832


johnynek:
@ianoc @erik-stripe @fwbrasil can someone review?

ianoc:
lgtm

PR #1836, opened 2018-03-03
Simplify the Memory platform using Dagon memoization

johnynek:
Using Dagon's Memoize.functionK, it is much easier to write recursive memoized functions without threading through the state yourself.

I changed the code here to do that since we will likely want to copy this approach in the spark backend (which I started to do in #1832 ).


johnynek:
@non @erik-stripe can you review? you have the most dagon familiarity.

cc @ianoc @fwbrasil

ianoc:
Erik's review would be better, but all looks pretty sane to me

non:
LGTM 👍 

Not having to explicitly thread the memoization through really improves readability!

PR #1841, opened 2018-03-07
Add in new test and filter handling

ianoc:
Expand the handling to filters too


ianoc:
This test fails and code fixes it

johnynek:
👍 

I am worried there are more cases now, but it seems like this is an improvement.

we can merge and add more tests.

PR #1842, opened 2018-03-07
Add more test cases for the parquet bug

johnynek:
There are a few more cases to deal with than just filter it turns out.

I wrote what I think is comprehensive test for all the cases that directly access the cascading API.

I changed the onComplete operation to be a filter so cascading can see it as a filter (that keeps everything). Also we have to deal with Checkpoint, which is not a filter.

The current code does not force the mapTo after all trailing Filters, but only when it has not already been forced by a previous mapping. I call these operations without maps "Passthrough" and test for seeing if a linear chain is a passthrough chain or not. If it is, we add a trailing map, otherwise we don't need to.

ptal @ianoc @fwbrasil

### DIFF
--- a/scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/PlanningTests.scala
+++ b/scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/PlanningTests.scala
@@ -8,10 +8,12 @@ import org.scalatest.FunSuite
 
 class PlanningTests extends FunSuite {
   // How many steps would this be in Hadoop on Cascading
-  def steps[A](p: TypedPipe[A]): Int = {
+  def steps[A](p: TypedPipe[A], opt: Boolean = true): Int = {
     val mode = Hdfs.default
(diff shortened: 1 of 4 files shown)


scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/PlanningTests.scala
```diff
@@ -8,10 +8,12 @@ import org.scalatest.FunSuite
 
 class PlanningTests extends FunSuite {
   // How many steps would this be in Hadoop on Cascading
-  def steps[A](p: TypedPipe[A]): Int = {
+  def steps[A](p: TypedPipe[A], opt: Boolean = true): Int = {
```
ianoc, line 11: small nit, but we should make this opt something longer, took me reading the impl to see it ment optimized and not some reference to optional

ianoc:
tiny nit, don't particularly care that much if you change it... (CI test times take so long..)

lgtm

PR #1843, opened 2018-03-08
Simplify AsyncFlowDefRunner

johnynek:
AsyncFlowDefRunner was factored out of Execution to support multiple backends (such as the new scala-only MemoryBackend as well as the nascent spark backend).

When doing this, we overlooked that the Mode of an execution is constant for the entire Execution. It is very hard to imagine ever changing that since when you `forceToDiskExecution` if you were to subsequently switch modes, how can you guarantee you can read the data in the new Mode? In any case, we don't have an API or any support for changing modes as an Execution runs currently.

Using this fact, we can simplify AsyncFlowDefRunner. Along the way I found two opportunities for bugs, although I doubt they were manifest anywhere.

Lastly, I split CascadingMode, the subtype of Mode which AsyncFlowDefRunner uses, into a separate file to make it easier to locate.


johnynek:
@ianoc can you take a look?

johnynek:
cc @fwbrasil @dieu @ttim @pankajroark 

Note, I expect this to be source compatible with previous releases since users should never know or have any way to even access the AsyncFlowDefRunner.

scalding-core/src/main/scala/com/twitter/scalding/CascadingMode.scala
```diff
@@ -0,0 +1,208 @@
+package com.twitter.scalding
```
fwbrasil, line 1: Are there changes to these classes or were they just moved?

fwbrasil:
Could you give some context on the two potential bugs you've fixed?

fwbrasil:
btw, I couldn't find references to `AsyncFlowDefRunner` in our code base

johnynek:
AsyncFlowDefRunner was introduced in 0.18

scalding-core/src/main/scala/com/twitter/scalding/CascadingMode.scala
johnynek: just moved.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -104,16 +104,15 @@ class AsyncFlowDefRunner extends Writer { self =>
    * SourcePipes or IterablePipes. These are for both toIterableExecution and forceToDiskExecution
    */
   private case class State(
-    filesToCleanup: Map[Mode, FilesToCleanUp],
-    initToOpt: HMap[TypedPipe, TypedPipe],
+    filesToCleanup: FilesToCleanUp,
+    initToOpt: HMap[StateKey, TypedPipe],
```
johnynek, line 108: @fwbrasil note this was a potential bug: we were ignoring the Config in the mapping from original to optimized pipe, but Config can control the optimizations, so this would not be safe in general. We need to include the Config as part of the key as we do in `forcedPipes`.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -122,37 +121,36 @@ class AsyncFlowDefRunner extends Writer { self =>
...
 
-  private[this] var state: State = State(Map.empty, HMap.empty, HMap.empty)
+  private[this] var state: State = State(FilesToCleanUp.empty, HMap.empty, HMap.empty)
 
   private def updateState[S](fn: State => (State, S)): S =
     mutex.synchronized {
-      val (st1, s) = fn(state)
+      val s0 = state
+      val (st1, s) = fn(s0)
+      require(state eq s0, "updateState has recursively modified state, programming error")
```
johnynek, line 153: @fwbrasil note this pattern, like the case of FlowStateMap could fail if fn also calls update. We never do that, and this is a private method, but to be paranoid we check that state has not changed during the call to fn.

fwbrasil (approved):

ianoc (approved):

PR #1844, opened 2018-03-11
Add more structure to the Spark backend

johnynek:
This follows the MemoryBackend pattern of introducing an Op type that we are planning onto. This Op in spark is basically calling a function with a SparkContext and ExecutionContext to produce a Future of an RDD.

This has the nice property that we don't take the SparkContext when we are planning, only when running.

Secondly, I filled in the other missing stuff: the SparkWriter, which manages writes as we evaluate Executions, and also the mapping of sources and sinks.

In I think 2-3 following PRs we can finish:

1. implement the writer
2. finish the planner

Note, the writer and planner implementation work can go on in parallel. So I can just fork myself and finish faster.

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
@@ -1,0 +1,17 @@
+package com.twitter.scalding.spark_backend
+
+import scala.concurrent.{ Future, ExecutionContext, Promise }
+import com.stripe.dagon.{ HMap, Rule }
+import com.twitter.scalding.typed._
+import com.twitter.scalding.{ Config, Execution, ExecutionCounters, Mode }
+
+import Execution.{ ToWrite, Writer }
+
+class SparkWriter(sparkMode: SparkMode) extends Writer {
+  def execute(conf: Config, writes: List[ToWrite])(implicit cec: ExecutionContext): Future[(Long, ExecutionCounters)] = ???
+  def finished(): Unit = ???
+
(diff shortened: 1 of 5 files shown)


johnynek:
@fwbrasil @ianoc can you take a look?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import Execution.{ ToWrite, Writer }
+
+class SparkWriter(sparkMode: SparkMode) extends Writer {
+  def execute(conf: Config, writes: List[ToWrite])(implicit cec: ExecutionContext): Future[(Long, ExecutionCounters)] = ???
+  def finished(): Unit = ???
+
+  def getForced[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[TypedPipe[T]] = ???
+  def getIterable[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[Iterable[T]] = ???
+  def start(): Unit = ???
```
ianoc, line 16: what does start do?

ianoc:
not sure what start and finish are for in the writer trait. but lgtm

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
johnynek: It’s on the base trait. They are called to begin and end the Execution. Obviously I didn’t implement them yet.

Start probably does nothing on this Writer but finish will likely need to call unpersist on some RDD.

PR #1845, opened 2018-03-14
 fix JobTest to use local mode

fwbrasil:
PR #1792 changed the way the configuration is created for `JobTest`. Instead of instantiating a `JobConf` it instantiates `Configuration`, which doesn't include a few configuration files. One of these files is `hadoop-mapreduce-client-core-2.6.0.t01.jar!/mapred-default.xml`, that sets `mapreduce.framework.name` as `local` by default.

cc/ @dieu @ttim @pankajroark @johnynek


johnynek (approved):
sorry for the break. Thanks for the test!

dieu (approved):

fwbrasil:
@johnynek @dieu thank you for the quick review! Could you merge it?

PR #1847, opened 2018-03-29
Fix macro hygiene problems

travisbrown:
I've been working with these macros lately and thought I should just go ahead and PR this fix.

Here's an example of the kind of bug lack of hygiene can cause:

```scala
import com.twitter.scalding.serialization.{OrderedSerialization, Serialization}
import com.twitter.scalding.serialization.macros.impl.BinaryOrdering.ordSer

val bytes = Serialization.toBytes[Either[Int, String]](Right("very important message"))

object Test {
  def Right(b: String): Either[Nothing, String] = util.Right(b + " ha ha")

  val result = Serialization.fromBytes[Either[Int, String]](bytes)
}
```
…and then:
```scala
scala> Test.result
res0: scala.util.Try[Either[Int,String]] = Success(Right(very important message ha ha))
```
…or a little less unpleasantly:
```scala
scala> object Test {
     |   class Int
     | 
     |   val result = Serialization.fromBytes[Either[scala.Int, String]](bytes)
     | }
<console>:17: error: could not find implicit value for evidence parameter of type com.twitter.scalding.serialization.Serialization[Either[Int,String]]
         val result = Serialization.fromBytes[Either[scala.Int, String]](bytes)
                                                                        ^
```

I think I caught all the unrooted names, but ideally we should have a separate hygiene test project (like [circe](https://github.com/circe/circe/blob/master/modules/hygiene/jvm/src/main/scala/io/circe/hygiene/HygieneTests.scala)'s) with `-Yno-imports` and `-Yno-predef` turned on. I'd be happy to add that in this PR if people want to see it.

### DIFF
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/ProductLike.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/ProductLike.scala
@@ -67,7 +67,7 @@ object ProductLike {
     }
 
     q"""
-      var $currentHash: Int = _root_.com.twitter.scalding.serialization.MurmurHashUtils.seed
+      var $currentHash: _root_.scala.Int = _root_.com.twitter.scalding.serialization.MurmurHashUtils.seed
       ..${hashUpdates}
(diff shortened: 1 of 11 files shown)


johnynek (approved):
This looks great. Thanks Travis.

PR #1848, opened 2018-04-03
Handle runtime exception from cascading for executions

dieu:
Cascading in case of runtime exception from listeners call onStopping (and don't call onThrowable).

https://github.com/Cascading/cascading/blob/2.6/cascading-core/src/main/java/cascading/flow/BaseFlow.java#L1578

### DIFF
--- a/scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
+++ b/scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
@@ -17,6 +17,8 @@ package com.twitter.scalding.cascading_interop;
 
 import cascading.flow.FlowListener;
 import cascading.flow.Flow;
+import cascading.stats.CascadingStats;
+
 import scala.concurrent.Promise$;
 import scala.concurrent.Promise;
 import scala.concurrent.Future;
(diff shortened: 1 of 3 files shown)


scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
```diff
@@ -34,17 +36,24 @@
     final Promise<T> result = Promise$.MODULE$.<T>apply();
     flow.addListener(new FlowListener() {
       public void onStarting(Flow f) { } // ignore
-      public void onStopping(Flow f) { } // ignore
+      public void onStopping(Flow f) { // in case of runtime exception cascading call onStopping
+        if (!result.isCompleted()) {
```
johnynek: let's use a single atomic call here and do `result.tryFailure`:
https://www.scala-lang.org/api/2.12.3/scala/concurrent/Promise.html#tryFailure(cause:Throwable):Boolean

scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
```diff
@@ -34,17 +36,24 @@
...
+          result.failure(new Exception("Flow was stopped"));
```
johnynek: can we exercise this in a test where we throw in a FlowListener?

scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
dieu: done, thanks

scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
dieu: done, thanks

johnynek (approved):
this is great!

PR #1850, opened 2018-04-17
Twitter release

fwbrasil:
@pankajroark @ttim @dieu @ktrivedisf @benpence could you review? I don't think the OSS tests will pass, but it should be ok since we've tested everything extensively internally

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/RichFlowDef.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/RichFlowDef.scala
@@ -17,7 +17,7 @@ package com.twitter.scalding
 
 import cascading.flow.FlowDef
 import cascading.pipe.Pipe
-
+import com.twitter.algebird.Monoid
 import java.util.{ Map => JMap, List => JList }
 
 /**
@@ -87,15 +87,8 @@ class RichFlowDef(val fd: FlowDef) {
 
     fd.mergeMiscFrom(o)
     // Merge the FlowState
-    FlowStateMap.get(o)
-      .foreach { oFS =>
-        FlowStateMap.mutate(fd) { current =>
-          // overwrite the items from o with current
-          (current.copy(sourceMap = oFS.sourceMap ++ current.sourceMap,
-            flowConfigUpdates = oFS.flowConfigUpdates ++ current.flowConfigUpdates,
-            pendingTypedWrites = oFS.pendingTypedWrites ::: current.pendingTypedWrites), ())
-        }
-      }
+    val oFS = FlowStateMap(o)
+    FlowStateMap.merge(fd, oFS)
   }
 
   /**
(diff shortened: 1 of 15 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -391,31 +392,29 @@ object CascadingBackend {
...
-      def doWrite[A](pair: (Id[A], TypedSink[A])): Unit = {
-        val optPipe = optDag.evaluate(pair._1)
-        val dest = pair._2
-        val cpipe = toPipeUnoptimized[A](optPipe, dest.sinkFields)(fd, mode, dest.setter)
-        dest.writeFrom(cpipe)(fd, mode)
-      }
-      todos.foreach(doWrite(_))
-      (st.copy(pendingTypedWrites = Nil), ())
+  def planTypedWrites(fd: FlowDef, mode: Mode): Unit = {
+    val writes = FlowStateMap.removeWrites(fd).pendingTypedWrites
```
dieu, line 396: nit: very confusing signature of `removeWrites`, without knowledge that `removeWrites` return original `flowState` is hard to understand this line.

scalding-core/src/main/scala/com/twitter/scalding/RichFlowDef.scala
```diff
@@ -146,17 +139,46 @@ class RichFlowDef(val fd: FlowDef) {
...
+        val withConfig = thisFS.copy(sourceMap = Map.empty, pendingTypedWrites = Nil)
+
+        /*
+         * Note that newFd was just allocated, so it has no
+         * FlowState at all, we verify that here to be defensive
+         * since this is not performance critical code
+         */
+        require(FlowStateMap(newFd) == FlowState.empty, s"FlowState is not empty: ${FlowStateMap(newFd)}")
+
+        FlowStateMap.merge(newFd, Monoid.plus(subFlowState, withConfig))
```
dieu, line 179: nit: again confusing name and usage, `FlowState.merge` is an immutable operation, but `FlowStateMap.merge` is mutable, maybe we need use different name or somehow be more clear that we mutate fd here.

dieu: nit: why we need here `Monoid.plus`? if `subFlowState.flowConfigUpdates` should be empty, we can do just `copy`.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -158,27 +158,27 @@ object TypedPipe extends Serializable {
         }
    }
    final case class DebugPipe[T](input: TypedPipe[T]) extends TypedPipe[T]
-   final case class FilterKeys[K, V](input: TypedPipe[(K, V)], fn: K => Boolean) extends TypedPipe[(K, V)]
-   final case class Filter[T](input: TypedPipe[T], fn: T => Boolean) extends TypedPipe[T]
-   final case class FlatMapValues[K, V, U](input: TypedPipe[(K, V)], fn: V => TraversableOnce[U]) extends TypedPipe[(K, U)]
-   final case class FlatMapped[T, U](input: TypedPipe[T], fn: T => TraversableOnce[U]) extends TypedPipe[U]
+   final case class FilterKeys[K, V](input: TypedPipe[(K, V)], @transient fn: K => Boolean) extends TypedPipe[(K, V)]
```
dieu, line 161: nit: why some arguments have `@transient`, another don't have it? do we serialize ast at all?

dieu:
let's fix compilation of test which you added.

dieu (changes requested):
let's fix compilation of test which you added.

fwbrasil:
@dieu these are actually changes already in develop. These are the commits that aren't in develop:

[disable cross build](https://github.com/twitter/scalding/pull/1850/commits/9b0de9a23598b0ff2d9cad4b5a752e960bc04534) - Necessary to make the internal release script work
[fix JobTest to use local mode](https://github.com/twitter/scalding/pull/1850/commits/3f577cc89dd7a14caf6b399465b0ac174acbfea8) - A port of a fix in develop
[fix hash joins to use multiple reducers](https://github.com/twitter/scalding/pull/1850/commits/afb03a49bf17b920ed69094a89b55e7f812e254f) - A port of a fix in develop
[expose flowstate](https://github.com/twitter/scalding/pull/1850/commits/86669ac044c8cce5ddac443ff1e3b0f72f48d09d) - FlowState is now `private[scalding]` but we use it from `scalding_internal`. I'll send a PR as part of the next release I'll work on
[force to disk before making the iterable execution](https://github.com/twitter/scalding/pull/1850/commits/d859aaccf25945e98d52580b304114c8332edae0) - Reverts `toIterableExecution` to the previous behavior so it forces the creation of a sink. This is what was making the e2e fail. I'll prepare a PR as part of the next release process

Sorry, I realized just now that I should have let you know this when I posted the PR

dieu (approved):

Issue #1851, opened 2018-04-19
None.get in AsyncFlowDefRunner


fwbrasil:
I'm trying to isolate the issue. Basically `ToWrite.optimizeWriteBatch(writes, phases)` returns fewer elements than the `writes` input and `AsyncFlowDefRunner. execute` expects that the map will have one entry for each element in `writes`. See:

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala#L269
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala#L313

Stack trace:

```
Exception in thread "main" java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:347)
	at scala.None$.get(Option.scala:345)
	at com.stripe.dagon.HMap.apply(HMap.scala:53)
	at com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner$$anonfun$com$twitter$scalding$typed$cascading_backend$AsyncFlowDefRunner$$prepareFD$1$1.apply(AsyncFlowDefRunner.scala:324)
	at com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner$$anonfun$com$twitter$scalding$typed$cascading_backend$AsyncFlowDefRunner$$prepareFD$1$1.apply(AsyncFlowDefRunner.scala:304)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner.com$twitter$scalding$typed$cascading_backend$AsyncFlowDefRunner$$prepareFD$1(AsyncFlowDefRunner.scala:304)
	at com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner$$anonfun$5.apply(AsyncFlowDefRunner.scala:330)
	at com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner$$anonfun$5.apply(AsyncFlowDefRunner.scala:330)
	at com.twitter.scalding.typed.cascading_backend.AsyncFlowDefRunner$$anonfun$validateAndRun$1.apply(AsyncFlowDefRunner.scala:242)
[...]

PR #1852, opened 2018-04-28
remove HMap from optimizeWriteBatch

johnynek:
This removes the use of HMap in the optimizeWriteBatch function, which is not needed, since we just need a list of writes and to keep them type-aligned.

This may be marginally faster at submission since we don't do the HMap lookup, which is a bit expensive on big TypedPipe graphs (although optimization requires a lot of that).

It may have given a clearer error message in the case of #1851 since we wouldn't have wrongly thought fetching from the list was broken (although, the optimizer would likely have given weird results in any case).

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -16,7 +16,7 @@ limitations under the License.
 package com.twitter.scalding
 
 import cascading.flow.{ FlowDef, Flow }
-import com.stripe.dagon.{ Dag, Id, Rule, HMap }
+import com.stripe.dagon.{ Dag, Id, Rule }
 import com.twitter.algebird.monad.Trampoline
 import com.twitter.algebird.{ Monoid, Monad, Semigroup }
 import com.twitter.scalding.cascading_interop.FlowListenerPromise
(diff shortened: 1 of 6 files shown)


johnynek:
@fwbrasil can you take a look?

johnynek:
@fwbrasil what about this one?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -607,35 +607,46 @@ object Execution {
...
+  }
   object ToWrite extends Serializable {
-    final case class Force[T](@transient pipe: TypedPipe[T]) extends ToWrite
-    final case class ToIterable[T](@transient pipe: TypedPipe[T]) extends ToWrite
-    final case class SimpleWrite[T](@transient pipe: TypedPipe[T], @transient sink: TypedSink[T]) extends ToWrite
+    final case class Force[T](@transient pipe: TypedPipe[T]) extends ToWrite[T]
+    final case class ToIterable[T](@transient pipe: TypedPipe[T]) extends ToWrite[T]
+    final case class SimpleWrite[T](@transient pipe: TypedPipe[T], @transient sink: TypedSink[T]) extends ToWrite[T]
+
+    case class OptimizedWrite[F[_], T](@transient original: F[T], toWrite: ToWrite[T])
```
fwbrasil: `final` for consistency?

fwbrasil:
LGTM!

PR #1853, opened 2018-04-28
Add a test for toIterableExecution with sources

johnynek:
@fwbrasil has seen some errors he thought could be related to this:

https://gitter.im/twitter/scalding?at=5ad65dc415c9b03114092490

I don't see an error yet. The code seems to force appropriately. We only need to run a hadoop job if there is a sink, and we seem to set that up correctly.

@fwbrasil can you add more context to suggest what the error is and if we can get a reproduction?


johnynek:
Ping again @fwbrasil

fwbrasil:
@johnynek I'm looking into this right now

fwbrasil:
@johnynek it seems that it was a problem with the base revision we used for our release. I can't reproduce on top of the latest develop

johnynek:
Should we merge this to add some more tests?

fwbrasil:
LGTM

PR #1857, opened 2018-06-02
[prototype] Introduce `KeyGrouping` to make it possible to enable `OrderedSerialization` without code changes

ttim:
We're considering to enable `OrderedSerialization` for most users at Twitter. Currently we have a blocker for that - users needs to change a source code to enable it (and not only pass a parameter, which is much more beneficial for us because we have auto-tuning infrastructure to make incremental land of such things possible).

I've tried to prototype how we can workaround this - basic idea is to introduce `KeyGrouping` class which holds implicitly provided `Ordering` **and** generated `OrderedSerialization`. With this approach we can choose which one of them to use in runtime.
Another pros of this approach: 
- users will not be needed to import `OrderedSerialization` if they want to use it - it would be provided by default
- Scalding's key grouping will not mix with general orderings in the code - right now when you import `OrderedSerialization` on top level you can accidentally break some other ordering use cases.

Another thing which is nice about this approach is the fact it's source backward compatible.

This is just prototype (it's not even holds both `OrderedSerialization` and `Ordering`, I used it to gather `Ordering`'s stats across our repo) but it illustrates the idea.

@johnynek what do you think?


scalding-core/src/main/scala/com/twitter/scalding/typed/KeyGroupingMacro.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    if (implicitInScope.toString() == "<empty>")
+      throw new IllegalArgumentException("Just can't")
+
+    val pos = c.macroApplication.pos
+
+    val desc = ordSerDesc(implicitInScope.toString())
+      .getOrElse(implicitInScope.toString())
+    val path =
+      if (pos.source.path.contains("/workspace/"))
```
johnynek, line 24: What is this?

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyGrouping.scala
```diff
@@ -0,0 +1,25 @@
...
+    KeyGrouping[(T1, T2)](
+      OrderedSerialization2.maybeOrderedSerialization2(ord1.ord, ord2.ord))
+
+  implicit def ordSerConvert[T](
+    ordSer: OrderedSerialization[T]): KeyGrouping[T] = KeyGrouping(ordSer)
+  // add unsafe convert as well to make everything compatible? add it somehow in a way with macro log?
+  //  implicit def unsafeConvert[T](ord: Ordering[T]): KeyGrouping[T] = KeyGrouping(ord)
+
+  implicit def convert[T]: KeyGrouping[T] = macro KeyGroupingMacro[T]
+}
```
johnynek, line 25: Can you comment what this macro is and why it is needed?

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.17.2"
+version in ThisBuild := "0.17.2"
```
johnynek, line 1: This version is out of date.

scalding-core/src/main/scala/com/twitter/scalding/typed/KeyGrouping.scala
```diff
@@ -0,0 +1,25 @@
+package com.twitter.scalding.typed
+
+import com.twitter.scalding.serialization.{
+  OrderedSerialization,
+  OrderedSerialization2
+}
+import scala.math.Ordering
+
+case class KeyGrouping[T](ord: Ordering[T])
```
johnynek, line 9: If we introduce a new type here I’d like to be a sealed trait and either be an Ordering or an OrderedSerialization.

johnynek (commented):
My knee jerk is reaction is that we should be very suspect of an unusual macro here. We have thought about having OrderedSerialization required but I want to explore this carefully. Another approach is a wrapper that requires OrderedSerialization.

If you have users with conflicting Orderings, can we examine that a bit more? How commonly? What are the use cases? Why does it conflict with OrderedSerialization which is a superset?

johnynek:
Thinking back at this...

We could have an alternative approach:
require OrderedSerialization everywhere we use Ordering, but have a low priority implicit that uses Kryo to supply the implicit if you have an ordering.

PR #1858, opened 2018-06-20
Fix `FileNotFoundException` in some cases in `VersionedStore#getAllVersions`

ttim:
Some jobs in twitter sometimes fail with `FileNotFoundException` due to our replication process - during replication temporary file is created in destination folder (and later renamed). If `getAllVersions` method executes `listDir` while temporary file exists and later executes `getFileSystem().getFileStatus(p)` while this temporary file renamed exception is thrown and job fails.

Solution is to move `getFileStatus` closer to block where it's used and where it's guarded by `try-catch`.

### DIFF
--- a/scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
+++ b/scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
@@ -152,7 +152,6 @@ public class VersionedStore {
             // have both version suffix and success flag below
             Set<Long> ret = new HashSet<Long>();
             for(Path p: listDir(getRoot())) {
-                final FileStatus status = getFileSystem().getFileStatus(p);
                 if (skipVersionSuffix) {
                     // backwards compatible if version suffix does not exist
                     if(Utils.isLong(p.getName())) {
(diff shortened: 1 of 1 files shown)


gerashegalov (approved):
LGTM

scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
```diff
@@ -152,7 +152,6 @@ public void cleanup(int versionsToKeep) throws IOException {
             // have both version suffix and success flag below
             Set<Long> ret = new HashSet<Long>();
             for(Path p: listDir(getRoot())) {
```
gerashegalov, line 154: Note you can use listStatus to avoid additional fileStatus RPC later on for directory contents.

johnynek (approved):

scalding-commons/src/main/java/com/twitter/scalding/commons/datastores/VersionedStore.java
ttim: I agree, but I don't want to intercept into current logic around `LocalFileSystem` in `listDir`.

PR #1862, opened 2018-07-04
Fix search of success files in FileSource

dieu:
Currently, we do `sumByKey` on `usedDirs` which has type `Iterable[(String, (String, OrVal, OrVal))]`, what ends up with summing of strings by monoid, what's unnecessary and waste a lot of time concatenation of strings.


ttim (approved):

johnynek (approved):

gerashegalov:
Thanks for fixing it @dieu !

PR #1863, opened 2018-07-20
Upgrade hbase version up to 1.2.4

dieu:
In Twitter, we are using 1.2.4 version of hbase to make compatible with internal sources make changes in upstream.


ttim (approved):
Sounds good but I want to hear what other stakeholders think? cc/ @johnynek

johnynek:
👍 

We don't use hbase or hraven, but this looks peaceful.

PR #1868, opened 2018-08-20
Spark improvements

johnynek:
This gets all the of the spark planner working, which is to say, we can convert TypedPipe into spark_backend.Op.

In a follow up PR I have the rest of it, which is the spark Writer which manages the Execution rendering.

this leverages #1867 which was highly useful to implement join/grouping related issues without materializing into memory (at the cost of instead sorting).

This could later be a Config option such that we control if we want to use the sort-based or memory-based approach to implementing join or mapGroup, but I chose to be conservative here and use the more scalable, less memory, version (sorting).

cc @ianoc @non

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
@@ -1,6 +1,6 @@
 package com.twitter.scalding.spark_backend
 
-import org.apache.spark.SparkContext
+import org.apache.spark.{HashPartitioner, SparkContext}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.StorageLevel
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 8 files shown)


scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
```diff
@@ -97,4 +155,36 @@ object Op {
...
+    @transient private val cache = new FutureCache[SparkContext, RDD[_ <: (A, D)]]
+
+    def run(ctx: SparkContext)(implicit ec: ExecutionContext): Future[RDD[_ <: (A, D)]] =
+      cache.getOrElseUpdate(ctx, {
+        // start running in parallel
+        val rrdd = right.run(ctx)
+        val lrdd = left.run(ctx)
+
+        rrdd.flatMap { rightRdd =>
+          // TODO: spark has some thing to send replicated data to nodes
```
johnynek, line 169: I solved this in #1869 using broadcast.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -130,19 +132,157 @@ object SparkPlanner {
...
+            if (evicted.nonEmpty) {
+              resultIterator = OnEmptyIterator(evicted.iterator, () => evicted.clear())
+            }
+            next
+          } else {
+            // time to flush the cache
+            import scala.collection.JavaConverters._
+            val cacheIter = currentCache.entrySet.iterator.asScala.map { e => (e.getKey, e.getValue) }
+            resultIterator = OnEmptyIterator(cacheIter, () => currentCache.clear())
+            next
```
ianoc, line 204: not sure if it particularly matters since calling next on an empty iterator usually just throws an exception -- but this will stack overflow on next called on empty?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -130,19 +132,157 @@ object SparkPlanner {
...
+    rec(TypedPipe.ReduceStepPipe(HashJoinable.toReduceStep(hj)))
+
+  private def planCoGroup[K, V](config: Config, cg: CoGrouped[K, V], rec: FunctionK[TypedPipe, Op]): Op[(K, V)] = {
+    import CoGrouped._
+
+    cg match {
+      case FilterKeys(cg, fn) =>
+        planCoGroup(config, cg, rec).filter { case (k, _) => fn(k) }
+      case MapGroup(cg, fn) =>
+        // don't need to repartition, just mapPartitions
```
ianoc, line 221: Do we not need to call the MapGroup operator here? or are we guaranteed that its already been cogrouped/shuffled?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
johnynek: `next` is tailrec, so it won't stack overflow. If it is empty it will loop forever. Good catch.

johnynek:
@ianoc take a look, I've addressed these two concerns.

johnynek:
PS: please don't click merge on this. I'll probably merge the second one into this then squash both when they are good.

ianoc:
lgtm

PR #1869, opened 2018-08-20
Make SparkWriter work

johnynek:
Note, this is a PR into #1868 (so, I will point it at develop after that merges).

This is part to support Execution in spark.

With this, you can run an `Execution[A]` in `SparkMode` which pretty much gets us to probably an alpha version of this code.

The last thing to do is provide standard SparkSources that convert the common sources/sinks to spark (e.g. TypedText.tsv, TextLine, SequenceFile, and maybe a few others).

cc @ianoc @non

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
@@ -175,11 +175,13 @@ object Op extends Serializable {
             .groupBy(_._1)
             .map { case (k, vs) => (k, vs.map(_._2)) }
 
+          val bcastMap = ctx.broadcast(rightMap)
           lrdd.map { leftrdd =>
 
             val localJoiner = joiner
 
             leftrdd.mapPartitions({ it: Iterator[(A, B)] =>
+              val rightMap = bcastMap.value
               it.flatMap { case (a, b) => localJoiner(a, b, rightMap.getOrElse(a, Nil)).map((a, _)) }
             }, preservesPartitioning = true)
           }
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkMode.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkMode.scala
@@ -3,18 +3,23 @@ package com.twitter.scalding.spark_backend
 import com.twitter.scalding.{ Config, Mode }
 import com.twitter.scalding.typed.{ Resolver, TypedSource, TypedSink }
 import org.apache.spark.rdd.RDD
-import org.apache.spark.SparkContext
+import org.apache.spark.{ SparkContext, SparkConf }
 import scala.concurrent.{ Future, ExecutionContext }
 
-case class SparkMode(sources: Resolver[TypedSource, SparkSource], sink: Resolver[TypedSink, SparkSink]) extends Mode {
+case class SparkMode(ctx: SparkContext, sources: Resolver[TypedSource, SparkSource], sink: Resolver[TypedSink, SparkSink]) extends Mode {
   def newWriter(): SparkWriter =
     new SparkWriter(this)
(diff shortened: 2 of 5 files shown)


scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
```diff
@@ -175,11 +175,13 @@ object Op extends Serializable {
             .groupBy(_._1)
             .map { case (k, vs) => (k, vs.map(_._2)) }
 
+          val bcastMap = ctx.broadcast(rightMap)
```
ianoc, line 178: you can/probably should destroy the broadcast when your done with it... not sure if you can easily know that here though...

(I've only found one case where its caused problems not destroying them, they usually are eventually GC'd anyway )

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkMode.scala
```diff
@@ -3,18 +3,23 @@ package com.twitter.scalding.spark_backend
 import com.twitter.scalding.{ Config, Mode }
 import com.twitter.scalding.typed.{ Resolver, TypedSource, TypedSink }
 import org.apache.spark.rdd.RDD
-import org.apache.spark.SparkContext
+import org.apache.spark.{ SparkContext, SparkConf }
```
ianoc, line 6: if your defining these things, these days i'd take a SparkSession than a spark context, you can get a context from a session but not vice versa

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
johnynek: yeah, adding clean ups it a bit challenging, I was thinking of punting for now.

Basically, the `Op[A]` would need to return ` Future[(RDD[_ <: A], () => Future[Unit])]` or something so the right hand function is a cleanup on finish?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkMode.scala
johnynek: wouldn't that suggest we should take the least power, not most power? For instance, if you have a session, you can make a SparkMode now, but if you have a context you couldn't if I require session.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
ianoc: Yeah for best optimization you'd want to some how gc this thing when the resulting stuff goes away, but how/when to invoke oncleanup seems hard to know it won't be referenced again

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkMode.scala
ianoc: Well except it would mean users need to thread both through, a bunch of API's say for doing I/O will want a session. So using any of the datasource v2 api's or to write out via a dataset. I think the view is normally that SparkSession is the one you hand about.


(Though there is a default global static one you can get access, so it is workaround able to grab that if a user doesn't have multiple sessions in flight)

johnynek: okay, can I make this change in a follow up PR? I didn't know that it was better to require SparkSession.

Also, I guess it is better since you can use sql first to produce smarter sources, which I think needs the Session.

Can I merge this and follow up?

ianoc: Yep on all fronts -- I think you'll get into the DatasourceV2 stuff to do custom stuff before long.

but totally do in a follow up, its all alpha anyway! :)

PR #1870, opened 2018-08-22
Use SparkSession instead of SparkContext

johnynek:
Move to using SparkSession instead of SparkContext. This allows you to use the dataset or sql APIs in sources or sinks, which can be interesting.

@ianoc please take a look.


ianoc:
lgtm

PR #1876, opened 2018-10-26
Introduce TypedPipe.toIterableExecutionWithClose and TypedPipe.useAsI…

dieu:
Port of #1874 to 0.18.x

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -239,6 +239,8 @@ sealed trait Execution[+T] extends Serializable { self: Product =>
               rec(RefPair(ex0, ex1))
             case (RefPair(TransformedConfig(ex0, fn0), TransformedConfig(ex1, fn1)), rec) =>
               (fn0 == fn1) && rec(RefPair(ex0, ex1))
+            case (RefPair(CleanupExecution(u0), CleanupExecution(u1)), _) =>
+              u0 == u1
             case (RefPair(UniqueIdExecution(fn0), UniqueIdExecution(fn1)), _) =>
               fn0 == fn1
             case (RefPair(WithNewCache(ex0), WithNewCache(ex1)), rec) =>
@@ -443,6 +445,14 @@ object Execution {
     }
   }
 
+  private final case class CleanupExecution(uuid: UUID) extends Execution[Unit] {
+    protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
+      Trampoline(cache.getOrElseInsert(conf, this,
+        cache.writer.getCleanupFor(uuid)(cec).map((_, Map.empty))
+      ))
+    }
+  }
+
   /**
    * This class allows running the passed execution with its own cache.
    * This will mean anything inside won't benefit from Execution's global attempts to avoid
@@ -610,14 +620,14 @@ object Execution {
     def pipe: TypedPipe[T]
     def replacePipe(p: TypedPipe[T]): ToWrite[T] =
       this match {
-        case ToWrite.Force(_) => ToWrite.Force(p)
-        case ToWrite.ToIterable(_) => ToWrite.ToIterable(p)
+        case ToWrite.Force(_, uuid) => ToWrite.Force(p, uuid)
+        case ToWrite.ToIterable(_, uuid) => ToWrite.ToIterable(p, uuid)
         case ToWrite.SimpleWrite(_, sink) => ToWrite.SimpleWrite(p, sink)
       }
   }
   object ToWrite extends Serializable {
-    final case class Force[T](@transient pipe: TypedPipe[T]) extends ToWrite[T]
-    final case class ToIterable[T](@transient pipe: TypedPipe[T]) extends ToWrite[T]
+    final case class Force[T](@transient pipe: TypedPipe[T], @transient uuid: UUID) extends ToWrite[T]
+    final case class ToIterable[T](@transient pipe: TypedPipe[T], @transient uuid: UUID) extends ToWrite[T]
     final case class SimpleWrite[T](@transient pipe: TypedPipe[T], @transient sink: TypedSink[T]) extends ToWrite[T]
 
     final case class OptimizedWrite[F[_], T](@transient original: F[T], toWrite: ToWrite[T])
@@ -690,6 +700,10 @@ object Execution {
       conf: Config,
       initial: TypedPipe[T]
       )(implicit cec: ConcurrentExecutionContext): Future[Iterable[T]]
+
+    private[Execution] def getCleanupFor(
+      uuid: UUID
+      )(implicit cec: ConcurrentExecutionContext): Future[Unit]
   }
 
   /**
@@ -806,6 +820,22 @@ object Execution {
       case Failure(err) => Future.failed(err)
     }
 
+  private case class RandomUUID() extends (ConcurrentExecutionContext => Future[UUID]) {
+    override def apply(context: ConcurrentExecutionContext): Future[UUID] = Future.successful(UUID.randomUUID())
+  }
+  private case class WithCleanup[A](uuid: UUID) extends (A => (A, CleanupExecution)) {
+    def apply(a: A): (A, CleanupExecution) = (a, CleanupExecution(uuid))
+  }
+  private case class ToIterableWithClose[A](t: TypedPipe[A]) extends (UUID => Execution[(Iterable[A], CleanupExecution)]) {
+    def apply(uuid: UUID): Execution[(Iterable[A], CleanupExecution)] = toIterable(t, uuid).map(WithCleanup(uuid))
+  }
+  private case class ToIterable[A](t: TypedPipe[A]) extends (UUID => Execution[Iterable[A]]) {
+    def apply(uuid: UUID): Execution[Iterable[A]] = toIterable(t, uuid)
+  }
+  private case class ForceToDisk[A](t: TypedPipe[A]) extends (UUID => Execution[TypedPipe[A]]) {
+    def apply(uuid: UUID): Execution[TypedPipe[A]] = forceToDisk(t, uuid)
+  }
+
   /**
    * This creates a definitely failed Execution.
    */
@@ -846,16 +876,26 @@ object Execution {
     FlowDefExecution(fn)
 
   def forceToDisk[T](t: TypedPipe[T]): Execution[TypedPipe[T]] =
+    fromFuture(RandomUUID()).flatMap(ForceToDisk(t))
+
+  def toIterable[T](t: TypedPipe[T]): Execution[Iterable[T]] =
+    fromFuture(RandomUUID()).flatMap(ToIterable(t))
+
+  def toIterableWithClose[T](t: TypedPipe[T]): Execution[(Iterable[T], Execution[Unit])] =
+    fromFuture(RandomUUID()).flatMap(ToIterableWithClose(t))
+
+  private[scalding] def forceToDisk[T](t: TypedPipe[T], uuid: UUID): Execution[TypedPipe[T]] =
     WriteExecution(
-      ToWrite.Force(t),
+      ToWrite.Force(t, uuid),
       Nil,
       { case (conf, _, w, cec) => w.getForced(conf, t)(cec) })
 
-  def toIterable[T](t: TypedPipe[T]): Execution[Iterable[T]] =
+  private[scalding] def toIterable[T](t: TypedPipe[T], uuid: UUID): Execution[Iterable[T]] =
     WriteExecution(
-      ToWrite.ToIterable(t),
+      ToWrite.ToIterable(t, uuid),
       Nil,
       { case (conf, _, w, cec) => w.getIterable(conf, t)(cec) })
+
 
   /**
    * The simplest form, just sink the typed pipe into the sink and get a unit execution back
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -373,6 +373,12 @@ object TypedPipe extends Serializable {
       case Left(cnt) => (Nil, ((group, cnt), 1L) :: Nil)
     }
   }
+  private case class ItClose[A, B](fn: Iterator[A] => Execution[B]) extends (((Iterable[A], Execution[Unit])) => Execution[B]) {
+    def apply(asClose: (Iterable[A], Execution[Unit])): Execution[B] = {
+      val (iterable, close) = asClose
+      fn(iterable.iterator).flatMap(b => close.map(_ => b))
+    }
+  }
 
   implicit class TallyEnrichment[A, B <: Iterable[((String, String), Long)]](val pipe: TypedPipe[(A, B)]) extends AnyVal {
     /**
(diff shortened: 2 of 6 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -851,12 +865,19 @@ object Execution {
       Nil,
       { case (conf, _, w, cec) => w.getForced(conf, t)(cec) })
 
-  def toIterable[T](t: TypedPipe[T]): Execution[Iterable[T]] =
+  def toIterable[T](t: TypedPipe[T], uuid: UUID = UUID.randomUUID()): Execution[Iterable[T]] =
```
johnynek: I don't like putting this `randomUUID` here actually.

Could we do `Execution.from(randomUUID()).flatMap { ...` since getting the randomUUID is actually an effect that we want to model? On top, why should we allow a user to pass in a UUID? shouldn't this be an implementation detail?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -851,12 +865,19 @@ object Execution {
...
     WriteExecution(
-      ToWrite.ToIterable(t),
+      ToWrite.ToIterable(t, uuid),
       Nil,
       { case (conf, _, w, cec) => w.getIterable(conf, t)(cec) })
 
+  def toIterableWithClose[T](t: TypedPipe[T]): Execution[(Iterable[T], Execution[Unit])] = {
+    val uuid = UUID.randomUUID()
+    toIterable(t, uuid).map { iter =>
```
johnynek: we should make a case class function so we don't break equality on Executions (which will harm the caching).

```scala
private case class WithCleanup[A](uuid: UUID) extends Function1[A, (A, CleanupExecution)] {
  def apply(a: A) = (A, CleanupExecution(uuid))
}
```
if we put this the Execution companion we can do `toIterable(t).map(WithCleanup())`

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -816,6 +816,53 @@ sealed abstract class TypedPipe[+T] extends Serializable with Product {
...
+   *
+   * Example:
+   *
+   * pipe.useAsIterator { iterator =>
+   *    val distinctId = iterable.map(_.id).toSet
+   *    Execution.from(distinctId)
+   * }
+   */
+  def useAsIterator[B](fn: Iterator[T] => Execution[B]): Execution[B] =
+    toIterableExecutionWithClose.flatMap { case (iterable, close) =>
```
johnynek: we need to make a case class function for this for the same reason as above (equality in caching).

```scala
private case class ItClose[A, B](fn: Iterator[A] => Execution[B]) extends Function1[(Iterable[A], Execution[Unit]), Execution[B]] {
  def apply(asClose: (Iterable[A], Execution[Unit])) = {
    val (as, close) = asClose
    fn(iterable.iterator).flatMap(b => close.map(_ =>b))
}
```

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -85,13 +85,31 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
+    private val onShutdownMap: Map[UUID, String]
+  ) {
+
+    def addFile(conf: Config, uuid: UUID, s: String): FilesToCleanUp =
+      if (conf.getExecutionCleanupOnFinish) copy(onFinishMap = onFinishMap + (uuid -> s))
+      else copy(onShutdownMap = onShutdownMap + (uuid -> s))
+
+    def popFile(uuid: UUID): (FilesToCleanUp, Option[String]) = {
+      if (onShutdownMap.contains(uuid)) {
+        val file = onShutdownMap.get(uuid)
```
johnynek: what if we do:
```scala
onShutdownMap.get(uuid) match {
  case file@Some(_) => (copy(onShutdownMap = onShutdownMap - uuid), file)
  case None =>
    val file = onFinishMap.get(uuid)
    (copy(onFinishMap = onFinishMap - uuid), file)
}
```
So we don't consult the onShutdownMap twice.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
```diff
@@ -137,4 +138,7 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
     case TypedPipe.SourcePipe(src) => getSource(src)
     case other => getForced(conf, other).flatMap(getIterable(conf, _))
   }
+
+  def getCleanupFor(uuid: UUID)(implicit cec: ConcurrentExecutionContext): Future[Unit] =
+    Future.successful(Unit)
```
johnynek: can we add support for removing the forced value here? That would help us exercise this code in more use cases.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
```diff
@@ -63,6 +63,50 @@ object ExecutionTestJobs {
...
+      .flatMap {
+        case (iter, close) =>
+          val list = iter.toList
+          close.map(_ => list)
+      }
+  }
+
+  def writeExecutionAsIteratorWithoutForce(tempFile: String, testData: List[String]): Execution[List[String]] = {
+    val exec =
+      TypedPipe.from(testData).map(s => s).useAsIterator { iter =>
```
johnynek, line 103: can we comment why you are doing `s => s`, I think it so they look like they are doing some work so scalding won't treat them as sources.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -127,6 +127,9 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
   def getIterable[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[Iterable[T]] =
     state.get().getIterable(conf, initial)
 
+  def getCleanupFor(uuid: UUID)(implicit cec: ExecutionContext): Future[Unit] =
+    Future.successful(Unit)
```
johnynek: we should at least add an issue here to support this correctly.

Note, you mean `Future.succesful(())`. `Unit` the value is the companion object to `()`. `Unit` the value has type `Unit.type`. It's a common mistake. This works because implicit conversion to `()` in scala, but it is better to write it correctly.

dieu: thanks.

if I right understand spark doesn't have this problem at all?

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
dieu: then we need to wire uuid in `Force` as well, do you fine with that?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
dieu: thanks!

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
dieu: yep, will do

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: yep, will do

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: sounds reasonable.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
dieu: I actually copy past from old functions.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -852,11 +870,22 @@ object Execution {
       { case (conf, _, w, cec) => w.getForced(conf, t)(cec) })
 
   def toIterable[T](t: TypedPipe[T]): Execution[Iterable[T]] =
+    from(UUID.randomUUID()).flatMap { uuid =>
```
johnynek: let's add a test that: `Execution.toIterable(p) == Execution.toIterable(p)`. I think we fail that currently.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -852,11 +870,22 @@ object Execution {
...
+      toIterable(t, uuid)
+    }
+
+  def toIterableWithClose[T](t: TypedPipe[T]): Execution[(Iterable[T], Execution[Unit])] =
+    from(UUID.randomUUID()).flatMap { uuid =>
+      toIterable(t, uuid).map(WithCleanup(uuid)(_))
```
johnynek: you need a function for the `flatMap` too in order to make sure `Execution` caching works.

Note, you are defeating your case class with `WithCleanup(uuid)(_)`. It is already a function, you should write: `WithCleanup(uuid)`

The test to add is: `Execution.toIterableWithClose(a) == Execution.toIterableWithClose(a)`. I think we will fail this test at the moment due to the closures.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -127,6 +127,9 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
   def getIterable[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[Iterable[T]] =
     state.get().getIterable(conf, initial)
 
+  def getCleanupFor(uuid: UUID)(implicit cec: ExecutionContext): Future[Unit] =
+    Future.successful(())
```
johnynek, line 131: I think spark has the same issue: giant data may stick around for a long time, but we can ignore for now.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: since we include uuid into `ToWrite.ToIterable` they can not be the same. is it a problem?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
dieu: ok, got it.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: okey, got what you mean.

dieu: added tests and fixed equality.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: done

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -816,6 +822,50 @@ sealed abstract class TypedPipe[+T] extends Serializable with Product {
...
+   * because after `fn` is done we will clean up resources and data will be no longer available.
+   *
+   * Example:
+   *
+   * pipe.useAsIterator { iterator =>
+   *    val distinctId = iterable.map(_.id).toSet
+   *    Execution.from(distinctId)
+   * }
+   */
+  def useAsIterator[B](fn: Iterator[T] => Execution[B]): Execution[B] =
```
johnynek, line 866: the only thing I have some anxiety about is that we force people to only go through it once, whereas if we did `useAsIterable` they could go through it multiple times, e.g. sometimes even 2 passes is a big win: go through and find the max, then take all items > x% of the max. Doing that in one pass is a pain or maybe a lot of memory.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -274,16 +298,15 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
 
-      def force[A](init: TypedPipe[A], opt: TypedPipe[A]): Unit = {
+      def force[A](init: TypedPipe[A], opt: TypedPipe[A], uuid: UUID): Unit = {
         val pipePromise = Promise[TypedPipe[A]]()
         val fut = pipePromise.future
         // This updates mutable state
         val sinkOpt = updateState { s =>
           val (nextState, added) = s.addForce(conf, init, opt, fut)
           if (added) {
-            val uuid = UUID.randomUUID
```
johnynek, line 284: I worry about this change in semantics. If we rerun the execution (say because we have withNewCache) will we correctly overwrite the data? Some taps fail on overwrite to protect data. Why do we want a UUID per Execution verses a new UUID each time we try to run?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -846,17 +876,27 @@ object Execution {
     FlowDefExecution(fn)
 
   def forceToDisk[T](t: TypedPipe[T]): Execution[TypedPipe[T]] =
+    fromFuture(RandomUUID()).flatMap(ForceToDisk(t))
+
+  def toIterable[T](t: TypedPipe[T]): Execution[Iterable[T]] =
+    fromFuture(RandomUUID()).flatMap(ToIterable(t))
+
+  def toIterableWithClose[T](t: TypedPipe[T]): Execution[(Iterable[T], Execution[Unit])] =
+    fromFuture(RandomUUID()).flatMap(ToIterableWithClose(t))
```
johnynek, line 885: Actually, I think we need something like: `Execution.withNewCache(fromFuture(RandomUUID())).flatMap...`

Because I think it is essential that we not cache the UUID generation during retries.

I think this is the answer to my question about the possible change in semantics of having the UUID.random moved out of the Execution evaluation earlier.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
dieu: ok, let's do `Iterable`, only things that are iterating over this iterator are expensive

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
dieu: thanks, nice.

dieu:
@johnynek I played more to make this PR workable and hit more problems with caching of `ToWrite` and then find out a new vulnerability with caching `force`. So in the end, looks like this feature is more complicated than needed, and option `scalding.execution.cleanup.onfinish` is good enough to achieve our goal with cleaning resources for long-running usage. 

at this point, I think we should close this feature and not increase complexity. what's do you think?

johnynek:
caching is hard...

I tend to agree with you.

PR #1879, opened 2018-11-07
Introducing ExecutionOptimizationRules for optimizing executing graph

dieu:
Hey,

We have in code some optimizations on Executions, like special implementation for zip on `WriteExecution`, to do this kind of optimization in the more generic way we can use dagon and describe optimizations in the same way and one place.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
@@ -1,0 +1,207 @@
+package com.twitter.scalding
+
+import com.stripe.dagon.{Dag, FunctionK, Literal, Memoize, PartialRule, Rule}
+import scala.annotation.tailrec
+import scala.concurrent.{Future, ExecutionContext => ConcurrentExecutionContext}
+
+object ExecutionOptimizationRules {
+  type LiteralExecution[T] = Literal[Execution, T]
+
+  /**
+   * Since our Execution is covariant, but the Literal is not
+   * this is actually safe in this context, but not in general
+   */
+  def widen[T](l: LiteralExecution[_ <: T]): LiteralExecution[T] = {
+    // to prove this is safe, see that if you have
+    // LiteralExecution[_ <: T] we can call .evaluate to get
+    // Execution[_ <: T] which due to covariance is
+    // Execution[T], and then using toLiteral we can get
+    // LiteralExecution[T]
+    //
+    // that would be wasteful to apply since the final
+    // result is identity.
+    l.asInstanceOf[LiteralExecution[T]]
+  }
+
+  def toLiteral: FunctionK[Execution, LiteralExecution] =
+    Memoize.functionK[Execution, LiteralExecution](
+      new Memoize.RecursiveK[Execution, LiteralExecution] {
+        override def toFunction[A] = {
+          case (e@Execution.ReaderExecution, _) =>
+            Literal.Const(e)
+          case (e: Execution.FutureConst[a], _) =>
+            Literal.Const(e)
+          case (e: Execution.UniqueIdExecution[a], _) =>
+            Literal.Const(e)
+          case (e: Execution.FlowDefExecution, _) =>
+            Literal.Const(e)
+          case (e: Execution.WriteExecution[a], _) =>
+            Literal.Const(e)
+          case (e: Execution.GetCounters[a], f) =>
+            widen(Literal.Unary[Execution, a, (a, ExecutionCounters)](f(e.prev), Execution.GetCounters(_: Execution[a])))
+          case (e: Execution.ResetCounters[a], f) =>
+            Literal.Unary(f(e.prev), Execution.ResetCounters(_: Execution[a]))
+          case (e: Execution.WithNewCache[a], f) =>
+            Literal.Unary(f(e.prev), Execution.WithNewCache(_: Execution[a]))
+          case (e: Execution.TransformedConfig[a], f) =>
+            Literal.Unary(f(e.prev), Execution.TransformedConfig(_: Execution[a], e.fn))
+          case (e: Execution.OnComplete[a], f) =>
+            Literal.Unary(f(e.prev), Execution.OnComplete(_: Execution[a], e.fn))
+          case (e: Execution.RecoverWith[a], f) =>
+            Literal.Unary(f(e.prev), Execution.RecoverWith(_: Execution[a], e.fn))
+          case (e: Execution.Mapped[a, b], f) =>
+            Literal.Unary(f(e.prev), Execution.Mapped(_: Execution[a], e.fn))
+          case (e: Execution.FlatMapped[a, b], f) =>
+            Literal.Unary(f(e.prev), Execution.FlatMapped(_: Execution[a], e.fn))
+          case (e: Execution.Zipped[a, b], f) =>
+            Literal.Binary(f(e.one), f(e.two), Execution.Zipped(_: Execution[a], _: Execution[b]))
+        }
+      }
+    )
+
+  /**
+   * If `Execution` is `FlowDefExecution` or `WriteExecution`,
+   * we are considering those executions as slow, since they will schedule some expensive work,
+   * like Hadoop or Spark Job.
+   *
+   * If `Execution` is `FlatMapped` or `UniqueIdExecution`,
+   * we are considering those executions as slow,
+   * since we don't know which execution they can produce.
+   *
+   * Everything else we are considering as fast execution compare to `FlowDefExecution` and `WriteExecution`.
+   */
+  def isFastExecution[A](e: Execution[A]): Boolean =
+    areFastExecution(List(e))
+
+  /**
+   * If `Execution` is `FlowDefExecution` or `WriteExecution`,
+   * we are considering those executions as slow, since they will schedule some expensive work,
+   * like Hadoop or Spark Job.
+   *
+   * If `Execution` is `FlatMapped` or `UniqueIdExecution`,
+   * we are considering those executions as slow,
+   * since we don't know which execution they can produce.
+   *
+   * Everything else we are considering as fast execution compare to `FlowDefExecution` and `WriteExecution`.
+   */
+  @tailrec
+  def areFastExecution(es: List[Execution[Any]]): Boolean =
+    es match {
+      case Nil => true
+      case h :: tail =>
+        h match {
+          case Execution.UniqueIdExecution(_) => false
+          case Execution.FlowDefExecution(_) => false
+          case Execution.WriteExecution(_, _, _) => false
+          case Execution.FlatMapped(_, _) => false
+
+          case Execution.ReaderExecution => areFastExecution(tail)
+          case Execution.FutureConst(_) => areFastExecution(tail)
+          case Execution.GetCounters(e) => areFastExecution(e :: tail)
+          case Execution.ResetCounters(e) => areFastExecution(e :: tail)
+          case Execution.WithNewCache(e) => areFastExecution(e :: tail)
+          case Execution.TransformedConfig(e, _) => areFastExecution(e :: tail)
+          case Execution.OnComplete(e, _) => areFastExecution(e :: tail)
+          case Execution.RecoverWith(e, _) => areFastExecution(e :: tail)
+          case Execution.Mapped(e, _) => areFastExecution(e :: tail)
+          case Execution.Zipped(one, two) => areFastExecution(one :: two :: tail)
+        }
+    }
+
+  object ZipWrite extends PartialRule[Execution] {
+    type InputType = (Config, Mode, Execution.Writer, ConcurrentExecutionContext)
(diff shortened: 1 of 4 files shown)


scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -0,0 +1,77 @@
...
+          case (e: Execution.FutureConst[a], _) =>
+            Literal.Const(e)
+          case (e: Execution.UniqueIdExecution[a], _) =>
+            Literal.Const(e)
+          case (e: Execution.FlowDefExecution, _) =>
+            Literal.Const(e)
+          case (e: Execution.WriteExecution[a], _) =>
+            Literal.Const(e)
+          case (e: Execution.GetCounters[a], f) => //Not sure how to treat them right
+            Literal.Unary[Execution, a, (a, ExecutionCounters)](f(e.prev), Execution.GetCounters(_: Execution[a])).asInstanceOf[LiteralExecution[A]]
```
dieu: @johnynek can we avoid `.asInstanceOf[LiteralExecution[A]]` here, somehow?

dieu:
@johnynek I have one problem:

```
"can hashCode, compare, and run a long sequence" in {
      val execution = Execution.sequence((1 to 100000).toList.map(Execution.from(_)))
      assert(execution.hashCode == execution.hashCode)

      assert(execution == execution)

      assert(execution.shouldSucceed() == (1 to 100000).toList)
    }
```
this test is falling with `StackOverflowError` on evaluating `ExecutionOptimizationRules.toLiteral`, do we have in mind something to prevent `StackOverflowError` on huge graph of executions?

johnynek:
@dieu we don't yet in dagon... :( a solution is to use something like Trampoline (or cats Eval).

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -0,0 +1,89 @@
+package com.twitter.scalding
+
+import com.stripe.dagon.{Dag, FunctionK, Literal, Memoize, PartialRule, Rule}
+import scala.concurrent.{ExecutionContext => ConcurrentExecutionContext}
+
+object ExecutionOptimizationRules {
+  type LiteralExecution[T] = Literal[Execution, T]
```
johnynek, line 8: one idea might be to do this:

```scala
type TrampLit[+A] = com.twitter.algebird.monad.Trampoline[LiteralExecution[A]]
```
and write a new `FunctionK[Execution, TrampLit]` which instead of recursing calls `Trampoline.call(f(...))` to recurse. Now you can make thke `FunctionK[Execution, LiteralExecution]` by calling the above then `.get` on the result. 

I think this will work....

And actually, if it does, we can probably change the TypedPipe example to do the same since we have had a few examples of stack overflows on that at Stripe.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -0,0 +1,89 @@
...
+      * be done in a single cascading flow if possible.
+      *
+      * If both sides are write executions then merge them
+      */
+      case Execution.Zipped(Execution.WriteExecution(oH, oT, oR), Execution.WriteExecution(tH, tT, tR)) =>
+        val newFn = { tup: (Config, Mode, Execution.Writer, ConcurrentExecutionContext) =>
+          Execution.failFastZip(oR(tup), tR(tup))(tup._4)
+        }
+        Execution.WriteExecution(oH, tH :: tT ::: oT, newFn)
+    }
```
johnynek, line 184: I think we want a few more rules:

1. pushing `Execution.Mapped(Execution.WriteExecution` into the `WriteExecution` function.
2. `zip(a.map(f), b)` == `zip(a, b).map { case (left, right) => (f(left), right) }` similarly on the right hand side.
3. there are some internal flatMaps that we can have case class functions to pattern match on optimize in some ways, e.g the random UUID ones, we are better doing `a.flatMap(b).zip(c)` to `a.flatMap { x => b(x).zip(c) }` if we know that `a` is cheap to compute. Probably all our FromFuture examples are like this by assumption, so we don't want to kick off any hadoop jobs until the futures we depend on succeed.

Does this make sense?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
johnynek: I'm pretty sure the above works, BUT I think that second function will also need to be memoized or else you may have a quadratic cost: since the first layer is just memoizing the computation of the Trampoline, not actually memoizing the result of the Trampoline.

An alternate approach is to consider using Promise/Future so you only compute each thing once Promise being a kind of memoized Trampoline. I'm not 100% sure standard scala futures are stack safe, but I think they are....

Anyway, I'd try the Trampoline first to keep it simple and only reach for promise of the perf is not working out.

dieu: I played a lot with Dagon, Dagon with Trampoline, etc. unfortunately, Dagon itself is not stack-safe for now. Let's have `try-catch` to catch `StackOverflowError` and fallback to original execution. 

And later we can improve Dagon to handle huge dags.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
dieu: done.

dieu:
@johnynek I write up simple optimizations for `Execution` and tests cases to check those optimizations. 

I'm also was able to optimize dag of executions to fix original problems with zip on writes, but unfortunately with keeping override `zip` method on `WriteExecution`, because I was able to push zip into flatMap/map, but `ZipWrite` optimization not applicable since after pushing, `zip` become part of internal lambda (what's make sense, because they depended on future computation).

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -0,0 +1,157 @@
...
+      private val fun = lfn.andThen { lr =>
+        Execution.FlatMapped(right, rfn.andThen(lr.zip))
+      }
+
+      override def apply(s: S): Execution[(T, A)] = fun(s)
+    }
+
+    override def applyWhere[T](on: Dag[Execution]) = {
+      case Execution.Zipped(Execution.FlatMapped(left, lfn), Execution.FlatMapped(right, rfn)) =>
+        Execution.FlatMapped(left, NestedZip(right, lfn, rfn))
```
johnynek, line 172: I think we are breaking a lot of parallelism here. I'm worried this will be a huge regression for us since this now runs in series.

I think we should probably only do this transformation if `left` is a "fast" Execution (say, fromFuture, getArgs, etc...), which we assume happens fast and we should see if it fails before we start any map reduce job.

Previously, left and right would be started at the same time here.

We use Executions pretty heavily and I don't think we have good test coverage on parallelism regressions like this.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -0,0 +1,157 @@
...
+
+    case class MapRight[S, T, B](fn: T => B) extends (((S, T)) => (S, B)) {
+      override def apply(st: (S, T)): (S, B) = (st._1, fn(st._2))
+    }
+
+    override def applyWhere[T](on: Dag[Execution]) = {
+      case Execution.Zipped(Execution.Mapped(left, fn), right) =>
+        Execution.Zipped(left, right).map(MapLeft(fn))
+      case Execution.Zipped(left, Execution.Mapped(right, fn)) =>
+        Execution.Zipped(left, right).map(MapRight(fn))
```
johnynek, line 145: this is safe because it doesn't decrease parallelism.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
johnynek: maybe we can write a function like: `def isFastExecution[A](ex: Execution[A])` which is something like:
```scala
case FutureConst(_) => true
case GetCounters(e) => isFastExecution(e)
case Mapped(e) => isFastExecution(e)
case WriteExecution(_, _, _, _) => false
case Zipped(a, b) => isFastExecution(a) && isFastExecution(b)
...
```
This can be made stack safe by using a list:
```scala
@annotation.tailrec
def areFastEx(es: List[Execution[Any]]): Boolean = 
  es match {
    case Nil => true
    case h :: tail =>
      h match {
        case WriteExecution(_, _, _, _) => false
        case Zipped(a, b) => areFastEx(a :: b :: tail)
        case FutureConst(_) => areFastEx(tail)
        ...
      }
  }
```

johnynek: We probably have to say a `FlatMapped` is not fast, since we have no idea what the inner execution is, but I think that is still okay, lots of examples we have will just be zips, writes, maps, and fromFuture/getArgs....

So, I guess the law we want is we never serialize two Writes that were previously parallel, but otherwise we don't mind serializing FutureConst (since we assume that is O(1-10000ms) whereas a WriteExecution is O(10m - 10h).

dieu: good, point. thanks to catching this.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
dieu: agree.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -0,0 +1,185 @@
...
+            Literal.Unary(f(e.prev), Execution.Mapped(_: Execution[a], e.fn))
+          case (e: Execution.FlatMapped[a, b], f) =>
+            Literal.Unary(f(e.prev), Execution.FlatMapped(_: Execution[a], e.fn))
+          case (e: Execution.Zipped[a, b], f) =>
+            Literal.Binary(f(e.one), f(e.two), Execution.Zipped(_: Execution[a], _: Execution[b]))
+        }
+      }
+    )
+
+  def areFastExecution[A](e: Execution[A]): Boolean =
```
johnynek: I'd call this `def isFastExecution` and add a comment on both: if this isn't a flatMap or a job requiring running a job on the hadoop cluster, we consider this a "fast" execution, which in optimizations we don't make great effort to avoid serializing.

By constrast, anything not fast, we want to maximize parallelism as much as possible.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionOptimizationRulesTest.scala
```diff
@@ -0,0 +1,267 @@
...
+      val optimized2 = ExecutionOptimizationRules.stdOptimizations(exec)
+      assert(optimized == optimized2)
+    }
+  }
+
+  def runAndCompare[A](left: Execution[A], right: Execution[A]) = {
+    val config = Config.unitTestDefault
+
+    assert(
+      left.waitFor(config, Local(true)).get ==
```
johnynek: isn't this test still applying the optimization rules inside of `def run`? I think we need a config flag to disable running optimizations, similar to TypedPipe case, and if the Config has that, do no optimization. Then in this test, we run the optimizations externally and do none inside.

What do you think?

johnynek (commented):
just a couple of minor comments, but I think this is very nearly ready.

Very nice work and possibly some real benefits to latency here.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionOptimizationRulesTest.scala
dieu: yep, nice.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
dieu: will do.

johnynek (approved):

johnynek:
This is good, and I merged it, but I wonder if we should apply optimizations again after FlatMap and UniqueIdExecution if the config is set to true?

Seems like we would want to apply the optimizations on the results in those cases.

dieu:
@johnynek I thinking about this, but then we need to bring optimizer in an implementation of `Execution`, if you don't see any problem with it, I can do it.

johnynek:
@dieu let's try a PR and see what it looks like if you have time.

PR #1881, opened 2018-11-08
Remove fold cases and sumAll from CoGroupable.atMostOneFn since folds and sumAll can return 1+ result from 0 input

dieu:
Hey,

We found the problem with `Join` and if something before can produce `non-empty` iterator for `empty-iterator`, like `fold`, `foldLeft`, crazy `Semigroup` in `sumLeft`.

While we are joining, we get an iterator from cascading and try to apply everything that's not yet applied to this value, some of the function can produce `non-empty` iterator even if don't have anything (aka cascading give us `empty` iterator).

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
@@ -84,9 +84,9 @@ object CoGroupable extends Serializable {
       case MapValueStream(Take(_)) => true
       case MapValueStream(TakeWhile(_)) => true
       case FilterGroup(_) => true
+      case EmptyGuard(fn) if atMostOneFn(fn) => true // since 0 always goes to 0 due to empty guard, and 1 -> 0 or 1 since atMostOne
       case EmptyGuard(fn) => atMostInputSizeFn(fn)
       case ComposedMapGroup(first, second) => atMostInputSizeFn(first) && atMostInputSizeFn(second)
-      case fn if atMostOneFn(fn) => true // since 0 always goes to 0, and 1 -> 1, atMostOne implies atMostInputSize
       case _ => false
     }
 }
--- a/scalding-core/src/test/scala/com/twitter/scalding/CoreTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/CoreTest.scala
@@ -15,13 +15,15 @@ limitations under the License.
 */
 package com.twitter.scalding
 
-import org.scalatest.{ WordSpec, Matchers }
-
+import org.scalatest.{Matchers, WordSpec}
 import cascading.tuple.Fields
 import cascading.tuple.TupleEntry
+import com.twitter.algebird.{Fold, Semigroup}
 import com.twitter.scalding.source.DailySuffixTsv
-
-import java.lang.{ Integer => JInt }
+import com.twitter.scalding.typed.TypedPipeGen
+import java.lang.{Integer => JInt}
+import org.scalacheck.{Arbitrary, Gen}
+import org.scalatest.prop.PropertyChecks
 
 class NumberJoinerJob(args: Args) extends Job(args) {
   val in0 = TypedTsv[(Int, Int)]("input0").read.rename((0, 1) -> ('x0, 'y0))
@@ -266,7 +268,7 @@ class JoinJob(args: Args) extends Job(args) {
     .write(Tsv(args("output")))
 }
 
-class JoinTest extends WordSpec with Matchers {
+class JoinTest extends WordSpec with Matchers with PropertyChecks {
   "A JoinJob" should {
     val input1 = List("a" -> 1, "b" -> 2, "c" -> 3)
     val input2 = List("b" -> -1, "c" -> 5, "d" -> 4)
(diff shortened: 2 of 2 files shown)


johnynek:
this is certainly scary.... can you also add one using mapGroup which should also show the bug, since fold is just calling mapGroup...

scalding-core/src/test/scala/com/twitter/scalding/CoreTest.scala
```diff
@@ -290,6 +290,35 @@ class JoinTest extends WordSpec with Matchers {
...
+          .groupBy(_._1)
+          .foldLeft(Value()) {
+            case (acc, (_, int)) => acc.copy(Some(int))
+          }
+
+      val right =
+        TypedPipe.from(List("c" -> 3, "b" -> 2))
+
+
+      val res = left.join(right).map {
```
johnynek: also interesting if this manifests only on the left. A join should be symmetric: we can make a law to test that `a.join(b).values == b.join(a).values.swap` I wonder if that law is being violated somehow.

We have random tests in OptimizationTest that we could add this to.

johnynek: also, is the .map needed to trigger it?

dieu:
so, the problem with `fold`, we doing fold on `empty-iterator` (when no data on left), but it's legit that fold on empty produce `init` as result.

johnynek:
last comment, does only fold trigger it? We don't often use fold, but we do very often use `.sum` before a join.

I can't believe we don't have test coverage for this case. That's really unfortunate.

dieu:
> last comment, does only fold trigger it? We don't often use fold, but we do very often use `.sum` before a join.

writing tests for others...

scalding-core/src/test/scala/com/twitter/scalding/CoreTest.scala
dieu: no, the map is not necessary to trigger.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -365,8 +365,14 @@ object Grouped extends Serializable {
 
   def addEmptyGuard[K, V1, V2](fn: (K, Iterator[V1]) => Iterator[V2]): (K, Iterator[V1]) => Iterator[V2] =
     fn match {
-      case alreadyGuarded@EmptyGuard(_) => alreadyGuarded
-      case ami if CoGroupable.atMostInputSizeFn(ami) => ami // already safe
```
johnynek: yeah, I think the bug is that atMostInputSizeFn is just wrong for folds, which as you point out always emit 1 thing, which can be greater than 1.

dieu: ok. let's go with this.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -63,11 +63,7 @@ object CoGroupable extends Serializable {
     fn match {
       case ComposedMapGroup(_, fn) if atMostOneFn(fn) => true
       case ComposedMapGroup(first, second) => atMostOneFn(first) && atMostInputSizeFn(second)
-      case MapValueStream(SumAll(_)) => true
```
johnynek: I think this one should not be removed.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -63,11 +63,7 @@ object CoGroupable extends Serializable {
...
       case MapValueStream(ToList()) => true
-      case MapValueStream(FoldIterator(_)) => true
-      case MapValueStream(FoldLeftIterator(_, _)) => true
-      case FoldWithKeyIterator(_) => true
```
johnynek: I think these should all be here, they do emit at most 1, no?

johnynek: line 85 below is the problem:

```
case fn if atMostOneFn(fn) => true // since 0 always goes to 0, and 1 -> 1, atMostOne implies atMostInputSize
```
that is not true. Since 0 does not always do go 0 if you are not guarded.

I think we should change to:
```
case EmptyGuard(fn) if atMostOneFn(fn) => true // then it becomes true due to the empty guard
```
we need to move that case above the current `case EmptyGuard(fn)`

dieu: Ok. got it.

scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala
```diff
@@ -63,7 +63,6 @@ object CoGroupable extends Serializable {
     fn match {
       case ComposedMapGroup(_, fn) if atMostOneFn(fn) => true
       case ComposedMapGroup(first, second) => atMostOneFn(first) && atMostInputSizeFn(second)
-      case MapValueStream(SumAll(_)) => true
```
johnynek: why isn't this true? Seems like it should be true.

dieu: agree.

johnynek (approved):
Thanks for running this down!

Great to see a few hundred lines of test to support a 1 line change! :)

PR #1885, opened 2018-11-27
Fix `planTypedWrites` when typed write happens in typed sink implementation

ttim:
During the release process in Twitter we found a customer who had a `TypedSink` implementation which did a casting of cascading's `Pipe` back to `TypedPipe` in `writeFrom` and doing typed `write` on it. This was working with 0.17.x branch but it's broken in 0.18.x since we do `planTypedWrites` only once.

This PR adds test for that and fixes it.
Even this type of `Sink`s implementation seems wrong it's better to keep scalding backward compatible with currently released version.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -422,7 +422,11 @@ object CascadingBackend {
       val cpipe = toPipeUnoptimized[A](optPipe, dest.sinkFields)(fd, mode, dest.setter)
       dest.writeFrom(cpipe)(fd, mode)
     }
-    todos.foreach(doWrite(_))
+    if (todos.nonEmpty) {
+      todos.foreach(doWrite(_))
+      // In case if during writes planning other typed writes happened.
+      planTypedWrites(fd, mode)
+    }
   }
 
   /**
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -23,8 +23,10 @@ import com.twitter.scalding.serialization.{
   WrappedSerialization
 }
 import java.util.WeakHashMap
+import org.slf4j.LoggerFactory
 
 object CascadingBackend {
+  private[this] val logger = LoggerFactory.getLogger(getClass)
 
   def areDefiniteInverse[A, B](t: TupleConverter[A], s: TupleSetter[B]): Boolean =
     (t, s) match {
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -422,7 +422,11 @@ object CascadingBackend {
       val cpipe = toPipeUnoptimized[A](optPipe, dest.sinkFields)(fd, mode, dest.setter)
       dest.writeFrom(cpipe)(fd, mode)
     }
-    todos.foreach(doWrite(_))
+    if (todos.nonEmpty) {
+      todos.foreach(doWrite(_))
+      // In case if during writes planning other typed writes happened.
+      planTypedWrites(fd, mode)
```
dieu: is `planTypedWrites` tailrec?

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
```diff
@@ -0,0 +1,86 @@
...
+      JobTest(new TypedSinkWithTypedImplementationJob(_))
+        .sink[String](TypedTsv[String]("output"))(_.toList == List("test"))
+        .runHadoop
+        .finish()
+    }
+  }
+
+  "A TypedSinkWithTypedImplementation" should {
+    "should work with .writeExecution" in {
+      val testMode = JobTest(new TypedSinkWithTypedImplementationJob(_))
```
dieu: maybe `val mode = Hdfs(true, new Configuration)` instead of `JobTest`?

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
```diff
@@ -0,0 +1,86 @@
...
+
+      val elements = List("test")
+      val elementsFromExecution = TypedPipe.from(elements)
+        .writeExecution(new TypedSinkWithTypedImplementation("output"))
+        .flatMap(_ => TypedPipe.from(TypedTsv[String]("output")).toIterableExecution)
+        .waitFor(Config.default, testMode)
+        .get
+        .toList
+
+      assert(elements == elementsFromExecution)
```
dieu, line 71: what's do you think about extending `TypedPipeChecker.checkOutput` to support `Hdfs` mode?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
ttim: I'm not particularly nitty on this but overall I prefer do not annotate methods with tailrec unless they require it.

johnynek: I think it is tailrec, but I do also worry about an infinite loop here with a buggy implementation of writeFrom.

Maybe we can do an info log if we see this and recommend users not do new writes in writes. We could have an inner method that keeps a count of how deep we go in the recursion, and if depth > 0 we do the info log message?

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.scalatest.{Matchers, WordSpec}
+
+class TypedSinkWithTypedImplementation(path: String) extends TypedSink[String] {
+  private val fields = new Fields(0)
+
+  override def setter[U <: String]: TupleSetter[U] = TupleSetter.singleSetter[U]
+
+  override def writeFrom(pipe: Pipe)(implicit flowDef: FlowDef, mode: Mode): Pipe = {
+    TypedPipe.from[String](pipe, fields).write(TypedTsv[String](path))
+    pipe
```
johnynek, line 31: oh god.... why.........

johnynek (commented):
Nice test coverage.

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
ttim: Changed to `HadoopTest`, good call, thanks!

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
ttim: I think it's out of scope and not really needed here.

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
ttim: Actual use case is VKVS which aggregates values on write.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
ttim: I did log warn message if you need to apply `planTypedWrites` twice and exception if you need to apply more.

johnynek (approved):
by the way, we could imagine implementing something like contramap for TypedSink which allows a more complex operation:
```scala
trait TypedSink[-A] {

  def preprocess[B](fn: TypedPipe[B] => TypedPipe[A]): TypedSink[B] = ...
}
```
which would allow you to put arbitrary static-dag (non-execution) before a TypedSink.

I guess this would satisfy their use case.

scalding-core/src/test/scala/com/twitter/scalding/TypedSinkWithTypedImplementationTest.scala
dieu: it will simplify your tests, but up to you.

dieu (approved):

johnynek:
similarly with:
```scala
trait TypedSource[+A] {
  def postprocess[B](fn: TypedPipe[A] => TypedPipe[B]): TypedSource[B]
}
```
if we had these APIs we could push logic into virtual sinks and sources...

Maybe that is making them too rich... The current thing only allows for mapping (no filtering, or even flattening by concatMap/flatMap operations).

ttim:
> similarly with:
> 
> ```scala
> trait TypedSource[+A] {
>   def postprocess[B](fn: TypedPipe[A] => TypedPipe[B]): TypedSource[B]
> }
> ```
> 
> if we had these APIs we could push logic into virtual sinks and sources...
> 
> Maybe that is making them too rich... The current thing only allows for mapping (no filtering, or even flattening by concatMap/flatMap operations).

In the case I'm referring to above we ended up with utility functions which wrap `TypedSink`s.

ttim:
> similarly with:
> 
> ```scala
> trait TypedSource[+A] {
>   def postprocess[B](fn: TypedPipe[A] => TypedPipe[B]): TypedSource[B]
> }
> ```
> 
> if we had these APIs we could push logic into virtual sinks and sources...
> 
> Maybe that is making them too rich... The current thing only allows for mapping (no filtering, or even flattening by concatMap/flatMap operations).

Overall I really like this approach thou.

Issue #1786, opened 2018-02-05
Potential planner bug in develop branch


johnynek:
```
[info] - all optimization rules don't change results *** FAILED ***
[info]   NullPointerException was thrown during property evaluation.
[info]     Message: "None"
[info]     Occurred when passed generated values (
[info]       arg0 = WithDescriptionTypedPipe(TrappedPipe(WithDescriptionTypedPipe(MergedTypedPipe(WithDescriptionTypedPipe(EmptyTypedPipe,lVscjjUwdswuhslob3wcryitdoprmphqbl,false),WithDescriptionTypedPipe(EmptyTypedPipe,p8buqblzwenawcsgy8ckij7rndez4xlpcyJPwpzicbj9ct9hlekkzMhro5ugym9keiikxrjs7sutjuOhk,false)),org.scalacheck.Gen$R.map(Gen.scala:237),true),com.twitter.scalding.source.FixedTypedText(qayWlj3wtbwSlnkdhrgUTwilmjusWwsOdv7qdreth1933gtnuqzeythlyl6mbktjoimervxhhoavLbcyxB4duzrjxmk8b),Single(com.twitter.scalding.TupleGetter$IntGetter$@78d30dcc)),org.scalacheck.Gen$R.map(Gen.scala:237),true),
[info]       arg1 = com.twitter.scalding.typed.OptimizationRules$FilterLocally$@342d6071.orElse(com.twitter.scalding.typed.OptimizationRules$ComposeFlatMap$@12db1063).orElse(com.twitter.scalding.typed.OptimizationRules$ComposeMap$@41a47a5c).orElse(com.twitter.scalding.typed.OptimizationRules$ComposeFilter$@52cbaab4).orElse(com.twitter.scalding.typed.OptimizationRules$ComposeWithOnComplete$@d3a7674).orElse(com.twitter.scalding.typed.OptimizationRules$RemoveDuplicateForceFork$@47a827b8).orElse(com.twitter.scalding.typed.OptimizationRules$ForceToDiskBeforeHashJoin$@253cd38b).orElse(com.twitter.scalding.typed.OptimizationRules$DeferMerge$@2a2d4392).orElse(com.twitter.scalding.typed.OptimizationRules$EmptyIterableIsEmpty$@747a0975).orElse(com.twitter.scalding.typed.OptimizationRules$HashToShuffleCoGroup$@726975a5)
[info]     )
```

The stack that caused the error:
```
[info]   Cause: java.lang.NullPointerException:
[...]

PR #1894, opened 2019-01-15
Add a guard on mutability in `Fill` function

ttim:
During internal testing in Twitter we found that `dedupMerge` optimization sometimes breaks jobs. Root cause of that turned out to be using of mutable values (to be precise `PriorityQueue` through topK monoid) in your job code and `Fill` function which duplicates this mutable values.

In this PR I've added guard against mutability in `Fill` function. It's not guaranteed to catch all mutability issues but at least some of them for almost no cost.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
   - export PATH=${PATH}:./vendor/bundle
 
 install:
-  - rvm use 2.2.6 --install --fuzzy
+  - rvm use 2.3.0 --install --fuzzy
   - gem update --system
   - gem install sass
   - gem install jekyll -f -v 3.2.1
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/functions/Functions.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/functions/Functions.scala
@@ -86,7 +86,16 @@ case class SumAll[T](sg: Semigroup[T]) extends Function1[TraversableOnce[T], Ite
 }
 
 case class Fill[A](size: Int) extends Function1[A, Iterator[A]] {
-  def apply(a: A) = Iterator.fill(size)(a)
+  def apply(a: A) = {
+    val originalHashCode = a.hashCode()
+    Iterator.fill(size)(a).map { value =>
+      if (value.hashCode() != originalHashCode) {
+        throw new IllegalStateException(s"Mutable value $value was passed and mutated during job execution. " +
+          s"This is prohibited.")
+      }
+      value
+    }
+  }
 }
 
 case class AggPrepare[A, B, C](agg: Aggregator[A, B, C]) extends Function1[A, B] {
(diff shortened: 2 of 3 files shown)


johnynek:
this hurts everyone's performance just because some jobs are using mutable values.

Can they instead disable this optimization on their end?

https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Config.scala#L272

johnynek:
computing hashCode has to traverse an entire object generally.

ttim:
> Can they instead disable this optimization on their end?

Unfortunately that's impossible since this particular optimization is enabled unconditionally in [`CascadingBackend`](https://github.com/twitter/scalding/blob/3242ecfad275609ec00fef36e9ea78f27e9056cb/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala#L511).
When we actually tried to disable it it broke our scalding job completely which feels related to [this code](https://github.com/twitter/scalding/blob/3242ecfad275609ec00fef36e9ea78f27e9056cb/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala#L504) but that's I'm unsure.

While this change has some perf overhead it's important to understand this is very rare - it happens only if you have merged identical typed pipes.

ttim:
@johnynek What do you think? Do you have this kind of graphs (`a ++ a`) in production at Stripe?

I have a feeling this optimization is more about making cascading to work with this kind of graphs (originally you added this optimization to fix https://github.com/twitter/scalding/issues/1786) and not to actually improve performance of the job. Since it's really hard to imagine this kind of graphs in real life (in fact we found this issue because of typo in a test) I don't think this performance overhead matters.

johnynek:
one principle I have long had for TypedPipe/Execution/etc... is that if it compiles, it should run.

I think there are probably legitimate cases of `a ++ a` that come up when people are writing functions on `TypedPipe`.

If we start trying to detect mutation of values, where do we stop? This isn't the only place that could go wrong.

We could keep adding more and other checks.

I'd really prefer we say: use mutable values at your own risk, and it is your responsibility to write tests to show that things work. They are not explicitly supported since it would hamstring our ability to do many optimizations.

ttim:
I think rule of thumb of mutations should be:
> if optimization might break in case of mutations then it *might* have sense to check for mutations

I don't think this is very common.

Unfortunately people write stupid code (and don't test it excessively) and I think if we have an ability to fail them fast with clear message it's a good idea to do that. Especially if cost of that is small and it happens rarely.

johnynek:
For instance, even map fusion is not safe if you do mutating side effects:

```
scala> (0 to 3).map { x => println(s"hey: $x"); x }.map { x => println(s"bye: $x"); x }
hey: 0
hey: 1
hey: 2
hey: 3
bye: 0
bye: 1
bye: 2
bye: 3

scala> (0 to 3).map { x => println(s"hey: $x"); println(s"bye: $x"); x }
hey: 0
bye: 0
hey: 1
bye: 1
hey: 2
bye: 2
hey: 3
bye: 3
```
that's a fundamental optimization we do, similarly with moving filter functions around.

ttim:
That's a good point. I guess I'm agree to abandon this PR then.

We still need to fix `topK` aggregation thou.

PR #1895, opened 2019-01-18
Fix a race condition in `JobTest`

ttim:
We found that when a lot of sink writes happen concurrently through `Execution`s using `JobTest` you can observe data race for `Buffer`s of tuples created. 

I've created a test to show this - if you do writes to 100 sinks in parallel it breaks internals of mutable map where we store tuple buffers for each mocked source and make test infinite.

I've added a lock to fix this race condition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
   - export PATH=${PATH}:./vendor/bundle
 
 install:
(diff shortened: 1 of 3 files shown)


johnynek (approved):
Change looks good. using a mutable buffer unguarded by a lock was a mistake. Merge when green.

PR #1896, opened 2019-01-25
Remove dependency on macro-compat

xeno-by:
In our experiments with supporting progressively bigger chunks of
Twitter codebase with Rsc, we stumbled upon usage of macro annotations
in Scalding.

At the moment, Rsc doesn't support macros (neither def macros, not
macro annotations), so macro annotations present a problem for our
experiments.

This pull request is specifically about the `@bundle` annotation from
macro-compat.

It looks like this annotation was added in order to cross-compile
between Scala 2.10 and Scala 2.11+ (since there are source-incompatible
changes in the macro API between these versions). However, since
this codebase no longer seems to support Scala 2.10, it looks like
usages of `@bundle` (and macro-compat) can be safely removed. That
would simplify our internal experiments with Rsc.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,7 +40,6 @@ val sparkVersion = "2.4.0"
 val slf4jVersion = "1.6.6"
 val thriftVersion = "0.9.3"
 val junitVersion = "4.10"
-val macroCompatVersion = "1.1.1"
 val jlineVersion = "2.14.3"
 
 val printDependencyClasspath = taskKey[Unit]("Prints location of the dependencies")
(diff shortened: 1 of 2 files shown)


johnynek:
did you mean we no longer support 2.10? We definitely do support 2.11.

xeno-by:
@johnynek Yes, I meant this codebase no longer seems to support Scala 2.10. Thank you for noticing! Updated the comment.

johnynek:
@xeno-by can rsc compile scalding now?

xeno-by:
@johnynek I didn't try that yet.

PR #1901, opened 2019-03-07
Updated/Removed invalid links

LorenzNickel:
Updated outdated links where possible, removed those where I  couldn't find a new location


johnynek (approved):
thanks!

PR #1902, opened 2019-03-20
Improve spark backend a bit, add some write tests

johnynek:
this is based on some work with @stephbian.

We are attempting to use spark-scalding for an internal library that compiles to scalding, vs make it compile to spark.

We do three things:
1. make the persist level configurable
2. support for TextLine and WritableSequenceFile sources out of the box.
3. add tests for those.

I didn't discover any bugs, but probably the pattern wasn't clear to people. It would be nice to add more built in types. I will try to make a follow up using csv/tsv which requires a bit more work (since scalding and spark both need typeclasses to describe the types).

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
@@ -10,6 +10,7 @@ import org.apache.spark.storage.StorageLevel
 import scala.collection.mutable.{ Map => MMap, ArrayBuffer }
 
 object SparkPlanner {
+  import SparkMode.SparkConfigMethods
   /**
    * Convert a TypedPipe to an RDD
    */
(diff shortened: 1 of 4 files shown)


scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -62,9 +63,17 @@ object SparkPlanner {
...
-          rec(pipe).persist(StorageLevel.DISK_ONLY)
+          val sparkPipe = rec(pipe)
+          config.getForceToDiskPersistMode.getOrElse(StorageLevel.DISK_ONLY) match {
+            case StorageLevel.NONE => sparkPipe
+            case notNone => sparkPipe.persist(notNone)
+          }
         case (Fork(pipe), rec) =>
-          rec(pipe).persist(StorageLevel.MEMORY_ONLY)
+          val sparkPipe = rec(pipe)
+          config.getForceToDiskPersistMode.getOrElse(StorageLevel.MEMORY_ONLY) match {
```
johnynek: obviously should be getForkPersistMode....

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -165,7 +177,7 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
     }
     def write[T](opt: TypedPipe[T], keyPipe: TypedPipe[T], sink: TypedSink[T], oldState: State): (State, Action) = {
       val promise = Promise[RDD[_ <: T]]()
-      val (newState, added) = oldState.addForce[T](conf, keyPipe, opt, promise.future)
+      val (newState, added) = oldState.addForce[T](conf, init = keyPipe, opt = opt, promise.future, None)
```
johnynek, line 180: there is no need to default force to disk first, if the user wanted that, they should have used .forceToDisk or forceToDiskExecution.

johnynek:
@ttim @dieu can you all take a look?

stephbian:
nice tests. i'm excited to see whether switching the persist mode will fix the maxResultSize exceptions i'm seeing.

johnynek:
Thanks for the review @stephbian ! I'll publish an internal version tomorrow and we can see if the persist changes are suitable for us.

PR #1903, opened 2019-03-22
Add configurable partition count for spark

stephbian:
@johnynek and I paired to add some configuration so that we can limit the number of partitions and reducers used when running a job using scalding-spark. Our hope is that this will allow us to reduce the total size of results that are being sent back to the driver, which is currently causing us some pain.

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
@@ -8,6 +8,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 import com.twitter.scalding.{Config, FutureCache}
 import com.twitter.scalding.typed.TypedSource
+import SparkPlanner.PartitionComputer
 
 sealed abstract class Op[+A] {
   import Op.{Transformed, fakeClassTag}
@@ -35,9 +36,9 @@ object Op extends Serializable {
       Transformed[(K, V), (K, U)](op, _.flatMapValues(fn))
     def mapValues[U](fn: V => U): Op[(K, U)] =
       Transformed[(K, V), (K, U)](op, _.mapValues(fn))
-    def mapGroup[U](fn: (K, Iterator[V]) => Iterator[U])(implicit ordK: Ordering[K]): Op[(K, U)] =
+    def mapGroup[U](partitionComputer: PartitionComputer)(fn: (K, Iterator[V]) => Iterator[U])(implicit ordK: Ordering[K]): Op[(K, U)] =
       Transformed[(K, V), (K, U)](op, { rdd: RDD[(K, V)] =>
-        val numPartitions = rdd.getNumPartitions
+        val numPartitions = partitionComputer(rdd.getNumPartitions)
         val partitioner = rdd.partitioner.getOrElse(new HashPartitioner(numPartitions))
         val partitioned = rdd.repartitionAndSortWithinPartitions(partitioner)
         partitioned.mapPartitions({ its =>
@@ -50,11 +51,11 @@ object Op extends Serializable {
     def hashJoin[U, W](right: Op[(K, U)])(fn: (K, V, Iterable[U]) => Iterator[W]): Op[(K, W)] =
       HashJoinOp(op, right, fn)
 
-    def sorted(implicit ordK: Ordering[K], ordV: Ordering[V]): Op[(K, V)] =
+    def sorted(partitionComputer: PartitionComputer)(implicit ordK: Ordering[K], ordV: Ordering[V]): Op[(K, V)] =
       Transformed[(K, V), (K, V)](op, { rdd: RDD[(K, V)] =>
         // The idea here is that we put the key and the value in
         // logical key, but partition only on the left part of the key
-        val numPartitions = rdd.getNumPartitions
+        val numPartitions = partitionComputer(rdd.getNumPartitions)
         val partitioner = rdd.partitioner.getOrElse(new HashPartitioner(numPartitions))
         val keyOnlyPartioner = KeyHashPartitioner(partitioner)
         val unitValue: RDD[((K, V), Unit)] = rdd.map { kv => (kv, ()) }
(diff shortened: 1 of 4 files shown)


scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -11,6 +11,38 @@ import scala.collection.mutable.{ Map => MMap, ArrayBuffer }
...
 object SparkPlanner {
   import SparkMode.SparkConfigMethods
+  sealed trait PartitionComputer {
+    def apply(currentNumPartitions: Int): Int
+  }
+
+  final case object IdentityPartitionComputer extends PartitionComputer {
+    def apply(currentNumPartitions: Int): Int = currentNumPartitions
+  }
+  final case class ConfigPartitionComputer(config: Config, scaldingReducers: Option[Int]) extends PartitionComputer {
```
oscar-stripe, line 31: we should write a test for this:
```scala
val pc = ConfigPartitionComputer(Config.empty, None)
forall { i: Int => if (i >= 1) assert(pc(i) == i) else assert(pc(i) == 1) }

val pc = ConfigPartitionComputer(Config.empty, Some(10))
forall { i: Int => pc(i) == 10 }

// test setReducerScaling, test setMaxPartitionCount...
```

oscar-stripe:
This looks great. Let's see if it solves our problems on the cluster, then maybe add the tests I suggest.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
```diff
@@ -155,7 +156,16 @@ object Op extends Serializable {
         for {
           l <- lrdd
           r <- rrdd
-        } yield widen[A](l) ++ widen[A](r)
+        } yield {
+          val merged = widen[A](l) ++ widen[A](r)
+          val partitions = merged.getNumPartitions
+          val newPartitions = pc(partitions)
+          if (partitions != newPartitions) {
```
ianoc-stripe: `partitions > newPartitions`
is likely the intent here, coalesce up doesn't make too much sense

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -11,6 +11,38 @@ import scala.collection.mutable.{ Map => MMap, ArrayBuffer }
...
+    def apply(currentNumPartitions: Int): Int = {
+      val maxPartitions = config.getMaxPartitionCount
+      val getReducerScaling = config.getReducerScaling.getOrElse(1.0D)
```
ianoc-stripe, line 34: This could do with some comments -- i'm not entirely sure how much sense it'll make here. Since we can hit one of these guards arbitrarily it could have little effect or lots depending a lot on the graph?

johnynek: yeah, the idea is that if you have a graph already that has explicit numbers of reducers set, you can optionally scale this since hadoop vs spark may behave differently wrt max number of tasks...

Comment is a good point.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -11,6 +11,38 @@ import scala.collection.mutable.{ Map => MMap, ArrayBuffer }
...
+      }
+      if (candidates > 0) {
+        maxPartitions match {
+          case Some(maxP) =>
+            Math.min(maxP, candidates)
+          case None =>
+            candidates
+        }
+      } else {
+        1
```
ianoc-stripe, line 53: The fall through should probably be to leave it unchanged?

johnynek: this is a case of `candidates <= 0` which can arise from bad configs from getReducerScaling.

Maybe `candidates == 0` goes to 1, else error? Negative shouldn't happen.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -11,6 +11,38 @@ import scala.collection.mutable.{ Map => MMap, ArrayBuffer }
...
+  }
+
+  final case object IdentityPartitionComputer extends PartitionComputer {
+    def apply(currentNumPartitions: Int): Int = currentNumPartitions
+  }
+  final case class ConfigPartitionComputer(config: Config, scaldingReducers: Option[Int]) extends PartitionComputer {
+    def apply(currentNumPartitions: Int): Int = {
+      val maxPartitions = config.getMaxPartitionCount
+      val getReducerScaling = config.getReducerScaling.getOrElse(1.0D)
+      val candidates = scaldingReducers match {
```
ianoc-stripe: candidates implies plural here, but from what i can see its just only one candidate possible as a result. so i'd probably update the name

stephbian:
I think I've addressed all comments here. PTAL @johnynek @ianoc

johnynek:
build failure, yay tests!

https://travis-ci.org/twitter/scalding/jobs/511118778#L3307

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBackendTests.scala
```diff
@@ -201,3 +203,34 @@ class SparkBackendTests extends FunSuite with BeforeAndAfter {
...
+    val pc = ConfigPartitionComputer(Config.empty, Some(10))
+    forAll { i: Int =>
+      pc(i) == 10
+    }
+  }
+
+  property("when reducer scaling factor given, scales the number of reducers") {
+    val pc = ConfigPartitionComputer(Config.empty.setReducerScaling(2.0), Some(10))
+    forAll { i: Int =>
+      pc(i) == 2.0D * 10
```
johnynek: I think this should be `20` not since `2.0D * 10` is a double, right, but pc is always returning an Int?

johnynek (approved):
Will merge when we get the tests green. I think the tests are good!

johnynek:
Thank you!

PR #1904, opened 2019-03-22
improve spark merge handling

johnynek:
This should be merged *after* #1903 on which this is based.

The important code here is the unrollMerges code, which results in a single Union vs a series of Unions (and coalesce) calls.

We noticed this looking at some generated graphs.

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
@@ -8,6 +8,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.reflect.ClassTag
 import com.twitter.scalding.{Config, FutureCache}
 import com.twitter.scalding.typed.TypedSource
+import SparkPlanner.PartitionComputer
 
 sealed abstract class Op[+A] {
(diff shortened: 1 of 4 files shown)


scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
```diff
@@ -144,18 +145,35 @@ object Op extends Serializable {
...
+            for {
+              l <- lrdd
+              rs <- rrdd
+            } yield {
+              val rdds = l :: rs
+              val rddAs = rdds.map(widen[A](_))
+              val merged = new UnionRDD(session.sparkContext, rddAs)
+              val partitions = merged.getNumPartitions
+              val newPartitions = pc(partitions)
+              if (partitions != newPartitions) {
```
stephanie-stripe: should this be `if (partitions > newPartitions)`? i'm intending to make this change as well.

johnynek: I'll just merge your change, but I think `newPartitions < partitions` reads better.

stephanie-stripe:
cool, this lgtm

PR #1906, opened 2019-03-27
Add a full de-diamonding optimization

johnynek:
one pattern that we see commonly on a system we have internally are graphs that look like:
```scala
val middle: TypedPipe[A] = ...
val out = (middle.map(f1) ++ middle.flatMap(f2) ++ .... ) // merging several transformations of middle
```
Currently planners will see this as a fork of middle, each are mapped, then merged back together.

But these map-only operations after the fork until the merge could be done in a single flatMap operation:

```scala
val optOut = middle.flatMap(mergeFn)
```
That is exactly what this current optimization rule does.

cc @stephbian @ianoc

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
@@ -65,7 +65,7 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
      * we will not materialize it, so both branches can't own it. Since we only emit Iterable
      * out, this may be okay because no external readers can modify, but worth thinking of
      */
-    val (id, acts) = state.update { s =>
+    val idActs: (Long, List[Action]) = state.update { s =>
       val (nextState, acts) = optimizedWrites.foldLeft((s, List.empty[Action])) {
         case (old @ (state, acts), write) =>
           write match {
@@ -108,6 +108,7 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
       }
       (nextState.copy(id = nextState.id + 1), (nextState.id, acts))
     }
+    val (id, acts) = idActs
     // now we run the actions:
     Future.traverse(acts) { fn => fn() }.map(_ => (id, ExecutionCounters.empty))
   }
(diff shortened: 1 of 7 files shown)


johnynek:
@ttim can you take a look at this?

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
```diff
@@ -644,4 +661,50 @@ class OptimizationRulesTest extends FunSuite with PropertyChecks {
 
     optimizedSteps(OptimizationRules.standardMapReduceRules, 1)(pipe)
   }
+
+  test("dediamond map only diamonds") {
```
dieu, line 665: could we have tests to cover all or most of cases (Filter, FlatMap, MapValues, etc)?

dieu (approved):

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
johnynek: yep. Happy to add more tests. That's a good call. Thank you.

johnynek: This was a good suggestion. There are some ordering issues that come into play that this detected. For instance, you don't want to push mappers up into reducers or filterKey above a group until after we have done the de-diamond, else it can make a diamond look not like a diamond by making the reduce look different.

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
```diff
@@ -65,7 +65,7 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
      * we will not materialize it, so both branches can't own it. Since we only emit Iterable
      * out, this may be okay because no external readers can modify, but worth thinking of
      */
-    val (id, acts) = state.update { s =>
+    val idActs: (Long, List[Action]) = state.update { s =>
```
johnynek, line 68: scala 2.12.7 didn't like the lack of a type annotation here.

scalding-core/src/test/scala/com/twitter/scalding/typed/OptimizationRulesTest.scala
dieu: nice!

johnynek:
@dieu on our internal test, this saved 30% CPU time and shuffle bytes. It was a job that clearly had two examples of this needless diamond, which motivated this, but you all may have a significant number of jobs at Twitter that will get a win by this optimization.

dieu:
@johnynek wow, that's awesome, but we need to encourage @ttim to another release :)

Issue #1905, opened 2019-03-25
Issue with data being materialized after `cross` and before subsequent `map` 


ttim:
In twitter we found a case when data got serialized after `cross` and before subsequent `map`. This might be critical since result of cross is usually really big and subsequent operations decrease size of it a lot. I created test for this: https://github.com/ttim/scalding/commit/85ddcff8067dc04418c975bb277ed4ccdc69a58c . Seems like it only reproducable when `.toPipe` is used and two branches happen after `.cross.map`.

Any ideas @johnynek ?

ttim:
Removing `.orElse(composeSame)` from default map reduce optimization rules fixes the test.

ttim:
I think I know what's wrong.

The test I wrote doing next:
```
 val crossedMapped = TypedPipe.from(source)
      .cross(mapPipe)
      .map { case (value, map) => map(value) }

  crossedMapped.toPipe('value).write(sink1)
  crossedMapped.map(identity).toPipe('value).write(sink2)
```
During optimization phase we optimize two maps going one after another in second fork and therefore only common part of this two graphs become cross join. Forcing this common part to disk (my guess) made by cascading itself.

I think with typed writes it works because during typed writes planning we have both writes at the same time and able to extract common part out of it to calculate it only once.

johnynek:
seems like a good guess that the interaction of toPipe is causing the issue.

That is really hiding from the optimizer what is going on.

We could put another rule that we should always avoid materializing a cross.

It's hard to say in-general when we should materialize vs not, only looking at the static graph.

We could imagine adding a new node to the graph to hint: even if we fan this out, don't recomputed it.

maybe `.ephemeral` or something, that is just a hint to the optimizer to ignore it when adding explicit Forks. I don't know how usable that will be though.

I would probably rather improve the optimization rules...

ttim:
So basically what happens is:
1. Two pipe writes get optimized separately
2. One of them has two maps subsequently which get composed by optimizer
3. The only common part of this two branches our `Pipe` `compiler` sees is `cross` result so it creates fork in cascading at this point.

Knowing this I also can make similar test without using `toPipe`:
```
    val crossed = TypedPipe.from(source).cross(mapPipe)
    crossed.map { case (value, map) => map(value) }.write(sink1)
    crossed.map { case (value, map) => map(identity(value)) }.write(sink2)
```

[...]

PR #1908, opened 2019-04-03
Always create different cascading pipes out of forked cross typed pipe

ttim:
As described in https://github.com/twitter/scalding/issues/1905 currently scalding might create forks in cascading graph after `cross` nodes. This is almost never wanted and broke one of production jobs in Twitter - job started to produce petabytes of intermediate data. 

In this PR:
- I've added two tests for the issue. One with old `.toPipe` api and another one with two different `map`s after `cross`.
- Changed `CascadingBackend` in a way it never caches cascading pipe created out of `cross` operation.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
@@ -4,7 +4,8 @@ import cascading.flow.FlowDef
 import cascading.operation.Debug
 import cascading.pipe.{ CoGroup, Each, Pipe, HashJoin }
 import cascading.tuple.{ Fields, Tuple => CTuple }
-import com.stripe.dagon.{ FunctionK, Id, Memoize, Rule, Dag }
+import com.stripe.dagon.{ FunctionK, HCache, Id, Rule, Dag }
+import com.stripe.dagon.Memoize.RecursiveK
 import com.twitter.scalding.TupleConverter.{ singleConverter, tuple2Converter }
 import com.twitter.scalding.TupleSetter.{ singleSetter, tup2Setter }
 import com.twitter.scalding.{
@@ -182,9 +183,35 @@ object CascadingBackend {
   }
   private[this] val cache = new CompilerCache
 
-  private def compile[T](mode: Mode): FunctionK[TypedPipe, CascadingPipe] =
-    Memoize.functionK[TypedPipe, CascadingPipe](
-      new Memoize.RecursiveK[TypedPipe, CascadingPipe] {
+  private def compile(mode: Mode): FunctionK[TypedPipe, CascadingPipe] =
+    new FunctionK[TypedPipe, CascadingPipe] {
+      private val cache = HCache.empty[TypedPipe, CascadingPipe]
+
+      override def toFunction[U]: TypedPipe[U] => CascadingPipe[U] = cached
+
+      private def cached[U]: TypedPipe[U] => CascadingPipe[U] = {
+        // Don't cache `CrossPipe`, but cache `left` and `right` side of it
+        case cp@CrossPipe(left, right) =>
+          notCached(excludes = Set(left, right))(cp)
+        // Don't cache `Fork` and `WithDescriptionTypedPipe`
+        // since if we do cache them `CrossPipe` will end up being cached as well
+        case tp@Fork(_) =>
+          transform(tp, this)
+        case tp@WithDescriptionTypedPipe(_, _) =>
+          transform(tp, this)
+        // Cache all other typed pipes
+        case tp =>
+          cache.getOrElseUpdate(tp, transform(tp, this))
+      }
+
+      private def notCached(excludes: Set[TypedPipe[_]]): FunctionK[TypedPipe, CascadingPipe] =
+        new FunctionK[TypedPipe, CascadingPipe] {
+          override def toFunction[U]: TypedPipe[U] => CascadingPipe[U] = { tp =>
+            if (excludes.contains(tp)) cached(tp) else transform(tp, this)
+          }
+        }
+
+      private val transform: RecursiveK[TypedPipe, CascadingPipe] = new RecursiveK[TypedPipe, CascadingPipe] {
         def toFunction[T] = {
           case (cp@CounterPipe(_), rec) =>
             def go[A](cp: CounterPipe[A]): CascadingPipe[A] = {
@@ -340,7 +367,8 @@ object CascadingBackend {
           case (CoGroupedPipe(cg), rec) =>
             planCoGroup(cg, rec)
         }
-      })
+      }
+    }
 
   private def applyDescriptions(p: Pipe, descriptions: List[(String, Boolean)]): Pipe = {
     val ordered = descriptions.collect { case (d, false) => d }.reverse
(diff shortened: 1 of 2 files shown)


ttim:
I've tried to write optimization rule to ensure the same, but:
- It's not going to cover `.toPipe` use case
- I don't know how to write it properly given current `Rule` abstraction. Basically we want to optimize `cross <- fork <- op1 & cross <- fork <- op2` to `fork <- cross <- op1 & fork <- cross <- op2`, so this rule cannot be over `fork` or `cross` nodes and should be over `op1` & `op2` nodes. Any ideas? cc/ @johnynek

johnynek:
I thought a rule like:
```
case Fork(Cross(a, b)) => Cross(Fork(a), Fork(b))
```
might be all that is needed: we always push Fork above a cross. Will that not work?

Note: we do apply optimization rules to toPipe:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala#L380

ttim:
> Note: we do apply optimization rules to toPipe

Yes, we do, but this is not going to help in this case. Issue arise because a) we cache `TypedPipe` => `CascadingPipe` mapping, b) we optimize two `.toPipe` calls independently.
First test case (https://github.com/twitter/scalding/pull/1908/files#diff-14795c9a6959be6381e99caaafba874dR1569) covers this case.
In this test case `.toPipe` called twice and after optimization the only common part of this two calls is `cross` and since it's cached it ends up being cascading's fork.


> I thought a rule like:
> ```case Fork(Cross(a, b)) => Cross(Fork(a), Fork(b))```

This I might misunderstand but I don't see how it helps. If you do this, subsequent `map`s (like in second test https://github.com/twitter/scalding/pull/1908/files#diff-14795c9a6959be6381e99caaafba874dR1615) will be dependent on the same `Cross(Fork(a), Fork(b))` so it will end up with being same fork on cascading graph.

ttim:
And I have related question - is there any way right now to force scalding to compute same typed pipe twice? Is it going to be always squashed? I.e. is it true that:
```
TypedPipe.from(source).map(fn1).map(fn2).write(sink1)
TypedPipe.from(source).map(fn1).map(fn3).write(sink2)
``` 
Always ends up with having fork after `.map(fn1)` and there is no way to force other behaviour (imagine use case where `fn1` is very cheap and doesn't shrink data)?

johnynek:
Currently no. There is no way for rerun. In this case you probably don’t want to materialize. We could choose to not materialize pure map operations if they could be helped by moving the fork above Map operations, then map fusion should do the thing we want.

johnynek:
flatMap is more dubious since it often used for filtering as well.

ttim:
> Currently no. There is no way for rerun. In this case you probably don’t want to materialize. We could choose to not materialize pure map operations if they could be helped by moving the fork above Map operations, then map fusion should do the thing we want.

I see. I think in future we might want to have `.opaque` which is opposite to `.forceToDisk` in a way.

ttim:
@johnynek tests are passing now so I think it's good to go, wdyt?

As a side note I found that some counters (like `org.apache.hadoop.mapreduce.FileSystemCounter#FILE_BYTES_WRITTEN`) aren't properly nullified between runs with `runHadoop` =(.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -182,9 +183,35 @@ object CascadingBackend {
...
+      }
+
+      private def notCached(excludes: Set[TypedPipe[_]]): FunctionK[TypedPipe, CascadingPipe] =
+        new FunctionK[TypedPipe, CascadingPipe] {
+          override def toFunction[U]: TypedPipe[U] => CascadingPipe[U] = { tp =>
+            if (excludes.contains(tp)) cached(tp) else transform(tp, this)
+          }
+        }
+
+      private val transform: RecursiveK[TypedPipe, CascadingPipe] = new RecursiveK[TypedPipe, CascadingPipe] {
```
johnynek: I’m worried this is causing nothing below here to be cached. Note the use of rec here used to cache. Now it has been disabled for everything in here.

I think this is a pretty large difference from the current behavior and basically only the outer most pipe will be cache. I don’t think that’s what we want.

ttim: I guess code isn't 100% clear, but everything apart from `Cross` implementation (cross itself, and not `left` and `right` arguments to it), `Fork` and `WithDescriptionTypedPipe` will be cached.

`transform` here is pure implementation of cascading backend tranformation without any caching logic applied. On top of that caching logic applied (in `cached`) which is:
- don't cache `Fork` and `WithDescriptionTypedPipe`
- don't cache `Cross`, but cache `left` and `right` branches of it
- cache everything else

johnynek: yeah, I think adding comments will help. There are two layers of caching here that is confusing.

1. the first cache is the explicit HCache
2. the second cache is the RecursiveK which caches each call.

So, I think you are right, the transform is caching, but I think nested Cross *will* be cached by transform.

so `a.cross(b).cross(c)` I think will hit the inner cross inside transform, which will be cached.

Do you think that's not right?

ttim: I might know where your confusion comes from: before this code was using `Memoize` over `RecursiveK`. Now this `transform` function isn't caching at all and I basically reimplemented it's caching ability explicitly with a bit different policy. And since core transforming function is still `RecursiveK` by essence I reused it from `Memoize`.

In the case you wrote - first cross will be not cached, then it will use `notCached(Set(a.cross(b)), c))` which leads to `cached(a.cross(b))` which leads to *not* caching inner cross as well. I guess I should rename `.cached` to `.withCachePolicy` or something like this.

johnynek: okay, yeah. I was getting confused with `new RecursiveK` thinking it was caching. It isn't. I see it is not controlling how the recursion is done.

Instead, you call it with `transform(tp, this)` which recurses using the entire outer strategy.

Yeah, I now think I understand and think it is correct, but I think it is a bit subtle. Can we add a comment that describes and summarizes the points of confusion and clears them up. We are going be confused later:

1. why are we doing this? We are avoiding caching certain node types because they make bad plans that are usually worse.
2. how are we achieving this? We are wiring up this kind of co-recursive check so that the recursion has a branch and is sure not to cache certain types of nodes...

something like this.

ttim: Removed `RecursiveK` usage. I think it's more straightforward and simpler this way. Also added comment explaining motivation for this logic.

ttim:
@johnynek do you mind to take a look on updated version?

johnynek (approved):
Sorry for the latency. The new version is clearer to me now, thanks.

It would really be ideal to have a better algorithm for estimating if we should materialize or not a particular pipe.

ttim:
@johnynek I have a feeling like at least all `hashJoin`s should end up being not materialized. And at most everything?

johnynek:
I don’t think it is clear. HashJoins can often be used to filter, and I think we generally would want to materialize filters that can out.

I think we certainly want to materialize joins that fan out, no? That would be crazy to recompute joins wouldn’t it be?

johnynek:
Maybe the rule is as simple as never materialize map-only operations.

PR #1911, opened 2019-04-17
Solve optimization of zipped executions

johnynek:
previously we could only merge two writes if they were directly zipped. If you zip with `Execution.unit` you ruin this behavior.

Secondly, we weren't applying execution optimizations after flatmap-like functions, which reduces the effectiveness if there are any flatMaps in your system.

We hit a bug that we really wanted to merge Executions but it was not happening. This should fix it for us.

@ttim @dieu can you all take a look?

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -728,6 +728,7 @@ object Execution {
      * Here we inline the map operation into the presentation function so we can zip after map.
      */
     override def map[U](mapFn: T => U): Execution[U] =
+
       WriteExecution(head,
         tail,
         ExecutionOptimizationRules.MapWrite.ComposeMap(result, mapFn))
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -151,13 +151,7 @@ sealed trait Execution[+T] extends Serializable { self: Product =>
     val ec = new EvalCache(writer)
     val confWithId = conf.setScaldingExecutionId(UUID.randomUUID.toString)
 
-    val exec =
-      if (conf.getExecutionOptimization) {
-        ExecutionOptimizationRules.stdOptimizations(this)
-      } else {
-        this
-      }
-
+    val exec = Execution.optimize(conf, this)
     // get on Trampoline
     val result = exec.runStats(confWithId, mode, ec)(cec).get.map(_._1)
     // When the final future in complete we stop the submit thread
@@ -305,6 +299,13 @@ object Execution {
     }
   }
 
+  private def optimize[A](conf: Config, ex: Execution[A]): Execution[A] =
+    if (conf.getExecutionOptimization) {
+      ExecutionOptimizationRules.stdOptimizations(ex)
+    } else {
+      ex
+    }
+
   /**
    * This is an instance of Monad for execution so it can be used
    * in functions that apply to all Monads
@@ -411,7 +412,9 @@ object Execution {
         cache.getOrElseInsert(conf, this,
           for {
             (s, st1) <- fut1
-            next = fn(s)
+            next0 = fn(s)
+            // next0 has not been optimized yet, we need to try
+            next = optimize(conf, next0)
             (t, st2) <- Trampoline.call(next.runStats(conf, mode, cache)).get
           } yield (t, st1 ++ st2))
       }
@@ -496,7 +499,11 @@ object Execution {
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
       Trampoline.call(prev.runStats(conf, mode, cache)).map { fut =>
         cache.getOrElseInsert(conf, this,
-          fut.recoverWith(fn.andThen(_.runStats(conf, mode, cache).get)))
+          fut.recoverWith(fn.andThen { ex0 =>
+            // we haven't optimized ex0 yet
+            val ex = optimize(conf, ex0)
+            ex.runStats(conf, mode, cache).get
+          }))
       }
   }
 
@@ -582,7 +589,10 @@ object Execution {
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) = {
       Trampoline(cache.getOrElseInsert(conf, this, {
         val (uid, nextConf) = conf.ensureUniqueId
-        fn(uid).runStats(nextConf, mode, cache).get
+        val next0 = fn(uid)
+        // next0 has not been optimized yet, we need to try
+        val next = optimize(conf, next0)
+        next.runStats(nextConf, mode, cache).get
       }))
     }
   }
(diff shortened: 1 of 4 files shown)


dieu:
looks good from my side.

cc @ttim

johnynek:
let's not merge just yet, I have some tests I want to add, I think this still may not be enough.

I'd like to finally get this solved for good.

johnynek:
@dieu @ttim please take another look.

I found a failing unit test with my previous approach, and I realized it was still playing whack a mole. So, I took a step back, and thought about how to totally solve the problem.

While much more complex, I believe the current PR totally solves the problem of merging WriteExecutions. I think we now have the property that (other than flatMaps) there is at most 1 WriteExecution in each Execution. Since we apply the optimizations after each flatMap application now, I think this is the best we can do.

I've tried to comment the code, but feel free to ask any questions.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -108,24 +109,368 @@ object ExecutionOptimizationRules {
         }
     }
 
-  object ZipWrite extends PartialRule[Execution] {
-    type InputType = (Config, Mode, Execution.Writer, ConcurrentExecutionContext)
-    type ResFun[T] =  InputType => Future[T]
+  sealed abstract class HList
```
johnynek: I didn't want to add a dependency to shapeless just to get a simple HList, so I implemented this one.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -108,24 +109,346 @@ object ExecutionOptimizationRules {
...
+       * is something scala has a hard time with.
+       */
+      final case class Many[+Ex[x] <: Execution[x], A, T, +C](head: Ex[A], rest: FlattenedZip[Execution, T], sub: SubTypes[(A, T), C]) extends FlattenedZip[Ex, C] {
+        type H = A :: rest.H
+        type ExH = Execution[A] :: rest.ExH
+        def executions = ::(head, rest.executions)
+        def toExecution(ex: Execution[A] :: rest.ExH): Execution[A :: rest.H] =
+          Mapped(Zipped(ex.head, rest.toExecution(ex.tail)), Cons2Fn())
+
+        def fn: (A :: rest.H) => C = { cons =>
```
johnynek: this is an anonymous function. We should remove this an use a case class function.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -777,19 +788,24 @@ object Execution {
 
 
     /*
+     * This is such an important optimization, that we apply it locally.
```
ianoc-stripe, line 799: could you say why we need to do it locally rather than the later rules having the same effect?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -108,24 +109,346 @@ object ExecutionOptimizationRules {
...
+    /*
+     * Some functions to deal with HLists
+     */
+    case class ConsFn[A, B <: HList](b: B) extends Function1[A, A :: B] {
+      def apply(a: A): A :: B = ::(a, b)
     }
+    case class Cons2Fn[A, B <: HList]() extends Function1[(A, B), A :: B] {
+      def apply(ab: (A, B)): A :: B = ::(ab._1, ab._2)
+    }
+    case class HeadFn[A]() extends Function1[A :: HNil.type, A] {
```
ianoc-stripe: this head would only work on a single element list right?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -108,24 +109,346 @@ object ExecutionOptimizationRules {
...
+      import HList._
+
+      /**
+       * Here is the simplest FlattenZip: just a single Execution
+       */
+      final case class Single[+Ex[x] <: Execution[x], A](ex: Ex[A]) extends FlattenedZip[Ex, A] {
+        type H = A :: HNil.type
+        type ExH = Execution[A] :: HNil.type
+        def executions = ::(ex, HNil)
+        def toExecution(ex: Execution[A] :: HNil.type): Execution[A :: HNil.type] =
```
ianoc-stripe: minor nit, but i'd use another variable name here since your shadowing the ex from the case class definition itself.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -108,24 +109,346 @@ object ExecutionOptimizationRules {
...
+                sub0.liftCo[F](flat0)
+              }
+              go(m)
+            case MapZip(left, fn) =>
+              zip(left).map(ZipMap.MapRight(fn))
+          }
+        def optimize: Option[FlattenedZip[WriteExecution, A]] =
+          ex match {
+            case w@WriteExecution(_, _, _) => Some(Single(w))
+            case _ => None
```
ianoc-stripe, line 269: For this my understanding is that you can never hit this optimize from the external call site so it returning none isn't lossy? (it looks like for many if it sees a None on a single it would keep it around as the rest).
If you didn't guard the outer with:
```
          wc = flat.writeCount
          // only optimize if there is more the 1 write, otherwise we create an infinite loop
          if (wc > 1)
```
this could be an issue? (just trying to make sure i understand whats happening)

johnynek: I'll add a comment, but we call optimize recursively, and some of those branches can hit None. Also note the type it returns: `Option[FlattenedZip[WriteExecution, A]]` so, you have no way of making `FlattenedZip[WriteExecution, A]` if there is not at least one `WriteExecution` in the zip, since that type requires the first execution be a write.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
johnynek: yep, will do.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
johnynek: yeah, I only use it that way, but I think it will be clearer actually if I make it generic on the tail too, and avoid the `HNil.type` which is a bit confusing.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: will do. The reason is Execution optimization is optional, but by default we do it. Very giant executions could blow the stack while optimizing, and I guess they saw some examples of that at Twitter. So, by config you can disable optimization. But I didn't want to regress on the old style local optimizations that we were doing before by removing this.

In dagon 0.3.2 I *think* I have fixed those stack overflows, but we don't use that yet (because I haven't published it...)

ianoc-stripe (approved):

oscar-stripe:
I think this is good to go. Any comments @ttim @dieu ?

johnynek:
Pinging @ttim you haven’t weighed in yet.

I think this is a big improvement which I hope can really improve plans for many users.

ttim:
@johnynek I'm going through it right now. It would be nice to separate change where you apply execution optimizations of `flatMap`s & co and where actual zip optimization happens, but this is optional.

johnynek:
Simplified after talking to @ttim -- I think @ttim may have an even simpler approach which may follow up with as a PR, but this seems like strictly better than our current situation.

PR #1913, opened 2019-04-24
Simplify Execution's write optimization

ttim:
This is follow up to https://github.com/twitter/scalding/pull/1911 to simplify merge of write executions.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
@@ -242,11 +242,18 @@ object ExecutionOptimizationRules {
      * into a single value. If ex is already optimal (0 or 1 write) return None
      */
     def optimize[A](ex: Execution[A]): Option[Execution[A]] = {
-      def writes(ex: Execution[_]): Int = ex match {
-        case Zipped(left, right) => writes(left) + writes(right)
-        case Mapped(that, _) => writes(that)
-        case WriteExecution(_, _, _) => 1
-        case _ => 0
+      def writes(execution: Execution[_]): Int = {
+        @tailrec
+        def loop(executions: List[Execution[_]], acc: Int): Int = executions match {
+          case Nil => acc
+          case head :: tail => head match {
+            case Zipped(left, right) => loop(left :: right :: tail, acc)
+            case Mapped(that, _) => loop(that :: tail, acc)
+            case WriteExecution(_, _, _) => loop(tail, acc + 1)
+            case _ => loop(tail, acc)
+          }
+        }
+        loop(execution :: Nil, 0)
       }
       // only optimize if there are 2 or more writes, otherwise we create an infinite loop
       if (writes(ex) > 1)
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
@@ -141,7 +141,10 @@ object ExecutionOptimizationRules {
       def apply(in: (A, (B, C))) =
         (in._2._1, (in._1, in._2._2))
     }
-    case class ComposedFn[A1, A2, A, B1, B2, B](fn1: ((A1, A2)) => A, fn2: ((B1, B2)) => B) extends Function1[((A1, B1), (A2, B2)), (A, B)] {
+    case class ComposedFn[A1, A2, A, B1, B2, B](
+      fn1: Function1[(A1, A2), A],
+      fn2: Function1[(B1, B2), B]
+    ) extends Function1[((A1, B1), (A2, B2)), (A, B)] {
       override def apply(v1: ((A1, B1), (A2, B2))): (A, B) = (fn1(v1._1._1, v1._2._1), fn2(v1._1._2, v1._2._2))
     }
 
--- a/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
@@ -1,7 +1,8 @@
 package com.twitter.scalding
 
 import com.stripe.dagon.{Dag, FunctionK, Literal, Memoize, PartialRule, Rule}
-import com.twitter.scalding.typed.functions.{ ComposedFunctions, SubTypes, Swap }
+import com.twitter.scalding.ExecutionOptimizationRules.ZipMap.{MapLeft, MapRight}
+import com.twitter.scalding.typed.functions.{ComposedFunctions, Identity, Swap}
 import scala.annotation.tailrec
 import scala.concurrent.{Future, ExecutionContext => ConcurrentExecutionContext}
 
@@ -139,6 +140,9 @@ object ExecutionOptimizationRules {
       def apply(in: (A, (B, C))) =
         (in._2._1, (in._1, in._2._2))
     }
+    case class ComposedFn[A1, A2, A, B1, B2, B](fn1: ((A1, A2)) => A, fn2: ((B1, B2)) => B) extends Function1[((A1, B1), (A2, B2)), (A, B)] {
+      override def apply(v1: ((A1, B1), (A2, B2))): (A, B) = (fn1(v1._1._1, v1._2._1), fn2(v1._1._2, v1._2._2))
+    }
 
     case class ComposeWriteFn[A, B, C, D, E](
       fn1: ((A, B, C, ConcurrentExecutionContext)) => Future[D],
@@ -147,10 +151,23 @@ object ExecutionOptimizationRules {
       def apply(tup: (A, B, C, ConcurrentExecutionContext)): Future[(D, E)] =
         (Execution.failFastZip(fn1(tup), fn2(tup))(tup._4))
     }
+
+    case class MapWriteFn[A, B, C, D, E](
+      fn1: ((A, B, C, ConcurrentExecutionContext)) => Future[D],
+      mapFn: D => E) extends Function1[(A, B, C, ConcurrentExecutionContext), Future[E]] {
+
+      def apply(tup: (A, B, C, ConcurrentExecutionContext)): Future[E] = fn1(tup).map(mapFn)(tup._4)
+    }
+
     def mergeWrite[A, B](w1: WriteExecution[A], w2: WriteExecution[B]): WriteExecution[(A, B)] = {
       val newFn = ComposeWriteFn(w1.result, w2.result)
       WriteExecution(w1.head, w1.tail ::: (w2.head :: w2.tail), newFn)
     }
+    def mapWrite[A, B](w: WriteExecution[A], fn: A => B): WriteExecution[B] = {
+      WriteExecution(w.head, w.tail, MapWriteFn(w.result, fn))
+    }
+
+    // either NonWrite(ex: Execution), Write(ex: Execution), composed: NonWrite, Write
 
     /**
      * This is the fundamental type we use to optimize zips, basically we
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -147,10 +151,23 @@ object ExecutionOptimizationRules {
...
+
     def mergeWrite[A, B](w1: WriteExecution[A], w2: WriteExecution[B]): WriteExecution[(A, B)] = {
       val newFn = ComposeWriteFn(w1.result, w2.result)
       WriteExecution(w1.head, w1.tail ::: (w2.head :: w2.tail), newFn)
     }
+    def mapWrite[A, B](w: WriteExecution[A], fn: A => B): WriteExecution[B] = {
+      WriteExecution(w.head, w.tail, MapWriteFn(w.result, fn))
+    }
+
+    // either NonWrite(ex: Execution), Write(ex: Execution), composed: NonWrite, Write
```
johnynek: This comment seems stray. Can we move it closer to where we want it?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -167,250 +184,69 @@ object ExecutionOptimizationRules {
...
-          Mapped(flat.toExecution(ex), mapFn)
+      private def zipNonWriteComposed[A, B1, B2, B](left: NonWrite[A], right: Composed[B1, B2, B]): Composed[B1, (B2, A), (A, B)] =
+        Composed(right.write, right.nonWrite.zip(left.nonWrite),
+          UnTwist().andThen(MapLeft[(B1, B2), A, B](right.compose)).andThen(Swap()))
 
-        def zip[C](that: FlattenedZip[Execution, C]): FlattenedZip[Execution, (B, C)] =
-          MapZip(flat.zip(that), ZipMap.MapLeft[A, C, B](mapFn))
+      private def zipWriteComposed[A, B1, B2, B](left: Write[A], right: Composed[B1, B2, B]): Composed[(A, B1), B2, (A, B)] =
+        Composed(mergeWrite(left.write, right.write), right.nonWrite,
+          Twist().andThen(MapRight[A, (B1, B2), B](right.compose)))
```
johnynek: andThen allocates a new function. Can we use one from the functions file?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -139,6 +140,9 @@ object ExecutionOptimizationRules {
       def apply(in: (A, (B, C))) =
         (in._2._1, (in._1, in._2._2))
     }
+    case class ComposedFn[A1, A2, A, B1, B2, B](fn1: ((A1, A2)) => A, fn2: ((B1, B2)) => B) extends Function1[((A1, B1), (A2, B2)), (A, B)] {
```
johnynek: This seems a bit complex. Do we need the tuples? Isn’t this just A => B with C => D to make ((A, C)) => (B, D) feels like the extra tuple types are distracting here.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -147,10 +151,23 @@ object ExecutionOptimizationRules {
       def apply(tup: (A, B, C, ConcurrentExecutionContext)): Future[(D, E)] =
         (Execution.failFastZip(fn1(tup), fn2(tup))(tup._4))
     }
+
+    case class MapWriteFn[A, B, C, D, E](
```
johnynek: I feel like we already have something like this. I’m on my phone so I can’t find it now easily but I thought we already had a case class function for map + write.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -167,250 +184,69 @@ object ExecutionOptimizationRules {
...
-       */
-      final case class MapZip[Ex[x] <: Execution[x], A, B](flat: FlattenedZip[Ex, A], mapFn: A => B) extends FlattenedZip[Ex, B] {
-        type ExH = flat.ExH
-
-        def executions: ExH = flat.executions
-        def toExecution(ex: ExH): Execution[B] =
-          Mapped(flat.toExecution(ex), mapFn)
+      private def zipNonWriteComposed[A, B1, B2, B](left: NonWrite[A], right: Composed[B1, B2, B]): Composed[B1, (B2, A), (A, B)] =
+        Composed(right.write, right.nonWrite.zip(left.nonWrite),
+          UnTwist().andThen(MapLeft[(B1, B2), A, B](right.compose)).andThen(Swap()))
```
johnynek: I think andThen allocates can we use a case class function here?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -167,250 +184,69 @@ object ExecutionOptimizationRules {
...
-              zip(left).map(ZipMap.MapRight(fn))
-          }
-        def optimize: Option[FlattenedZip[WriteExecution, A]] =
-          ex match {
-            case w@WriteExecution(_, _, _) => Some(Single(w))
-            case _ => None
-          }
+      final case class NonWrite[T](nonWrite: Execution[T]) extends FlattenedZip[T]
+      final case class Write[T](write: WriteExecution[T]) extends FlattenedZip[T]
+      final case class Composed[T1, T2, T](write: WriteExecution[T1], nonWrite: Execution[T2], compose: (((T1, T2)) => T)) extends FlattenedZip[T]
```
johnynek: Why the extra parens around compose? Can we remove?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -167,250 +184,69 @@ object ExecutionOptimizationRules {
...
+          Composed(right.write, left.nonWrite, Swap[B, A]())
+        case (left@NonWrite(_), right@Composed(_, _, _)) =>
+          zipNonWriteComposed(left, right)
+
+        case (left@Write(_), right@NonWrite(_)) =>
+          Composed(left.write, right.nonWrite, Identity[(A, B)]())
+        case (left@Write(_), right@Write(_)) =>
+          Write(mergeWrite(left.write, right.write))
+        case (left@Write(_), right@Composed(_, _, _)) =>
+          zipWriteComposed(left, right)
```
johnynek, line 209: Can we inline these? I think it is harder to read to have to jump over to read the methods but they are one liners.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Removed

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Replaced with `ComposedMapFn`

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Replaced by `ComposedMapFn`

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Changed to Function1

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Inlined everything apart from `zipWriteComposed` & `zipNonWriteComposed`. Two reasons: it used twice and inlined version doesn't go through scala's type system

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Not exactly because it's also do twist around the arguments.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Didn't find it too, but I might missed something.

johnynek: https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L742

Looks like it is defined in the MapWrite optimization rule.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -147,270 +155,92 @@ object ExecutionOptimizationRules {
...
+      def apply(tup: (A, B, C, ConcurrentExecutionContext)): Future[E] = fn1(tup).map(mapFn)(tup._4)
+    }
+
     def mergeWrite[A, B](w1: WriteExecution[A], w2: WriteExecution[B]): WriteExecution[(A, B)] = {
       val newFn = ComposeWriteFn(w1.result, w2.result)
       WriteExecution(w1.head, w1.tail ::: (w2.head :: w2.tail), newFn)
     }
+    def mapWrite[A, B](w: WriteExecution[A], fn: A => B): WriteExecution[B] = {
+      WriteExecution(w.head, w.tail, MapWriteFn(w.result, fn))
+    }
```
johnynek: no need for the `{ }`, maybe remove?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -147,270 +155,92 @@ object ExecutionOptimizationRules {
...
-        type ExH = flat.ExH
-
-        def executions: ExH = flat.executions
-        def toExecution(ex: ExH): Execution[B] =
-          Mapped(flat.toExecution(ex), mapFn)
-
-        def zip[C](that: FlattenedZip[Execution, C]): FlattenedZip[Execution, (B, C)] =
-          MapZip(flat.zip(that), ZipMap.MapLeft[A, C, B](mapFn))
+      private def zipNonWriteComposed[A, B1, B2, B](left: NonWrite[A], right: Composed[B1, B2, B]): Composed[B1, (B2, A), (A, B)] =
+        Composed(right.write, right.nonWrite.zip(left.nonWrite),
```
johnynek, line 221: I wonder if the function get simpler if we do `left.nonWrite.zip(right.nonWrite)` since you have to move the `A` fewer steps to the left?

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
```diff
@@ -419,16 +249,17 @@ object ExecutionOptimizationRules {
...
     def optimize[A](ex: Execution[A]): Option[Execution[A]] = {
-      val flat = FlattenedZip(ex)
-      val wc = flat.writeCount
-      // only optimize if there are 2 or more writes, otherwise we create an infinite loop
-      if (wc > 1) {
-        // unless there is a bug, this optimized will be defined
-        // but internally, optimize can return None. Just to be
-        // cleaner we use flatMap here rather than getOrElse(sys.error("unreachable"))
-        flat.optimize.map(_.execution)
+      def writes(ex: Execution[_]): Int = ex match {
```
johnynek: can we make this `@annotation.tailrec` so we can check this without blowing the stack?

e.g.

```scala
@annotation.tailrec
def writes(ex: List[Execution[Any]], acc: Int): Int =
  ex match {
    case Nil => acc
    case h :: tail =>
     ...
```

johnynek (commented):
A few more comments. This is a good simplification. Nice work!

My design had two phases: constructing FlattenedZip, then optimizing it. Here, we combine the two by making the AST always keep Write distinct, and it makes it a bit simpler to implement.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Done

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Done

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: I've tried it but it doesn't work because of `zip` - it sums two branches so `tailrec` isn't applicable.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
ttim: Then you'll have `Composed[B1, (A, B2), (A, B)]` and i.e. `(B1, (A, B2))` => `(A, B)` which is unclear how to untwist simpler.

scalding-core/src/main/scala/com/twitter/scalding/ExecutionOptimizationRules.scala
johnynek: I think this works:
```scala
@annotation.tailrec
def writes(ex: List[Execution[Any]], acc: Int): Int =
  ex match {
    case Nil => acc
    case h :: tail =>
      h match {
        case Zipped(l, r) => writes(l :: r :: tail, acc)
        case Mapped(e, _) => writes(e :: tail, acc)
        case WriteExecution(_, _, _) => writes(tail, acc + 1)
        case _ => writes(tail, acc)
  }
```
then we can define `def writes(ex: Execution[Any]): Int = writes(ex :: Nil, 0)`

Won't that work?

ttim: It works, thanks! This is interesting way of making recursion tail. Really liked it!

johnynek (approved):

PR #1914, opened 2019-04-25
Use elements hash code in `RandomNextInt` generator

ttim:
In twitter we use `.shard` to make reshuffle data. Sometimes most of the input files for `.shard` are extremely small (like 100 items) and since all mappers reuse random generator for shard all this items ends up on the same reducers and led to skew in output files.

In this review I changed `RandomNextInt` to xor with elements hashCode and therefore while better handle this kind of cases also be more or less deterministic.


athuras:
the pain that lead to this was unspeakable.

johnynek (commented):
Previously this only returned a number between 0 and modulus. Now it can be anything. Will that cause a problem? Seems a bit worrying.

ttim:
@johnynek changed it to return from 0 to modulus

johnynek (approved):

PR #1915, opened 2019-05-02
Adds support for blob type columns in TypedJdbcSource

navinvishy:
This patch adds support for `BLOB` columns in `TypedJdbcSource`. 
* Blob type columns are mapped to `Array[Byte]`. 
* This follows the convention used in other ORM tools such as Hibernate that maps blobs to a Java `byte[]`.

### DIFF
--- a/scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
+++ b/scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
@@ -58,6 +58,8 @@ object ColumnDefinitionProviderImpl {
       oTpe match {
         // String handling
         case tpe if tpe =:= typeOf[String] => StringTypeHandler(c)(accessorTree, fieldName, defaultValOpt, annotationInfo, nullable)
+        case tpe if tpe =:= typeOf[Array[Byte]] => BlobTypeHandler(c)(accessorTree, fieldName, defaultValOpt, annotationInfo, nullable)
+        case tpe if tpe =:= typeOf[Byte] => NumericTypeHandler(c)(accessorTree, fieldName, defaultValOpt, annotationInfo, nullable, "TINYINT")
         case tpe if tpe =:= typeOf[Short] => NumericTypeHandler(c)(accessorTree, fieldName, defaultValOpt, annotationInfo, nullable, "SMALLINT")
         case tpe if tpe =:= typeOf[Int] => NumericTypeHandler(c)(accessorTree, fieldName, defaultValOpt, annotationInfo, nullable, "INT")
         case tpe if tpe =:= typeOf[Long] => NumericTypeHandler(c)(accessorTree, fieldName, defaultValOpt, annotationInfo, nullable, "BIGINT")
@@ -208,6 +210,7 @@ object ColumnDefinitionProviderImpl {
           case "BIGINT" =>
             q"""List("INTEGER", "INT", "BIGINT", "INT8", "SMALLINT",
                "TINYINT", "SMALLINT", "MEDIUMINT").contains($typeNameTerm)"""
+          case "DATETIME" => q"""List("DATE","DATETIME","TIMESTAMP").contains($typeNameTerm)"""
           case f => q"""$f == $typeNameTerm"""
         }
         val typeAssert = q"""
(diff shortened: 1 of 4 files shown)


scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/handler/BlobTypeHandler.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.twitter.scalding.db.macros.impl.FieldName
+
+object BlobTypeHandler {
+  def apply[T](c: Context)(implicit accessorTree: List[c.universe.MethodSymbol],
+    fieldName: FieldName,
+    defaultValue: Option[c.Expr[String]],
+    annotationInfo: List[(c.universe.Type, Option[Int])],
+    nullable: Boolean): scala.util.Try[List[ColumnFormat[c.type]]] = {
+
+    assert(defaultValue.isEmpty)
```
ttim: nit: add message that default values and annotation info aren't supported for blob fields

scalding-db/src/test/scala/com/twitter/scalding/db/macros/MacrosUnitTests.scala
```diff
@@ -64,8 +63,8 @@ case class ExhaustiveJdbcCaseClass(
   @size(2051)@varchar forcedVarChar: String, // Forced inline to table -- only some sql version support > 255 for varchar
   myDateWithTime: Date, // Default goes to MySQL DateTime/Timestamp so its not lossy
   @date myDateWithoutTime: Date,
```
ttim, line 65: Add `timestamp` and `tinyint` to test case

ttim (approved):
Couple of nits but looks good overall! Thanks!

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
```diff
@@ -240,8 +243,10 @@ object ColumnDefinitionProviderImpl {
         val (box: Option[Tree], primitiveGetter: Tree) = cf.fieldType match {
           case "VARCHAR" | "TEXT" =>
             (None, q"""$rsTerm.getString($fieldName)""")
-          case "BOOLEAN" | "TINYINT" =>
```
gerashegalov: the reason for this is probably that in mysql `boolean` is the synonym of `tinyint(1)`.

navinvishy: Yes, but shouldn't this be the right way to handle a case class with a `Byte` field?

navinvishy: @gerashegalov wanted to follow up with you on this: If the field in the case class is declared as `Boolean` and the column in the DB is declared as a `TINYINT` things should work as before. However, if the case class declares the same field as a `Byte` instead, I've changed the implementation to handle the column in the DB as a byte since a `TINYINT` column can hold values in the range -128 to +127 (signed).

gerashegalov: @navinvishy tbh I was just giving my 2c for what I think the background was for the common treatment of BOOLEAN and TINYINT. If I can make a suggestion, I'd recommend to leave this change to another PR . It does not align with the title and intent to add support for blobs.

navinvishy: @gerashegalov agreed. I've updated the PR to include only support for blobs. I'll add the support for tinyint and timestamp fields in a separate PR.

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/handler/BlobTypeHandler.scala
```diff
@@ -0,0 +1,20 @@
...
+import com.twitter.scalding.db.macros.impl.FieldName
+
+object BlobTypeHandler {
+  def apply[T](c: Context)(implicit accessorTree: List[c.universe.MethodSymbol],
+    fieldName: FieldName,
+    defaultValue: Option[c.Expr[String]],
+    annotationInfo: List[(c.universe.Type, Option[Int])],
+    nullable: Boolean): scala.util.Try[List[ColumnFormat[c.type]]] = {
+
+    assert(defaultValue.isEmpty, "Default values are not supported for blob fields")
```
gerashegalov: should these be `require` . Not only because it's about params rather than state, but it sounds like you want this checks even if we compile assertions away

gerashegalov (approved):
LGTM, minor comment

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/handler/BlobTypeHandler.scala
navinvishy: Yes, I think `require` is more appropriate in this case. I'll make that change.

johnynek:
closed by #1915

PR #1918, opened 2019-05-17
Type independent gens for TypedPipe/Execution

dieu:
Generators of TypedPipe/Execution inspired by @erik-stripe talk http://plastic-idolatry.com/erik/oslo2019.pdf

### DIFF
--- a/scalding-core/src/test/scala/com/twitter/scalding/ExecutionOptimizationRulesTest.scala
+++ b/scalding-core/src/test/scala/com/twitter/scalding/ExecutionOptimizationRulesTest.scala
@@ -8,6 +8,8 @@ import cascading.tuple.{Fields, Tuple}
 import com.stripe.dagon.{Dag, Rule}
 import com.twitter.maple.tap.MemorySourceTap
 import com.twitter.scalding.typed.TypedPipeGen
+import com.twitter.scalding.typed.gen
+import com.twitter.scalding.typed.gen.{ExecutionGen, TypeGen, TypeWith}
 import java.io.{InputStream, OutputStream}
 import java.util.UUID
 import org.scalacheck.{Arbitrary, Gen}
@@ -98,8 +100,18 @@ class ExecutionOptimizationRulesTest extends FunSuite with PropertyChecks {
   def write(pipe: Gen[TypedPipe[Int]]): Gen[Execution[TypedPipe[Int]]] =
     pipe.map(_.writeThrough(new MemorySource[Int]()))
 
+  def write(t: TypeWith[TypeGen])(pipe: Gen[TypedPipe[_]]): Gen[Execution[TypedPipe[_]]] = {
+    implicit val tc = t.evidence.tupleConverter
+    pipe.map(_.writeThrough(new MemorySource()))
+  }
+
   val mappedOrFlatMapped =
     Gen.oneOf(mapped(pipe), flatMapped(pipe))
+
+  val stdZippedWrites =
+    Gen.zip(gen.TypedPipeGen.pipeOf, gen.TypedPipeGen.pipeOf).flatMap { case ((l, tl), (r, tr)) =>
+      zipped(write(tl)(l), write(tr)(r))
+    }
 
   val zippedWrites =
     zipped(write(TypedPipeGen.genWithIterableSources), write(TypedPipeGen.genWithIterableSources))
@@ -175,7 +187,7 @@ class ExecutionOptimizationRulesTest extends FunSuite with PropertyChecks {
   }
 
   test("randomly generated executions trees are invertible") {
-    forAll(genExec) { exec =>
+    forAll(ExecutionGen.executionOf) { case (exec, _) =>
       invert(exec)
     }
   }
@@ -183,7 +195,7 @@ class ExecutionOptimizationRulesTest extends FunSuite with PropertyChecks {
   test("optimization rules are reproducible") {
     implicit val generatorDrivenConfig: PropertyCheckConfiguration = PropertyCheckConfiguration(minSuccessful = 500)
 
-    forAll(genExec, genRule) { (exec, rule) =>
+    forAll(ExecutionGen.executionOf, genRule) { case ((exec, _), rule) =>
       val optimized = ExecutionOptimizationRules.apply(exec, rule)
       val optimized2 = ExecutionOptimizationRules.apply(exec, rule)
       assert(optimized == optimized2)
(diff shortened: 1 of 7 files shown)


johnynek:
This is great!

I’ll review in detail this week. Exciting to have this level of test coverage.

Cc @non

scalding-core/src/test/scala/com/twitter/scalding/ExecutionOptimizationRulesTest.scala
```diff
@@ -98,9 +100,19 @@ class ExecutionOptimizationRulesTest extends FunSuite with PropertyChecks {
   def write(pipe: Gen[TypedPipe[Int]]): Gen[Execution[TypedPipe[Int]]] =
     pipe.map(_.writeThrough(new MemorySource[Int]()))
 
+  def write(pipe: Gen[TypedPipe[TypeWith[TypeGen]#Type]], withT: TypeWith[TypeGen]): Gen[Execution[TypedPipe[TypeWith[TypeGen]#Type]]] = {
```
non: I'm somewhat surprised that this type checks.

It seems like the following signature would be just as good or better:

```scala
def write(t: TypeWith[TypeGen])(pipe: Gen[TypedPipe[t.Type]]: Gen[Execution[TypedPipe[t.Type]]] = ...
```

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/ExecutionGen.scala
```diff
@@ -0,0 +1,61 @@
...
+
+object ExecutionGen {
+  import TypedPipeGen._
+
+  private[this] def cogen(t: TypeWith[TypeGen]): Cogen[TypedPipe[t.Type]] =
+    Cogen[TypedPipe[t.Type]] { pipe: TypedPipe[t.Type] =>
+      pipe.hashCode.toLong
+    }
+
+  def executionOfStd: Gen[Execution[TypedPipe[TypeWith[TypeGen]#Type]]] =
```
non: Does the type projection here add anything over `Gen[Execution[TypedPipe[_]]]`?

I have generally avoided these kinds of projections since I understand Dotty doesn't consider them sound. That said, if this works there may not be a big reason to change it. I was just curious if the type projection added anything?

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/StdGen.scala
```diff
@@ -0,0 +1,29 @@
+package com.twitter.scalding.typed.gen
+
+import org.scalacheck.Gen
+import org.scalacheck.Gen._
+
+object StdGen {
+  implicit def num[A](implicit n: Numeric[A], c: Choose[A]): Gen[A] =
+    Gen.sized(s => c.choose(n.zero, n.max(n.fromInt(s), n.zero)))
```
non: Is there a reason to avoid testing negative numbers? It's probably worth having a chance to generate those as well.

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/TypeGen.scala
```diff
@@ -0,0 +1,59 @@
...
+  def apply[A](g: Gen[A], c: Cogen[A], o: Ordering[A], s: Semigroup[A], t: TupleConverter[A]): TypeGen[A] =
+    new TypeGen[A] {
+      val gen: Gen[A] = g
+      val cogen: Cogen[A] = c
+      val ordering: Ordering[A] = o
+      val semigroup: Semigroup[A] = s
+      val tupleConverter: TupleConverter[A] = t
+    }
+
+  def apply[A, B](a: TypeGen[A], b: TypeGen[B]): TypeGen[(A, B)] =
```
non, line 29: This is really nice!

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/TypedPipeGen.scala
```diff
@@ -0,0 +1,260 @@
...
+import com.twitter.scalding.TypedPipe
+import org.scalacheck.{Cogen, Gen}
+
+object TypedPipeGen {
+  def pipeOfStd: Gen[(TypeWith[TypeGen], TypedPipe[TypeWith[TypeGen]#Type])] =
+    TypeGen.std.flatMap { t =>
+      pipeOf(t).map(t -> _)
+    }
+
+  def pipeOf(a: TypeWith[TypeGen])(implicit tg: Gen[TypeWith[TypeGen]]): Gen[TypedPipe[a.Type]] =
```
non, line 12: Do you find that this sometimes blows up and generates gigantic graphs?

I often find I need to put a max depth or max size parameter on these kinds of recursive generators. But if this is working for you then it's definitely a lot cleaner to do it this way.

non (commented):
Hey, thanks for tagging me on this.

I had a few questions about the use of type projections in this code, although they don't necessarily indicate problems. Some of the primitive generators seemed a bit weak (no negative values). I'd also consider adding at least one or two case classes which might be generated, just to exercise code paths that aren't dealing with simple types.

The overall approach to type generation looks great!

non:
You may also decide you want a way to make your "key type" generators a bit denser (to ensure you are likely to generate repeated values, allowing your joins to succeed); currently your joins are fairly likely to be empty.

I don't think that's needed in this PR but it's just something else that might be useful later on.

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/StdGen.scala
dieu: no reason, just wrong approach to have generic gen :) thanks.

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/ExecutionGen.scala
dieu: not really any reason behind type projection. didn't know about dotty and type projections, thanks.

scalding-core/src/test/scala/com/twitter/scalding/typed/gen/TypedPipeGen.scala
dieu: Yeh, at first run it always blow up with StackOverFlow, with `frequency` and bigger weight on terminated leaf works now, but we can do it smarter if you like to propose.

dieu:
> You may also decide you want a way to make your "key type" generators a bit denser (to ensure you are likely to generate repeated values, allowing your joins to succeed); currently your joins are fairly likely to be empty.
> 
> I don't think that's needed in this PR but it's just something else that might be useful later on.

That's would be nice.

dieu:
Forgot to mention, this PR is more proof of concept to see if we want to invest in this more and migrate all tests on this approach, I'm generally like this but want to see your thoughts.

non:
@dieu Got it, that makes sense. I'll defer to others but (maybe obvious) I'm in favor of moving things to this style.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionOptimizationRulesTest.scala
dieu: that's will be hard to use with output from anything like `Gen[(TypedPipe[_], TypeWith[TypeGen])]`, but later I would love to see some API on generators to generate a graph with some requirements, like have `write` at the end. 

PS: will remove type projection.

PR #1919, opened 2019-06-06
Use https instead of http in urls

alanbato:
This PR changes every instance of an http url to its https version for security reasons.


johnynek (approved):
This is really great.  Thank you!

PR #1920, opened 2019-09-19
Cancellable Executions

stephbian:
This change tries to enable cancellation of executions by stopping the underlying flow.

Prior to this change, parallel executions that have started running could not be cancelled, even if one execution failed. This is problematic because if one parallel execution fails, others will continue running and the overall job will not fail until all executions complete.

In this change, I define a `CancellationHandler` which is built up alongside its corresponding `Future`.


scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
```diff
@@ -0,0 +1,37 @@
+package com.twitter.scalding
+
+import scala.concurrent.{ Future, ExecutionContext => ConcurrentExecutionContext, Promise }
+
+trait CancellationHandler { outer =>
+  def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit]
+  def compose(other: CancellationHandler): CancellationHandler = new CancellationHandler {
+    override def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit] = {
+      other.stop().flatMap(_ => outer.stop())
```
johnynek: I think we can do `Future.zip(other.stop(), outer.stop())` I don't think we need to cancel in order.

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }
+  def compose(other: Future[CancellationHandler]): CancellationHandler = new CancellationHandler {
+    override def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit] = {
+      other.flatMap(_.stop).flatMap(_ => outer.stop())
```
johnynek: same comment about using `zip` here.

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
```diff
@@ -0,0 +1,37 @@
...
+  def empty: CancellationHandler = new CancellationHandler {
+    def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit] = Future(())
+  }
+
+  def cancellable[T](f: Future[T], onCancel: => Unit = ())(implicit ec: ConcurrentExecutionContext): (Future[T], CancellationHandler) = {
+    val p = Promise[T]
+    val first = Future.firstCompletedOf(List(p.future, f))
+    val cancel = new CancellationHandler {
+      override def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit] = Future {
+        first.onFailure { case e => onCancel }
```
johnynek: I don't see why we need this. Can you add a comment?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -69,6 +71,12 @@ object AsyncFlowDefRunner {
       }
     }
   }
+
+  case class FlowCancellationHandler(flow: Promise[Flow[_]]) extends CancellationHandler {
+    def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit] = {
+      flow.future.map(_.stop)
```
johnynek: I think you actually want to use `onComplete` rather than `map` because `.stop` is a side effect. But maybe scala futures aren't so uptight.

Also, I bet `stop` can throw an exception. What should we do then? I don't think we should fail the future because it could fail other calls to cancel. i think we want something like:
`flow.future.onComplete { flow => Try(flow.stop()) }`

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -242,17 +255,20 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
+    val fut = for {
       flowDef <- Future(fn(conf))
       _ = FlowStateMap.validateSources(flowDef, mode)
-      (id, jobStats) <- runFlowDef(conf, flowDef)
+      (resultFut, cancelHandler) = runFlowDef(conf, flowDef) // how to pass cancelHandler through?
+      (id, jobStats) <- resultFut
       _ = FlowStateMap.clear(flowDef)
-    } yield (id, ExecutionCounters.fromJobStats(jobStats))
+    } yield ((id, ExecutionCounters.fromJobStats(jobStats)), cancelHandler)
+    (fut.map(_._1), CancellationHandler.empty.compose(fut.map(_._2)))
```
johnynek: rather than make a `compose` that takes `Future[CancellationHandler]` maybe just have a method like `def handleFromFuture(f: Future[CancelationHandle]): CancelationHandle`, which I think is what you want here.

I think the way you have done it is fine, although I agree it is awkward to lift out of the Future.

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
```diff
@@ -1,45 +1,46 @@
 package com.twitter.scalding
 
 import java.util.concurrent.ConcurrentHashMap
-import scala.concurrent.{ Future, Promise }
+import scala.concurrent.{ Future, Promise, ExecutionContext => ConcurrentExecutionContext }
 /**
  * This is a map for values that are produced in futures
  * as is common in Execution
  */
 class FutureCache[-K, V] {
```
johnynek: it's unfortunate how invasive this is.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -380,70 +387,78 @@ object Execution {
...
-      Trampoline(cache.getOrElseInsert(conf, this,
-        for {
+      Trampoline {
+        lazy val fut = for {
           futt <- toFuture(Try(get(cec)))
           t <- futt
-        } yield (t, Map.empty)))
+        } yield (t, Map.empty[Long, ExecutionCounters])
+
+        cache.getOrElseInsert(conf, this, (fut, CancellationHandler.empty))
```
johnynek: could we not use:

https://github.com/twitter/scalding/pull/1920/files#diff-122dc0ec8aa7eec7186fd18a43ec7d67R24

here?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -380,70 +387,78 @@ object Execution {
...
+          (s, st1) <- fut1
+          next0 = fn(s)
+          // next0 has not been optimized yet, we need to try
+          next = optimize(conf, next0)
+          (fut, cancelHandler2) = Trampoline.call(next.runStats(conf, mode, cache)).get
+          (t, st2) <- fut
+        } yield ((t, st1 ++ st2), cancelHandler2)
+
+        val futCancel = cache.getOrElseInsert(conf, this,
+          (uncachedFutCancel.map(_._1), CancellationHandler.empty.compose(uncachedFutCancel.map(_._2))))
```
johnynek: again we are using empty.compose. i think it is better to have a method to lift a handle from a future.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -475,35 +490,42 @@ object Execution {
...
             // we haven't optimized ex0 yet
             val ex = optimize(conf, ex0)
-            ex.runStats(conf, mode, cache).get
-          }))
+            val (f, c) = ex.runStats(conf, mode, cache).get
+            f.map { v => (v, c) }
+          })
+        }
+
+        val recoveredFut = cache.getOrElseInsert(conf, this, (uncachedFut.map(_._1), CancellationHandler.empty.compose(uncachedFut.map(_._2))))
```
johnynek: this line is very long. Can we introduce some `defs` to deal with it?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -475,35 +490,42 @@ object Execution {
...
               finished.complete(tryT)
             }
           }
-          finished.future
+          (finished.future, cancelHandler)
         })
       }
   }
 
   private[scalding] final case class RecoverWith[T](prev: Execution[T], fn: PartialFunction[Throwable, Execution[T]]) extends Execution[T] {
```
johnynek, line 515: I haven't look in detail at this, but I think we don't want to recover cancellations. We may need a special exception and pattern match and make sure we never pass that to fn.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -741,57 +766,68 @@ object Execution {
...
+        val cacheLookup: List[(ToWrite[_], (Either[Promise[Map[Long, ExecutionCounters]], Future[Map[Long, ExecutionCounters]]], CancellationHandler))] =
           (head :: tail).map{ tw => (tw, cache.getOrLock(conf, tw)) }
         val (weDoOperation, someoneElseDoesOperation) = unwrapListEither(cacheLookup)
 
         val otherResult = failFastSequence(someoneElseDoesOperation.map(_._2))
+        val otherCancellationHandler = someoneElseDoesOperation.map(_._3).foldLeft(CancellationHandler.empty) { case (acc, cancel) => acc.compose(cancel) }
+
         otherResult.value match {
-          case Some(Failure(e)) => Future.failed(e)
+          case Some(Failure(e)) => (Future.failed(e), CancellationHandler.empty)
```
johnynek: should we call the `otherCancellationHandler` here? Can you add a comment in any case.

johnynek (commented):
some comments. I didn't inspect the tests yet.

Good work. We should be able to get this going.

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
stephanie-stripe: i initially thought that maybe we would like to stop a Flow as part of a callback on the failed future. but i don't think this is really needed currently. although, tbh i'm not sure if i should even bother with wrapping Futures with this function, as we aren't really able to cancel the execution. lmk if you think it's worth keeping this around.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -153,14 +153,15 @@ sealed trait Execution[+T] extends Serializable { self: Product =>
...
+    val CFuture(fut, cancelHandler) = exec.runStats(confWithId, mode, ec)(cec).get
     val result = fut.map(_._1)
     // When the final future in complete we stop the submit thread
     result.onComplete { t =>
       writer.finished()
       if (t.isFailure) {
-        // cancel running upstream executions if this was a failure
-        cancelHandler.stop() // need to block until cancelled?
+        // cancel running downstream executions if this was a failure
+        // TODO consider shortening the timeout
```
johnynek: yeah, I think like 30 seconds would still be a long time...

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -410,44 +412,46 @@ object Execution {
           t <- futt
         } yield (t, Map.empty[Long, ExecutionCounters])
 
-        cache.getOrElseInsert(conf, this, (fut, CancellationHandler.empty))
+        lazy val cfut = CFuture(fut, CancellationHandler.empty)
```
johnynek: I think you can use your cancelation wrapper on Future here, but maybe it doesn't add much due to the way we compose cancelation. Not sure.... but at least it could help downstreams fail rather than ever starting....

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -410,44 +412,46 @@ object Execution {
...
           // next0 has not been optimized yet, we need to try
           next = optimize(conf, next0)
-          (fut, cancelHandler2) = Trampoline.call(next.runStats(conf, mode, cache)).get
+          CFuture(fut, cancelHandler2) = Trampoline.call(next.runStats(conf, mode, cache)).get
           (t, st2) <- fut
         } yield ((t, st1 ++ st2), cancelHandler2)
 
         val futCancel = cache.getOrElseInsert(conf, this,
-          (uncachedFutCancel.map(_._1), CancellationHandler.empty.compose(uncachedFutCancel.map(_._2))))
+          CFuture(uncachedFutCancel.map(_._1), CancellationHandler.fromFuture(uncachedFutCancel.map(_._2))))
```
johnynek: looks like a method on `CFuture.fromFuture[A](f: Future[CFuture[A]]): CFuture[A]` or something....

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -410,44 +412,46 @@ object Execution {
...
       }
   }
 
   private[scalding] final case class GetCounters[T](prev: Execution[T]) extends Execution[(T, ExecutionCounters)] {
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      Trampoline.call(prev.runStats(conf, mode, cache)).map { case (fut, cancelHandler) =>
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { case CFuture(fut, cancelHandler) =>
         cache.getOrElseInsert(conf, this,
-          (fut.map {
+          CFuture(fut.map {
```
johnynek: can we put `map` directly on `CFuture` so this can be `.map { cfut => cfut.map { } )` and not have to deal with unpacking and repacking the CFuture?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -456,9 +460,9 @@ object Execution {
   }
   private[scalding] final case class ResetCounters[T](prev: Execution[T]) extends Execution[T] {
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      Trampoline.call(prev.runStats(conf, mode, cache)).map { case (fut, cancelHandler) =>
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { case CFuture(fut, cancelHandler) =>
         cache.getOrElseInsert(conf, this,
-          (fut.map { case (t, _) => (t, Map.empty[Long, ExecutionCounters]) }, cancelHandler))
+          CFuture(fut.map { case (t, _) => (t, Map.empty[Long, ExecutionCounters]) }, cancelHandler))
```
johnynek: same comment about using `CFuture.map` here.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -490,7 +494,7 @@ object Execution {
 
   private[scalding] final case class OnComplete[T](prev: Execution[T], fn: Try[T] => Unit) extends Execution[T] {
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      Trampoline.call(prev.runStats(conf, mode, cache)).map { case (fut, cancelHandler) =>
+      Trampoline.call(prev.runStats(conf, mode, cache)).map { case CFuture(fut, cancelHandler) =>
```
johnynek: maybe also have `CFuture.mapFuture(fn: Future[A] => Future[B]): CFuture[B]` so you don't need to mention cancelHandler here.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: I still think this comment is in play. We should check if this is a cancelation exception, and if it is, don't pass it to the handler.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -505,37 +509,42 @@ object Execution {
...
+        CFuture(recoveredFut.future, cancelHandler.compose(recoveredFut.cancellationHandler))
       }
   }
 
   /**
    * Use our internal faster failing zip function rather than the standard one due to waiting
    */
-  def failFastSequence[T](t: Iterable[Future[T]])(implicit cec: ConcurrentExecutionContext): Future[List[T]] = {
-    t.foldLeft(Future.successful(Nil: List[T])) { (f, i) =>
+  def failFastSequence[T](t: Iterable[CFuture[T]])(implicit cec: ConcurrentExecutionContext): CFuture[List[T]] = {
```
johnynek: we can move this method to `CFuture` I think.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -602,11 +611,12 @@ object Execution {
         futCancel1 <- Trampoline.call(one.runStats(conf, mode, cache))
         futCancel2 <- Trampoline.call(two.runStats(conf, mode, cache))
       } yield {
-        val (f1, cancelHandler1) = futCancel1
-        val (f2, cancelHandler2) = futCancel2
+        val CFuture(f1, cancelHandler1) = futCancel1
+        val CFuture(f2, cancelHandler2) = futCancel2
```
johnynek: can we move this to `CFuture.zip`? Seems like we can simplify the code if we push it there.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -628,14 +638,14 @@ object Execution {
       Trampoline(cache.getOrElseInsert(conf, this, {
         cache.writer match {
           case ar: AsyncFlowDefRunner =>
-            val (fut, cancelHandler) = ar.validateAndRun(conf)(result(_, mode))
-            (fut.map { m => ((), Map(m)) }, cancelHandler)
+            val CFuture(fut, cancelHandler) = ar.validateAndRun(conf)(result(_, mode))
+            CFuture(fut.map { m => ((), Map(m)) }, cancelHandler)
```
johnynek: can we use `CFuture.map` here?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -628,14 +638,14 @@ object Execution {
...
           case other =>
             val fut = Future.failed(
               new IllegalArgumentException(
                 s"requires cascading Mode producing AsyncFlowDefRunner, found mode: $mode and writer ${other.getClass}: $other"))
 
-            (fut, CancellationHandler.empty)
+            CFuture(fut, CancellationHandler.empty)
```
johnynek: maybe `CFuture.failed(ex: Exception): CFuture[Nothing]` here? and move this empty logic there?

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
```diff
@@ -2,45 +2,73 @@ package com.twitter.scalding
 
 import java.util.concurrent.ConcurrentHashMap
 import scala.concurrent.{ Future, Promise, ExecutionContext => ConcurrentExecutionContext }
+
+/**
+ * Represents a cancellable future.
+ */
+case class CFuture[+T](future: Future[T], cancellationHandler: CancellationHandler)
```
johnynek: can we move this to a separate file to make it easier to find for non-IDE users?

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
johnynek: my only question is whether we want to wrap `fromFuture` with this so that the downstreams can be failed without waiting.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -166,20 +167,27 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
           try {
             val ctx = ExecutionContext.newContext(conf.setScaldingFlowCounterValue(id))(fd, mode)
             ctx.buildFlow match {
               case Success(Some(flow)) =>
                 val future = FlowListenerPromise
                   .start(flow, { f: Flow[_] => (id, JobStats(f.getFlowStats)) })
-
-                promise.completeWith(future)
+                // we want to stop the flow when the execution is cancelled
+                val cancel = new CancellationHandler {
```
johnynek: I kind of think `CancellationHandler` should be sealed... maybe this can be `CancellationHandler.fromFn(fn: () => Unit): CancelationHandler` then this can be `val cancel = CancellationHander.fromFn(() => flow.stop())`

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -202,17 +210,17 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
+      // fill in flow once it is built
+      messageQueue.put(RunFlowDef(conf, fd, cpromise))
       // Don't do any work after the .put call, we want no chance for exception
       // after the put
-      fut
+      cpromise.cfuture
     } catch {
       case NonFatal(e) =>
-        Future.failed(e)
+        CFuture(Future.failed(e), CancellationHandler.empty)
```
johnynek: `CFuture.failed`

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -242,17 +250,25 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
-      _ = FlowStateMap.validateSources(flowDef, mode)
-      (id, jobStats) <- runFlowDef(conf, flowDef)
-      _ = FlowStateMap.clear(flowDef)
-    } yield (id, ExecutionCounters.fromJobStats(jobStats))
+    implicit cec: ConcurrentExecutionContext): CFuture[(Long, ExecutionCounters)] = {
+    Try(fn(conf)) match {
+      case Success(flowDef) =>
+        FlowStateMap.validateSources(flowDef, mode)
+        val cfuture = runFlowDef(conf, flowDef)
+        val fut = cfuture.future.map {
```
johnynek: `cfuture.map`

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -242,17 +250,25 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
+          case (id, jobStats) =>
+            FlowStateMap.clear(flowDef)
+            (id, ExecutionCounters.fromJobStats(jobStats))
+        }
+        CFuture(fut, cfuture.cancellationHandler)
+      case Failure(e) =>
+        CFuture(Future.failed(e), CancellationHandler.empty)
```
johnynek: `CFuture.failed`

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -330,11 +346,11 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
       fd
     }
 
-    val resultFuture = validateAndRun(conf)(prepareFD _)
+    val CFuture(resultFuture, cancelHandler) = validateAndRun(conf)(prepareFD _)
 
     // When we are done, the forced pipes are ready:
     done.completeWith(resultFuture.map(_ => ()))
-    resultFuture
+    CFuture(resultFuture, cancelHandler)
```
johnynek: `CFuture.mapFuture`

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
```diff
@@ -110,7 +111,8 @@ class MemoryWriter(mem: MemoryMode) extends Writer {
     }
     val (id, acts) = idActs
     // now we run the actions:
-    Future.traverse(acts) { fn => fn() }.map(_ => (id, ExecutionCounters.empty))
+    val fut = Future.traverse(acts) { fn => fn() }.map(_ => (id, ExecutionCounters.empty))
+    CFuture(fut, CancellationHandler.empty)
```
johnynek: can we comment how returning an empty cancelation handler from all the executions is working? To me, it seems like `type Action = () => CFuture[Unit]` and we should be composing cancelation. I'm confused how this can work.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
```diff
@@ -142,14 +142,14 @@ object Op extends Serializable {
 
     def run(session: SparkSession)(implicit ec: ExecutionContext): Future[RDD[_ <: A]] =
       cache.getOrElseUpdate(session,
-        input.run(session).map { rdd => fn(widen(rdd)) })
+        CFuture(input.run(session).map { rdd => fn(widen(rdd)) }, CancellationHandler.empty)).future
```
johnynek: I think we should have two FutureCache implementations or use a typeclass approach to make FutureCache generic. Copying FutureCache to CFutureCache and not changing the original might work here.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -213,6 +213,6 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
       (nextState.copy(id = nextState.id + 1), (nextState.id, acts))
     }
     // now we run the actions:
-    Future.traverse(acts) { fn => fn() }.map(_ => (id, ExecutionCounters.empty))
+    CFuture(Future.traverse(acts) { fn => fn() }.map(_ => (id, ExecutionCounters.empty)), CancellationHandler.empty)
```
johnynek: I think we want to use `CFuture.uncancellable` or something here.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍 added this logic and added some tests

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

stephanie-stripe: actually, `result.completeWith` needs a future, so i just did `cfuture.future.map`

scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryWriter.scala
stephanie-stripe: it's still the case that `type Action = () => Future[Unit]` -- i just wrapped the result in a `CFuture` at the end

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
stephanie-stripe: ah, the typeclass makes things nicer, thanks

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
```diff
@@ -1,45 +1,67 @@
...
-class FutureCache[-K, V] {
-  private[this] val cache = new ConcurrentHashMap[K, Future[V]]()
+class FutureCache[-K, V, P[_], F[_]](implicit pl: PromiseLike[P, F]) {
+  private[this] val cache = new ConcurrentHashMap[K, F[V]]()
 
-  def get(k: K): Option[Future[V]] = Option(cache.get(k))
+  def get(k: K): Option[F[V]] = Option(cache.get(k))
 
-  def getOrElseUpdate(k: K, res: => Future[V]): Future[V] =
+  def getOrElseUpdate(k: K, res: => F[V])(implicit cec: ConcurrentExecutionContext): F[V] =
```
johnynek: do we still need the ConcurrentExecutionContext on these methods? It seems like not?

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
```diff
@@ -1,45 +1,67 @@
 package com.twitter.scalding
 
 import java.util.concurrent.ConcurrentHashMap
-import scala.concurrent.{ Future, Promise }
+import scala.concurrent.{ Future, Promise, ExecutionContext => ConcurrentExecutionContext }
+
+trait PromiseLike[P[_], F[_]] {
```
johnynek, line 6: this is nice!

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -369,70 +377,79 @@ object Execution {
    * This is a mutable state that is kept internal to an execution
    * as it is evaluating.
    */
-  private[scalding] class EvalCache(val writer: Execution.Writer) {
+  private[scalding] class EvalCache(val writer: Execution.Writer)(implicit cec: ConcurrentExecutionContext) {
```
johnynek: do we need cec here? I don't see why we need it now, but not before?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -369,70 +377,79 @@ object Execution {
...
-      Trampoline(cache.getOrElseInsert(conf, this,
-        for {
+      Trampoline {
+        lazy val fut = for {
           futt <- toFuture(Try(get(cec)))
           t <- futt
-        } yield (t, Map.empty)))
+        } yield (t, Map.empty[Long, ExecutionCounters])
+
+        lazy val cfut = CFuture(fut, CancellationHandler.empty)
```
johnynek: isn't this `CFuture.uncancellable(fut)`?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -755,43 +773,50 @@ object Execution {
...
-                    case (toWrite, promise) =>
-                      promise.completeWith(futCounters)
-                  }
-                  futCounters
-                case Nil =>
-                  // No work to do, provide a fulled set of 0 counters to operate on
-                  Future.successful(Map.empty)
+        val otherResult = CFuture.failFastSequence(someoneElseDoesOperation.map(_._2))
+
+        val localFlowDefCountersFuture: CFuture[Map[Long, ExecutionCounters]] =
```
johnynek: shouldn't this be lazy? There is no need to start this future if `otherResult.value` is failed as we check below, no?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -755,43 +773,50 @@ object Execution {
...
+                  cpromise.completeWith(CFuture(futCounters, cancelHandler))
               }
+              CFuture(futCounters, cancelHandler)
+            case Nil =>
+              // No work to do, provide a fulled set of 0 counters to operate on
+              CFuture(Future.successful(Map.empty), CancellationHandler.empty)
+          }
+
+        otherResult.future.value match {
+          case Some(Failure(e)) => CFuture(Future.failed(e), localFlowDefCountersFuture.cancellationHandler) // the other future failed, we need to cancel the other side
```
johnynek: can't we just avoid starting `localFlowDefCountersFuture` above with lazy val, and here use `CFuture.failed(e)`?

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -828,7 +853,7 @@ object Execution {
    */
   private[scalding] case object ReaderExecution extends Execution[(Config, Mode)] {
     protected def runStats(conf: Config, mode: Mode, cache: EvalCache)(implicit cec: ConcurrentExecutionContext) =
-      Trampoline(Future.successful(((conf, mode), Map.empty)))
+      Trampoline(CFuture(Future.successful(((conf, mode), Map.empty)), CancellationHandler.empty))
```
johnynek: can we use `CFuture.successful` here?

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
```diff
@@ -1,45 +1,67 @@
...
+  }
+}
+
 /**
  * This is a map for values that are produced in futures
  * as is common in Execution
  */
-class FutureCache[-K, V] {
-  private[this] val cache = new ConcurrentHashMap[K, Future[V]]()
+class FutureCache[-K, V, P[_], F[_]](implicit pl: PromiseLike[P, F]) {
```
johnynek: maybe to not break any users that were using this elsewhere, could we do:
```scala
class FutureCacheGeneric[-K, V, P[_], F[_]](implicit pl: PromiseLike[P, F]) ...

class FutureCache[-K, V] extends FutureGeneric[K, V, Promise, Future](PromiseLike.PromiseLikePromise)
```
This should be source compatible for them.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -166,20 +167,21 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
                   .start(flow, { f: Flow[_] => (id, JobStats(f.getFlowStats)) })
-
-                promise.completeWith(future)
+                // we want to stop the flow when the execution is cancelled
+                val cancel = CancellationHandler.fromFn(() => flow.stop())
+                cpromise.completeWith(CFuture(future, cancel))
               case Success(None) =>
                 // These is nothing to do:
-                promise.success((id, JobStats.empty))
+                cpromise.promise.success((id, JobStats.empty))
```
johnynek, line 190: don't we need to do `cpromise.cancellationHandler.success(CancellationHandler.empty)` here as well?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -166,20 +167,21 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
               case Failure(err) =>
-                promise.failure(err)
+                cpromise.promise.failure(err)
```
johnynek, line 193: don't we need to do `cpromise.cancellationHandler.success(CancellationHandler.empty)` here as well?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -191,7 +193,7 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
               // In a sense, this thread does not exist logically and
               // must forward all exceptions to threads that requested
               // this work be started.
-              promise.tryFailure(t)
+              cpromise.promise.tryFailure(t)
```
johnynek, line 206: don't we need to do `cpromise.cancellationHandler.success(CancellationHandler.empty)` here as well?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -242,17 +244,23 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
   def validateAndRun(conf: Config)(fn: Config => FlowDef)(
-    implicit cec: ConcurrentExecutionContext): Future[(Long, ExecutionCounters)] =
-    for {
-      flowDef <- Future(fn(conf))
-      _ = FlowStateMap.validateSources(flowDef, mode)
-      (id, jobStats) <- runFlowDef(conf, flowDef)
-      _ = FlowStateMap.clear(flowDef)
-    } yield (id, ExecutionCounters.fromJobStats(jobStats))
+    implicit cec: ConcurrentExecutionContext): CFuture[(Long, ExecutionCounters)] = {
+    Try(fn(conf)) match {
```
johnynek: this changes the logic. Previously `Future(fn(conf))` made running that asynchronous, now that is synchronous. It is possibly fine, but if validateSources, which throws if things are not valid, fails, nothing will catch the exception. Previously, that was going to happen inside a Future where it would be caught.

So, i think something like `Future(fn(conf)).transform` or something will work, so you can transform both arms.

Alternatively, you could do `Try(fn(conf)).map { flowDef => FlowStateMap.validateSources(flowDef, mode); flowDef }` so if that fails it fails in the Try.map.

johnynek: PS: IMO, all this gets simpler using IO because there is not such careful reasoning about when things have been started or not.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
```diff
@@ -212,6 +214,131 @@ class ExecutionTest extends WordSpec with Matchers {
...
+      val mapCountDownLatch2 = new CountDownLatch(1)
+
+      val blockingTp2: TypedPipe[Int] = TypedPipe.from(Seq(1)).groupAll.sum.values.map { i =>
+        // block until we are done
+        mapCountDownLatch2.await()
+        i
+      }
+
+      val onCompleteCountDownLatch2 = new CountDownLatch(1)
+      val otherEx2: Execution[Iterable[Int]] = blockingTp2.toIterableExecution.onComplete { t =>
```
johnynek, line 268: can you also make a copy of this test that instead of using `.onComplete` uses `.liftToTry.map`. This has different semantics slightly from onComplete, but should still pass in my opinion.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/Op.scala
```diff
@@ -138,15 +138,14 @@ object Op extends Serializable {
     r.asInstanceOf[RDD[A]]
 
   final case class Transformed[Z, A](input: Op[Z], fn: RDD[Z] => RDD[A]) extends Op[A] {
-    @transient private val cache = new FutureCache[SparkSession, RDD[_ <: A]]
+    @transient private val cache = new FutureCache[SparkSession, RDD[_ <: A], Promise, Future]
```
johnynek: if we make my change to FutureCacheGeneric and leave FutureCache this diff should go away i think.

johnynek (commented):
This looks very close. A few minor changes.

cc @ttim

johnynek:
I think travis is unhappy: 

https://travis-ci.org/twitter/scalding/builds/589230369?utm_source=github_status&utm_medium=notification

maybe set `oraclejdk8` to `openjdk8` in the `.travis.yml` and I think CI will run?

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
stephanie-stripe: yep, removed.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: removed

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: yep, changed

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: ah, i think you're right. i pulled this out of the `case _ =>` because i thought i needed the cancellation handler. but since we are starting the work for these futures in this method, there isn't anything to cancel in the failed case.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: yes. i changed this back to define localFlowDefCountersFuture in the `case _ =>`

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/FutureCache.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: good catch. i opted for the second option to avoid unpacking another future.

scalding-core/src/test/scala/com/twitter/scalding/ExecutionTest.scala
stephanie-stripe: i think there is a problem here where `liftToTry` uses `recoverWith` to construct the failure case, but we skip the `recoverWith` function when an execution is cancelled. as a result, we end up throwing instead of returning a failed Try.

do you think it is worth implementing a `LiftToTry` execution type, or do you have other suggestions?

johnynek: Ahh, yeah. I think just remove the test then. 

If you want to test more cases, you might consider exercising `Execution.flatMap` to make sure cancelation propagates inside if a job is started by the flatMap function.

Like if we had `Exection.unit.flatMap { _ => foo.toIterableExecution.map(waitingFn) }` we should expect that cancellation still works.

johnynek (commented):
this looks good.

One more test request and I think we can merge once it is green.

johnynek:
@ttim I assume you have no concerns since you haven't commented. I'll merge tomorrow morning unless you have some concerns.

The goal here is to abort large and expensive jobs when another part of the flow fails.

ttim:
@johnynek I'll take a look until tomorrow's evening. Seems like a great change!

scalding-core/src/main/java/com/twitter/scalding/cascading_interop/FlowListenerPromise.java
```diff
@@ -67,6 +69,17 @@ public boolean onThrowable(Flow f, Throwable t) {
         return true;
       }
     });
+    flow.addStepListener(new FlowStepListener() {
+      public void onStepStarting(FlowStep flowStep) { } // ignore
+      public void onStepRunning(FlowStep flowStep) { } // ignore
+      public void onStepCompleted(FlowStep flowStep) { } // ignore
+      public void onStepStopping(FlowStep f) { result.tryFailure(new FlowStopException("Flow step was stopped")); }
+      public boolean onStepThrowable(FlowStep f, Throwable t) {
+        result.failure(t);
```
johnynek: Since there can be more than one failure don’t we need tryFailure or something? I think promise errors if two calls are made to fill it unless the try variants are used.

stephanie-stripe: ah, you're right. fixed.

stephanie-stripe:
i would hold off on merging this for now... i am doing some testing and noticing that jobs are killed before retries can happen, so trying to debug that now.

stephanie-stripe:
i've tested this on the cluster, and it looks like things are being cancelled as expected (both in separate parallel executions and within the same one)! one strange thing though: since adding the `FlowStepListener`, i've been seeing a NPE from the main process where i kick off the job:
```
Exception in thread "main" java.lang.Throwable: If you know what exactly caused this error, please consider contributing to GitHub via following link.
https://github.com/twitter/scalding/wiki/Common-Exceptions-and-possible-reasons#javalangnullpointerexception
        at com.twitter.scalding.Tool$.main(Tool.scala:153)
        at com.twitter.scalding.Tool.main(Tool.scala)
Caused by: java.lang.NullPointerException
        at scala.util.Failure.get(Try.scala:207)
        at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:227)
        at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:190)
        at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
        at scala.concurrent.Await$.result(package.scala:190)
        at com.twitter.scalding.ExecutionJob$$anonfun$6.apply(Job.scala:526)
        at com.twitter.scalding.ExecutionJob$$anonfun$6.apply(Job.scala:525)
        at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
        at scala.util.Try$.apply(Try.scala:192)
        at scala.util.Success.map(Try.scala:237)
        at com.twitter.scalding.ExecutionJob.run(Job.scala:525)
        at com.twitter.scalding.Tool.start$1(Tool.scala:124)
        at com.twitter.scalding.Tool.run(Tool.scala:141)
        at com.twitter.scalding.Tool.run(Tool.scala:68)
        at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
        at com.twitter.scalding.Tool$.main(Tool.scala:149)
```
i'm still not sure where this is coming from, but it doesn't seem to be affecting shutdown afaict.

stephanie-stripe:
@johnynek and i paired to fix the npe with the latest commit, so this is good to go from my perspective!

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+
+  def fromFuture(f: Future[CancellationHandler]): CancellationHandler = new CancellationHandler {
+    override def stop()(implicit ec: ConcurrentExecutionContext): Future[Unit] = {
+      f.flatMap(_.stop())
+    }
+  }
+
+  def cancellable[T](f: Future[T], onCancel: => Unit = ())(implicit ec: ConcurrentExecutionContext): (Future[T], CancellationHandler) = {
```
johnynek: If we didn’t wind up using this can we remove?

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -166,20 +167,23 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
       @annotation.tailrec
       def go(id: Long): Unit = messageQueue.take match {
         case Stop => ()
```
johnynek, line 170: Can we add the StopFlow stuff here? I think I know what we did wrong: we need to call `go(id)` after we do `flow.stop()` because we need to keep processing messages. We only stop the loop on the Stop message.

johnynek (commented):
One minor request. This seems like a great change.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
stephanie-stripe: ahhh yep that fixed it.

scalding-core/src/main/scala/com/twitter/scalding/CancellationHandler.scala
stephanie-stripe: 👍

stephanie-stripe: 👍

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/AsyncFlowDefRunner.scala
```diff
@@ -167,6 +168,9 @@ class AsyncFlowDefRunner(mode: CascadingMode) extends Writer {
...
+        case StopFlow(flow) =>
```
johnynek: sorry to be a pain in the ass... I'm worried that cascading can throw here on `flow.stop()` and if it did we have to swallow the exception.

Maybe we should do something like:
```scala

case class StopFlow(flow: Flow[_], result: Promise[Unit]) extends FlowDefAction
//...
case StopFlow(flow, promise) =>
  promise.complete(Try(flow.stop()))
  go(id)

// then change CancellationHandler:
def fromFn(fn: ConcurrentExecutionContext => Future[Unit]): CancellationHandler =
  new CancellationHandler {
    def stop()(implicit cec: ConcurrentExecutionContext): Future[Unit] =
      fn(cec)
    }

// finally:

val cancel = fromFn { implicit cec =>
  val done = Promise[Unit]()
  messageQueue.put(StopFlow(flow, done))
  done.future
}
```

stephanie-stripe: makes sense, done

johnynek (approved):

johnynek:
I'll merge when it's green!

Great work on this!

PR #1921, opened 2019-09-27
Make projected parquet collection schema forward compatible with the given file schema

mickjermsurawong-stripe:
- Problem: projection schema is created by [ThriftSchemaConvertVisitor](https://github.com/apache/parquet-mr/blob/master/parquet-thrift/src/main/java/org/apache/parquet/thrift/ThriftSchemaConvertVisitor.java#L180) which suffix fields with `_tuple` 
https://github.com/twitter/scalding/blob/59d932731f7396eaaf6024624d1ce6660534ca77/scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ScroogeReadSupport.java#L98-L101
- When reading parquet file with standard format 3-level or even legacy ones 2-level `array`, projection fails because it has name `_tuple` which is incompatible. For example, given file schemas to represent `List<String> country_codes`
```
message compliant_schema {
  required group country_codes (LIST) {
    repeated binary array (UTF8);
  }
  required int32 to_be_removed;
}
```
will fail with the generated projection 
```
message thrift_generated {
  optional group country_codes (LIST) {
    repeated binary country_codes_tuple (UTF8);
  }
}
```
- This PR introduces `ParquetListFormatForwardCompatibility` to convert thrift-generated schema to compliant ones. The difficulties here are:
  - we can not simply just traverse the two graphs and swap node one-for-one. The node has to be from the projection, because it has subset of fields and different optional/required info.
  - there are at least 4 variations of compliant parquet types, and we want to convert from non-compliant to compliant, and among the compliant. 

- The strategy here is to implement different rules for each format. The rule tell us how to decompose the list schema, and to reconstruct it again.
  - For example, we match projected schema with Rule 1, and match file schema with Rule 2.
  - Rule 1 will decompose the projected schema, and Rule 2 will take that information to construct in light of the file schema.


scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+        groupProjection.getFields().get(0).isRepetition(Type.Repetition.REPEATED);
+  }
+
+  /**
+   * Resolve list format in forward compatible way.
+   * @param fileType   file type which has new format
+   * @param projection projection type which has legacy format
+   * @return projection schema in the new format.
+   */
+  public Type resolveTypeFormat(Type fileType, Type projection) {
```
mickjermsurawong-stripe: Main logic of this PR

scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/ScroogeReadSupportTests.scala
```diff
@@ -90,6 +90,384 @@ class ScroogeReadSupportTests extends WordSpec with Matchers with HadoopSharedPl
     }
   }
 
+  "ScroogeReadSupport resolving list format" should {
```
mickjermsurawong-stripe: Note that in projected schema, the field is optional, and that is maintained in the result.

scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/ScroogeReadSupportTests.scala
```diff
@@ -90,6 +90,384 @@ class ScroogeReadSupportTests extends WordSpec with Matchers with HadoopSharedPl
...
+          |    }
+          |  }
+          |}
+        """.stripMargin)
+      val requestedProjection = MessageTypeParser.parseMessageType(
+        """
+          |message SampleProjection {
+          |  optional group foo (LIST) {
+          |    repeated group foo_tuple (LIST) {
+          |      repeated binary foo_tuple_tuple (UTF8);
```
mickjermsurawong-stripe: Got this schema from `ThriftSchemaConverter` which is what `ScroogeReadSupport` does when projection given thrift class.
```
      val innerListType = new ListType(new ThriftField("list", 2, Requirement.REQUIRED, new ThriftType.StringType))
      val listType = new ListType(new ThriftField("list", 2, Requirement.REQUIRED, innerListType))
      val children = new ThriftField("foo", 1, Requirement.REQUIRED, listType)
      val message = new ThriftSchemaConverter().convert(new StructType(util.Arrays.asList(children), StructOrUnionType.STRUCT))
```

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+      new RuleGroupArray(),
+
+      new RuleGroupTuple(),
+      new RuleStandardThreeLevel());
+
+  /**
+   * Rule describes how to match a repeated type, how to decompose them, and reconstruct a
+   * repeated type.
+   */
+  abstract static public class Rule {
```
mickjermsurawong-stripe: This and bulk of below are implementation of different rules.

mickjermsurawong-stripe:
cc @moulimukherjee

mickjermsurawong-stripe:
Hi @johnynek,
We ran into a problem where Scalding job fails reading parequet files produced by another Spark job. The problem we discovered is due to projected schema which is generated from Thrift class has different list format than our Spark job.

Just would like to ask for a quick look whether this is the right strategy, and if this is something that we would like to push it to the public one as well.

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+/**
+ * Compatibility class to convert parquet schema of legacy type to standard one
+ * namely 3-level list structure as recommended in
+ * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists
+ *
+ * More specifically this handles converting from parquet file created by
+ * {{@code org.apache.parquet.thrift.ThriftSchemaConvertVisitor}} which always suffix
+ * list element with "_tuple".
+ */
+public class ParquetListFormatForwardCompatibility {
```
johnynek: we generally prefer scala in this repo. Was there any reason to implement this in Java?

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+
+    @Override
+    public Boolean check(Type repeatedType) {
+      return repeatedType.isPrimitive() && repeatedType.getName().equals(this.constantElementName());
+    }
+
+    @Override
+    public Type createCompliantRepeatedType(Type type, String name, Boolean isElementRequired, OriginalType originalType) {
+      if (!isElementRequired) {
+        throw new IllegalArgumentException("Rule 1 can only take required element");
```
johnynek: you use the string "Rule 1" here, but the class name is `RulePrimitiveElement`. Can you explain the difference?

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
johnynek: can't this be a static method?

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+      }
+    }
+    if (matchedRule == null) {
+      throw new RuntimeException(String.format(
+          "Unable to find matching rule for %s schema:\n%s", debuggingTypeSource, repeatedType));
+    }
+    return matchedRule;
+  }
+
+  private Type makeForwardCompatible(Type repeatedFileType, Type repeatedProjectedType) {
```
johnynek: can't this be a static method?

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+          fields.add(resolveTypeFormat(fileField.asGroupType(), projected.asGroupType()));
+        } else {
+          fields.add(projected);
+        }
+      }
+      return groupProjection.withNewFields(fields);
+    }
+  }
+
+  private Rule findFirstRule(Type repeatedType, String debuggingTypeSource) {
```
johnynek: can't this be a static method?

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
```diff
@@ -0,0 +1,349 @@
...
+   * Resolve list format in forward compatible way.
+   * @param fileType   file type which has new format
+   * @param projection projection type which has legacy format
+   * @return projection schema in the new format.
+   */
+  public Type resolveTypeFormat(Type fileType, Type projection) {
+    if (projection.isPrimitive() || fileType.isPrimitive()) {
+      return projection;
+    }
+    ParquetListFormatForwardCompatibility compatibility = new ParquetListFormatForwardCompatibility();
```
johnynek: why do we need to allocate a new instance?

johnynek (commented):
Thanks for the PR.

Have we run this internally to do some smoke tests that it can read old data that was previously written?

I added some general comments. I really don't know the latest with parquet. Maybe @isnotinvain or @julienledem can comment?

cc @ttim

johnynek:
Also, you need to bump the `.travis.yml` jdk setting to `openjdk8` since `oraclejdk8` I think no longer works on travis.

mickjermsurawong-stripe:
Thanks @johnynek. Ah I haven't done any tests internally on actual data. Just purely unit tests now. Will follow up with that. 

I can port all these to Scala, and will make the resolver stateless; yup the object instantiation isn't neccessary.

scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ParquetListFormatForwardCompatibility.java
mickjermsurawong-stripe: Sorry, that's bad naming. Will clean up.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+import org.apache.parquet.schema.Type.Repetition
+import org.apache.parquet.schema.{GroupType, MessageType, Type}
+import org.apache.parquet.thrift.DecodingSchemaMismatchException
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable
+
+object ParquetCollectionFormatForwardCompatibility {
+
+  private val LOGGER = LoggerFactory.getLogger(getClass)
```
joshrosen-stripe: Nit: `private[this] val logger = LoggerFactory.getLogger(getClass)` seems to be the convention used elsewhere in the existing code.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * Create a forward-compatible schema, using content from source type with format from target type.
+   * @param sourceType source type with legacy format
+   * @param targetType target type to which source is converted to
+   */
+  def formatForwardCompatibleMessage(sourceType: MessageType, targetType: MessageType): MessageType = {
+    val groupResult = formatForwardCompatibleType(sourceType, targetType).asGroupType()
+    LOGGER.info("Making source schema to be compatible with target" +
```
joshrosen-stripe: Should this be `info` level by default? Or will that be too verbose? I took a look at the existing code to try to find precedent but couldn't identify an obvious existing convention.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+        }
+        sourceGroup.withNewFields(resultFields.asJava)
+    }
+  }
+
+  private case class GroupUnwrapped(wrappers: Seq[GroupType],
+                                    repeatedListType: Option[Type] = None,
+                                    repeatedMapType: Option[Type] = None)
+
+  private def unwrapGroup(typ: Type, wrappers: Seq[GroupType] = Seq()): GroupUnwrapped = {
```
joshrosen-stripe: Perhaps a naive question, but what does "wrapped" or "wrapper" refer to in this case? Could you add a comment explaining this? That might make the following code easier to review.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
```diff
@@ -0,0 +1,262 @@
...
+            isElementRequired = sourceRule.isElementRequired(repeatedSourceType),
+            originalType = sourceRule.elementOriginalType(repeatedSourceType)
+          )
+        }
+      }
+      case _ => repeatedSourceType
+    }
+  }
+
+  def isGroupList(projection: Type): Boolean = {
```
joshrosen-stripe: To confirm my understanding:

This is matching on things like 

```
required group my_list (LIST) {
  repeated group list {
    optional binary element (UTF8);
  }
}
```

where `groupProjection` here refers to the top-most type, and the `groupProjection.getFields.get(0).isRepetition(Type.Repetition.REPEATED)` check would apply to the second-level element (`repeated group list`)?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+    if (typ.isPrimitive) {
+      GroupUnwrapped(
+        wrappers,
+        repeatedListType=None,
+        repeatedMapType=None
+      )
+    } else if (typ.asGroupType.getFieldCount != 1) {
```
joshrosen-stripe: In this `else if` branch and the `else` branch I noticed we have `typ.asGroupType` calls everywhere. To avoid that and to make it clear that `!typ.isPrimitive` implies that `typ` is a group type, could we instead do something like

```scala
if (typ.isPrimitive) {
   // ...
} else {
  val groupTyp = typ.asGroupTyp
  if (groupTyp.getFieldCount != 1) {
     //...
  } else {
     //...
  }
}
```

We might be able to further improve clarity by changing the sign of the second condition, doing

```scala
if (typ.isPrimitive) {
   // ...
} else {
  val groupTyp = typ.asGroupTyp
  if (groupTyp.getFieldCount == 1) {  // <--- inverted condition
     //...
  } else {
     //...
  }
}
```

which then simplifies or clarifies the

```
// Note the field count is strictly 1 here
```

comment present in the current code.

We could then add a comment in the `else` branch saying something like

```
// The group contains more than one field, so ... <whatever that implies>
```

This is maybe nitpicky, but I'm trying to walk through each case here to make sure that I understand what all of the branches correspond to, so I'm considering whether a different ordering or presentation of the cases can make it easier to understand and verify.

mickjermsurawong-stripe:
hi  @isnotinvain or @julienledem, I've addressed the first round of general feedback, and added support for map schema compat as well. Would appreciate your review here.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+    new MessageType(groupResult.getName, groupResult.getFields)
+  }
+
+  /**
+   * Traverse source/target schemas and format nodes of list or map.
+   * The formatting is not to one-to-one node swapping from source to target,
+   * this is because the subset fields of source node and its optional/required must
+   * be maintained in the formatted result.
+   */
+  private def formatForwardCompatibleType(sourceType: Type, targetType: Type): Type = {
```
joshrosen-stripe: I find this method's pattern match to be a bit difficult to intuitively understand and I think that maybe it's because of the `GroupUnwrapped`: there, we seem to have a big cross product of potential possibilities, but it looks like we only have explicit cases for some of them.

By construction, it looks like we never wind up with `repeatedListType` and `repeatedMapType` both non-None.

To simplify the pattern match and `unwrapGroup`, what do you think about instead defining a `sealed trait` of case-classes corresponding to the different cases that we might need to handle? What that approach, I think we can avoid the need to do this more complicated positional pattern-matching. 

Given that, could it make sense to have methods like `ParquetMapFormatRule.isGroupMap()` instead be "extractor"-type methods which return `Option[GroupMapSomethingIHaventNamedYet]`? That could help to avoid separate definition and extraction code that we have now (where the "is X" and "extract X" code for the same logical concept are in separate places).
 
This would also have the advantage of avoiding the need to have comments like `// Format list` to explain the cases, since that semantic information could be instead carried in the sealed trait class names.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+        // describing field name and whether it's optional/required
+        sourceWrappers.foldRight(formattedRepeated) {
+          (wrapper, group) => wrapper.withNewFields(group)
+        }
+      case _ =>
+        // Field projection
+        val sourceGroup = sourceType.asGroupType
+        val targetGroup = targetType.asGroupType
+
+        import scala.collection.JavaConverters._
```
joshrosen-stripe: Import at the top of the file? That should be safe because this is `Converters` (not `Conversions`, which is much more implicit).

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  def findFirstListRule(repeatedType: Type,
+                        sourceOrTarget: SourceOrTarget): Option[ParquetListFormatRule] = {
+    val ruleFound = sourceOrTarget.rules.find(rule => rule.check(repeatedType))
+    if (ruleFound.isEmpty) LOGGER.warn(s"Unable to find matching rule for ${sourceOrTarget.name} schema:\n$repeatedType")
+    ruleFound
+  }
+
+  def wrapElementAsRepeatedType(rule: ParquetListFormatRule, repeatedType: Type, elementType: Type): Type = rule.createCompliantRepeatedType(
```
joshrosen-stripe: This is formatted a bit funny right now; putting the `rule.create*` on the next line?

joshrosen-stripe (commented):
I left some initial review comments, primarily focused on understanding (and potentially simplifying) two of the larger and more complex pattern matches at the heart of the converter.

I also left a couple of minor style nits, but those can be addressed later (in a final polish pass) once we're confident in the rest of the design.

I haven't finished looking at this yet, though: I still need to go through `ParquetListFormatRule` and also need to take a closer look at the test cases.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
```diff
@@ -0,0 +1,262 @@
...
+private[scrooge] sealed trait ParquetListFormatRule {
+  def elementType(repeatedType: Type): Type
+
+  def elementName(repeatedType: Type): String = this.elementType(repeatedType).getName
+
+  def elementOriginalType(repeatedType: Type): OriginalType = this.elementType(repeatedType).getOriginalType
+
+  private[scrooge] def isElementRequired(repeatedType: Type): Boolean
+
+  private[scrooge] def check(typ: Type): Boolean
```
joshrosen-stripe: From the name, it's a bit non-obvious what this method does. It this actually something like `appliesToType(typ: Type)` which returns `true` if the type is a list of this format and `false` otherwise?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
mickjermsurawong-stripe: yup that's correct on `groupProjection`, and the repeated is the second level.

The part after `repeated` could be different depending on formats eg.
```
required group my_list (LIST) {
  repeated repeated int32 [element|array];
}
```

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Helper to specify supported source/target conversion.
+ */
+private[scrooge] sealed trait SourceOrTarget {
+  def rules: Seq[ParquetListFormatRule]
+  def name: String
+}
+
+private[scrooge] object Source extends SourceOrTarget {
```
thomaschow: what's the advantage of distinguishing between Source and Target? is it so we can define supported directionalities in terms of parquet spec conversion? I'm wondering if we're being too specific here and wind up seeing a direction that we didn't engineer for, and we fail

thomaschow: I guess since scalding is going to be stuck on the older parquet spec, this would not happen.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
```diff
@@ -0,0 +1,262 @@
...
+
+private[scrooge] object GroupArrayRule extends GroupListRule {
+  override def constantElementName: String = "array"
+}
+
+private[scrooge] object TupleRule extends ParquetListFormatRule {
+  override def check(repeatedType: Type): Boolean = repeatedType.getName.endsWith("_tuple")
+
+  override def elementName(repeatedType: Type): String = {
+    repeatedType.getName.substring(0, repeatedType.getName.length - 6)
```
thomaschow: nit: would like docs or variable/method renaming to explain this substring indexing a little more clearly

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
```diff
@@ -0,0 +1,262 @@
...
+
+  override def createCompliantRepeatedType(originalElementType: Type, name: String, isElementRequired: Boolean, originalType: OriginalType): Type = {
+
+    val repetition = if (isElementRequired) Type.Repetition.REQUIRED else Type.Repetition.OPTIONAL
+    val elementType = if (originalElementType.isPrimitive) {
+      new PrimitiveType(repetition, originalElementType.asPrimitiveType.getPrimitiveTypeName, "element", originalType)
+    } else {
+      new GroupType(
+        repetition,
+        "element",
```
thomaschow: nit: DRY with https://github.com/twitter/scalding/pull/1921/files#diff-f1236da4ace7eaf46e33ad1d12d599b8R241 ?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
thomaschow: +1 on naming this something more descriptive

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
thomaschow: nvm, doesn't seem necessary

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
isnotinvain: I think too verbose, the schemas can often be very large. Maybe debug level?

scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibilityTests.scala
```diff
@@ -0,0 +1,919 @@
...
+
+import org.apache.parquet.schema.MessageTypeParser
+import org.apache.parquet.thrift.{DecodingSchemaMismatchException, ThriftSchemaConverter}
+import org.apache.parquet.thrift.struct.ThriftField.Requirement
+import org.apache.parquet.thrift.struct.{ThriftField, ThriftType}
+import org.apache.parquet.thrift.struct.ThriftType.StructType.StructOrUnionType
+import org.apache.parquet.thrift.struct.ThriftType.{ListType, MapType, StructType}
+import org.scalatest.{Matchers, WordSpec}
+
+class ParquetCollectionFormatForwardCompatibilityTests extends WordSpec with Matchers {
```
thomaschow: ooc how hard would it be to add some actual thrift fixtures based tests in here?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,153 @@
...
+import java.util
+
+import org.apache.parquet.schema.Type.Repetition
+import org.apache.parquet.schema.{GroupType, MessageType, Type}
+import org.apache.parquet.thrift.DecodingSchemaMismatchException
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable
+
+object ParquetCollectionFormatForwardCompatibility {
```
isnotinvain: Parquet is complicated because there are always 4 schemas involved: the write-time schema, the read-time schema, and a projected schema in terms of the write-schema, and a projected schema in terms of the read-time schema. I find this really hard to keep track of, so I think it'd be a huge help to me and future readers to put a really detailed comment on this class explaining what it's for, what it makes compatible, what the problem is, how it solves it, etc. I think for something like this, the more detailed the better.

isnotinvain (commented):
Thanks for doing this, parquet's interop story between thrift and other object models often doesn't work for reasons like this. It might be worth making sure that this is all that's required to make it work, it could be that after fixing this there will still be more things that don't interop between parquet-thrift and other parquet integrations. 

I will try to dig into the details of this review, but right away I can say that I think all this new code should come with new tests. The tests should also help us review by demonstrating how it ought to work.

mickjermsurawong-stripe:
hi @isnotinvain thank you for the note. 
Will go through comments by my colleagues and improve docs as you suggested. 
The test class is in `ParquetCollectionFormatForwardCompatibilityTests.scala` not sure if that's what you are looking for, but that demonstrates the format conversion here.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Addressed.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Updated to debug.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: I now simplified to just extract for only the repeated type, and let the recursive call in main method `formatForwardCompatibleType` handle multiple layers instead.
This now becomes a sealed trait for the group type itself, and a helper method to get its repeated type
```
private[scrooge] sealed trait CollectionGroup {
  def groupType: GroupType

  def repeatedType: Type
}
```

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Simplified this to two methods to extract group list and maplist instead.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
mickjermsurawong-stripe: Updated as suggested

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
mickjermsurawong-stripe: ~~ack. Will see how I can do end-to-end testing in this repo.~~
Ah the only thing i don't support here is the target type of `_tuple` format which then would be compatible with the thrift-generated projection source

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatRule.scala
mickjermsurawong-stripe: I now made `_tuple` as a variable where length is calculated from it more explicitly.

scalding-parquet-scrooge/src/test/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibilityTests.scala
mickjermsurawong-stripe: ack. Will see how I can do end-to-end testing in this repo.

mickjermsurawong-stripe:
@joshrosen-stripe, @tchow-stripe 
Thanks for the really helpful review. I've simplified many parts of the code, and addressed the PR feedback.
Pending issues are figuring out to do end-to-end testing here.
@isnotinvain this is ready for another round of review.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,130 @@
...
+    val formattedRepeated = formatter.formatForwardCompatibleRepeatedType(
+      sourceGroup.repeatedType,
+      targetGroup.repeatedType,
+      formatForwardCompatibleType(_, _))
+    // Wrap the formatted repeated type in its original group.
+    // This maintains the field name, and optional/required information
+    sourceGroup.groupType.withNewFields(formattedRepeated)
+  }
+
+  private def findCollectionGroup(typ: Type): Option[CollectionGroup] = {
```
xton: Naming nit:

```suggestion
  private def extractCollectionGroup(typ: Type): Option[CollectionGroup] = {
```

"find" implies that the return values will be a subset of the input, while "extract" (which the underlying functions use) suggests that the whole input will be considered.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,130 @@
...
+   * this is because the subset fields of source node and its optional/required must
+   * be maintained in the formatted result.
+   */
+  private def formatForwardCompatibleType(sourceType: Type, targetType: Type): Type = {
+    (findCollectionGroup(sourceType), findCollectionGroup(targetType)) match {
+      case (Some(sourceGroup: ListGroup), Some(targetGroup: ListGroup)) =>
+        formatForwardCompatibleCollectionGroup[ListGroup](sourceGroup, targetGroup)
+      case (Some(sourceGroup: MapGroup), Some(targetGroup: MapGroup)) =>
+        formatForwardCompatibleCollectionGroup[MapGroup](sourceGroup, targetGroup)
+      case _ if sourceType.isPrimitive || targetType.isPrimitive => // Base case
```
xton: I would explicitly check and throw if one is a primitive but not the other. As written this will silently pass through certain schema mismatches.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,130 @@
...
+    new MessageType(groupResult.getName, groupResult.getFields)
+  }
+
+  /**
+   * Traverse source/target schemas and format nodes of list or map.
+   * The formatting is not to one-to-one node swapping from source to target,
+   * this is because the subset fields of source node and its optional/required must
+   * be maintained in the formatted result.
+   */
+  private def formatForwardCompatibleType(sourceType: Type, targetType: Type): Type = {
```
xton:
```suggestion
  private def formatForwardCompatibleType(fileType: Type, projectionType: Type): Type = {
```

"source" and "target" are very generic names and this function is really only used for one thing. It might make sense to rename them accordingly to make reasoning about the correctness here easier.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,130 @@
...
+      case (Some(sourceGroup: MapGroup), Some(targetGroup: MapGroup)) =>
+        formatForwardCompatibleCollectionGroup[MapGroup](sourceGroup, targetGroup)
+      case _ if sourceType.isPrimitive || targetType.isPrimitive => // Base case
+        sourceType
+      case _ => // Field projection
+        val sourceGroup = sourceType.asGroupType
+        val targetGroup = targetType.asGroupType
+        val resultFields = sourceGroup.getFields.asScala.map { sourceField =>
+          if (!targetGroup.containsField(sourceField.getName)) {
+            if (!sourceField.isRepetition(Repetition.OPTIONAL)) {
```
xton: This check is incorrect, I think. "source" == the file's schema, yes (see above renaming suggestion)? If so, then the optionality of a field which is not included in the projection is not important. We should be iterating over the fields in the projected (target) schema and checking the optionality of the projected fields if they're missing from the file.

xton (commented):
This review is incomplete, I'll finish later this evening. I've got some rename suggestions and some possible bugs so far.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Ah no, source here is actually the projection, and target is the file's scheme.
By that i mean we want to convert projection source schema to be the in the format of the target file schema. 
Given this confusion, i think makes sense that I should rename this.

isnotinvain: on this point, I think naming convention along the lines of: fileSchema, readSchema, projectedFileSchema, projectedReadSchema all help clarify which is which

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: From conversation thread below: 
> By that i mean we want to convert projection source schema to be the in the format of the target file schema.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Ack

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,146 @@
...
+private[scrooge] trait ParquetCollectionFormatter {
+  /**
+   * Format source repeated type in the structure of target repeated type.
+   * @param sourceRepeatedType repeated type from which the formatted result get content
+   * @param targetRepeatedType repeated type from which the formatted result get the structure
+   * @param recursiveSolver solver for the inner content of the repeated type
+   * @return formatted result
+   */
+  def formatForwardCompatibleRepeatedType(sourceRepeatedType: Type,
+                                          targetRepeatedType: Type,
```
xton: Another rename candidate?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatter.scala
```diff
@@ -0,0 +1,270 @@
...
+ * More specifically this handles converting from parquet file created by
+ * [[org.apache.parquet.thrift.ThriftSchemaConvertVisitor]] which always suffix
+ * list element with "_tuple".
+ */
+private[scrooge] object ParquetListFormatter extends ParquetCollectionFormatter {
+
+  private val LOGGER = LoggerFactory.getLogger(getClass)
+
+  def formatForwardCompatibleRepeatedType(repeatedSourceType: Type,
+                                          repeatedTargetType: Type,
```
xton: another rename candidate?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,146 @@
...
+    val formatter = t.runtimeClass.asInstanceOf[Class[T]] match {
+      case c if c == classOf[MapGroup] => ParquetMapFormatter
+      case c if c == classOf[ListGroup] => ParquetListFormatter
+    }
+    val projectedFileRepeatedType = formatter.formatForwardCompatibleRepeatedType(
+      projectedReadGroup.repeatedType,
+      fileGroup.repeatedType,
+      projectFileType(_, _))
+    // Respect optional/required from the projected read group.
+    projectedReadGroup.groupType.withNewFields(projectedFileRepeatedType)
```
xton: Could this end up letting through cases where the file's field is optional and the projection's field is required? If so we should explicitly detect and throw for that (probably in `projectFileType`).

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,146 @@
...
+      case _ if projectedReadType.isPrimitive != fileType.isPrimitive =>
+        throw new DecodingSchemaMismatchException(
+          s"Found schema mismatch between projected read type:\n$projectedReadType\n" +
+            s"and file type:\n${fileType}"
+        )
+      case (Some(projectedReadGroup: ListGroup), Some(fileGroup: ListGroup)) =>
+        projectFileGroup[ListGroup](projectedReadGroup, fileGroup)
+      case (Some(projectedReadGroup: MapGroup), Some(fileGroup: MapGroup)) =>
+        projectFileGroup[MapGroup](projectedReadGroup, fileGroup)
+      case _ => // Field projection
```
xton: Comment nit: I might call this "struct projection". We're projecting through many fields.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetListFormatter.scala
```diff
@@ -0,0 +1,270 @@
...
+private[scrooge] sealed trait SourceOrTarget {
+  def rules: Seq[ParquetListFormatRule]
+  def name: String
+}
+
+private[scrooge] object Source extends SourceOrTarget {
+  override val rules: Seq[ParquetListFormatRule] = Seq(
+    PrimitiveElementRule, PrimitiveArrayRule,
+    GroupElementRule, GroupArrayRule,
+    TupleRule, StandardRule
```
xton: Soooo many cases... That said, I like this approach. Makes it painfully clear the variants that we're supporting.

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetMapFormatter.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Handle legacy type when
+   * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps
+   *
+   * @param sourceRepeatedMapType
+   * @param targetRepeatedMapType
+   */
+  def formatForwardCompatibleRepeatedType(sourceRepeatedMapType: Type,
+                                          targetRepeatedMapType: Type,
```
xton: Candidate for rename?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
```diff
@@ -0,0 +1,159 @@
...
+  private def projectFileGroup[T <: CollectionGroup](projectedReadGroup: T,
+                                                     fileGroup: T,
+                                                     fieldContext: FieldContext)(implicit t: ClassTag[T]): GroupType = {
+
+    val (formatter, updatedFieldContext) = t.runtimeClass.asInstanceOf[Class[T]] match {
+      case c if c == classOf[MapGroup] =>
+        (ParquetMapFormatter, fieldContext.copy(nestedListLevel = fieldContext.nestedListLevel + 1))
+      case c if c == classOf[ListGroup] =>
+        (ParquetListFormatter, fieldContext.copy(nestedListLevel = fieldContext.nestedListLevel + 1))
+    }
```
xton: Let's refactor this to avoid reflection and implicit magic. E.g.:

```suggestion
private def projectFileGroup(projectedReadGroup: T, fileGroup: T, fieldContext: FieldContext, formatter: ParquetCollectionFormatter): GroupType = {
```

xton (changes requested):
One potential simplification and some rename/docs nits. Looking good!

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Changed to `fileRepeatedType` and `readRepeatedType`

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Added the assertion at the "struct projection" which will have access of field-level optional/required

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatForwardCompatibility.scala
mickjermsurawong-stripe: Updated to
```
def projectFileGroup(fileGroup: CollectionGroup,
                               projectedReadGroup: CollectionGroup,
                               fieldContext: FieldContext,
                               formatter: ParquetCollectionFormatter)
```

mickjermsurawong-stripe:
Thanks @xton, addressed the feedback.
@isnotinvain could you help to take a look on this please?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatCompatibility.scala
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Main recursion to get projected file type. Traverse given schemas, filter out unneeded
+   * fields, and format read schema's list/map node to file schema's structure.
+   * The formatting of repeated type is not to one-to-one node swapping because we also have to
+   * handle projection and possible nested collection types in the repeated type.
+   */
+  private def projectFileType(fileType: Type, projectedReadType: Type, fieldContext: FieldContext): Type = {
+    (extractCollectionGroup(projectedReadType), extractCollectionGroup(fileType)) match {
+      case _ if projectedReadType.isPrimitive && fileType.isPrimitive =>
+        projectedReadType
```
joshrosen-stripe, line 62: Do we need to check that both types are the _same_ primitive type? Or is that handled elsewhere? I imagine that certain types of primitive-type-changing projections might be supported (e.g. widening an int to a long) but others might not be (e.g. converting a string to an int).

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatCompatibility.scala
```diff
@@ -0,0 +1,163 @@
...
+   *
+   * @param readRepeatedType repeated type from which the formatted result get content
+   * @param fileRepeatedType repeated type from which the formatted result get the structure
+   * @param recursiveSolver solver for the inner content of the repeated type
+   * @return formatted result
+   */
+  def formatCompatibleRepeatedType(fileRepeatedType: Type,
+                                   readRepeatedType: Type,
+                                   fieldContext: FieldContext,
+                                   recursiveSolver: (Type, Type, FieldContext) => Type): Type
```
joshrosen-stripe: In practice, I think this is always invoked with the `projectFileType` method defined above?

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatCompatibility.scala
mickjermsurawong-stripe: That type compatibility is handled in a separate place here, so by the time this is called we should already have type compatibility.
https://github.com/twitter/scalding/blob/73ce124f3c0156bd41f5394d2b609a5cd74f370e/scalding-parquet-scrooge/src/main/java/com/twitter/scalding/parquet/scrooge/ScroogeReadSupport.java#L133-L136

scalding-parquet-scrooge/src/main/scala/com/twitter/scalding/parquet/scrooge/ParquetCollectionFormatCompatibility.scala
mickjermsurawong-stripe: Yup it is. My thought is to make this recursive call explicit from the caller side so it's clear from the reading within this format compatbility layer instead of jumping into the respective formatter which then cyclically refers back to the caller's method.

mickjermsurawong-stripe:
hi @isnotinvain, would like to follow-up on this if you have more feedback here please.

mickjermsurawong-stripe:
Hi  @isnotinvain, another bump on this please.

Updates from our side: we actually have implemented this in our internally (via subclassing of `ParquetScroogeScheme`), and it does cover production use cases of read Spark job schema.
- list having 3 level, and 2-level in legacy mode, compared to `_tuple` suffixed via generated schema from thrift class.
- map having repeated group named `key_value` instead of `map` 
This will also cover more conversion types as outlined in `ParquetMapFormatter` and `ParquetListFormatter`

Without this patch, scalding job reading Spark job output can either fail hard or falsely read empty collection.

PR #1629, opened 2016-12-10
Upgrade build to sbt 0.13.13

sritchie:
This PR:

- Upgrades to 0.13.13
- fixes all deprecation warnings
- removes the unused `Unidoc.scala` file
- formats plugins

PR #1922, opened 2019-10-04
Re-enable code coverage in CI and switch from Coveralls to Codecov

joshrosen-stripe:
This PR re-enables code coverage in CI and switches from Coveralls to Codecov.

It looks like Coveralls reporting was removed in #1629 (via the removal of the `sbt-coveralls` plugin).

Codecov is already in use in other Twitter OSS repositories, including Algebird.


johnynek:
Thanks for doing this! would be really great to have coverage back on.

I didn't notice we had lost it.

johnynek:
replaced by #1981

PR #1924, opened 2020-01-03
[build] bump cascading to latest 2.x: 2.7.1

grimreaper:


johnynek:
What’s the motivation and what testing has been done on this? You can change the version in your local repo but this change could be risky.

Just being on the latest version (of an unmaintained library) doesn’t seem like a strong enough counter to the risk to me.

grimreaper:
@johnynek - I'm awaiting the tests to run for this branch. Motivation is that the underlying cascading library itself depends on older version of various libraries - and I'm trying to upgrade them.  Further this will only help us if we do decide to eventually resolve #1465. 

(also this change was prompted by @ttim after asking about upgrading internally)

PR #1926, opened 2020-04-02
Add type ascriptions to serialization code

Duhemm:
Scalding  macros expand into a large amount of code, most of which
contained no or very little type ascriptions, leaving a lot of
unnecessary work to the compiler. By explicitly adding these type
ascriptions in the generated code, we can reduce compilation times.

### DIFF
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
@@ -120,8 +120,8 @@ object TraversablesOrderedBuf {
     val innerOrd = q"""
       new _root_.scala.math.Ordering[${innerBuf.tpe}] {
         def compare(a: ${innerBuf.tpe}, b: ${innerBuf.tpe}) = {
-          val $ioa = a
-          val $iob = b
+          val $ioa: ${innerBuf.tpe} = a
+          val $iob: ${innerBuf.tpe} = b
           ${innerBuf.compare(ioa, iob)}
         }
       }
@@ -136,8 +136,8 @@ object TraversablesOrderedBuf {
         val b = freshT("b")
         q"""
         val $innerCompareFn = { (a: _root_.java.io.InputStream, b: _root_.java.io.InputStream) =>
-          val $a = a
-          val $b = b
+          val $a: _root_.java.io.InputStream = a
+          val $b: _root_.java.io.InputStream = b
           ${innerBuf.compareBinary(a, b)}
         };
         _root_.com.twitter.scalding.serialization.macros.impl.ordered_serialization.runtime_helpers.TraversableHelpers.rawCompare($inputStreamA, $inputStreamB)($innerCompareFn)
--- a/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ByteBufferOrderedBuf.scala
+++ b/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/ByteBufferOrderedBuf.scala
@@ -50,9 +50,9 @@ object ByteBufferOrderedBuf {
       val $lenA: _root_.scala.Int = $inputStreamA.readPosVarInt
       val $lenB: _root_.scala.Int = $inputStreamB.readPosVarInt
 
-      val $queryLength = _root_.scala.math.min($lenA, $lenB)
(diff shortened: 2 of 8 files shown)


scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
```diff
@@ -136,8 +136,8 @@ object TraversablesOrderedBuf {
         val b = freshT("b")
         q"""
         val $innerCompareFn = { (a: _root_.java.io.InputStream, b: _root_.java.io.InputStream) =>
-          val $a = a
-          val $b = b
+          val $a: _root_.java.io.InputStream = a
```
johnynek, line 139: I wonder if we even need this binding. It looks unused doesn’t it? Same with `$b`

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
```diff
@@ -120,8 +120,8 @@ object TraversablesOrderedBuf {
     val innerOrd = q"""
       new _root_.scala.math.Ordering[${innerBuf.tpe}] {
         def compare(a: ${innerBuf.tpe}, b: ${innerBuf.tpe}) = {
-          val $ioa = a
-          val $iob = b
+          val $ioa: ${innerBuf.tpe} = a
+          val $iob: ${innerBuf.tpe} = b
```
johnynek, line 124: Why do we need these bindings? Why not just pass a and b directly?

Cc @ianoc do you recall?

johnynek (commented):
This looks great! Just a couple of questions about bindings it looks like we don’t need (and in one case use).

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
ianoc: This was around getting the passed variable names into unique ones that are supplied in the macro context. Why we couldn't just put it on the method signature in these cases though is a good question. Maybe a bug with an old version of this/scala or something. It does seem wasteful. This has good test coverage though so I'd be inclined to try collapsing `$ioa` into the `a: ${innerBuf.tpe}` part above?

ianoc:
ooc even roughly do you know of an impact this had on compilation speed for something?

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
ianoc: we don't pass a and b themselves I believe since the lower section expands inline to more code (And so having unique names to avoid shadowing of some sort is the goal). But why the args don't just have the names idk

Duhemm:
> ooc even roughly do you know of an impact this had on compilation speed for something?

We have some thrift-generated classes where this change helped reducing the compilation time from 90 minutes to 30 seconds. I don't have any actual benchmarks though.

scalding-serialization/src/main/scala/com/twitter/scalding/serialization/macros/impl/ordered_serialization/providers/TraversablesOrderedBuf.scala
johnynek: ahh yeah, i forgot that aspect of it.

johnynek:
holy crap! 90 minutes to 30 seconds is kind of unreal....

johnynek (approved):

johnynek:
the CI is having a weird issue. most of the jobs can't fetch from maven. Not sure if it is travis or what.

mosesn:
> holy crap! 90 minutes to 30 seconds is kind of unreal....

yeah, we were unable to add a new trait to a scrooge-generated class because it made SQ time out ¯\\\_(ツ)\_/¯

johnynek:
the CI failed again... it seems that the CI is not in a working order at the moment.

I haven't had time to look into this.

johnynek:
Can you merge develop to get #1930 which may get CI green?

ttim:
@johnynek I ran CI with this PR on top of `develop` (https://travis-ci.org/github/twitter/scalding/builds/684520554) and it passed, so I think we're good to merge this one.

PR #1931, opened 2020-06-01
Add additional guards in `ReferencedClassFinder`

ttim:
During upgrade to Scala 2.12.10 in Twitter we discovered `illegal cyclic reference` exception failing from `scalaType.member(universe.stringToTermName(field.getName)).typeSignature` code in `ReferencedClassFinder`.

In this PR I've guarded this code with the same check as we do for `getClassType`.

Unfortunately I wasn't able to replicate this failure is OSS easily so added this guard without failing test.


johnynek:
Thanks for sending the PR!

PR #1934, opened 2020-09-28
TypedPipe.groupRandomly() use Ordering with EquivSerialization

tlazaro:
Seen `"Scalding's ordered serialization logic exhausted the finite supply of boxed classes."` error when creating many parallel Executions from the same job. We traced it to groupRandomly() not using an EquivSerialization in its groupBy(), taking up new slots of boxed classes for every Execution.

Converted `identityOrdering` to a case object and extended EquivSerialization following the pattern of `com.twitter.scalding.serialization.UnitOrderedSerialization`.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
@@ -22,7 +22,7 @@ import com.twitter.algebird.{ Aggregator, Batched, Monoid, Semigroup }
 import com.twitter.scalding.TupleConverter.singleConverter
 import com.twitter.scalding._
 import com.twitter.scalding.typed.functions.{ AsLeft, AsRight, Constant, ConstantKey, DropValue1, Identity, MakeKey, GetKey, GetValue, RandomFilter, RandomNextInt, Swap, TuplizeFunction, WithConstant, PartialFunctionToFilter, SubTypes }
-import com.twitter.scalding.serialization.{ OrderedSerialization, UnitOrderedSerialization }
+import com.twitter.scalding.serialization.{ EquivSerialization, OrderedSerialization, UnitOrderedSerialization }
 import com.twitter.scalding.serialization.OrderedSerialization.Result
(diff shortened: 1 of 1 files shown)


scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -143,17 +143,16 @@ object TypedPipe extends Serializable {
...
-      override def hash(x: Int): Int = x
-    }
+    
+    override def compareBinary(a: InputStream, b: InputStream): Result = delegate.compareBinary(a, b)
+    override def compare(x: Int, y: Int): Int = delegate.compare(x, y)
+    override def dynamicSize(t: Int): Option[Int] = delegate.dynamicSize(t)
+    override def write(out: OutputStream, t: Int): Try[Unit] = delegate.write(out, t)
+    override def read(in: InputStream): Try[Int] = delegate.read(in)
+    override def staticSize: Option[Int] = delegate.staticSize
+    override def hash(x: Int): Int = x
```
johnynek, line 155: small note: why don't we do `delegate.hash(x)` here? seems weird that we delegate through the others. Maybe a good reason, we have had weird bugs related to non-uniform partitioning on this method before, and this may be related to fixing that, but it would be nice if we had commented that.

scalding-core/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -143,17 +143,16 @@ object TypedPipe extends Serializable {
       }
   }
 
-  private val identityOrdering: OrderedSerialization[Int] = {
+  private case object IdentityOrdering extends OrderedSerialization[Int] with EquivSerialization[Int] {
     val delegate = BinaryOrdering.ordSer[Int]
-    new OrderedSerialization[Int] {
```
johnynek, line 148: I *think* this would have worked with just adding `new OrderedSerialization[Int] with EquivSerialization[Int]` but I think this is fine too.

johnynek (approved):

PR #1936, opened 2020-12-22
Remove mapreduce.input.fileinputformat.inputdir setting in memory source

navinvishy:
Memory source sets the mapreduce.input.fileinputformat.inputdir property to
a random UUID value. Often in clusters with HDFS federation, paths like that are
not valid namespaces.
While this path is not usually checked since this is a memory source, in clusters
where Kerberos is enabled, Hadoop lists the input dirs to a job to get delegation
tokens. Since this path is not valid, this results in a FileNotFoundException on
a Kerberized cluster.

This patch removes this setting in Scalding memory sources since they are not valid anyway.


ttim:
I think this should be fine and shouldn't break existing code,

@johnynek we're not suppose to rely on `inputPaths` for in memory sources in our code, right?

PR #1938, opened 2021-04-01
Fix race between jvm shutdown and `writer.finished`

ttim:
Currently `writer.finished` happens in `onComplete` callback on `Future` result in `Exucution`. However since `onComplete` isn't being called before future being resolved and called asynchroniously after future being resolved, it leads to a race and runtime error:
- User's code as last operation in `main` executes `Execution`
- `onComplete` with `writer.finished` is being scheduled
- result `Future` gets resolved and jvm starts to shutdown
- `writer.finished` starts to execute and in case of cascading backend adds shutdown hook
- which is not permitted during jvm shutdown and breaks

To fix this behaviour I made `onComplete` logic to happen before result future get resolved by changing `onComplete` to `andThen`

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -155,9 +155,8 @@ sealed trait Execution[+T] extends Serializable { self: Product =>
     val exec = Execution.optimize(conf, this)
     // get on Trampoline
     val CFuture(fut, cancelHandler) = exec.runStats(confWithId, mode, ec)(cec).get
-    val result = fut.map(_._1)
     // When the final future in complete we stop the submit thread
-    result.onComplete { t =>
+    val result = fut.map(_._1).andThen { case t =>
       if (t.isFailure) {
         // cancel running executions if this was a failure
         Await.ready(cancelHandler.stop(), scala.concurrent.duration.Duration(30, scala.concurrent.duration.SECONDS))
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -15,21 +15,22 @@ limitations under the License.
  */
 package com.twitter.scalding
 
-import cascading.flow.{ FlowDef, Flow }
-import com.stripe.dagon.{ Dag, Id, Rule }
+import cascading.flow.{Flow, FlowDef}
+import com.stripe.dagon.{Dag, Id, Rule}
 import com.twitter.algebird.monad.Trampoline
-import com.twitter.algebird.{ Monoid, Monad, Semigroup }
+import com.twitter.algebird.{Monad, Monoid, Semigroup}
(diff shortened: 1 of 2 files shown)


tlazaro:
LGTM, both onComplete vs andThen are for performing side-effects when the Future is done. But andThen makes it so the side-effect runs before anything else. Seems like exactly what we wanted in this case.

ttim:
@johnynek does it seem reasonable to you? I'm wondering is there something I missing

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -155,9 +155,8 @@ sealed trait Execution[+T] extends Serializable { self: Product =>
     val exec = Execution.optimize(conf, this)
     // get on Trampoline
     val CFuture(fut, cancelHandler) = exec.runStats(confWithId, mode, ec)(cec).get
-    val result = fut.map(_._1)
     // When the final future in complete we stop the submit thread
-    result.onComplete { t =>
+    val result = fut.map(_._1).andThen { case t =>
       if (t.isFailure) {
         // cancel running executions if this was a failure
         Await.ready(cancelHandler.stop(), scala.concurrent.duration.Duration(30, scala.concurrent.duration.SECONDS))
```
johnynek: I think we are technically supposed to do:
```scala
blocking {
  Await.ready(cancelHandler.stop(), ...)
```
here:

https://docs.scala-lang.org/overviews/core/futures.html#blocking-inside-a-future

Other than that, it looks good.

Small nit, we should probably import Duration and SECOND from `scala.concurrent.duration`

ttim: Can't import `Duration` because of scalding `Duration`, imported `scala.concurrent.duration` instead.

ttim: Added blocking block around `Await`

scalding-parquet/src/test/java/com/twitter/scalding/parquet/example/TestInputOutputFormat.java
```diff
@@ -79,7 +79,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(TestInputOutputFormat.class);
 
   final Path parquetPath = new Path("target/test/example/TestInputOutputFormat/parquet");
-  final Path inputPath = new Path("src/test/java/org/apache/parquet/hadoop/example/TestInputOutputFormat.java");
```
johnynek, line 82: Why this change?

johnynek (approved):
One minor question but other than that this looks good.

scalding-parquet/src/test/java/com/twitter/scalding/parquet/example/TestInputOutputFormat.java
ttim: This test was copied by @tlazaro in previous PR, and it relies on reading itself to run properly. Since path of copied file is different it started to fail, so I changed it.

johnynek (approved):

PR #1939, opened 2021-06-30
Update sbt to 1.5.6

ttim:
This PR updates sbt version to latest available one: `1.5.6`.

I've updated sbt plugins versions to be compatible with newer version of sbt, as well as updates `build.sbt` to work successfully.

I'll follow up with two PRs: to move to scalafmt and mdoc plugins


build.sbt
```diff
@@ -44,14 +40,14 @@ val jlineVersion = "2.14.3"
 
 val printDependencyClasspath = taskKey[Unit]("Prints location of the dependencies")
 
-val sharedSettings = assemblySettings ++ scalariformSettings ++ Seq(
+val sharedSettings = com.typesafe.sbt.SbtScalariform.scalariformSettings ++ Seq(
```
johnynek: can we just remove scalariform? scalafmt is the winner here.

build.sbt
```diff
@@ -476,13 +470,21 @@ lazy val scaldingHRaven = module("hraven").settings(
...
     "org.apache.hbase" % "hbase-client" % hbaseVersion % "provided",
     "org.apache.hbase" % "hbase-common" % hbaseVersion % "provided",
     "org.apache.hbase" % "hbase-server" % hbaseVersion % "provided",
     "org.slf4j" % "slf4j-api" % slf4jVersion,
-    "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided"
+    "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
+
+    "com.twitter" % "util-registry_2.10" % "6.34.0",
+    "com.twitter" % "util-core_2.10" % "6.34.0",
+    "com.twitter" % "util-jvm_2.10" % "6.34.0",
```
johnynek: can we make a val holding this version constant for this?

Also, fixing to a particular scala version here is really scary. That sounds like a great way for runtime errors if someone tries `2.11` for this artifact.

If it only runs for 2.10, we should have the build filter this out when scalaVersion is not 2.10.

build.sbt
```diff
@@ -720,15 +722,17 @@ lazy val docSettings = Seq(
...
+  .enablePlugins(ScalaUnidocPlugin)
+  .enablePlugins(GhpagesPlugin)
   .settings(moduleName := "scalding-docs")
   .settings(sharedSettings)
   .settings(noPublishSettings)
-  .settings(unidocSettings)
-  .settings(ghpages.settings)
   .settings(docSettings)
-  .settings(tutScalacOptions ~= (_.filterNot(Set("-Ywarn-unused-import", "-Ywarn-dead-code"))))
+  .settings(scalacOptions in Tut ~= (_.filterNot(Set("-Ywarn-unused-import", "-Ywarn-dead-code"))))
```
johnynek: side note: we should migrate to mdoc, tut is EOL I believe.

also, I though `Tut / scalacOption` is the new syntax?

build.sbt
tlazaro: Looking at the rest of the code here I think the pattern would be to capture the version in a variable, like hbaseVersion and hadoopVersion.

Why are these Scala 2.10?

project/plugins.sbt
```diff
@@ -1,21 +1,21 @@
...
 addSbtPlugin("com.typesafe"       % "sbt-mima-plugin"     % "0.1.14")
-addSbtPlugin("com.typesafe.sbt"   % "sbt-ghpages"         % "0.5.4")
-addSbtPlugin("com.typesafe.sbt"   % "sbt-git"             % "0.6.2")
-addSbtPlugin("com.typesafe.sbt"   % "sbt-scalariform"     % "1.3.0")
-addSbtPlugin("org.scoverage"      % "sbt-scoverage"       % "1.5.0")
-addSbtPlugin("org.xerial.sbt"     % "sbt-sonatype"        % "1.0")
-addSbtPlugin("org.wartremover"    % "sbt-wartremover"     % "2.1.1")
+addSbtPlugin("com.typesafe.sbt"   % "sbt-ghpages"         % "0.6.3")
+addSbtPlugin("com.typesafe.sbt"   % "sbt-git"             % "1.0.0")
+addSbtPlugin("org.scalariform"    % "sbt-scalariform"     % "1.8.3")
```
johnynek, line 16: think we can just nuke this.

build.sbt
```diff
@@ -720,15 +722,17 @@ lazy val docSettings = Seq(
...
   .dependsOn(scaldingCore)
+*/
```
tlazaro: New line at EOF

build.sbt
tlazaro: Why not keeping the import?

ttim: Was failing with compilation error with duplicate imports, will try to return it back

ttim: Should we follow up with this separately?

build.sbt
ttim: Good question. I believe currently hraven is compiled only against scala 2.10, hence this. By making this deps explicit I made behaviour to be the same as before (resolve changed and it started to fail with resolving to non existent version).

I can see 2 alternatives:
- remove module completely
- don't add any explicit dependencies here, or make them provided

ttim: Because that's what hraven depends on... See more context in comment above.

build.sbt
ttim: I'll change it. Copied from `Tut` docs.

project/plugins.sbt
ttim: And change to scalafmt? I can follow up with that.

build.sbt
ttim: Removed it for now, it compiles with this excludes successfully, will be up to users of this module to add this deps

ttim: Decided to remove this deps

build.sbt
ttim: Commented this line for now,  will migrate to mdoc in separate PR

build.sbt
ttim: Fixed

johnynek:
question: why not use the latest sbt?

ttim:
> question: why not use the latest sbt?

I've used last version which works with our `sbt` script, will follow up with upgrading it to most recent version.

ttim:
> question: why not use the latest sbt?

Updated to the latest available sbt version

johnynek (approved):

johnynek:
`error: File /home/travis/build/twitter/scalding/docs/docs does not exist.`

I think this may have to do with tut being disabled.

ttim:
> `error: File /home/travis/build/twitter/scalding/docs/docs does not exist.`
> 
> I think this may have to do with tut being disabled.

Yes. I'll follow up with docs fix, since it's already broken in `develop`, and that's not a regression.

There is one repl test which started to fail which I investigate currently.

ttim:
@johnynek finally was able to figure out why repl test started to fail - travis enables batch mode for sbt runs...

I'm planning to follow up with mdoc & scalafmt in separate PRs, let me know if that's ok with you.

johnynek (approved):

PR #1940, opened 2021-07-30
Beam-Dataflow support for Scalding

Franco1010:
This is a first PR for beam support to scalding

It now extends support to basic operations in SimpleWrite only:
- Filter
- FilterKeys
- Mapped
- FlatMapped
- MapValues
- FlatMapValues

Added support for reading from Source path and Iterable
Added support for writing into Sink path
Added KryoCoder sharing kryo instances and buffers
Added unit tests for BeamMode end to end

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,6 +35,7 @@ val scalaCheckVersion = "1.13.4"
 val scalaTestVersion = "3.0.1"
 val scroogeVersion = "18.9.0"
 val sparkVersion = "2.4.0"
+val beamVersion = "2.29.0"
 val slf4jVersion = "1.6.6"
 val thriftVersion = "0.9.3"
 val junitVersion = "4.10"
@@ -231,6 +232,7 @@ lazy val scalding = Project(
   executionTutorial,
   scaldingSerialization,
   scaldingSpark,
+  scaldingBeam,
   scaldingThriftMacros
 )
 
@@ -358,6 +360,17 @@ lazy val scaldingSpark = module("spark").settings(
     "org.apache.spark" %% "spark-sql" % sparkVersion
     )
   ).dependsOn(scaldingCore)
+
+lazy val scaldingBeam = module("beam").settings(
+  libraryDependencies ++= Seq(
+    "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-google-cloud-platform-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-sorter" % beamVersion,
+    "org.apache.beam" % "beam-runners-direct-java" % beamVersion % "test",
+    // Including this dependency since scalding configuration depends on hadoop
+    "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
+  )
+).dependsOn(scaldingCore)
 
 lazy val scaldingCommons = module("commons").settings(
   libraryDependencies ++= Seq(
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
@@ -1,0 +1,87 @@
+package com.twitter.scalding.beam_backend
+
+import com.stripe.dagon.{ FunctionK, Memoize, Rule }
+import com.twitter.chill.KryoInstantiator
+import com.twitter.chill.config.ScalaMapConfig
+import com.twitter.scalding.Config
+import com.twitter.scalding.serialization.KryoHadoop
+import com.twitter.scalding.typed.functions.{ FilterKeysToFilter, FlatMapValuesToFlatMap, MapValuesToMap }
+import com.twitter.scalding.typed.{ OptimizationRules, Resolver, TypedPipe, TypedSource }
+
+object BeamPlanner {
+  def plan(config: Config, srcs: Resolver[TypedSource, BeamSource]): FunctionK[TypedPipe, BeamOp] = {
+    implicit val kryoCoder: KryoCoder = new KryoCoder(defaultKryoCoderConfiguration(config))
+    Memoize.functionK(f = new Memoize.RecursiveK[TypedPipe, BeamOp] {
+
+      import TypedPipe._
+
+      def toFunction[A] = {
+        case (f @ Filter(_, _), rec) =>
+          def go[T](f: Filter[T]): BeamOp[T] = {
+            val Filter(p, fn) = f
+            rec[T](p).filter(fn)
+          }
+          go(f)
+        case (fk @ FilterKeys(_, _), rec) =>
+          def go[K, V](node: FilterKeys[K, V]): BeamOp[(K, V)] = {
+            val FilterKeys(pipe, fn) = node
+            rec(pipe).filter(FilterKeysToFilter(fn))
+          }
+          go(fk)
+        case (Mapped(input, fn), rec) =>
+          val op = rec(input)
+          op.map(fn)
+        case (FlatMapped(input, fn), rec) =>
+          val op = rec(input)
+          op.flatMap(fn)
+        case (f @ MapValues(_, _), rec) =>
+          def go[K, V, U](node: MapValues[K, V, U]): BeamOp[(K, U)] = {
+            val MapValues(pipe, fn) = node
+            rec(pipe).map(MapValuesToMap[K, V, U](fn))
+          }
+          go(f)
+        case (f @ FlatMapValues(_, _), rec) =>
+          def go[K, V, U](node: FlatMapValues[K, V, U]): BeamOp[(K, U)] = {
+            val FlatMapValues(pipe, fn) = node
+            rec(pipe).flatMap(FlatMapValuesToFlatMap[K, V, U](fn))
+          }
+          go(f)
+        case (SourcePipe(src), _) =>
+          BeamOp.Source(config, src, srcs(src))
+        case (IterablePipe(iterable), _) =>
+          BeamOp.FromIterable(iterable, kryoCoder)
+        case (wd: WithDescriptionTypedPipe[a], rec) =>
+          rec[a](wd.input)
+      }
+    })
+  }
+
+  def defaultKryoCoderConfiguration(config: Config): KryoInstantiator = {
+    config.getKryo match {
+      case Some(kryoInstantiator) => kryoInstantiator
+      case None => new KryoHadoop(new ScalaMapConfig(Map.empty))
+    }
+  }
+
+  def defaultOptimizationRules(config: Config): Seq[Rule[TypedPipe]] = {
+    def std(forceHash: Rule[TypedPipe]) =
+      (OptimizationRules.standardMapReduceRules :::
+        List(
+          OptimizationRules.FilterLocally, // after filtering, we may have filtered to nothing, lets see
+          OptimizationRules.simplifyEmpty,
+          // add any explicit forces to the optimized graph
+          Rule.orElse(List(
+            forceHash,
+            OptimizationRules.RemoveDuplicateForceFork))))
+
+    config.getOptimizationPhases match {
+      case Some(tryPhases) => tryPhases.get.phases
+      case None =>
+        val force =
+          if (config.getHashJoinAutoForceRight) OptimizationRules.ForceToDiskBeforeHashJoin
+          else Rule.empty[TypedPipe]
+        std(force)
+    }
+  }
+}
+
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
@@ -1,0 +1,84 @@
+package com.twitter.scalding.beam_backend
+
+import com.twitter.scalding.Execution.Writer
+import com.twitter.scalding.typed.{ Resolver, TypedSink, TypedSource }
+import com.twitter.scalding.{ Config, Mode, TextLine }
+import org.apache.beam.sdk.Pipeline
+import org.apache.beam.sdk.io.TextIO
+import org.apache.beam.sdk.options.PipelineOptions
+import org.apache.beam.sdk.values.PCollection
+
+case class BeamMode(
+  pipelineOptions: PipelineOptions,
+  sources: Resolver[TypedSource, BeamSource],
+  sink: Resolver[TypedSink, BeamSink]) extends Mode {
+  def newWriter(): Writer = new BeamWriter(this)
+}
+
+object BeamMode {
+  def empty(pipelineOptions: PipelineOptions): BeamMode =
+    BeamMode(pipelineOptions, Resolver.empty, Resolver.empty)
+  def default(pipelineOptions: PipelineOptions): BeamMode =
+    BeamMode(pipelineOptions, BeamSource.Default, BeamSink.Default)
+}
+
+trait BeamSource[+A] extends Serializable {
+  def read(pipeline: Pipeline, config: Config): PCollection[_ <: A]
+}
+
+object BeamSource extends Serializable {
+  val Default: Resolver[TypedSource, BeamSource] = {
+    new Resolver[TypedSource, BeamSource] {
+      def apply[A](source: TypedSource[A]): Option[BeamSource[A]] = {
+        source match {
+          case tl: TextLine =>
+            tl.localPaths match {
+              case path :: Nil => Some(textLine(path))
+              case _ => throw new Exception("Can not accept multiple paths to BeamSource")
+            }
+          case _ => None
+        }
+      }
+    }
+  }
+
+  def textLine(path: String): BeamSource[String] =
+    new BeamSource[String] {
+      override def read(
+        pipeline: Pipeline,
+        config: Config): PCollection[_ <: String] = {
+        pipeline.apply(TextIO.read().from(path))
+      }
+    }
+}
+
+trait BeamSink[-A] extends Serializable {
+  def write(pipeline: Pipeline, config: Config, pc: PCollection[_ <: A]): Unit
+}
+
+object BeamSink extends Serializable {
+  val Default: Resolver[TypedSink, BeamSink] = {
+    new Resolver[TypedSink, BeamSink] {
+      def apply[A](sink: TypedSink[A]): Option[BeamSink[A]] = {
+        sink match {
+          case tl: TextLine =>
+            tl.localPaths match {
+              case path :: Nil => Some(textLine(path).asInstanceOf[BeamSink[A]])
+              case _ => throw new Exception("Can not accept multiple paths to BeamSink")
+            }
+          case _ => None
+        }
+      }
+    }
+  }
+
+  def textLine(path: String): BeamSink[String] =
+    new BeamSink[String] {
+      override def write(
+        pipeline: Pipeline,
+        config: Config,
+        pc: PCollection[_ <: String]): Unit = {
+        pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
+      }
+    }
+}
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
@@ -1,0 +1,59 @@
+package com.twitter.scalding.beam_backend
+
+import com.twitter.scalding.Config
+import com.twitter.scalding.beam_backend.BeamFunctions._
+import com.twitter.scalding.typed.TypedSource
+import org.apache.beam.sdk.Pipeline
+import org.apache.beam.sdk.transforms._
+import org.apache.beam.sdk.values.PCollection
+import scala.collection.JavaConverters._
+
+sealed abstract class BeamOp[+A] {
+  import BeamOp.TransformBeamOp
+
+  def run(pipeline: Pipeline): PCollection[_ <: A]
+
+  def map[B](f: A => B)(implicit kryoCoder: KryoCoder): BeamOp[B] =
+    parDo(MapFn(f))
+
+  def parDo[C >: A, B](f: DoFn[C, B])(implicit kryoCoder: KryoCoder): BeamOp[B] = {
+    val pTransform = new PTransform[PCollection[C], PCollection[B]]() {
+      override def expand(input: PCollection[C]): PCollection[B] = input.apply(ParDo.of(f))
+    }
+    applyPTransform(pTransform)
+  }
+
+  def filter(f: A => Boolean)(implicit kryoCoder: KryoCoder): BeamOp[A] =
+    applyPTransform(Filter.by[A, ProcessFunction[A, java.lang.Boolean]](ProcessPredicate(f)))
+
+  def applyPTransform[C >: A, B](f: PTransform[PCollection[C], PCollection[B]])(implicit kryoCoder: KryoCoder): BeamOp[B] =
+    TransformBeamOp(this, f, kryoCoder)
+
+  def flatMap[B](f: A => TraversableOnce[B])(implicit kryoCoder: KryoCoder): BeamOp[B] =
+    parDo(FlatMapFn(f))
+}
+
+object BeamOp extends Serializable {
+  final case class Source[A](
+    conf: Config,
+    original: TypedSource[A],
+    input: Option[BeamSource[A]]) extends BeamOp[A] {
+    def run(pipeline: Pipeline): PCollection[_ <: A] =
+      input match {
+        case None => throw new IllegalArgumentException(s"source $original was not connected to a beam source")
+        case Some(src) => src.read(pipeline, conf)
+      }
+  }
+
+  final case class FromIterable[A](iterable: Iterable[A], kryoCoder: KryoCoder) extends BeamOp[A] {
+    override def run(pipeline: Pipeline): PCollection[_ <: A] = {
+      pipeline.apply(Create.of(iterable.asJava).withCoder(kryoCoder))
+    }
+  }
+
+  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder)() extends BeamOp[B] {
+    def run(pipeline: Pipeline): PCollection[B] = {
+      source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
+    }
+  }
+}
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
@@ -1,0 +1,64 @@
+package com.twitter.scalding.beam_backend
+
+import com.stripe.dagon.Rule
+import com.twitter.scalding.Execution.{ ToWrite, Writer }
+import com.twitter.scalding.typed._
+import com.twitter.scalding.{ CFuture, CancellationHandler, Config, Execution, ExecutionCounters }
+import java.util.concurrent.atomic.AtomicLong
+import org.apache.beam.sdk.Pipeline
+import scala.annotation.tailrec
+import scala.concurrent.{ ExecutionContext, Future }
+
+class BeamWriter(val beamMode: BeamMode) extends Writer {
+  private val state = new AtomicLong()
+
+  override def start(): Unit = ()
+
+  override def finished(): Unit = ()
+
+  def getForced[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[TypedPipe[T]] = ???
+
+  def getIterable[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[Iterable[T]] = ???
+
+  override def execute(
+    conf: Config,
+    writes: List[ToWrite[_]])(implicit cec: ExecutionContext): CFuture[(Long, ExecutionCounters)] = {
+    import Execution.ToWrite._
+    val planner = BeamPlanner.plan(conf, beamMode.sources)
+    val phases: Seq[Rule[TypedPipe]] = BeamPlanner.defaultOptimizationRules(conf)
+    val optimizedWrites = ToWrite.optimizeWriteBatch(writes, phases)
+    val pipeline = Pipeline.create(beamMode.pipelineOptions)
+
+    @tailrec
+    def rec(optimizedWrites: List[OptimizedWrite[TypedPipe, _]]): Unit = {
+      optimizedWrites match {
+        case Nil => ()
+        case x :: xs =>
+          x match {
+            case OptimizedWrite(pipe, ToWrite.SimpleWrite(opt, sink)) => {
+              val pcoll = planner(opt).run(pipeline)
+              beamMode.sink(sink) match {
+                case Some(ssink) =>
+                  ssink.write(pipeline, conf, pcoll)
+                case _ => throw new Exception(s"unknown sink: $sink when writing $pipe")
+              }
+              rec(xs)
+            }
+            //TODO: handle Force and ToIterable
+            case _ => ???
+          }
+      }
+    }
+    rec(optimizedWrites)
+    val result = pipeline.run
+    val runId = state.getAndIncrement()
+    CFuture(
+      Future{
+        result.waitUntilFinish()
+        (runId, ExecutionCounters.empty)
+      },
+      CancellationHandler.fromFn{ ec =>
+        Future{ result.cancel(); () }(ec)
+      })
+  }
+}
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
@@ -1,0 +1,29 @@
+package com.twitter.scalding.beam_backend
+
+import com.esotericsoftware.kryo.io.Input
+import com.twitter.chill.{ KryoInstantiator, KryoPool }
+import com.twitter.scalding.serialization.JavaStreamEnrichments.{ RichInputStream, RichOutputStream }
+import java.io.{ InputStream, OutputStream }
+import org.apache.beam.sdk.coders.AtomicCoder
+import scala.language.implicitConversions
+
+final class KryoCoder(kryoInstantiator: KryoInstantiator) extends AtomicCoder[Any] {
+  @transient private[this] lazy val kryoPool: KryoPool = KryoPool.withByteArrayOutputStream(Runtime
+    .getRuntime
+    .availableProcessors, kryoInstantiator)
+  override def encode(value: Any, os: OutputStream): Unit = {
+    val bytes = kryoPool.toBytesWithClass(value)
+    os.writePosVarInt(bytes.length)
+    os.write(bytes)
+    os.flush()
+  }
+  override def decode(is: InputStream): Any = {
+    val size = is.readPosVarInt
+    val input = new Input(is, size)
+    kryoPool.fromBytes(input.readBytes(size))
+  }
+}
+object KryoCoder {
+  implicit def castType[T](kryoCoder: KryoCoder): AtomicCoder[T] =
+    kryoCoder.asInstanceOf[AtomicCoder[T]]
+}
--- a/scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
+++ b/scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
@@ -1,0 +1,103 @@
+package com.twitter.scalding.beam_backend
+
+import com.twitter.scalding.{Config, TextLine, TypedPipe}
+import java.io.File
+import java.nio.file.Paths
+import org.apache.beam.sdk.options.{PipelineOptions, PipelineOptionsFactory}
+import org.scalatest.{BeforeAndAfter, FunSuite}
+import scala.io.Source
+
+class BeamBackendTests extends FunSuite with BeforeAndAfter {
+
+  private var pipelineOptions: PipelineOptions = _
+  private var testPath: String = _
+
+  def tmpPath(suffix: String): String = {
+    Paths.get(testPath, suffix).toString
+  }
+
+  before {
+    testPath = Paths.get(
+      System.getProperty("java.io.tmpdir"),
+      "scalding",
+      "beam_backend").toString
+    pipelineOptions = PipelineOptionsFactory.create()
+  }
+
+  after {
+    removeDir(testPath)
+  }
+
+  test("map & flatMap operations"){
+    val outRoute = tmpPath("out1")
+    val ex = TypedPipe
+      .from(0 to 10)
+      .map(x => x * 2)
+      .flatMap(x => 0 to x)
+      .map(x => (x, x))
+      .mapValues(x => x * 2)
+      .flatMapValues(x => 0 to x)
+      .map(_.toString)
+      .writeExecution(TextLine(outRoute))
+
+    val bmode = BeamMode.default(pipelineOptions)
+    ex.waitFor(Config.empty, bmode).get
+    val expectedResult = List
+      .range(0, 11)
+      .map(x => x * 2)
+      .flatMap(x => 0 to x)
+      .map(x => (x, x))
+      .map(x => (x._1, x._2 * 2))
+      .flatMap(x => (0 to x._2).map(nx => (x._1, nx)))
+      .map(_.toString)
+      .sorted
+    val result = getContents(testPath, outRoute).sorted
+
+    assert(result == expectedResult)
+
+    removeDir(testPath)
+  }
+
+  test("filter operations"){
+    val outRoute = tmpPath("out1")
+    val ex = TypedPipe
+      .from(0 to 100)
+      .filter(x => x % 2 == 0)
+      .map(x => (x, x))
+      .filterKeys(x => x % 4 == 0)
+      .map(_.toString)
+      .writeExecution(TextLine(outRoute))
+
+    val bmode = BeamMode.default(pipelineOptions)
+    ex.waitFor(Config.empty, bmode).get
+    val expectedResult = List
+      .range(0, 101)
+      .filter(x => x % 2 == 0)
+      .map(x => (x, x))
+      .filter(x => x._1 % 4 == 0)
+      .map(_.toString)
+      .sorted
+    val result = getContents(testPath, outRoute).sorted
+
+    assert(result == expectedResult)
+
+    removeDir(testPath)
+  }
+
+  private def getContents(path: String, prefix: String): List[String] = {
+    new File(path).listFiles.flatMap(file => {
+      if(file.getPath.startsWith(prefix)){
+        Source.fromFile(file).getLines().flatMap(line => line.split("\\s+").toList)
+      }else List.empty[String]
+    }).toList
+  }
+
+  private def removeDir(path: String): Unit = {
+    def deleteRecursively(file: File): Unit = {
+      if (file.isDirectory) file.listFiles.foreach(deleteRecursively)
+      if (file.exists && !file.delete)
+        sys.error(s"Unable to delete ${file.getAbsolutePath}")
+    }
+    deleteRecursively(new File(path))
+  }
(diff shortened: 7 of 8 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import com.esotericsoftware.kryo.Kryo
+import com.twitter.chill.config.ScalaMapConfig
+import com.twitter.scalding.serialization
+import java.io.{ByteArrayOutputStream, InputStream, OutputStream}
+import org.apache.beam.sdk.coders.AtomicCoder
+
+class KryoCoder[T] extends AtomicCoder[T] {
+  override def encode(value: T, outStream: OutputStream): Unit = {
+    @transient lazy val kryo: Kryo = new serialization.KryoHadoop(new ScalaMapConfig(Map.empty)).newKryo
```
johnynek: I don't think you want to make a kryo for each item, that will be very slow. I think you want a private member variable.

scalding-beam/src/test/scala/com.twitter.scalding.beam_backend/BeamBackendTests.scala
```diff
@@ -0,0 +1,49 @@
...
+    val pipe2 = TypedPipe
+      .from(1 to 20)
+      .map(x => (x, x))
+      .flatMapValues(x => 1 to x)
+      .map(_.toString())
+    val ex2 = pipe2.writeExecution(TextLine(outRoute2))
+
+    val bmode = BeamMode.default(pipeline)
+    val ex3 = Execution.zip(ex1, ex2)
+    ex3.waitFor(Config.empty, bmode).get
```
johnynek: there is no assertion here. Can we check something? I guess that's todo.

One thing that would be really nice is eventually implement this test:

https://github.com/twitter/scalding/blob/b6d465f6fc308d66e5a53c62662d5860846d8dc4/scalding-core/src/test/scala/com/twitter/scalding/typed/memory_backend/MemoryTest.scala#L74

johnynek (commented):
this is a really exciting start!

Very happy to see a new backend being implemented!

scalding-beam/src/test/scala/com.twitter.scalding.beam_backend/BeamBackendTests.scala
```diff
@@ -0,0 +1,49 @@
...
+  private var pipeline: Pipeline = _
+
+  before {
+    val options = PipelineOptionsFactory.create()
+    pipeline = Pipeline.create(options)
+  }
+
+  after {
+    pipeline = null
+  }
```
nownikhil: We can just remove this code block

scalding-beam/src/test/scala/com.twitter.scalding.beam_backend/BeamBackendTests.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import com.twitter.scalding.{Config, Execution, TextLine, TypedPipe}
+import org.apache.beam.sdk.Pipeline
+import org.apache.beam.sdk.options.PipelineOptionsFactory
+import org.scalatest.{BeforeAndAfter, FunSuite}
+
+class BeamBackendTests extends FunSuite with BeforeAndAfter {
+
+  private val master = "local[2]"
+  private val appName = "beam-backend-tests"
```
nownikhil: Remove these as well

scalding-beam/src/test/scala/com.twitter.scalding.beam_backend/BeamBackendTests.scala
```diff
@@ -0,0 +1,49 @@
...
+    val options = PipelineOptionsFactory.create()
+    pipeline = Pipeline.create(options)
+  }
+
+  after {
+    pipeline = null
+  }
+
+  test("test") {
+    val pref = "scalding-beam/src/test/scala/com.twitter.scalding.beam_backend/BeamBackendTests.scala"
```
nownikhil: Create a temp dir for this.
You can use java.nio.file.Files.createTempDirectory

build.sbt
```diff
@@ -359,6 +361,18 @@ lazy val scaldingSpark = module("spark").settings(
     )
   ).dependsOn(scaldingCore)
 
+lazy val scaldingBeam = module("beam").settings(
+  libraryDependencies ++= Seq(
+    "org.apache.beam" % "beam-runners-direct-java" % beamVersion,
```
ttim: I would prefer this to be a dependency only in `test` scope

build.sbt
```diff
@@ -359,6 +361,18 @@ lazy val scaldingSpark = module("spark").settings(
...
+    "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-google-cloud-platform-core" % beamVersion,
+    "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
```
ttim: We should fix this before landing

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+import com.twitter.scalding.Execution.Writer
+import com.twitter.scalding.typed.{Resolver, TypedSink, TypedSource}
+import com.twitter.scalding.{Config, Mode, TextLine}
+import org.apache.beam.sdk.Pipeline
+import org.apache.beam.sdk.io.TextIO
+import org.apache.beam.sdk.values.PCollection
+import scala.concurrent.ExecutionContext
+
+case class BeamMode(
+  pipeline: Pipeline,
```
ttim: This can't be correct, because this means you keep running same `pipeline` again and again with every `execute()` call

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+      override def read(
+        pipeline: Pipeline,
+        config: Config)(implicit ec: ExecutionContext): PCollection[_ <: String] = {
+        pipeline.apply(TextIO.read().from(path))
+      }
+    }
+
+  val Default: Resolver[TypedSource, BeamSource] = {
+    new Resolver[TypedSource, BeamSource] {
+      def apply[A](i: TypedSource[A]): Option[BeamSource[A]] = {
```
ttim: `i` name seems not very descriptive

ttim: This applies to other places as well

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,47 @@
...
+              case _ => throw new Exception(s"unknown sink: $sink when writing $pipe")
+            }
+            rec(l.tail)
+          }
+          case _ => rec(l.tail)
+        }
+      }
+    }
+    rec(optimizedWrites)
+    pipeline.run().waitUntilFinish()
```
ttim: You don't want to block `execute`, instead you want to create `Promise`, and resolve it when corresponding pipeline finishes

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
ttim: +1, we want to keep same kryo instance but somehow made it not to reuse tokens / classIds

ttim (commented):
Initial comments, out of big ones: usage of `Pipeline` in mode, and blocking in `execute`

scalding-beam/src/test/scala/com.twitter.scalding.beam_backend/BeamBackendTests.scala
Franco1010: We are just implementing SimpleWrite for the moment, I will add assertions in next steps when getForced and getIterable are implemented

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,48 @@
...
+                case _ => throw new Exception(s"unknown sink: $sink when writing $pipe")
+              }
+              rec(xs)
+            }
+            case _ => rec(xs)
+          }
+      }
+    }
+    rec(optimizedWrites)
+    CFuture.uncancellable(Future(pipeline.run()).map(_ => (0L, ExecutionCounters.empty)))
```
nownikhil: I wonder if we still need to do `.waitUntilFinish`. This will be needed when we support chaining of executions.

Same for returning 0 as the unique id. We would need to keep some state which can increment this counter. For now we can keep it like this and address this in the follow up patch.

build.sbt
```diff
@@ -359,6 +361,18 @@ lazy val scaldingSpark = module("spark").settings(
...
+lazy val scaldingBeam = module("beam").settings(
+  libraryDependencies ++= Seq(
+    "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-google-cloud-platform-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-sorter" % beamVersion,
+    "org.apache.beam" % "beam-runners-direct-java" % beamVersion % "test",
+    "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
+  )
+).dependsOn(scaldingCore)
+
```
ttim, line 374: nit: extra new lines, should be only one

build.sbt
```diff
@@ -359,6 +361,18 @@ lazy val scaldingSpark = module("spark").settings(
     )
   ).dependsOn(scaldingCore)
 
+lazy val scaldingBeam = module("beam").settings(
+  libraryDependencies ++= Seq(
+    "org.apache.beam" % "beam-sdks-java-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-google-cloud-platform-core" % beamVersion,
+    "org.apache.beam" % "beam-sdks-java-extensions-sorter" % beamVersion,
+    "org.apache.beam" % "beam-runners-direct-java" % beamVersion % "test",
+    "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
```
ttim, line 371: Add comment indicating we need that because scalding config depends on hadoop

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+        pipeline.apply(TextIO.read().from(path))
+      }
+    }
+
+  val Default: Resolver[TypedSource, BeamSource] = {
+    new Resolver[TypedSource, BeamSource] {
+      def apply[A](source: TypedSource[A]): Option[BeamSource[A]] = {
+        source match {
+          case tl: TextLine =>
+            Some(textLine(tl.localPaths.head))
```
ttim: `.head` doesn't seem correct here, would be great to fix it or add `todo`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+
+object BeamMode {
+  def empty(pipelineOptions: PipelineOptions): BeamMode =
+    BeamMode(pipelineOptions, Resolver.empty, Resolver.empty)
+  def default(pipelineOptions: PipelineOptions): BeamMode =
+    BeamMode(pipelineOptions, BeamSource.Default, BeamSink.Default)
+}
+
+trait BeamSource[+A] extends Serializable {
+  def read(pipeline: Pipeline, config: Config)(implicit ec: ExecutionContext): PCollection[_ <: A]
```
ttim: I don't think you really need `ExecutionContext` here

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+            Some(textLine(tl.localPaths.head))
+          case _ => None
+        }
+      }
+    }
+  }
+}
+
+trait BeamSink[-A] extends Serializable {
+  def write(pipeline: Pipeline, config: Config, pc: PCollection[_ <: A])(implicit ec: ExecutionContext): Unit
```
ttim: Same as `Source` - no need for execution context

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+}
+
+object BeamSink extends Serializable {
+  def textLine(path: String): BeamSink[String] =
+    new BeamSink[String] {
+      override def write(
+        pipeline: Pipeline,
+        config: Config,
+        pc: PCollection[_ <: String])(implicit ec: ExecutionContext): Unit = {
+        pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
```
ttim: It doesn't compile without `asInstanceOf`?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.beam.sdk.Pipeline
+import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+import org.apache.beam.sdk.transforms._
+import org.apache.beam.sdk.values.PCollection
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+sealed abstract class BeamOp[+A] {
+  import BeamOp.TransformBeamOp
+  def run(pipeline: Pipeline)(implicit ec: ExecutionContext): PCollection[_ <: A]
```
ttim: `ExecutionContext` doesn't seem necessary

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def map[C >: A, B](f: C => B): BeamOp[B] = {
```
ttim: nit: No need to have `{}` if it's one expression
nit: `this.` isn't needed

ttim: same applies to other operations

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+    override def run(pipeline: Pipeline)
+      (implicit ec: ExecutionContext): PCollection[_ <: A] = {
+      pipeline.apply(Create.of(iterable.asJava).withCoder(new KryoCoder[A]))
+    }
+  }
+
+  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]]) extends BeamOp[B] {
+    def run(pipeline: Pipeline)(implicit ec: ExecutionContext): PCollection[B] = {
+      source.run(pipeline).asInstanceOf[PCollection[A]].apply(f)
+        .setCoder(new KryoCoder[B])
```
ttim: The fact we use `KryoCoder` here which we might want to have customizable indicates we want to have access to config in `run()` method, would be great to add it

ttim: Alternatively (and maybe better) to add `Coder` as an argument to these classes, and pass `KryoCoder` in planner

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+
+}
+class ProcessFn[A, B](f: A => B) extends ProcessFunction[A, B] {
```
ttim: nit: fix spaces

ttim: Would be great to move this functions into separate file / `BeamFunctions` object

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,50 @@
...
+              }
+              rec(xs)
+            }
+            case _ => rec(xs)
+          }
+      }
+    }
+    rec(optimizedWrites)
+    CFuture.uncancellable(Future{
+      pipeline.run().waitUntilFinish()
```
ttim: Given `cancel` method on `PipelineResult` object I don't think this future should be uncancellable. You might want to add `todo` or fix it.

ttim: Given `cancel` method on `PipelineResult` object I don't think this future should be uncancellable. You might want to add `todo` or fix it.

You don't need to do `map` separately instead you can return it as is as part of your Future block:
```
Future {
  pipeline.run().waitUntilFinish()
  (0L, ExecutionCounters.empty)
}
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,42 @@
...
+    kryo.readClassAndObject(input).asInstanceOf[T]
+  }
+}
+
+object KryoCoder{
+  val KryoState: ThreadLocal[Kryo] = {
+    new ThreadLocal[Kryo] {
+      override def initialValue(): Kryo = {
+        val newKryo = new serialization.KryoHadoop(new ScalaMapConfig(Map.empty)).newKryo
+        newKryo.setAutoReset(true)
```
ttim: Put comments explaining what we need this config

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,42 @@
...
+import com.twitter.chill.Kryo
+import com.twitter.chill.config.ScalaMapConfig
+import com.twitter.scalding.serialization
+import java.io.{ByteArrayOutputStream, InputStream, OutputStream}
+import org.apache.beam.sdk.coders.{AtomicCoder, CoderException => BCoderException}
+
+final class KryoCoder[T] extends AtomicCoder[T] {
+  override def encode(value: T, os: OutputStream): Unit = {
+    val kryo = KryoCoder.KryoState.get()
+    if (value == null) {
```
ttim: I'm not sure that's good default, I think we permit customers to pass nulls around

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,42 @@
...
+      throw new BCoderException("cannot encode a null value")
+    }
+    val bos = new ByteArrayOutputStream()
```
ttim: todo: on caching `bos` and `bosOutput` in thread local

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,42 @@
...
+    val bosOuotput = new com.esotericsoftware.kryo.io.Output(bos)
```
ttim: nit: typo

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -0,0 +1,62 @@
...
+    deleteRecursively(new File(path))
+  }
+
+  private var pipelineOptions: PipelineOptions = _
+
+  before {
+    pipelineOptions = PipelineOptionsFactory.create()
+  }
+
+  after {}
```
ttim: nit: remote

ttim (commented):
Great job and great beginning. There is bunch of comments but I'm fine with either fixing them in place or taking them to next reviews.

@johnynek do we mark experimental API somehow? I'm not sure how BeamSource/Sink apis are gonna look like until we implement them further. But also it might be not necessary until release is out...

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
johnynek: Is there some kind of merged `BeamSource` that can union several together? If so, you could use that I guess.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+
+  val Default: Resolver[TypedSink, BeamSink] = {
+    new Resolver[TypedSink, BeamSink] {
+      def apply[A](sink: TypedSink[A]): Option[BeamSink[A]] = {
+        sink match {
+          case tl: TextLine =>
+            Some(textLine(tl.localPaths.head).asInstanceOf[BeamSink[A]])
```
johnynek: here you probably want to make an error if the localPaths have more than one item in them.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: I don't think you need `C >: A` here since function is contravariant in the first parameter.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.beam.sdk.values.PCollection
+import scala.collection.JavaConverters._
+import scala.concurrent.ExecutionContext
+
+sealed abstract class BeamOp[+A] {
+  import BeamOp.TransformBeamOp
+  def run(pipeline: Pipeline)(implicit ec: ExecutionContext): PCollection[_ <: A]
+
+  def map[C >: A, B](f: C => B): BeamOp[B] = {
+    this.parDo(new MapFn(f))
```
johnynek: I think you could write this as `parDo(new MapFn(f))` without the `this`.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+  def parDo[C >: A, B](f: DoFn[C, B]): BeamOp[B] = {
+    this.applyPTransform(ParDo.of(f).asInstanceOf[PTransform[PCollection[C], PCollection[B]]])
```
johnynek: no need for `this`

can we remove the cast?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+  def applyPTransform[C >: A, B](f: PTransform[PCollection[C], PCollection[B]]): BeamOp[B] = {
+    TransformBeamOp(this, f)
+  }
+  def filter(f: A => Boolean): BeamOp[A] = {
+    val fn = new ProcessFn(f.asInstanceOf[A => java.lang.Boolean])
+    this.applyPTransform(Filter.by[A, ProcessFunction[A, java.lang.Boolean]](fn))
```
johnynek: no need for `this` Do we need the cast and the type annotations here? seems like the cast alone should be good enough.

That said, I think doing `f.andThen(java.lang.Boolean.valueOf(_))` would safer than the cast which is relying on the scala implementation details which could potentially change in some versions (e.g. there is no specialization to boolean).

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+  def flatMap[C >: A, B](f: C => TraversableOnce[B]): BeamOp[B] = {
+    this.parDo(new FlatMapFn(f))
```
johnynek: no need for `this` and I think we don't need `C >: A`.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: also probably better to make a `case class` here just to have some chance of plumbing through the equality and potentially getting a clearer toString if `f` is one of our other internal functions.

How is this different from `MapFn`? Why do we need both? Can you make a comment?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+
+}
+class ProcessFn[A, B](f: A => B) extends ProcessFunction[A, B] {
+  @throws[Exception]
+  override def apply(input: A): B = f(input)
+}
+
+class FlatMapFn[A, B](f: A => TraversableOnce[B]) extends DoFn[A, B] {
```
johnynek: case class...

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,80 @@
...
+  @ProcessElement
+  def processElement(c: DoFn[A, B]#ProcessContext): Unit = {
+    val it = f(c.element()).toIterator
+    while (it.hasNext) c.output(it.next())
+  }
+}
+
+class MapFn[A, B](f: A => B) extends DoFn[A, B] {
```
johnynek: case class.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,50 @@
...
+  def getForced[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[TypedPipe[T]] = ???
+
+  def getIterable[T](conf: Config, initial: TypedPipe[T])(implicit cec: ExecutionContext): Future[Iterable[T]] = ???
+
+  override def execute(
+    conf: Config,
+    writes: List[ToWrite[_]])(implicit cec: ExecutionContext): CFuture[(Long, ExecutionCounters)] = {
+    import Execution.ToWrite._
+    val planner = BeamPlanner.plan(conf, beamMode.sources)
+    val phases: Seq[Rule[TypedPipe]] = OptimizationRules.standardMapReduceRules
```
johnynek: you might want to add a TODO to allow overriding these optimization rules based on the conf. This is done in cascading mode.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,50 @@
...
+              rec(xs)
+            }
+            case _ => rec(xs)
+          }
+      }
+    }
+    rec(optimizedWrites)
+    CFuture.uncancellable(Future{
+      pipeline.run().waitUntilFinish()
+    }.map(_ => (0L, ExecutionCounters.empty)))
```
johnynek: are there no counters in Beam?

you don't want to return `0L` for each item. The contract is to return a new unique Long for each call to this function. You can keep an atomicLong and increment on each call, or something like that.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,42 @@
...
+  override def decode(is: InputStream): T = {
+    val kryo = KryoCoder.KryoState.get()
+    val size = is.read()
+    val input = new com.esotericsoftware.kryo.io.Input(is, size)
+    kryo.readClassAndObject(input).asInstanceOf[T]
+  }
+}
+
+object KryoCoder{
+  val KryoState: ThreadLocal[Kryo] = {
```
johnynek: I am not a fan of this.

1. note that when you are building the `BeamOp` you have access to the `Config`.
2. we have traditionally configured kryo by using the `Config` to get a `KryoHadoop` instance.

What I would prefer is in the BeamPlanner you build the `KryoInstantiator` and then pass that as needed.

Note: the KyroInsantiator is the abstraction recommended for what you are doing here (building new instances) and KryoPool can safely share Kryo instances and buffers: https://github.com/twitter/chill/blob/develop/chill-java/src/main/java/com/twitter/chill/KryoPool.java

Having kryo be unconfigurable isn't great since not that uncommonly we needed to configure it (e.g. set up thrift/proto/avro support).

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
Franco1010: We need ProcessFn just because Beam Filter needs it as a lambda function that extends an `apply` method, we could also use ParDo - DoFn with MapFn but then it will be just duplicated code from Beam Filter

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
Franco1010: No, it won't compile

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def parDo[C >: A, B](f: DoFn[C, B])(implicit kryoCoder: KryoCoder): BeamOp[B] = {
+    val pTransform = new PTransform[PCollection[C], PCollection[B]]() {
+      override def expand(input: PCollection[C]): PCollection[B] = input.apply(ParDo.of(f))
+    }
+    applyPTransform(pTransform)
+  }
+
+  def filter(f: A => Boolean)(implicit kryoCoder: KryoCoder): BeamOp[A] = {
+    val fn = ProcessFn(f.andThen(java.lang.Boolean.valueOf))
```
johnynek: this is the only place ProcessFn is used, Why not make:
```scala
case class ProcessPredicate[A](fn: A => Boolean) extends ProcessFunction[A, java.lang.Boolean] {
  def apply(a: A) = java.lang.Boolean.valueOf(fn(a))
}
```
This reduces unused abstraction and also reduces slightly the indirection at runtime (you avoid andThen)'

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.beam.sdk.transforms._
+import org.apache.beam.sdk.values.PCollection
+import scala.collection.JavaConverters._
+
+sealed abstract class BeamOp[+A] {
+  import BeamOp.TransformBeamOp
+
+  def run(pipeline: Pipeline): PCollection[_ <: A]
+
+  def map[C >: A, B](f: C => B)(implicit kryoCoder: KryoCoder): BeamOp[B] =
```
johnynek: I am still pretty sure: `def map[B](f: A => B)...` works here because `Function1` is contravariant in the first parameter.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def filter(f: A => Boolean)(implicit kryoCoder: KryoCoder): BeamOp[A] = {
+    val fn = ProcessFn(f.andThen(java.lang.Boolean.valueOf))
+    applyPTransform(Filter.by[A, ProcessFunction[A, java.lang.Boolean]](fn))
+  }
+
+  def applyPTransform[C >: A, B](f: PTransform[PCollection[C], PCollection[B]])(implicit kryoCoder: KryoCoder): BeamOp[B] =
+    TransformBeamOp(this, f, kryoCoder)
+
+  def flatMap[C >: A, B](f: C => TraversableOnce[B])(implicit kryoCoder: KryoCoder): BeamOp[B] =
```
johnynek: same comment here: `def flatMap[B](f: A => ...)` should work because `Function1` is contravariant in the first parameter.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,60 @@
...
+    writes: List[ToWrite[_]])(implicit cec: ExecutionContext): CFuture[(Long, ExecutionCounters)] = {
+    import Execution.ToWrite._
+    val planner = BeamPlanner.plan(conf, beamMode.sources)
+    val phases: Seq[Rule[TypedPipe]] = BeamPlanner.defaultOptimizationRules(conf)
+    val optimizedWrites = ToWrite.optimizeWriteBatch(writes, phases)
+    val pipeline = Pipeline.create(beamMode.pipelineOptions)
+
+    def rec(optimizedWrites: List[OptimizedWrite[TypedPipe, _]]): Unit = {
+      optimizedWrites match {
+        case Nil => Unit
```
johnynek: `Unit` is the type and companion object for the type, `()` is the value. This is working because scala is casting the companion object into the value unit.

It is better to write: `case Nil => ()` here.

that is not a shortcut, is actually the correct thing that works without relying on Any to Unit casting.

It's a bit subtle, but the key is the distinction between the type and companion object name `Unit` and the value `()`. Honestly, I think scala messed this up, and I think it would have been better for the unit value to be called `Unit`... but oh well.

that would have been something like this:
```scala
// here the type is defined
sealed trait Unit

// here is a value, note the name space of values and types are different, so they can somewhat confusingly have
// the same name.
object Unit extends Unit
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,60 @@
...
+            case OptimizedWrite(pipe, ToWrite.SimpleWrite(opt, sink)) => {
+              val pcoll = planner(opt).run(pipeline)
+              beamMode.sink(sink) match {
+                case Some(ssink) =>
+                  ssink.write(pipeline, conf, pcoll)
+                case _ => throw new Exception(s"unknown sink: $sink when writing $pipe")
+              }
+              rec(xs)
+            }
+            case _ => rec(xs)
```
johnynek: I think you should at least log or something if you are going to ignore items.

Honestly, I would prefer: `???` so you don't forget to implement it later.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -0,0 +1,60 @@
...
+          }
+      }
+    }
+    rec(optimizedWrites)
+    val result = pipeline.run
```
johnynek, line 53: we should get the runId *before* we start the job, not after:
```scala
val runId = state.getAndIncrement() // do this before returning
Future {
  result.waitUntilFinish()
  (runId, ...)
}
```
honestly the return type from the method should probably be `(Long, Future[ExecutionCounters])` but that it wasn't implemented that way.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,28 @@
+package com.twitter.scalding.beam_backend
+
+import com.esotericsoftware.kryo.io.Input
+import com.twitter.chill.{ KryoInstantiator, KryoPool }
+import java.io.{ InputStream, OutputStream }
+import org.apache.beam.sdk.coders.AtomicCoder
+import scala.language.implicitConversions
+
+final class KryoCoder(kryoInstantiator: KryoInstantiator) extends AtomicCoder[Any] {
+  @transient lazy val kryoPool = KryoPool.withByteArrayOutputStream(Runtime
```
johnynek: maybe make this `private[this]` as well.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,28 @@
...
+    .getRuntime
+    .availableProcessors, kryoInstantiator)
+  override def encode(value: Any, os: OutputStream): Unit = {
+    val bytes = kryoPool.toBytesWithClass(value)
+    os.write(bytes.length)
```
johnynek: this isn't safe... this can only write 1 byte (even though it takes an int). so if you have more than 255 bytes of data it will fail.

See:
https://github.com/twitter/chill/blob/19daefccbcda7d50f3130985de6e016be6eafa01/chill-hadoop/src/main/java/com/twitter/chill/hadoop/KryoDeserializer.java#L49

you might want to use this class:
https://github.com/twitter/scalding/blob/b6d465f6fc308d66e5a53c62662d5860846d8dc4/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/JavaStreamEnrichments.scala#L238

which can write positive ints to streams for you (and read them).

We use this class already for OrderedSerialization work.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,28 @@
...
+    os.write(bytes)
+    os.flush()
+  }
+  override def decode(is: InputStream): Any = {
+    val size = is.read()
```
johnynek: same comment about this not actually being safe.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,28 @@
...
+    val input = new Input(is, size)
+    kryoPool.fromBytes(input.readBytes(size))
+  }
+}
+object KryoCoder {
+  implicit def inferType[T](atomicCoder: AtomicCoder[Any]): AtomicCoder[T] =
```
johnynek: 1. can we not call this `inferType` but perhaps `castType`
2. can we accept a `KryoCoder` as the argument not an `AtomicCoder`? It isn't safe in general, it is safe(ish) for kryo because it can handle any type at runtime.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -0,0 +1,60 @@
...
+      .from(1 to 20)
+      .map(x => (x, x))
+      .flatMapValues(x => 1 to x)
+      .map(_.toString())
+    val ex2 = pipe2.writeExecution(TextLine(outRoute2))
+
+    val bmode = BeamMode.default(pipelineOptions)
+    val ex3 = Execution.zip(ex1, ex2)
+    ex3.waitFor(Config.empty, bmode).get
+
```
johnynek, line 86: Can we open the files using normal Java IO and make some assertion about them?

Like read all the numbers and put them in a set and assert it matches the expected Set?

I'd like a bit more assurance than it just didn't crash.

johnynek (changes requested):
One comment but then I think it is ready to merge as a first step.

Good work.

johnynek (approved):
This is exciting!

Would love to see you post some benchmarks and make a blog post once you can do groupBy and cogroup.

johnynek:
Looks like the microsite test failed:
```
error: File /home/travis/build/twitter/scalding/docs/docs does not exist.

Exception: sbt.TrapExitSecurityException thrown from the UncaughtExceptionHandler in thread "run-main-0"

[error] Nonzero exit code: 1

[error] (docs / Compile / runMain) Nonzero exit code: 1

[error] Total time: 168 s (02:48), completed Aug 18, 2021 2:58:18 AM

The command "./sbt ++$TRAVIS_SCALA_VERSION clean docs/makeMicrosite" exited with 1.
```

ttim:
@johnynek it's independent failure and was failing since I upgraded sbt (and fails on `develop` too =( ).
I'll tackle it this week.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -0,0 +1,59 @@
...
+      }
+  }
+
+  final case class FromIterable[A](iterable: Iterable[A], kryoCoder: KryoCoder) extends BeamOp[A] {
+    override def run(pipeline: Pipeline): PCollection[_ <: A] = {
+      pipeline.apply(Create.of(iterable.asJava).withCoder(kryoCoder))
+    }
+  }
+
+  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder)() extends BeamOp[B] {
```
ttim, line 54: nit: no need to accept concrete `KryoCoder` here, accepting `Coder[B]` would be better

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -0,0 +1,87 @@
...
+          BeamOp.Source(config, src, srcs(src))
+        case (IterablePipe(iterable), _) =>
+          BeamOp.FromIterable(iterable, kryoCoder)
+        case (wd: WithDescriptionTypedPipe[a], rec) =>
+          rec[a](wd.input)
+      }
+    })
+  }
+
+  def defaultKryoCoderConfiguration(config: Config): KryoInstantiator = {
```
ttim, line 59: nit: make it private

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -0,0 +1,87 @@
...
+    config.getKryo match {
+      case Some(kryoInstantiator) => kryoInstantiator
+      case None => new KryoHadoop(new ScalaMapConfig(Map.empty))
+    }
+  }
+
+  def defaultOptimizationRules(config: Config): Seq[Rule[TypedPipe]] = {
```
ttim, line 66: nit: private

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,29 @@
+package com.twitter.scalding.beam_backend
+
+import com.esotericsoftware.kryo.io.Input
+import com.twitter.chill.{ KryoInstantiator, KryoPool }
+import com.twitter.scalding.serialization.JavaStreamEnrichments.{ RichInputStream, RichOutputStream }
+import java.io.{ InputStream, OutputStream }
+import org.apache.beam.sdk.coders.AtomicCoder
+import scala.language.implicitConversions
+
+final class KryoCoder(kryoInstantiator: KryoInstantiator) extends AtomicCoder[Any] {
```
ttim, line 10: Let's make it `private[beam_backend]` for now? Applies to other classes around too

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -0,0 +1,29 @@
...
+    os.flush()
+  }
+  override def decode(is: InputStream): Any = {
+    val size = is.readPosVarInt
+    val input = new Input(is, size)
+    kryoPool.fromBytes(input.readBytes(size))
+  }
+}
+object KryoCoder {
+  implicit def castType[T](kryoCoder: KryoCoder): AtomicCoder[T] =
```
ttim, line 27: I would prefer don't have `castType` to be implicit, but it's up to you

ttim (approved):
Bunch of nits, feel free to fix them where appropriate in next review! Thank you!

PR #1942, opened 2021-08-25
Adding Beam-Dataflow support for group functions in Scalding 

Franco1010:
This PR adds support for scalding group functions by handling:

- Grouped
  - IdentityReduce
- UnsortedGrouped
  - UnsortedIdentityReduce
  - IteratorMappedReduce
- SortedGrouped
  - IdentityValueSortedReduce
  - ValueSortedReduce

via

- mapGroup
- sortedMapGroup
- sorted

methods in BeamOp.

- Optimized sum functions in mapGroup through planMapGroup

- Functions using priorityQueue are not supported in this PR, because they mutate elements passed as input, but Beam does not allow to mutate input between transformations


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -52,6 +52,36 @@ object BeamPlanner {
           BeamOp.FromIterable(iterable, kryoCoder)
         case (wd: WithDescriptionTypedPipe[a], rec) =>
           rec[a](wd.input)
+        case (SumByLocalKeys(pipe, _), rec) =>
+          rec(pipe)
```
johnynek: this is shame for this to be a no-op...

Can you not do some kind of stateful map only operation that processes all the keys on the mapper and uses a cache of a fixed size to aggregate?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +55,101 @@ object BeamOp extends Serializable {
     }
   }
 
-  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder)() extends BeamOp[B] {
+  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder) extends BeamOp[B] {
     def run(pipeline: Pipeline): PCollection[B] = {
       source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
     }
   }
+
```
johnynek, line 93: none of this code is handling `OrderedSerialization` which I think will be a significant deoptimization relative to scalding.

You might want to pattern match on Ordering and if it is an OrderedSerialization, then serialize to bytes and use the binary comparator.

If nothing else, you can use the OrderedSerialization to get a non-kryo based Coder which will be faster and use less data.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +55,101 @@ object BeamOp extends Serializable {
...
+        kryoCoder)
+    }
+  }
+
+  /**
+   * @note we are just ignoring value ordering and using just default ordering provided by
+   *       [[BufferedExternalSorter]], so we can't take custom ordering with it
+   * @see [[org.apache.beam.sdk.extensions.sorter.Sorter]]
+   * @see [[https://beam.apache.org/documentation/sdks/java-extensions/#sorter]]
+   * @todo accept ordering of values, maybe sort them in memory, but it could cause out-of-memory exceptions
```
johnynek: how does this work if ordV is being ignored? in what way is it a sort?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -28,60 +35,102 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+        .groupAll
+        .sorted
+        .limit(3),
+      Seq(((),0), ((),1), ((),2))
+    )
+  }
+
+  test("test"){
+    /**
+     * @note we are not extending support for `sortedTake` and `sortedReverseTake`, since both of them uses
```
johnynek, line 108: these could be changed to use an immutable heap which would work on hadoop as well as beam.

I am curious how beam can wind up erroring here though, how does it know an item has been modified?

Franco1010: It seems that they do coder before and after transformation and compare bytes

johnynek: How is that not insanely slow?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +55,101 @@ object BeamOp extends Serializable {
...
 
-  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder)() extends BeamOp[B] {
+  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder) extends BeamOp[B] {
     def run(pipeline: Pipeline): PCollection[B] = {
       source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
     }
   }
+
+  implicit class KVOp[K, V](val op: BeamOp[(K, V)]) extends AnyVal {
+    def mapGroup[U](reduceFn: (K, Iterator[V]) => Iterator[U])(implicit ordK: Ordering[K], kryoCoder: KryoCoder): BeamOp[(K, U)] = {
```
johnynek, line 95: so, here you can also match on the reduceFn:
```scala
reduceFn match {
  case MapValueStream(SumAll(semigroup)) =>
    // you can use beam combine here
}
```
there may be other functions you can optimize if you look in here:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/functions/Functions.scala

Note, they can be combined, so you may want something like:
```scala
def planMapGroup[K, V, U](op: PCollection[KV[K, JIterable[V]], fn: (K, Iterator[V]) => Iterator[U]): PCollection[KV[K, JIterable[U])] =
  fn match {
    case ComposedMapGroup(left, right) =>
      planMapGroup(planMapGroup(op left), right)
    case MapValueStream(SumAll(sg)) =>
      // use Combine.GroupedValues here
    case notComposedOrSum =>
      // we just apply it directly
``` 

I don't know if beam really has any core operators that it can be faster with, or if it is just an API that is doing what scalding is already doing. It would be nice to know what the core operators are that are optimized vs what are just built on top of those as conveniences.

I was looking at:

https://beam.apache.org/documentation/transforms/java/aggregation/combine/

and https://beam.apache.org/releases/javadoc/2.0.0/org/apache/beam/sdk/transforms/Combine.GroupedValues.html

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: to be concrete:

https://github.com/twitter/scalding/blob/132f9bd3767feab76f7284f73fc9deb0cec9116e/scalding-serialization/src/main/scala/com/twitter/scalding/serialization/OrderedSerialization.scala#L29

you want to implement:
```scala
def keyCoder[K](ord: Ordering[K], kryoCoder: KryoCoder): AtomicCoder[K] =
  ord match {
    case ordSer: OrderedSerialization[K] @unchecked =>
      // use the Serialization[K] to make a coder
      ???
    case _ =>
      // fall back to KryoCoder here
     kryoCoder
}
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +55,101 @@ object BeamOp extends Serializable {
...
+              .apply(GroupByKey.create[K, V]()).setCoder(KvCoder.of(kryoCoder, kryoCoder))
+              .apply(ParDo.of(FlatMapFn[KV[K, java.lang.Iterable[V]], KV[K, U]] { elem =>
+                reduceFn(elem.getKey, elem.getValue.asScala.toIterator).map(KV.of(elem.getKey, _))
+              })).setCoder(KvCoder.of(kryoCoder, kryoCoder))
+              .apply(new KVToTuple[K, U])
+          }
+        },
+        kryoCoder)
+    }
+    def sortedMapGroup[U](reduceFn: (K, Iterator[V]) => Iterator[U])(implicit ordK: Ordering[K], ordV: Ordering[V], kryoCoder: KryoCoder): BeamOp[(K, U)] = {
```
johnynek, line 114: we can pattern match on the `Ordering[V]` to see if it is a `OrderedSerialization` to recover a way to serialize.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: If beam requires the `V` to be comparable, here is a solution:

```scala
case class WrappedComparable[A](value: A, ord: Ordering[A]) extends Comparable[WrappedComparable[A]] {
  def compareTo(that: WrappedComparable[A]) =
    ord.compare(value, that.value)
}

class WrappedAtomicCoder[A](ord: Ordering[A], prev: AtomicCoder[A]) extends AtomicCoder[WrappedComparable[A]] {
  def serialize(....)
    // unwrap the case class, then call prev
  def deserialize(...
    // deserialize with prev then wrap in a case class since you have ord.
}
```

This should avoid serializing the V with an ordering, but will respect the ordering passed by the user.

So, before you send to the SortValues you will map all the values into this Wrapped form with the given AtomicCoder.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +55,101 @@ object BeamOp extends Serializable {
...
+          KV.of(elem.getKey, elem.getValue.asScala.map(KV.of(_, ())).asJava)
+        })).setCoder(KvCoder.of(kryoCoder, IterableCoder.of(KvCoder.of(kryoCoder, kryoCoder))))
+        .apply(SortValues.create[K, V, Unit](BufferedExternalSorter.options()))
+        .apply(ParDo.of(MapFn[KV[K, java.lang.Iterable[KV[V, Unit]]], KV[K, java.lang.Iterable[V]]]{ elem =>
+          KV.of(elem.getKey, elem.getValue.asScala.map(_.getKey).asJava)
+        })).setCoder(KvCoder.of(kryoCoder, IterableCoder.of(kryoCoder)))
+    }
+  }
+
+  case class TupleToKV[K, V](implicit kryoCoder: KryoCoder) extends PTransform[PCollection[(K, V)], PCollection[KV[K, V]]] {
```
johnynek: this should probably take two coders, one for K and one for V since you can construct some of them using `Ordering[K]` or `Ordering[V]`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
johnynek: looking at this:
https://beam.apache.org/releases/javadoc/2.0.0/index.html?org/apache/beam/sdk/transforms/ParDo.html

it seems we could use `@StartBundle` and `@FinishBundle` to set up a local cache which we use to do map-side aggregation, eg.
https://github.com/twitter/algebird/blob/de336f3818a6fb34f0c27302873772c6e48cd9fc/algebird-core/src/main/scala/com/twitter/algebird/SummingCache.scala#L34
```scala
  // in setup make:
config.getMapSideAggregationThreshold match 
  case None => rec(pipe)
  case Some(count) =>
    def makeSummingCache = new SummingCache(count)(semigroup)
    // now, in each StartBundle make a new SummingCache, in each FinishBundle, flush the cache and emit all values.
```

nownikhil: @johnynek Should it be done in a way that we create a map from bundle(atomic int) -> SummingCache? We need new SummingCache for every bundle and if we have a field in the class it can be used by multiple threads.

johnynek: yes, I think so. This was just a sketch. I would use a thread-local perhaps and put a new cache after the call to StartBundle, and then in FinishBundle you evict everything.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +25,31 @@ object BeamFunctions {
...
+    def startBundle(): Unit = {
+      cache = new SummingCache[K, V](size)(semigroup)
+    }
+    @ProcessElement
+    def processElement(c: DoFn[KV[K, V], KV[K, V]]#ProcessContext): Unit = {
+      val key = c.element().getKey
+      val value = c.element().getValue
+      val evicted = cache.put(Map((key, value)))
+      evicted match {
+        case Some(m) => m.foreach(x => c.output(KV.of(x._1, x._2)))
```
johnynek: nit: using a while loop and `m.iterator` is faster (no closure):
```scala
val mit = m.iterator
while (mit.hasNext) {
  val x = mit.next()
  c.output(KV.of(x._1, x._2))
}
```
A bit of a pain to write things like that out, but avoiding closures can be measurable wins in some of the internals.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +25,31 @@ object BeamFunctions {
...
+        case None => ()
+      }
+    }
+    @FinishBundle
+    def finishBundle(c: DoFn[KV[K, V], KV[K, V]]#FinishBundleContext): Unit = {
+      val evicted = cache.flush
+      evicted match {
+        case Some(m) =>
+          m.foreach(x => c.output(KV.of(x._1, x._2), BoundedWindow.TIMESTAMP_MIN_VALUE, GlobalWindow.INSTANCE))
```
johnynek: same comment about avoiding the closure.

Franco1010: Outside processElement Beam has no notion of window or timestamp to output elements, but also Beam does not set windows if a windowFn is not specified, so everything should remain in GlobalWindow, also Beam set all elements to TimestampMinValue if a timestamp is not provided manually to each element, so we are also setting everting to min value here

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -27,3 +29,21 @@ object KryoCoder {
   implicit def castType[T](kryoCoder: KryoCoder): AtomicCoder[T] =
     kryoCoder.asInstanceOf[AtomicCoder[T]]
 }
+
+case class OrderedSerializationCoder[T](ord: Ordering[T])(implicit kryoCoder: KryoCoder) extends AtomicCoder[T] {
+  override def encode(value: T, outStream: OutputStream): Unit = {
+    ord match {
+      case ordSer: OrderedSerialization[T] @unchecked =>
```
johnynek: this is going to pattern match on each value. I don't think we want this. I think we should do:

```scala
object OrderedSerializationCoder {
  def apply[T](ord: Ordering[T])(implicit kryoCoder: KryoCoder): AtomicCoder[T] =
   ord match {
     case ordSer: OrderedSerialization[T] @unchecked => // make an orderedserializationcoder that ALWAYS uses ordSer
    case _ => kryoCoder
  }
}
```

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -132,11 +122,25 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+    )
+  }
+
   private def getContents(path: String, prefix: String): List[String] = {
     new File(path).listFiles.flatMap(file => {
-      if (file.getPath.startsWith(prefix)) {
+      if(file.getPath.startsWith(prefix)){
         Source.fromFile(file).getLines().flatMap(line => line.split("\\s+").toList)
-      } else List.empty[String]
+      }else List.empty[String]
```
johnynek: why the unusual spacing?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -101,45 +128,52 @@ object BeamOp extends Serializable {
...
+        kryoCoder)
+    }
+
+    def mapSideAggregator(size: Int, semigroup: Semigroup[V])(implicit kryoCoder: KryoCoder): BeamOp[(K, V)] = {
+      TransformBeamOp[(K, V), (K, V)](
+        op,
+        new PTransform[PCollection[(K, V)], PCollection[(K, V)]]() {
+          override def expand(input: PCollection[(K, V)]): PCollection[(K, V)] = {
+            input
+              .apply(new TupleToKV[K, V](kryoCoder, kryoCoder))
```
johnynek: no need for `new` here.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -28,16 +39,28 @@ sealed abstract class BeamOp[+A] {
...
+}
 
-  def parDo[C >: A, B](f: DoFn[C, B])(implicit kryoCoder: KryoCoder): BeamOp[B] = {
-    val pTransform = new PTransform[PCollection[C], PCollection[B]]() {
-      override def expand(input: PCollection[C]): PCollection[B] = input.apply(ParDo.of(f))
+object BeamOp extends Serializable {
+  def planMapGroup[K, V, U](
+    pcoll: PCollection[KV[K, java.lang.Iterable[V]]],
+    reduceFn: (K, Iterator[V]) => Iterator[U])(implicit ordK: Ordering[K], kryoCoder: KryoCoder): PCollection[KV[K, java.lang.Iterable[U]]] = {
+    reduceFn match {
```
johnynek, line 49: by the way, my suggestion is that maybe Beam can do better optimization if we follow this pattern, but that could be incorrect... so it could be always doing what we are doing in `notComposedOrSum` is the optimal thing most of the time.

For instance by using `Combine.groupedValues` we are missing the chance to use `sa.sg.sumOption` which can be a huge win for some cases (since it can use mutability inside that method to speed up aggregating large groups)...

I don't know very much about Beam so I don't know the answer... this should probably be something that benchmarks answer the question of (presumably with cases where there are at least some examples where there are MANY values for each key).

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -151,5 +185,18 @@ object BeamOp extends Serializable {
         })).setCoder(kryoCoder)
     }
   }
+
+  case class WrappedComparable[A](value: A, ord: Ordering[A]) extends Comparable[WrappedComparable[A]] {
+    override def compareTo(o: WrappedComparable[A]): Int = ord.compare(value, o.value)
```
johnynek: we should probably override:
```scala
def hashCode = value.hashCode
def equals(that: Any) =
  that match {
    case WrappedComparable(thatValue, _) => value.equals(thatValue)
    case _ => false
  }
```

johnynek (commented):
some minor comments.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
nownikhil: We plan to do some performance testing eventually. For instance trying to use `Combine.perKey` inside sumByLocalKeys. This adds a shuffle step for sure so in theory it shouldn't perform better. I am not sure if Dataflow runner can anyhow optimize the next shuffle.

Interface of CombineFn is similar to that of an accumulator so I would guess that `Combine.groupedValues` is doing some map side aggregation but we would need to verify it.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
Franco1010: I use two different things for formatting the code, the first one is the reformat tool from IntelliJ, but that doesn't really work to format the code (sometimes add spaces), this one just optimize the imports, so before update the PR's I'm running scalariformFormat command (this one doesn't optimize imports) in sbt, and this one always format the code but always in a different way, reordering functions or removing/adding spaces etc... That's the reason for all weird unnecessary changes between updates

johnynek: we should use scalafmt.... but that's another issue.

johnynek (approved):

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +25,40 @@ object BeamFunctions {
...
+          }
+        case None => ()
+      }
+    }
+    @FinishBundle
+    def finishBundle(c: DoFn[KV[K, V], KV[K, V]]#FinishBundleContext): Unit = {
+      val evicted = cache.flush
+      evicted match {
+        case Some(m) =>
+          val mit = m.iterator
```
ttim, line 53: Do we need this iterator thing here? Why not
```
m.foreach { case (k, v) => 
  c.output(KV.of(k, v), BoundedWindow.TIMESTAMP_MIN_VALUE, GlobalWindow.INSTANCE)
}
```
?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +25,40 @@ object BeamFunctions {
...
+          while (mit.hasNext) {
+            val x = mit.next()
+            c.output(KV.of(x._1, x._2))
+          }
+        case None => ()
+      }
+    }
+    @FinishBundle
```
ttim, line 48: nit: empty lines between methods

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -34,6 +42,25 @@ sealed abstract class BeamOp[+A] {
 }
 
 object BeamOp extends Serializable {
+  def planMapGroup[K, V, U](
+    pcoll: PCollection[KV[K, java.lang.Iterable[V]]],
+    reduceFn: (K, Iterator[V]) => Iterator[U])(implicit ordK: Ordering[K], kryoCoder: KryoCoder): PCollection[KV[K, java.lang.Iterable[U]]] = {
```
ttim, line 48: I think it's better to keep `coder` as an argument around, and don't make it concrete to `KryoCoder` anywhere

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +25,40 @@ object BeamFunctions {
     def processElement(c: DoFn[A, B]#ProcessContext): Unit =
       c.output(f(c.element()))
   }
+
+  case class MapSideAggregator[K, V](size: Int, semigroup: Semigroup[V]) extends DoFn[KV[K, V], KV[K, V]] {
```
ttim: Do we really want to keep it over `KV`, and not just `()`?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +78,131 @@ object BeamOp extends Serializable {
...
+            planMapGroup[K, V, U](groupedSortedValues, reduceFn)
+              .apply(ParDo.of(FlatMapFn[KV[K, java.lang.Iterable[U]], KV[K, U]]{ elem =>
+                elem.getValue.asScala.map(KV.of(elem.getKey, _))
+              })).setCoder(KvCoder.of(OrderedSerializationCoder(ordK), kryoCoder))
+              .apply(KVToTuple[K, U])
+          }
+        },
+        kryoCoder)
+    }
+    def sorted(implicit ordK: Ordering[K], ordV: Ordering[V], kryoCoder: KryoCoder): BeamOp[(K, V)] = {
```
ttim, line 134: nit: empty lines between methods

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -27,3 +29,15 @@ object KryoCoder {
   implicit def castType[T](kryoCoder: KryoCoder): AtomicCoder[T] =
     kryoCoder.asInstanceOf[AtomicCoder[T]]
 }
+
+case class OrderedSerializationCoder[T](ordSer: OrderedSerialization[T])(implicit kryoCoder: KryoCoder) extends AtomicCoder[T] {
```
ttim: Why `kryoCoder` is being passed here?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -27,3 +29,15 @@ object KryoCoder {
...
+  override def encode(value: T, outStream: OutputStream): Unit = ordSer.write(outStream, value)
+  override def decode(inStream: InputStream): T = ordSer.read(inStream).get
+}
+object OrderedSerializationCoder {
+  def apply[T](ord: Ordering[T])(implicit kryoCoder: KryoCoder): AtomicCoder[T] =
```
ttim: nit: not a `KryoCoder`, and better to call it fallback. Also no need to have it as implicit parameter here

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +78,131 @@ object BeamOp extends Serializable {
...
+              .apply(TupleToKV[K, V](kryoCoder, kryoCoder))
+              .apply(ParDo.of(MapSideAggregator[K, V](size, semigroup)))
+              .apply(KVToTuple[K, V])
+          }
+        },
+        kryoCoder)
+    }
+  }
+
+  case class SortGroupedValues[K, V](implicit ordV: Ordering[V], kryoCoder: KryoCoder) extends PTransform[PCollection[KV[K, java.lang.Iterable[V]]], PCollection[KV[K, java.lang.Iterable[V]]]] {
```
ttim: Please don't supply `KryoCoder` here, `Coder[V]` is better

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +78,131 @@ object BeamOp extends Serializable {
...
+  case class KVToTuple[K, V](implicit kryoCoder: KryoCoder) extends PTransform[PCollection[KV[K, V]], PCollection[(K, V)]] {
+    override def expand(input: PCollection[KV[K, V]]): PCollection[(K, V)] = {
+      input
+        .apply(MapElements.via[KV[K, V], (K, V)](new SimpleFunction[KV[K, V], (K, V)]() {
+          override def apply(input: KV[K, V]): (K, V) = (input.getKey, input.getValue)
+        })).setCoder(kryoCoder)
+    }
+  }
+
+  case class WrappedComparable[A](value: A, ord: Ordering[A]) extends Comparable[WrappedComparable[A]] {
```
ttim: Why do we need this class?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +78,131 @@ object BeamOp extends Serializable {
...
+    }
+  }
+
+  case class SortGroupedValues[K, V](implicit ordV: Ordering[V], kryoCoder: KryoCoder) extends PTransform[PCollection[KV[K, java.lang.Iterable[V]]], PCollection[KV[K, java.lang.Iterable[V]]]] {
+    override def expand(input: PCollection[KV[K, lang.Iterable[V]]]): PCollection[KV[K, lang.Iterable[V]]] = {
+      input
+        .apply(ParDo.of(MapFn[KV[K, java.lang.Iterable[V]], KV[K, java.lang.Iterable[KV[WrappedComparable[V], Unit]]]]{ elem =>
+          KV.of(elem.getKey, elem.getValue.asScala.map(x => KV.of(WrappedComparable(x, ordV), ())).asJava)
+        })).setCoder(KvCoder.of(kryoCoder, IterableCoder.of(KvCoder.of(WrappedAtomicCoder[V](ordV, OrderedSerializationCoder(ordV)), kryoCoder))))
+        .apply(SortValues.create[K, WrappedComparable[V], Unit](BufferedExternalSorter.options()))
```
ttim: I don't think `SortValues` cares about comparability of values, so unfortunately we need to implement it ourselves...

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +78,131 @@ object BeamOp extends Serializable {
...
+      obj match {
+        case WrappedComparable(objValue, _) => value.equals(objValue)
+        case _ => false
+      }
+  }
+
+  case class WrappedAtomicCoder[A](ord: Ordering[A], prev: AtomicCoder[A]) extends AtomicCoder[WrappedComparable[A]] {
+    override def encode(
+      value: WrappedComparable[A],
+      outStream: OutputStream): Unit = prev.encode(value.value, outStream)
```
ttim: Add `assert` on `ord` being same as wrapped ordering

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +78,131 @@ object BeamOp extends Serializable {
...
+    override def compareTo(o: WrappedComparable[A]): Int = ord.compare(value, o.value)
+    override def hashCode(): Int = value.hashCode()
+    override def equals(obj: Any): Boolean =
+      obj match {
+        case WrappedComparable(objValue, _) => value.equals(objValue)
+        case _ => false
+      }
+  }
+
+  case class WrappedAtomicCoder[A](ord: Ordering[A], prev: AtomicCoder[A]) extends AtomicCoder[WrappedComparable[A]] {
```
ttim: Nice idea, love it!

ttim (commented):
Good job! I have doubts about `SortValues` being a solution for our case, would love to see links which show `SortValues` uses `Comparable` between values to sort them

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
johnynek: I agree that if beam doesn't require it, you are probably better staying with `Tuple2` and making a concrete `Tuple2Coder` so you don't have to allocate wasted tuple or KV instances.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: I suggested it because my reading of SortValues was that is assumes items are `Comparable` and without using it we are discarding the `Ordering[A]` which might happen to be right for some cases, but it will fail with `reverse` and other use cases.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: how did the test pass then? He has a test that sorts ints.

johnynek: ahh...

https://beam.apache.org/releases/javadoc/2.5.0/org/apache/beam/sdk/extensions/sorter/SortValues.html
```
This transform sorts by lexicographic comparison of the byte representations of the secondary keys and may write secondary key-value pairs to disk. In order to retrieve the byte representations it requires the input PCollection to use a KvCoder for its input, an IterableCoder for its input values and a KvCoder for its secondary key-value pairs.
```

yeah, so it sorts by the binary representation, which for integers is the same as the value sorting... which is how it passed.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: I think we can remove the whole class since we need a different strategy to do secondary sorts anyway...

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
Franco1010: https://github.com/twitter/scalding/pull/1942#discussion_r698812349

ttim: Got it

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -51,9 +82,106 @@ object BeamOp extends Serializable {
...
+        },
+        kryoCoder)
+    }
+  }
+
+  case class SortGroupedValues[K, V](implicit ordK: Ordering[K], ordV: Ordering[V], kryoCoder: KryoCoder) extends PTransform[PCollection[KV[K, java.lang.Iterable[V]]], PCollection[KV[K, java.lang.Iterable[V]]]] {
+    override def expand(input: PCollection[KV[K, lang.Iterable[V]]]): PCollection[KV[K, lang.Iterable[V]]] = {
+      input
+        .apply(ParDo.of(MapFn[KV[K, java.lang.Iterable[V]], KV[K, java.lang.Iterable[V]]]{ elem =>
+          KV.of(elem.getKey, elem.getValue.asScala.toSeq.sorted.toIterable.asJava)
```
ttim: Making it `toArray` would be more performant I think, using raw java here would be even better. But I'm fine to keep any way, since I hope we'll rewrite this eventually.

Please put `TODO` here documenting this needs to be changed to external sorter like the one in beam

ttim (approved):

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +24,42 @@ object BeamFunctions {
...
+    @StartBundle
+    def startBundle(): Unit = {
+      cache = new SummingCache[K, V](size)(semigroup)
+    }
+
+    @ProcessElement
+    def processElement(c: DoFn[(K, V), (K, V)]#ProcessContext): Unit = {
+      val key = c.element()._1
+      val value = c.element()._2
+      val evicted = cache.put(Map((key, value)))
```
johnynek: I just noticed you are breaking the pair just to allocate again:
```scala
val pair = c.element()
val evicted = cache.put(Map(pair))
...
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +24,42 @@ object BeamFunctions {
...
+      evicted match {
+        case Some(m) =>
+          val mit = m.iterator
+          while (mit.hasNext) {
+            val x = mit.next()
+            c.output((x._1, x._2))
```
johnynek: also here:
```scala
c.output(mit.next())
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -22,4 +24,42 @@ object BeamFunctions {
...
+
+    @FinishBundle
+    def finishBundle(c: DoFn[(K, V), (K, V)]#FinishBundleContext): Unit = {
+      val evicted = cache.flush
+      evicted match {
+        case Some(m) =>
+          val mit = m.iterator
+          while (mit.hasNext) {
+            val x = mit.next()
+            c.output((x._1, x._2), BoundedWindow.TIMESTAMP_MIN_VALUE, GlobalWindow.INSTANCE)
```
johnynek: `c.output(mit.next(), ...)`

ttim (approved):

PR #1943, opened 2021-08-30
Migrate scalding docs to mdoc 

ttim:
Tut was deprecated so I migrated scalding to mdoc. As part of this effort I discovered Jekyll being broken as part of our doc generation because of `-f` parameter, and I fixed that.


johnynek:
LGTM but one of the jobs timed.

Presumably unrelated.

PR #1945, opened 2021-09-15
Add support for cogroups in beam-backend

nownikhil:
In this change we are adding support for `HashCoGroup` and `CoGroupedPipe`.
For evaluating HashCoGroup we are creating a ParDo transformation on the larger pipe with smaller pipe as side input.
For evaluating CoGroupedPipe we are using the `MultiJoinFunction` to evaluate the final output iterator.

Also as part of this change we are doing a minor refactor for code > 100 lines.

TESTS: Added unit tests for both HashCoGroup and CoGroupedPipe.

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
@@ -1,9 +1,11 @@
 package com.twitter.scalding.beam_backend
 
-import com.twitter.algebird.{ Semigroup, SummingCache }
-import org.apache.beam.sdk.transforms.DoFn.{ FinishBundle, ProcessElement, StartBundle }
-import org.apache.beam.sdk.transforms.windowing.{ BoundedWindow, GlobalWindow }
-import org.apache.beam.sdk.transforms.{ DoFn, ProcessFunction }
+import com.twitter.algebird.{Semigroup, SummingCache}
+import org.apache.beam.sdk.transforms.DoFn.{FinishBundle, ProcessElement, StartBundle}
+import org.apache.beam.sdk.transforms.windowing.{BoundedWindow, GlobalWindow}
+import org.apache.beam.sdk.transforms.{DoFn, ProcessFunction}
+import org.apache.beam.sdk.values.{PCollection, PCollectionView}
+import scala.collection.JavaConverters._
 
 object BeamFunctions {
   case class ProcessPredicate[A](f: A => Boolean) extends ProcessFunction[A, java.lang.Boolean] {
@@ -25,7 +27,9 @@ object BeamFunctions {
       c.output(f(c.element()))
   }
 
-  case class MapSideAggregator[K, V](size: Int, semigroup: Semigroup[V]) extends DoFn[(K, V), (K, V)] {
+  case class MapSideAggregator[K, V](
+    size: Int, semigroup: Semigroup[V]
+  ) extends DoFn[(K, V), (K, V)] {
     var cache: SummingCache[K, V] = _
     @StartBundle
     def startBundle(): Unit = {
@@ -58,4 +62,35 @@ object BeamFunctions {
       }
     }
   }
+
+  case class HashJoinFn[K, V, U, W](
+    joiner: (K, V, Iterable[U]) => Iterator[W],
+    sideInput: PCollectionView[java.util.Map[K, java.lang.Iterable[U]]]
+  ) extends DoFn[(K, V), (K, W)] {
+    private[this] var mapRight: java.util.Map[K, java.lang.Iterable[U]] = null
+    private[this] val emptyUs: Iterable[U] = Seq.empty[U]
+
+    @ProcessElement
+    def processElement(c: DoFn[(K, V), (K, W)]#ProcessContext): Unit = {
+      if (mapRight == null) {
+        mapRight = c.sideInput(sideInput)
+      }
+      val key = c.element()._1
+      val value = c.element()._2
+      val it = mapRight.get(key) match {
+        case null => joiner(key, value, emptyUs)
+        case notEmpty => joiner(key, value, notEmpty.asScala)
+      }
+      while (it.hasNext) {
+        c.output((key, it.next()))
+      }
+    }
+
+    @FinishBundle
+    def finishBundle(c: DoFn[(K, V), (K, W)]#FinishBundleContext): Unit = {
+      mapRight = null
+    }
+  }
+
+  def widenPCollection[A, B >: A](p: PCollection[_ <: A]): PCollection[B] = p.asInstanceOf[PCollection[B]]
 }
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
@@ -107,7 +107,8 @@ object BeamOp extends Serializable {
     kryoCoder: KryoCoder
   ) extends BeamOp[B] {
     def run(pipeline: Pipeline): PCollection[B] = {
-      source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
+      val pCollection: PCollection[A] = widenPCollection(source.run(pipeline))
+      pCollection.apply(f).setCoder(kryoCoder)
     }
   }
 
@@ -165,9 +166,10 @@ object BeamOp extends Serializable {
         )((keyed, colWithTag) => keyed.and[Any](colWithTag._2, colWithTag._1))
 
       keyedPCollectionTuple
-        .apply(CoGroupByKey.create()).setCoder(KvCoder.of(kryoCoder, coGroupResultCoder))
-        .apply(ParDo.of(new CoGroupDoFn[K, V](cg, tupleTags))).setCoder(kryoCoder)
-        .apply(KVToTuple[K, V]).setCoder(TupleCoder(keyCoder, kryoCoder))
+        .apply(CoGroupByKey.create()).setCoder(KvCoder.of(keyCoder, coGroupResultCoder))
+        .apply(ParDo.of(new CoGroupDoFn[K, V](cg, tupleTags)))
+        .setCoder(KvCoder.of(keyCoder, kryoCoder))
+        .apply(KVToTuple[K, V](keyCoder, kryoCoder))
     }
   }
 
@@ -201,19 +203,18 @@ object BeamOp extends Serializable {
         op,
         new PTransform[PCollection[(K, V)], PCollection[(K, U)]]() {
           override def expand(input: PCollection[(K, V)]): PCollection[(K, U)] = {
+            val keyCoder: Coder[K] = OrderedSerializationCoder(ordK, kryoCoder)
+
             val groupedValues = input
-              .apply(TupleToKV[K, V](OrderedSerializationCoder(ordK, kryoCoder), kryoCoder))
+              .apply(TupleToKV[K, V](keyCoder, kryoCoder))
               .apply(GroupByKey.create[K, V]())
-              .setCoder(KvCoder.of(
-                OrderedSerializationCoder(ordK, kryoCoder),
-                IterableCoder.of(kryoCoder))
-              )
+              .setCoder(KvCoder.of(keyCoder, IterableCoder.of(kryoCoder)))
 
             planMapGroup[K, V, U](groupedValues, reduceFn)
               .apply(ParDo.of(FlatMapFn[KV[K, java.lang.Iterable[U]], KV[K, U]]{ elem =>
                 elem.getValue.asScala.map(KV.of(elem.getKey, _))
-              })).setCoder(KvCoder.of(OrderedSerializationCoder(ordK, kryoCoder), kryoCoder))
-              .apply(KVToTuple[K, U])
+              })).setCoder(KvCoder.of(keyCoder, kryoCoder))
+              .apply(KVToTuple[K, U](keyCoder, kryoCoder))
           }
         },
         kryoCoder)
@@ -226,23 +227,20 @@ object BeamOp extends Serializable {
         op,
         new PTransform[PCollection[(K, V)], PCollection[(K, U)]]() {
           override def expand(input: PCollection[(K, V)]): PCollection[(K, U)] = {
+            val keyCoder: Coder[K] = OrderedSerializationCoder(ordK, kryoCoder)
+            val valueCoder: Coder[V] = OrderedSerializationCoder(ordV, kryoCoder)
+
             val groupedSortedValues = input
-              .apply(TupleToKV[K, V](
-                OrderedSerializationCoder(ordK, kryoCoder),
-                OrderedSerializationCoder(ordV, kryoCoder))
-              )
+              .apply(TupleToKV[K, V](keyCoder, valueCoder))
               .apply(GroupByKey.create[K, V]())
-              .setCoder(KvCoder.of(
-                OrderedSerializationCoder(ordK, kryoCoder),
-                IterableCoder.of(OrderedSerializationCoder(ordV, kryoCoder)))
-              )
+              .setCoder(KvCoder.of(keyCoder, IterableCoder.of(valueCoder)))
               .apply(SortGroupedValues[K, V])
 
             planMapGroup[K, V, U](groupedSortedValues, reduceFn)
               .apply(ParDo.of(FlatMapFn[KV[K, java.lang.Iterable[U]], KV[K, U]]{ elem =>
                 elem.getValue.asScala.map(KV.of(elem.getKey, _))
-              })).setCoder(KvCoder.of(OrderedSerializationCoder(ordK, kryoCoder), kryoCoder))
-              .apply(KVToTuple[K, U])
+              })).setCoder(KvCoder.of(keyCoder, kryoCoder))
+              .apply(KVToTuple[K, U](keyCoder, kryoCoder))
           }
         },
         kryoCoder)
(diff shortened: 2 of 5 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -58,4 +62,20 @@ object BeamFunctions {
...
+  case class HashJoinFn[K, V, U, W](
+    joiner: (K, V, Iterable[U]) => Iterator[W],
+    sideInput: PCollectionView[java.util.Map[K, java.lang.Iterable[U]]]
+  ) extends DoFn[(K, V), (K, W)] {
+    @ProcessElement
+    def processElement(c: DoFn[(K, V), (K, W)]#ProcessContext): Unit = {
+      val sideInputMap = c.sideInput(sideInput)
+      val key = c.element()._1
+      val value = c.element()._2
+      val it = joiner(key, value, sideInputMap.asScala.getOrElse(key, Seq.empty[U].asJava).asScala)
```
johnynek: maybe allocate `private[this] val emptyUs: java.lang.Iterable[U] = Seq.empty[U].asJava` so you don't need to allocate multiple times.

keep in mind this gets called for each element on the left hand side. Each allocation avoided could be a win. I would do:

```scala
val it = sideInputMap.get(key) match {
  case null => joiner(key, value, emptyUs)
  case notEmpty => joiner(key, value, notEmpty.asScala)
}
while(it.hasNext()) {
  c.output((key, it.next()))
}
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,110 @@ object BeamOp extends Serializable {
...
+  ) extends BeamOp[B] {
     def run(pipeline: Pipeline): PCollection[B] = {
       source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
     }
   }
 
+  final case class HashJoinOp[K, V, U, W](
+    left: BeamOp[(K, V)],
+    right: BeamOp[(K, U)], joiner: (K, V, Iterable[U]) => Iterator[W]
+  )(implicit kryoCoder: KryoCoder) extends BeamOp[(K, W)] {
```
johnynek: I think `HashJoin` has access to `Ordering[K]` which in some cases will be a serialization of `K` if it is an `OrderedSerialization[K]`. I think you should accept the ordering here to potentially use that serialization.

https://github.com/twitter/scalding/blob/e445ae8d931845ddccec0795f2a0cb61c57d31d7/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala#L37

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,110 @@ object BeamOp extends Serializable {
...
+    override def run(pipeline: Pipeline): PCollection[_ <: (K, W)] = {
+      val leftPCollection = left.run(pipeline)
+      val rightPCollectionView = right.run(pipeline).asInstanceOf[PCollection[(K, U)]]
```
johnynek: rather than use `asInstanceOf`, which is dangerous: if scomeone refactors we won't get a compilation error at any cast site. Instead have one safe cast function:
```scala
def widenPCollection[A, B >: A](p: PCollection[A]): PCollection[B] = p.asInstanceOf[PCollection[B]]
```
if we have that one function, then we see that we have a very specific cast, not a totally general one.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,110 @@ object BeamOp extends Serializable {
...
+        .apply(GroupByKey.create[K, U]()).setCoder(KvCoder.of(kryoCoder, kryoCoder))
+        .apply(View.asMap[K, java.lang.Iterable[U]]())
+
+      leftPCollection
+        .apply(
+          ParDo
+            .of(HashJoinFn[K, V, U, W](joiner, rightPCollectionView))
+            .withSideInputs(rightPCollectionView)
+        )
+        .setCoder(kryoCoder)
```
johnynek: I think we should have a `TupleCoder` which uses the coder for the key and the value. The key coder may be found from the Ordering.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,110 @@ object BeamOp extends Serializable {
...
+
+  final case class CoGroupedOp[K, V](
+    cg: CoGrouped[K, V],
+    inputOps: Seq[BeamOp[(K, Any)]]
+  )(implicit kryoCoder: KryoCoder) extends BeamOp[(K, V)] {
+    override def run(pipeline: Pipeline): PCollection[_ <: (K, V)] = {
+      val inputOpsSize = inputOps.size
+      val pcols = inputOps.map { inputOp =>
+        inputOp.op
+          .run(pipeline).asInstanceOf[PCollection[(K, Any)]]
```
johnynek: we should use the `widenPCollection` function I mentioned.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,110 @@ object BeamOp extends Serializable {
...
+            .withSideInputs(rightPCollectionView)
+        )
+        .setCoder(kryoCoder)
+    }
+  }
+
+  final case class CoGroupedOp[K, V](
+    cg: CoGrouped[K, V],
+    inputOps: Seq[BeamOp[(K, Any)]]
+  )(implicit kryoCoder: KryoCoder) extends BeamOp[(K, V)] {
```
johnynek, line 142: cg has `Ordering[K]`. We should use that to potentially get an `Coder[K]` 

https://github.com/twitter/scalding/blob/e445ae8d931845ddccec0795f2a0cb61c57d31d7/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala#L37

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -136,6 +136,23 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+      TypedPipe.from(Seq((0, 1), (0, 2), (1, 0))).hashJoin(smaller)
+    },
+      Seq((0, (1, 0)), (0, (2, 0)))
+    )
+  }
+
+  test("CoGroup"){
+    beamMatchesSeq({
+      val tp1: TypedPipe[(Int, Int)] = TypedPipe.from(Seq((1, 2), (1, 3)))
+      val tp2: TypedPipe[(Int, Int)] = TypedPipe.from(Seq((1, 7)))
```
johnynek: can we add tests with multiple values per keys on left, right, both... and also filtering: where some keys only appear on the right or left.

Also, can we test join, leftJoin, rightJoin, outerJoin since those are the most common join functions?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -136,6 +136,23 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
     )
   }
 
+  test("HashJoin"){
+    beamMatchesSeq({
+      val smaller = TypedPipe.from(Seq((0, 0), (2, 1)))
```
johnynek: can we test with multiple values per key?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -58,4 +62,20 @@ object BeamFunctions {
...
     }
   }
+
+  case class HashJoinFn[K, V, U, W](
+    joiner: (K, V, Iterable[U]) => Iterator[W],
+    sideInput: PCollectionView[java.util.Map[K, java.lang.Iterable[U]]]
+  ) extends DoFn[(K, V), (K, W)] {
+    @ProcessElement
+    def processElement(c: DoFn[(K, V), (K, W)]#ProcessContext): Unit = {
+      val sideInputMap = c.sideInput(sideInput)
```
johnynek: are we not materializing this more than one? Should we not do something like:
```scala
private[this] var mapRight: java.util.Map[K, java.lang.Iterable[U]] = null

    @ProcessElement
    def processElement(c: DoFn[(K, V), (K, W)]#ProcessContext): Unit = {
      if (mapRight == null) {
        mapRight = c.sideInput(sideInput)
      }

      ...

```
Then in FinishBundle set it back to `null`.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -136,6 +136,23 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+      TypedPipe.from(Seq((0, 1), (0, 2), (1, 0))).hashJoin(smaller)
```
johnynek: how about a test with hashLeftJoin also?

johnynek:
small note:

Rebasing every time makes reviewers have to look through the entire commit again since we can't see what changed.

We use squash merge at the end, so there is no benefit to rebasing in terms of the final commit history. Would you be willing to just add commits so we can more easily review?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,113 @@ object BeamOp extends Serializable {
...
   }
 
-  final case class TransformBeamOp[A, B](source: BeamOp[A], f: PTransform[PCollection[A], PCollection[B]], kryoCoder: KryoCoder) extends BeamOp[B] {
+  final case class TransformBeamOp[A, B](
+    source: BeamOp[A],
+    f: PTransform[PCollection[A], PCollection[B]],
+    kryoCoder: KryoCoder
+  ) extends BeamOp[B] {
     def run(pipeline: Pipeline): PCollection[B] = {
       source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
```
johnynek: could we use `widenPCollection` here?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,113 @@ object BeamOp extends Serializable {
...
+
+      val keyedPCollectionTuple: KeyedPCollectionTuple[K] = pcols
+        .zip(tupleTags)
+        .foldLeft(
+          KeyedPCollectionTuple.empty[K](pipeline)
+        )((keyed, colWithTag) => keyed.and[Any](colWithTag._2, colWithTag._1))
+
+      keyedPCollectionTuple
+        .apply(CoGroupByKey.create()).setCoder(KvCoder.of(kryoCoder, coGroupResultCoder))
+        .apply(ParDo.of(new CoGroupDoFn[K, V](cg, tupleTags))).setCoder(kryoCoder)
```
johnynek: can't we use `.setCoder(KvCoder.of(keyCoder, kryoCoder))` here? if not, can we comment why?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,21 +101,113 @@ object BeamOp extends Serializable {
...
+      )
+
+      val keyedPCollectionTuple: KeyedPCollectionTuple[K] = pcols
+        .zip(tupleTags)
+        .foldLeft(
+          KeyedPCollectionTuple.empty[K](pipeline)
+        )((keyed, colWithTag) => keyed.and[Any](colWithTag._2, colWithTag._1))
+
+      keyedPCollectionTuple
+        .apply(CoGroupByKey.create()).setCoder(KvCoder.of(kryoCoder, coGroupResultCoder))
```
johnynek: can't we use `keyCoder` here at all?

nownikhil:
Yeah that makes sense, my bad. Will add a new commit next time.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -183,7 +331,9 @@ object BeamOp extends Serializable {
     }
   }
 
-  case class KVToTuple[K, V](implicit kryoCoder: KryoCoder) extends PTransform[PCollection[KV[K, V]], PCollection[(K, V)]] {
+  case class KVToTuple[K, V](
```
johnynek, line 325: seems like this should take a `Coder[K]` and `Coder[V]` and build a `TupleCoder[K, V]` from those rather than use kryo for the whole thing. At least we can save the serialization tag that the value is a tuple, since we know that statically.

johnynek (commented):
the new tests look good.

Just a few minor comments about serializations.

johnynek (approved):
:shipit:

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/KryoCoder.scala
```diff
@@ -33,10 +36,22 @@ case class OrderedSerializationCoder[T](ordSer: OrderedSerialization[T]) extends
   override def encode(value: T, outStream: OutputStream): Unit = ordSer.write(outStream, value)
```
johnynek, line 36: I just noticed an issue here. This should be: `ordSer.write(outStream, value).get` since the method returns a `Try[Unit]` and we want to see that exception. The current code will be hiding an exception if we overrun a buffer, for instance.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -58,4 +62,35 @@ object BeamFunctions {
...
+  ) extends DoFn[(K, V), (K, W)] {
+    private[this] var mapRight: java.util.Map[K, java.lang.Iterable[U]] = null
+    private[this] val emptyUs: Iterable[U] = Seq.empty[U]
+
+    @ProcessElement
+    def processElement(c: DoFn[(K, V), (K, W)]#ProcessContext): Unit = {
+      if (mapRight == null) {
+        mapRight = c.sideInput(sideInput)
+      }
+      val key = c.element()._1
```
ttim, line 78: nit: `val (kev, value) = c.element()`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
```diff
@@ -58,4 +62,35 @@ object BeamFunctions {
...
+      val value = c.element()._2
+      val it = mapRight.get(key) match {
```
ttim, line 80: nit: `.getOrElse(key, emptyUs)` or something like that

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -57,14 +68,17 @@ object BeamOp extends Serializable {
...
+        )
       case notComposedOrSum =>
         pcoll.apply(ParDo.of(
           MapFn[KV[K, java.lang.Iterable[V]], KV[K, java.lang.Iterable[U]]] { elem =>
             KV.of(
               elem.getKey,
               notComposedOrSum(elem.getKey, elem.getValue.asScala.toIterator).toIterable.asJava)
-          })).setCoder(KvCoder.of(OrderedSerializationCoder(ordK, kryoCoder), IterableCoder.of(kryoCoder)))
+          })).setCoder(
+          KvCoder.of(OrderedSerializationCoder(ordK, kryoCoder), IterableCoder.of(kryoCoder)))
```
ttim, line 81: I think it should be only one (maybe two) place/s where we do creation of `OrderedSerializationCoder`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,47 +101,146 @@ object BeamOp extends Serializable {
...
     def run(pipeline: Pipeline): PCollection[B] = {
-      source.run(pipeline).asInstanceOf[PCollection[A]].apply(f).setCoder(kryoCoder)
+      val pCollection: PCollection[A] = widenPCollection(source.run(pipeline))
+      pCollection.apply(f).setCoder(kryoCoder)
+    }
+  }
+
+  final case class HashJoinOp[K, V, U, W](
+    left: BeamOp[(K, V)],
+    right: BeamOp[(K, U)], joiner: (K, V, Iterable[U]) => Iterator[W]
```
ttim, line 117: nit: next line

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -85,47 +101,146 @@ object BeamOp extends Serializable {
...
+  )(implicit kryoCoder: KryoCoder, ordK: Ordering[K]) extends BeamOp[(K, W)] {
```
ttim, line 118: This is confusing signature, what `kryoCoder` is coder for?

ttim (approved):
Good job, thanks! Couple of nits you can handle in follow up reviews

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
johnynek: I may have suggested this. Actually here where we are in the core, I'm actually pretty supportive of coding it like Java.

Reducing a few allocations inside a big job could be measurable.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamFunctions.scala
johnynek: I suggested this. This code compiles to an if/else. The getOrElse has to allocate a closure to model the lazy nature of getOrElse. Then the else case is a function call, not the other side of a branch.

Inside the hot path of the backends I would use pretty low level scala.

PR #1948, opened 2021-09-21
Adds support for blob, tinyint and timestamp to TypedJDBCSource

tlazaro:


tlazaro:
Trying to do some cleanup and have all Twitter internal changes pushed to open source. Please advise on whether this change makes sense as-is and whether there could be consequences for other users.

Will try to add unit tests.

tlazaro:
Just found we had #1915 with this work already.

johnynek:
is there any reason why the previous one was not merged @navinvishy ?

Looks like it just got lost somehow...

navinvishy:
> is there any reason why the previous one was not merged @navinvishy ?
> 
> Looks like it just got lost somehow...

Yeah, my bad - I think this just got lost. fwiw we've had these changes at Twitter for a while now.

tlazaro:
Updated PR with unit tests from #1915. 

I think we were hesitant because of some comments around TINYINT(1) being a BOOLEAN and DATETIME (this one was removed the patch) and both out of topic from the PR.

This code in this PR is what we have had in production for the past 2 years, happy to make changes if it makes sense but otherwise we are satisfied with these changes.

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/handler/BlobTypeHandler.scala
```diff
@@ -0,0 +1,20 @@
...
+object BlobTypeHandler {
+  def apply[T](c: Context)(implicit
+    accessorTree: List[c.universe.MethodSymbol],
+    fieldName: FieldName,
+    defaultValue: Option[c.Expr[String]],
+    annotationInfo: List[(c.universe.Type, Option[Int])],
+    nullable: Boolean): scala.util.Try[List[ColumnFormat[c.type]]] = {
+
+    assert(defaultValue.isEmpty)
+    assert(annotationInfo.isEmpty)
```
johnynek: why use `assert` here but return `Success`?

Why not:
```scala
if (defaultValue.nonEmpty || annotationInfo.nonEmpty) Failure(new Exception(s"invalid args: defaultValue = $defaultValue annotationInfo = $annotationInfo"))
else Success(List(ColumnFormat(c)(accessorTree, "BLOB", None)))
```

johnynek (changes requested):

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
```diff
@@ -208,6 +210,7 @@ object ColumnDefinitionProviderImpl {
           case "BIGINT" =>
```
tlazaro, line 210: GitHub doesn't allow me to leave a comment on  `case "BOOLEAN" | "TINYINT"` above...

Should we also change it this?:

```
case "BOOLEAN" => q"""List("BOOLEAN", "BOOL").contains($typeNameTerm)"""
case "TINYINT" => q"""List("BOOLEAN", "BOOL", "TINYINT").contains($typeNameTerm)"""
```

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/handler/BlobTypeHandler.scala
tlazaro: Just changed it to `required` as was done in the original PR, would that be ok or `if else` better?

tlazaro: I will do if else of Success Failure makes more sense than throwing

johnynek: well, to a method that returns a `Try` shouldn't also throw.

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
johnynek: I really don't know. I think maybe @ianoc wrote this code initially. I don't know if he has time to glance at it.

navinvishy: Here's what Hibernate uses for mapping basic types: https://docs.jboss.org/hibernate/stable/core.old/reference/en/html/mapping-types.html

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/handler/BlobTypeHandler.scala
tlazaro: Done

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
tlazaro: I think we can leave this as-is.

tlazaro:
@johnynek do you think we can land this?

johnynek (approved):
sorry for the latency.

scalding-db/src/main/scala/com/twitter/scalding/db/macros/impl/ColumnDefinitionProviderImpl.scala
ianoc: I tried looking through this yesterday, but at this point its been so long, i've mostly forgotten at all. I'm mostly shocked its still in use tbh! :)

tlazaro:
> sorry for the latency.

I think the complete opposite is the truth, you have been incredibly responsive and helping out so much in such short notice. To the point I worry we are taking too much of your time.

Issue #1947, opened 2021-09-21
Solving `sortedTake` in beam runner


nownikhil:
Hi folks,
We just implemented Beam Runner for scalding. One open issue is with using sortedTake. Since it uses PriorityQueue Monoid we mutate the input elements and that causes the job to fail. This check is enabled for Direct Runner to catch issues during testing and might not be same for Dataflow runner.

We might break consistency guarantees if we mutate input elements.
https://stackoverflow.com/questions/43142900/apache-beam-returns-input-values-must-not-be-mutated-in-any-way-when-using-lo

What would be the best way to solve this?

Maybe extend PQ Monoid and overwrite method plus to not mutate bigger PQ and just return a new one and then use it, or have SortedTake as a member of AST and implement it for every runner.

Happy to hear a clever solution to this.

tlazaro:
Regarding having the `SortedTake` as first class AST element would allow us to use `Top.of` from Beam to implement: https://beam.apache.org/releases/javadoc/2.32.0/org/apache/beam/sdk/transforms/Top.html.

If we could have a way of keeping current behavior for other backends but Beam using Top would be the safest from the POV of Beam.

tlazaro:
But Nikhil and I aren't sure this change to the AST is easy or viable.

johnynek:
I think a way to solve it would be to pattern match on the Monoid:

https://github.com/twitter/scalding/blob/767a9457cf763f84b02f50a687bf6f75d67cb9db/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala#L62

```scala
  case EmptyGuard(MapValueStream(SumAll(pq: PriorityQueueMonoid[V])) =>
    val ordering = pq.zero.comparator()
    // now use beam's built in take-of...
```

johnynek:
but an additional approach would be to use an immutable heap, and that is really a better approach:

https://github.com/typelevel/cats-collections/blob/master/core/src/main/scala/cats/collections/PairingHeap.scala

and we could implement that in scalding and not call out to algebird's mutable implementation. I really regret the mutable implementation in algebird. We should have implemented an immutable heap there.

nownikhil:
Should we add dependency of Cats in scalding or copy paste the implementation ?

johnynek:
I would copy it to avoid the dependency, and copy the tests. (I wrote pairing heap, btw). Just leave the copyright headers and the license is the same, so it would be fine.

[...]

PR #1949, opened 2021-09-21
Support `sortedTake` in beam runner

nownikhil:
We flatten all the input PriorityQueues and construct a new PQ using the monoid provided.
TESTS: Updated failing unit test

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
@@ -1,12 +1,14 @@
 package com.twitter.scalding.beam_backend
 
 import com.twitter.algebird.Semigroup
+import com.twitter.algebird.mutable.PriorityQueueMonoid
 import com.twitter.scalding.Config
 import com.twitter.scalding.beam_backend.BeamFunctions._
 import com.twitter.scalding.typed.functions.ComposedFunctions.ComposedMapGroup
 import com.twitter.scalding.typed.functions.{EmptyGuard, MapValueStream, SumAll}
 import com.twitter.scalding.typed.{CoGrouped, TypedSource}
 import java.lang
+import java.util.PriorityQueue
 import org.apache.beam.sdk.Pipeline
 import org.apache.beam.sdk.coders.{Coder, IterableCoder, KvCoder}
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
@@ -59,6 +61,25 @@ object BeamOp extends Serializable {
   )(implicit ordK: Ordering[K], kryoCoder: KryoCoder): PCollection[KV[K, java.lang.Iterable[U]]] = {
     reduceFn match {
       case ComposedMapGroup(f, g) => planMapGroup(planMapGroup(pcoll, f), g)
+      case EmptyGuard(MapValueStream(SumAll(pqm: PriorityQueueMonoid[V]))) =>
+        pcoll.apply(MapElements.via(
+          new SimpleFunction[KV[K, java.lang.Iterable[V]], KV[K, java.lang.Iterable[U]]]() {
+            override def apply(input: KV[K, lang.Iterable[V]]): KV[K, java.lang.Iterable[U]] = {
+              // We are not using plus method defined in PriorityQueueMonoid as it is mutating
+              // input Priority Queues. We create a new PQ from the individual ones.
+              // We didn't use Top PTransformation in beam as it is not needed, also
+              // we cannot access `max` defined in PQ monoid.
+              val flattenedValues = input.getValue.asScala.flatMap { value =>
+                value.asInstanceOf[PriorityQueue[V]].iterator().asScala
+              }
+              val mergedPQ = pqm.build(flattenedValues)
+              KV.of(input.getKey, Iterable(mergedPQ.asInstanceOf[U]).asJava)
+            }
+          })
+        ).setCoder(KvCoder.of(
+          OrderedSerializationCoder(ordK, kryoCoder),
+          IterableCoder.of(kryoCoder))
+        )
       case EmptyGuard(MapValueStream(sa: SumAll[V])) =>
         pcoll
           .apply(Combine.groupedValues(
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
@@ -3,6 +3,7 @@ package com.twitter.scalding.beam_backend
 import com.twitter.algebird.Semigroup
 import com.twitter.scalding.Config
 import com.twitter.scalding.beam_backend.BeamFunctions._
+import com.twitter.scalding.serialization.Externalizer
 import com.twitter.scalding.typed.functions.ComposedFunctions.ComposedMapGroup
 import com.twitter.scalding.typed.functions.{
   EmptyGuard,
@@ -58,7 +59,8 @@ sealed abstract class BeamOp[+A] {
 }
 
 private final case class SerializableComparator[T](comp: Comparator[T]) extends Comparator[T] {
-  override def compare(o1: T, o2: T): Int = comp.compare(o1, o2)
+  private[this] val extCmp = Externalizer(comp)
+  override def compare(o1: T, o2: T): Int = extCmp.get.compare(o1, o2)
 }
 
 object BeamOp extends Serializable {
@@ -75,17 +77,17 @@ object BeamOp extends Serializable {
 
         vCollection.apply(MapElements.via(
           new SimpleFunction[KV[K, java.lang.Iterable[PriorityQueue[v]]], KV[K, java.lang.Iterable[U]]]() {
+            private final val topCombineFn = new TopCombineFn[v, SerializableComparator[v]](
+              pqm.count,
+              SerializableComparator[v](pqm.ordering.reverse)
+            )
+
             override def apply(input: KV[K, lang.Iterable[PriorityQueue[v]]]): KV[K, java.lang.Iterable[U]] = {
-
-              val topCombineFn = new TopCombineFn[v, SerializableComparator[v]](
-                pqm.count,
-                SerializableComparator[v](pqm.ordering.reverse)
-              )
-
               @inline def flattenedValues: Stream[v] =
                 input.getValue.asScala.toStream.flatMap(_.asScala.toStream)
 
               val outputs: java.util.List[v] = topCombineFn.apply(flattenedValues.asJava)
+              // We are building the PriorityQueue back as output type U is PriorityQueue[v]
               val pqs = pqm.build(outputs.asScala)
               KV.of(input.getKey, Iterable(pqs.asInstanceOf[U]).asJava)
             }
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
@@ -1,17 +1,22 @@
 package com.twitter.scalding.beam_backend
 
 import com.twitter.algebird.Semigroup
-import com.twitter.algebird.mutable.PriorityQueueMonoid
 import com.twitter.scalding.Config
 import com.twitter.scalding.beam_backend.BeamFunctions._
 import com.twitter.scalding.typed.functions.ComposedFunctions.ComposedMapGroup
-import com.twitter.scalding.typed.functions.{EmptyGuard, MapValueStream, SumAll}
+import com.twitter.scalding.typed.functions.{
+  EmptyGuard,
+  MapValueStream,
+  ScaldingPriorityQueueMonoid,
+  SumAll
+}
 import com.twitter.scalding.typed.{CoGrouped, TypedSource}
 import java.lang
-import java.util.PriorityQueue
+import java.util.{Comparator, PriorityQueue}
 import org.apache.beam.sdk.Pipeline
 import org.apache.beam.sdk.coders.{Coder, IterableCoder, KvCoder}
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+import org.apache.beam.sdk.transforms.Top.TopCombineFn
 import org.apache.beam.sdk.transforms._
 import org.apache.beam.sdk.transforms.join.{
   CoGbkResult,
@@ -52,6 +57,10 @@ sealed abstract class BeamOp[+A] {
     parDo(FlatMapFn(f))
 }
 
+private final case class SerializableComparator[T](comp: Comparator[T]) extends Comparator[T] {
+  override def compare(o1: T, o2: T): Int = comp.compare(o1, o2)
+}
+
 object BeamOp extends Serializable {
   implicit private def fakeClassTag[A]: ClassTag[A] = ClassTag(classOf[AnyRef]).asInstanceOf[ClassTag[A]]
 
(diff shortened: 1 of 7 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -59,6 +61,25 @@ object BeamOp extends Serializable {
   )(implicit ordK: Ordering[K], kryoCoder: KryoCoder): PCollection[KV[K, java.lang.Iterable[U]]] = {
     reduceFn match {
       case ComposedMapGroup(f, g) => planMapGroup(planMapGroup(pcoll, f), g)
+      case EmptyGuard(MapValueStream(SumAll(pqm: PriorityQueueMonoid[V]))) =>
```
johnynek: I don't *think* this should either compile correctly. I probably wrote some bad code in my example. The `SumAll` type is `TraversableOnce[A] => Iterator[A]` where in this case `A` = `V`. So when you pattern match on `PriorityQueueMonoid[T]` that means `V = PriorityQueue[T]`, so, it shouldn't be `PriorityQueueMonoid[V]`. A little known fact is that you can use a lower case to bind an unknown type: `pqm: PriorityQueueMonoid[v]` and then use `v` lower down.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -59,6 +61,25 @@ object BeamOp extends Serializable {
...
+        pcoll.apply(MapElements.via(
+          new SimpleFunction[KV[K, java.lang.Iterable[V]], KV[K, java.lang.Iterable[U]]]() {
+            override def apply(input: KV[K, lang.Iterable[V]]): KV[K, java.lang.Iterable[U]] = {
+              // We are not using plus method defined in PriorityQueueMonoid as it is mutating
+              // input Priority Queues. We create a new PQ from the individual ones.
+              // We didn't use Top PTransformation in beam as it is not needed, also
+              // we cannot access `max` defined in PQ monoid.
```
johnynek: why not extract the comparator: `val cmp = pqm.zero.comparator()` and then use Top.of as @tlazaro suggested here:

https://github.com/twitter/scalding/issues/1947#issuecomment-923522362

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -59,6 +61,25 @@ object BeamOp extends Serializable {
...
+              val flattenedValues = input.getValue.asScala.flatMap { value =>
```
johnynek: I think this is going to materialize everything in memory. I think you should do `input.getValue.asScala.toStream.flatMap`... to make sure this is a lazy Iterable.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -59,6 +61,25 @@ object BeamOp extends Serializable {
   )(implicit ordK: Ordering[K], kryoCoder: KryoCoder): PCollection[KV[K, java.lang.Iterable[U]]] = {
     reduceFn match {
       case ComposedMapGroup(f, g) => planMapGroup(planMapGroup(pcoll, f), g)
+      case EmptyGuard(MapValueStream(SumAll(pqm: PriorityQueueMonoid[V]))) =>
+        pcoll.apply(MapElements.via(
+          new SimpleFunction[KV[K, java.lang.Iterable[V]], KV[K, java.lang.Iterable[U]]]() {
+            override def apply(input: KV[K, lang.Iterable[V]]): KV[K, java.lang.Iterable[U]] = {
+              // We are not using plus method defined in PriorityQueueMonoid as it is mutating
```
johnynek: I think you also need to match on:

https://github.com/twitter/scalding/blob/e1988110f61a2849a077b92250e1e1d58511cd90/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala#L271

since I guess this restriction is also enforced on the mapside operations... 

You'll need a slightly different approach there (basically a custom cache).

But maybe they don't do their mutation detection on the mappers?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -102,23 +101,16 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
-     */
-    val test = Try {
-      beamMatchesSeq(
-        TypedPipe
-          .from(Seq(5, 3, 2, 0, 1, 4))
-          .map(x => x.toDouble)
-          .groupAll
-          .sortedReverseTake(3),
-        Seq(5, 4, 3)
+  test("sortedTake"){
```
johnynek, line 106: note that `bufferedTake` also uses the problematic monoid internally. Worth testing:

https://github.com/twitter/scalding/blob/e1988110f61a2849a077b92250e1e1d58511cd90/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala#L662

johnynek:
Good lesson here: mutation, NOT EVEN ONCE!

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
nownikhil: Top requires `max` no of elements in its constructor. But we cannot access it from the monoid.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
nownikhil: `input.getValue.asScala` gives an Iterator only and `flatmap` also works lazily. Which call would materialize it?

johnynek: `getValue` returns the `Iterable[_]` no? so, asScala on that is going to give you an `Iterable` if I'm not mistaken. You could call `.iterator` on that before you call .asScala and that will certainly be lazy, which is better than the `toStream` since we would rather not materialize the whole set.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: ahh... man that PriorityQueueMonoid is really bad. :(

ahh, but there is a way out!

https://github.com/twitter/algebird/blob/0c45f9395020ceecd28899fe156b7e3c92b29814/algebird-core/src/main/scala/com/twitter/algebird/mutable/PriorityQueueMonoid.scala#L27

you could subclass it! `class ScaldingPriorityQueueMonoid[K](val count: Int)(implicit val ordering: Ordering[K]) extends PriorityQueueMonoid(count)(ordering)` and then you can access the count and the ordering.

johnynek: I made this quick PR: https://github.com/twitter/algebird/pull/1008

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: actually... `build` requires an `Iterable` and since `flattenedValues` is a `val` it will pin things in memory.

To take this suggestion, I think you need `@inline def flattenedValues = input.getValue.asScala.toStream.flatMap...`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
nownikhil: Tried with a unit test and it failed because of mutation. Instead of overwriting the `put` method in the `SummingCache` I was hoping to pattern match on Semigroup and overwrite the plus method.

```
case class ImmutablePQMonoid[T](
    size: Int
  )(implicit ord: Ordering[T]) extends PriorityQueueMonoid[T](size) {
    override def plus(left: PriorityQueue[T], right: PriorityQueue[T]): PriorityQueue[T] = {
      super.build(left.iterator().asScala.toIterable ++ right.iterator().asScala.toIterable)
    }
  }
```
But the same issue is there. I cannot access the size field.

nownikhil: Or maybe this?

```
  case class ImmutablePQMonoid[T](
    pqm: PriorityQueueMonoid[T]
  )(implicit ord: Ordering[T]) extends Monoid[PriorityQueue[T]] {
    override def zero: PriorityQueue[T] = pqm.zero

    override def plus(
      l: PriorityQueue[T],
      r: PriorityQueue[T]
    ): PriorityQueue[T] = ???
  }
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
tlazaro: I think I understand all the pieces but would really appreciate if you could clarify how to make it properly lazy.

In the first part are you referring to `Iterable` keeping references to all elements, as opposed to `Iterator` which wouldn't?

Then when you mention using inline, is it to avoid a closure over that variable keeping the reference alive for longer or landing in more expensive generation? Or more about escape analysis or similar?

johnynek: In scala, most Iterable methods (like flatMap) are not lazy and will materialize the entire result in many cases (basically the only exception is Stream or LazyList in scala 2.13).

An Iterator generally drops references to the items it has iterated past, although you could imagine making one that doesn't. Consider a List iterator:
```scala
class ListIterator[A](lst: List[A]) extends Iterator[A] {
  private[this] var current: List[A] = lst
  def hasNext: Boolean = current.nonEmpty
  def next(): A = {
    val result = current.head
    current = current.tail
    result
  }
}
```
The way the scala compiler works, `lst` is only referenced by the constructor, so you won't keep a reference to that beyond the constructor which just uses it to initialize.

Finally, the `@inline` is just a hint, but the `def` means there is nothing keeping the reference around so when the GC runs it could colllect anything not pointed to by the Stream at that moment.

BUT, all of this is a bit moot because looking inside `PriorityQueueMonoid` we see that `build` traverses the Iterable twice... so if it is a large `Stream` you will materialize the whole thing...

I think the best idea would be just subclass PriorityQueueMonoid so you can access the max size and the ordering, and use that call Top.of.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: why not just pattern match on this monoid and not use it at all? Like, as a short fix, you could just make sumByLocalKeys a no-op for this particular monoid in the short term, but for a better fix you could implement a summingcache for this particular monoid.

I think making a full copy of the priority queue each time will be a perf killer and I doubt it will be better than just not using a mapside cache at all...

The more we talk about this, the more it seems like maybe copying the immutable heap in from cats-collections is the right move (or send a PR to algebird and we can publish a new version and use it from there, either way...)

Seems like we keep getting paper cuts here.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -52,6 +57,10 @@ sealed abstract class BeamOp[+A] {
     parDo(FlatMapFn(f))
 }
 
+private final case class SerializableComparator[T](comp: Comparator[T]) extends Comparator[T] {
```
johnynek, line 61: note, you may want to use 
https://github.com/twitter/scalding/blob/a9500fd154fa3a318e2fad063e6d3d15e70e8b17/scalding-core/src/main/scala/com/twitter/scalding/serialization/Externalizer.scala#L29

(and in other locations), e.g.
```scala
private class SerializableComparator[A](comp: Comparator[A]) extends Comparator[A] {
  private[this] val extCmp = Externalizer(comp)
  override def compare(o1: T, o2: T): Int = extCmp.get.compare(o1, o2)
}
```
The `Externalizer` tries both java serialization and kryo serialization. You may want to use it for some of the beamops as well with the functions (scalding on cascading does use Externalizer).

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -61,19 +70,24 @@ object BeamOp extends Serializable {
...
+              val topCombineFn = new TopCombineFn[v, SerializableComparator[v]](
+                pqm.count,
+                SerializableComparator[v](pqm.ordering.reverse)
+              )
+
+              @inline def flattenedValues: Stream[v] =
+                input.getValue.asScala.toStream.flatMap(_.asScala.toStream)
+
+              val outputs: java.util.List[v] = topCombineFn.apply(flattenedValues.asJava)
+              val pqs = pqm.build(outputs.asScala)
```
johnynek, line 91: why do we use both `TopCombineFn` and `pqm`? I'm confused why one isn't enough? Seems like we should be able to return `output` to me.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -113,6 +115,18 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
       )
   }
 
+  test("bufferedTake"){
+    beamMatchesSeq(
+      TypedPipe
+        .from(1 to 50)
+        .groupAll
+        .bufferedTake(100)
```
johnynek, line 123: just a note: this is going to be really bad in a real job without map-side aggregation: the key is `Unit` so there is only one key, so this would have each mapper send 100, then have the reducers pick 100 of those.

But with no mapside aggregation, all the data will be sent to the reducers, and they will throw away all but 100.

But we can add an issue and come back and address this.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -59,6 +70,30 @@ object BeamOp extends Serializable {
...
     reduceFn match {
       case ComposedMapGroup(f, g) => planMapGroup(planMapGroup(pcoll, f), g)
+      case EmptyGuard(MapValueStream(SumAll(pqm: ScaldingPriorityQueueMonoid[v]))) =>
+        val vCollection = pcoll.asInstanceOf[PCollection[KV[K, java.lang.Iterable[PriorityQueue[v]]]]]
+
+        vCollection.apply(MapElements.via(
+          new SimpleFunction[KV[K, java.lang.Iterable[PriorityQueue[v]]], KV[K, java.lang.Iterable[U]]]() {
+            override def apply(input: KV[K, lang.Iterable[PriorityQueue[v]]]): KV[K, java.lang.Iterable[U]] = {
+
+              val topCombineFn = new TopCombineFn[v, SerializableComparator[v]](
```
johnynek: I think you can lift this out of the `apply` method to reduce allocations. We shouldn't allocate a new one on every input should we?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
nownikhil: `U` here is `PriorityQueue[v]` so casting to `U` fails without converting the `outputs` to a PriorityQueue.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
nownikhil: Opened a ticket for this.
https://github.com/twitter/scalding/issues/1952

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: Ahh. Okay. It makes sense now. Can you comment?

johnynek (approved):

tlazaro (approved):

tlazaro:
I'm fine with landing this, how do we usually do it? 'Squash and merge' ?

johnynek:
yes, squash and merge if you think it is ready. Looks good to me too.

PR #1951, opened 2021-09-23
Fixes Execution#recoverWith when partial function doesn't match

tlazaro:
When a recoverWith was used and the function didn't match a MatchError would be thrown. This is due to calling apply on the partial function instead of passing as a partial to the recoverWith of Future.

The provided unit test fails without this fix.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
@@ -521,16 +521,17 @@ object Execution {
             fut
               .map {v => (v, CancellationHandler.empty) } // map this to the right shape
               .recoverWith {
-                case t: FlowStopException => // do not recover when the flow was stopped
-                  Future.failed(t)
-                case t =>
-                  val chainedFn = fn.andThen { ex0 =>
-                    // we haven't optimized ex0 yet
-                    val ex = optimize(conf, ex0)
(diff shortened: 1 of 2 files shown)


scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -521,16 +521,17 @@ object Execution {
...
-                case t: FlowStopException => // do not recover when the flow was stopped
-                  Future.failed(t)
-                case t =>
-                  val chainedFn = fn.andThen { ex0 =>
-                    // we haven't optimized ex0 yet
-                    val ex = optimize(conf, ex0)
-                    val CFuture(f, c) = ex.runStats(conf, mode, cache).get
-                    f.map { v => (v, c) }
-                  }
-                chainedFn(t)
```
tlazaro, line 533: This here was causing the MatchError to be thrown.

tlazaro:
Please advise on the way I tested this, even if a simple test there might be a more convenient or idiomatic way.

scalding-core/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: ahh, so the problem is that we should be calling `chainedFn.isDefinedAt(t)` before using... 

good catch.

johnynek (approved):

PR #1957, opened 2021-12-14
scalding-beam improvements on resulting DAG

tlazaro:
- Implements missing Merge and Fork
- Improvement to have names and better grouping of some operations so these show as fewer boxes on the Dataflow UI. These boxes can be expanded if needed to show the inner steps.
- Caching of runs of BeamOp now means PCollection used by multiple steps don't recompute.

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
@@ -119,9 +119,9 @@ object BeamPlanner {
           rec(input)
         case (m @ MergedTypedPipe(_, _), rec) =>
           OptimizationRules.unrollMerge(m) match {
-            case Nil        => rec(EmptyTypedPipe)
-            case one :: Nil => rec(one)
-            case many       => MergedBeamOp(many.map(rec(_)))
+            case Nil                     => rec(EmptyTypedPipe)
+            case single :: Nil           => rec(single)
+            case first :: second :: tail => MergedBeamOp(rec(first), rec(second), tail.map(rec(_)))
           }
       }
     })
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
@@ -4,7 +4,7 @@ import com.stripe.dagon.{FunctionK, Memoize, Rule}
 import com.twitter.chill.KryoInstantiator
 import com.twitter.chill.config.ScalaMapConfig
 import com.twitter.scalding.Config
-import com.twitter.scalding.beam_backend.BeamOp.CoGroupedOp
+import com.twitter.scalding.beam_backend.BeamOp.{CoGroupedOp, MergedBeamOp}
 import com.twitter.scalding.serialization.KryoHadoop
 import com.twitter.scalding.typed._
 import com.twitter.scalding.typed.functions.{
@@ -115,6 +115,14 @@ object BeamPlanner {
             CoGroupedOp(cg, ops)
           }
           go(cg)
+        case (Fork(input), rec) =>
+          rec(input)
+        case (m @ MergedTypedPipe(_, _), rec) =>
+          OptimizationRules.unrollMerge(m) match {
+            case Nil        => rec(EmptyTypedPipe)
+            case one :: Nil => rec(one)
+            case many       => MergedBeamOp(many.map(rec(_)))
+          }
       }
     })
   }
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
@@ -1,38 +1,37 @@
 package com.twitter.scalding.beam_backend
 
+import com.stripe.dagon.Memoize
 import com.twitter.algebird.Semigroup
 import com.twitter.scalding.Config
 import com.twitter.scalding.beam_backend.BeamFunctions._
 import com.twitter.scalding.serialization.Externalizer
 import com.twitter.scalding.typed.functions.ComposedFunctions.ComposedMapGroup
 import com.twitter.scalding.typed.functions.{EmptyGuard, MapValueStream, ScaldingPriorityQueueMonoid, SumAll}
-import com.twitter.scalding.typed.{CoGrouped, TypedSource}
-import java.lang
+import com.twitter.scalding.typed.{CoGrouped, MultiJoinFunction, TypedSource}
 import java.util.{Comparator, PriorityQueue}
 import org.apache.beam.sdk.Pipeline
 import org.apache.beam.sdk.coders.{Coder, IterableCoder, KvCoder}
 import org.apache.beam.sdk.transforms.DoFn.ProcessElement
 import org.apache.beam.sdk.transforms.Top.TopCombineFn
 import org.apache.beam.sdk.transforms._
-import org.apache.beam.sdk.transforms.join.{
-  CoGbkResult,
-  CoGbkResultSchema,
-  CoGroupByKey,
-  KeyedPCollectionTuple,
-  UnionCoder
-}
+import org.apache.beam.sdk.transforms.join.{CoGbkResult, CoGroupByKey, KeyedPCollectionTuple}
 import org.apache.beam.sdk.values.PCollectionList
 import org.apache.beam.sdk.values.PCollectionTuple
 import org.apache.beam.sdk.values.{KV, PCollection, TupleTag}
 import scala.collection.JavaConverters._
-import scala.language.implicitConversions
 import scala.reflect.ClassTag
 
 sealed abstract class BeamOp[+A] {
 
   import BeamOp.TransformBeamOp
 
-  def run(pipeline: Pipeline): PCollection[_ <: A]
+  protected lazy val cachedRun = Memoize.function[Pipeline, PCollection[_ <: A]] { case (pipeline, _) =>
+    runNoCache(pipeline)
+  }
+
+  final def run(pipeline: Pipeline): PCollection[_ <: A] = cachedRun(pipeline)
+
+  protected def runNoCache(p: Pipeline): PCollection[_ <: A]
 
   def map[B](f: A => B)(implicit kryoCoder: KryoCoder): BeamOp[B] =
     parDo(MapFn(f), "map")
@@ -87,7 +86,7 @@ object BeamOp extends Serializable {
                 )
 
                 override def apply(
-                    input: KV[K, lang.Iterable[PriorityQueue[v]]]
+                    input: KV[K, java.lang.Iterable[PriorityQueue[v]]]
                 ): KV[K, java.lang.Iterable[U]] = {
                   @inline def flattenedValues: Stream[v] =
                     input.getValue.asScala.toStream.flatMap(_.asScala.toStream)
@@ -108,7 +107,7 @@ object BeamOp extends Serializable {
           }))
           .setCoder(KvCoder.of(OrderedSerializationCoder(ordK, kryoCoder), kryoCoder))
           .apply(MapElements.via(new SimpleFunction[KV[K, V], KV[K, java.lang.Iterable[U]]]() {
-            override def apply(input: KV[K, V]): KV[K, lang.Iterable[U]] =
+            override def apply(input: KV[K, V]): KV[K, java.lang.Iterable[U]] =
               KV.of(input.getKey, Seq(input.getValue.asInstanceOf[U]).toIterable.asJava)
           }))
           .setCoder(
@@ -127,7 +126,7 @@ object BeamOp extends Serializable {
 
   final case class Source[A](conf: Config, original: TypedSource[A], input: Option[BeamSource[A]])
       extends BeamOp[A] {
-    def run(pipeline: Pipeline): PCollection[_ <: A] =
+    override def runNoCache(pipeline: Pipeline): PCollection[_ <: A] =
       input match {
         case None =>
           throw new IllegalArgumentException(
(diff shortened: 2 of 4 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -176,6 +180,54 @@ object BeamOp extends Serializable {
...
+      val rightPCollection: PCollection[(K, U)] = widenPCollection(right.run(pipeline))
+
+      val tuple: PCollectionTuple = PCollectionTuple.of[(K, _)](
+        "left",
+        widenPCollection(leftPCollection): PCollection[(K, _)],
+        "right",
+        widenPCollection(rightPCollection): PCollection[(K, _)]
+      )
+
+      tuple.apply(HashJoinTransform(keyCoder, joiner))
```
johnynek, line 200: we don't need the explicit `apply` do we ? `tuple(HashJoinTransform(keyCoder, joiner))` would work, no?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -176,6 +180,54 @@ object BeamOp extends Serializable {
...
+    }
+  }
+
+  final case class MergedBeamOp[A](operations: Seq[BeamOp[A]]) extends BeamOp[A] {
```
johnynek: do we want to add `require(operations.size > 1, s"expected at least two items in Merged: $operations")` just to guard against a refactoring creating suboptimal plans here?

johnynek (commented):
this looks good.

a few comments.

Also, I think toIterableExecution and forceToDiskExecution aren't implemented yet. Is that right?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -115,6 +115,14 @@ object BeamPlanner {
             CoGroupedOp(cg, ops)
           }
           go(cg)
+        case (Fork(input), rec) =>
+          rec(input)
```
johnynek, line 119: btw: if you could link to any documentation here that says that beam can handle fanouts well or under what circumstances you need to annotate them in any way, that would be useful.

A fork node is added by the planner if a node has more than one downstream consumer.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
tlazaro: Yes that is true but for all the `apply` from Beam's Pipeline, PTransforms and similar we write it explicitly to signal it's Beam API calls and that's what it would look like called from Java or from Beam examples.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
tlazaro: Sounds reasonable, or we could make it `first`, `second`, `tail` ?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
tlazaro: I assumed it would but let me dig into that.

tlazaro:
We haven't tackled toIterableExecution and forceToDiskExecution yet. Still trying to get a successful run for a specific pretty large job.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
tlazaro: Just to be clear, is it ok to assume that the nodes leading up to this Fork will be processed only once by the backend and that their resulting PCollection instance will be used for all downstream nodes? So if there is re-computation it's entirely on Beam/Dataflow for re-computing a PCollection?

johnynek: Yes. That's what the memoization is for. The whole `rec` bit is about caching.

tlazaro: Cool, just wanted to confirm. I'm observing the following:

- Handwritten Beam code running on Dataflow does reuse PCollections and does a fan-out to multiple outputs. Not sure there are rules or hints can be passed but seems to just work. Will try to find out if there are limitations or if always reuses.
- The Scalding to Beam conversion for a real job is not doing reuse and is recomputing everything
  - One of our sources implementation doesn't have equality implemented and we though this was messing up memoization, we think we fixed it but still no reuse.
  - Trying even simpler cases, with TypedPipe.from(List(...)) and still can't get any reuse.
  
I will continue experimenting and narrowing it down, trying to understand if there is an issue with the conversion or in some assumption about Beam/Dataflow.

tlazaro: I think we have found the problem, while we take care to end up with a single BeamOp instance per transform, the `run` in a BeamOp is called by every other transform that depends on it, causing recomputation. We could cache the result of the `run` locally in the BeamOp which seems easier or have a shared cache.

tlazaro: Ok something like this does the trick:

```scala
  final case class TransformBeamOp[A, B](
    source: BeamOp[A],
    f: PTransform[PCollection[A], PCollection[B]],
    kryoCoder: KryoCoder,
    name: String
  ) extends BeamOp[B] {

    private lazy val memo = Memoize.function[Pipeline, PCollection[B]] {
      case (pipeline, _) =>
        val pCollection: PCollection[A] = widenPCollection(source.run(pipeline))
        pCollection.apply(name, f).setCoder(kryoCoder)
    }

    def run(pipeline: Pipeline): PCollection[B] = memo(pipeline)
  }
  ```

@johnynek do you think this approach makes sense? There are only a handful of BeamOp's so it isn't too much work. We could do the memo in the parent trait but the type of the elements wouldn't be as precise. We could cast or use some abstract types if feeling fancy, or just keep this duplication.

Comment (johnynek) [reply scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala:119]: yeah, I would probably do something like below inside BeamOp:

```scala

protected def runNoCache(p: Pipeline): PCollection[_ <: A]
private val resultCache = scala.collection.mutable.Map.empty[Pipeline, PCollection[_ <: A]]
  
final def run(p: Pipeline): PCollection[_ <: A] =
  resultCache.getOrElseUpdate(p, runNoCache(p))
```

and then change all the current `def run` into `def runNoCache`

tlazaro: This works by the way, got reuse on Dataflow.

johnynek: Okay, good. Sorry my message missed yours. I would still do the `def runNoCache` on the base trait (but you are right, using `Memoize.function` is safer actually we found some bugs trying to do memoization with mutable Map in scala, so internally it uses a var + immutable map).

tlazaro:
Added some improvements to CoGroup so the TupleToKv doesn't show in the top level ui and rather lives inside CoGroup outer PTransform.

Added caching to BeamOp following @johnynek's advice, new implementations should by default be built using the cache preventing mistakes.

Added very shallow tests for the caching. Our goal should be testing the structure of the Pipeline in code, instead of visually in a Dataflow UI. The Pipeline DAG is a bit tedious to work with, the way would be to use the visitor pattern it has and perform mutations to build a better DAG instance where could perform better analysis.

tlazaro:
Can't afford to do proper structural testing now, though I'm sure will come back as a problem later.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -35,6 +42,63 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
   def tmpPath(suffix: String): String =
     Paths.get(testPath, suffix).toString
 
+  test("BeamOp caching: FromIterable") {
+    val a = TypedPipe.from(0 to 5)
+
+    val (pipeline, op) = beamPlan(a)
+
+    assert(op.isInstanceOf[FromIterable[Int]])
+    assert(op.run(pipeline) == op.run(pipeline))
```
johnynek: do you want `eq` here? to test reference equality?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -35,6 +42,63 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+  }
+
+  test("BeamOp caching: CoGroupedOp") {
+    val leftPipe: TypedPipe[(Int, Int)] = TypedPipe.from(Seq((0, 0), (0, 1), (1, 1), (3, 3)))
+    val rightPipe: TypedPipe[(Int, Int)] = TypedPipe.from(Seq((0, 0), (0, 3), (2, 2), (2, 3)))
+
+    val (pipeline, op) = beamPlan(leftPipe.join(rightPipe))
+
+    assert(op.isInstanceOf[CoGroupedOp[Int, Int]])
+    assert(op.run(pipeline) == op.run(pipeline))
```
johnynek: do you want `eq` here?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -35,6 +42,63 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+  }
+
+  test("BeamOp caching: MergedBeamOp") {
+    val a = TypedPipe.from(0 to 5)
+    val b = TypedPipe.from(6 to 10)
+
+    val (pipeline, op) = beamPlan(a ++ b)
+
+    assert(op.isInstanceOf[MergedBeamOp[Int]])
+    assert(op.run(pipeline) == op.run(pipeline))
```
johnynek: eq?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -35,6 +42,63 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+  }
+
+  test("BeamOp caching: Source") {
+    val source = TypedPipe.from(TextLine("/"))
+
+    val (pipeline, op) = beamPlan(source)
+
+    assert(op.isInstanceOf[BeamOp.Source[String]])
+    assert(op.run(pipeline) == op.run(pipeline))
```
johnynek: `eq`?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -249,6 +352,28 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
     )
   }
 
+  test("Merge (++) two pipes") {
+    val a = TypedPipe.from(Seq(5, 3, 2, 6, 1, 4))
+    val b = TypedPipe.from(Seq(15, 13, 12, 16, 11, 14))
```
johnynek: maybe try with some duplicates to make sure `++` is uniquing the results?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -249,6 +352,28 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+
+    beamMatchesSeq(
+      a ++ b,
+      Seq(1, 2, 3, 4, 5, 6, 11, 12, 13, 14, 15, 16)
+    )
+  }
+
+  test("Merge (++) many pipes") {
```
johnynek, line 367: testing a `a ++ a` example is interesting since it should be optimized into a `flatMap` that just duplicates each element. Same with `a ++ a ++ a` I think.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: yes my bad

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: for sure

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: sure

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: I added a test to check merging a pipe to itself produces the right result but had a hard time trying to test the shape of the conversion, in a way that includes the optimizations. Do you think we can leave that work to a separate PR? I don't want to grow this one too much more.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: done

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: done

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: done

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
tlazaro: done

johnynek (approved):

Issue #1958, opened 2021-12-29
Scalding on Beam discussion about Joins


tlazaro:
Multi joins for Scalding on Beam are not performing well on Dataflow, or getting stuck on multi joins (many inputs) with very large collections where many keys with many values are common. i.e. a very large, production Scalding job with a 7-way join with billions of elements in a few inputs and millions in others, with many hot keys performs well on MapReduce but can't complete the join when running on Dataflow using the Beam backend.

It's possible the use of Iterators in the joins is causing unnecessary materialization when running on Beam or that the order of the iteration is not favorable to Dataflow as discovered by Scio.

The Scio project offers a very similar user facing API as Scalding and was used to convert Scalding jobs to run on Beam with little modifications. The convention in Scalding and Scio for joins assumes the larger collection or collection with more values for per key should be on the left `big.join(small)`. For Beam and Dataflow the convention seems to be the reverse as explored in: https://github.com/spotify/scio/issues/289. Following the same approach we could reverse the order in which we do the joins.

We could do a custom implementation of MultiJoinFunction specific for the Beam backend. The one in Scalding ([here](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/MultiJoinFunction.scala) and [here](https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/Joiner.scala)) currently depends on Iterable (which can be turned to Iterator) on one side and Iterator on the other (which can only be iterated once). When implementing the joins in Beam we have Iterables for the values coming from all inputs and could be using Iterables for everything in MultiJoinFunction and Joiner, making the functions symmetrical and making the reversal simpler.

_I'm creating this issue so we can ground follow-up code changes on this discussion._

tlazaro:
To fully map MultiJoinFunction and Joiner to a Beam backend version using only Iterable, we have a problem:

While JoinFunction is a sealed trait, we still accept arbitrary functions from users when creating a MultiJoinFunction:
https://github.com/twitter/scalding/blob/develop/scalding-core/src/main/scala/com/twitter/scalding/typed/Grouped.scala#L121

Also for JoinFunction: FlatMappedJoin, MappedGroupJoin and JoinFromHashJoin take arbitrary functions depending on Iterator.

Issue #1639, opened 2017-02-13
group randomly is not balanced


ianoc-stripe:
Group randomly uses nextInt(numReducers), this means we get huge reducer skew

ulyssepence:
@ianoc-stripe Not sure I totally understand. Are you saying we should just use `nextInt()` to distribute it evenly over the `Integer` space?

ianoc-stripe:
Exactly! -- otherwise we are hashing the reduced integer space and then overload a few keys

johnynek:
I think this is a birthday problem issue. I think we want at least something like `numReducers * numReducers` or something.

ianoc-stripe:
Is there any downside to the full integer space?

johnynek:
well, in the groups, they will have fewer entries. Where, ideally, we could apply a `mapValueStream` to a bigger block of them for some use cases.

I think the ideal would be 1 group per reducer, but doing this randomly creates skew.

ianoc-stripe:
Fair,  not sure how the distribution over reducers would look at n^2 or how common that use case is. If n^2 covers it might aswell use that for the bound

johnynek:
yeah, I'm not too uptight about it, but having fewer groups seems like it *could* be useful. Actually, I can't replicate the skew in a simple repl session:
```scala
scala> def bucket(n: Int, k: Int, recs: Int, seed: Int): Vector[Int] = {
     |   val rng = new java.util.Random(seed)
     |   val m = (0 until recs).iterator.map { _ => rng.nextInt(k) }.toList.groupBy(_ % n).mapValues(_.size)
     |   (0 until n).map(m(_)).toVector
     | }

scala> def spread(vs: Vector[Int]) = vs.max - vs.min
spread: (vs: Vector[Int])Int

scala> spread(bucket(10, 10, 1000, 123))
res1: Int = 47

scala> spread(bucket(10, 10, 10000, 123))
res2: Int = 127

scala> spread(bucket(10, 100, 10000, 123))
res3: Int = 127
```
did I make some silly error?

ianoc-stripe:
Yep -- well your code relies on the fact that integer hash codes on the jvm are the identity.

```
def bucket(n: Int, k: Int, recs: Int, seed: Int): Vector[Int] = {
       val rng = new java.util.Random(seed)
[...]

PR #1959, opened 2021-12-31
scalding-beam custom join functions

tlazaro:
Following the discussion #1958, attempting a custom join implementation for Beam backend.

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
@@ -23,10 +23,7 @@ object BeamJoiner {
       case f: Joiner.FilteredHashJoin[K, v1, v2, r]       => FilteredHashJoin(beamHashJoin(f.jf), f.fn)
       case f: Joiner.MappedHashJoin[K, v1, v2, r, r1]     => MappedHashJoin(beamHashJoin(f.jf), f.fn)
       case f: Joiner.FlatMappedHashJoin[K, v1, v2, r, r1] => FlatMappedHashJoin(beamHashJoin(f.jf), f.fn)
-      case f =>
-        throw new UnsupportedOperationException(
-          s"Beam backend does not support arbitrary Hash Join functions but got $f"
-        )
+      case f                                              => ArbitraryHashJoin(f)
     }
 
   private def beamJoin[K, A, B, C](
@@ -43,10 +40,7 @@ object BeamJoiner {
       case join: Joiner.MappedGroupJoin[K, v1, v2, r, r1] =>
         MappedGroupJoin(beamJoin(join.jf), beamMapGroupJoin(join.fn))
       case join: Joiner.JoinFromHashJoin[K, v1, v2, r] => JoinFromHashJoin(beamHashJoin(join.hj))
-      case join =>
-        throw new UnsupportedOperationException(
-          s"Beam backend does not support arbitrary cogroup functions. Replace cogroup with join, outerJoin, leftJoin or rightJoin"
-        )
+      case join                                        => ArbitraryJoin(join)
     }
 
   def beamMultiJoin[A, B](m: typed.MultiJoinFunction[A, B]): MultiJoinFunction[A, B] =
@@ -212,6 +206,14 @@ object BeamJoiner {
     def apply(k: K, left: V1, right: Iterable[V2]) =
       jf.apply(k, left, right).flatMap(fn)
   }
+  final case class ArbitraryHashJoin[K, V1, V2, R](
+      hj: (K, V1, Iterable[V2]) => Iterator[R]
+  ) extends HashJoinFunction[K, V1, V2, R] {
+    def apply(k: K, left: V1, right: Iterable[V2]): Iterable[R] =
+      new Iterable[R] {
+        def iterator: Iterator[R] = hj.apply(k, left, right)
+      }
+  }
 
   /**
    * As opposed to Scalding's JoinFunction, in Beam we make 'right' be the one iterated once and 'left' many
@@ -268,6 +270,14 @@ object BeamJoiner {
       extends JoinFunction[K, V1, V2, R] {
     def apply(k: K, left: Iterable[V1], right: Iterable[V2]): Iterable[R] =
       left.flatMap(hj(k, _, right))
+  }
+
+  final case class ArbitraryJoin[K, V1, V2, R](fn: (K, Iterator[V1], Iterable[V2]) => Iterator[R])
+      extends JoinFunction[K, V1, V2, R] {
+    def apply(k: K, left: Iterable[V1], right: Iterable[V2]): Iterable[R] =
+      new Iterable[R] {
+        def iterator: Iterator[R] = fn(k, left.iterator, right)
+      }
   }
 
 }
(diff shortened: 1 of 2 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
```diff
@@ -0,0 +1,283 @@
+package com.twitter.scalding.beam_backend
+
+import com.twitter.scalding.serialization.Externalizer
+import com.twitter.scalding.typed.Joiner
+import com.twitter.scalding.typed
+import java.io.Serializable
+
+object BeamJoiner {
+  private def beamMapGroupJoin[K, R, R1](
```
tlazaro, line 9: Questionable, please review.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
```diff
@@ -0,0 +1,283 @@
...
+  private def beamHashJoin[K, V1, V2, R](
+      hj: (K, V1, Iterable[V2]) => Iterator[R]
+  ): HashJoinFunction[K, V1, V2, R] =
+    hj match {
+      case _: Joiner.HashInner[K, v, u]                   => HashInner()
+      case _: Joiner.HashLeft[K, v, u]                    => HashLeft()
+      case f: Joiner.FilteredHashJoin[K, v1, v2, r]       => FilteredHashJoin(beamHashJoin(f.jf), f.fn)
+      case f: Joiner.MappedHashJoin[K, v1, v2, r, r1]     => MappedHashJoin(beamHashJoin(f.jf), f.fn)
+      case f: Joiner.FlatMappedHashJoin[K, v1, v2, r, r1] => FlatMappedHashJoin(beamHashJoin(f.jf), f.fn)
+      case f                                              => ArbitraryHashJoin(f)
```
tlazaro, line 26: Questionable, please review.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
```diff
@@ -0,0 +1,283 @@
...
+      case _: Joiner.RightJoin[K, v1, v2]                => RightJoin()
+      case _: Joiner.OuterJoin[K, v1, v2]                => OuterJoin()
+      case _: Joiner.InnerJoin[K, v1, v2]                => InnerJoin()
+      case join: Joiner.FilteredJoin[K, v1, v2, r]       => FilteredJoin(beamJoin(join.jf), join.fn)
+      case join: Joiner.MappedJoin[K, v1, v2, r, r1]     => MappedJoin(beamJoin(join.jf), join.fn)
+      case join: Joiner.FlatMappedJoin[K, v1, v2, r, r1] => FlatMappedJoin(beamJoin(join.jf), join.fn)
+      case join: Joiner.MappedGroupJoin[K, v1, v2, r, r1] =>
+        MappedGroupJoin(beamJoin(join.jf), beamMapGroupJoin(join.fn))
+      case join: Joiner.JoinFromHashJoin[K, v1, v2, r] => JoinFromHashJoin(beamHashJoin(join.hj))
+      case join                                        => ArbitraryJoin(join)
```
tlazaro, line 43: Questionable, please review.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
```diff
@@ -0,0 +1,283 @@
...
+import com.twitter.scalding.serialization.Externalizer
+import com.twitter.scalding.typed.Joiner
+import com.twitter.scalding.typed
+import java.io.Serializable
+
+object BeamJoiner {
+  private def beamMapGroupJoin[K, R, R1](
+      fn: (K, Iterator[R]) => Iterator[R1]
+  ): (K, Iterable[R]) => Iterable[R1] = { (k: K, iter: Iterable[R]) =>
+    new Iterable[R1] {
```
johnynek, line 12: I think this is right, but the only risk is that it is unclear the default semantics of `Iterable`. Like, I think `map` will wind up materializing everything.

You could also experiment with using `Stream` instead, which will be lazy, but if someone holds the head it will possibly blow up the memory. I'm not sure the constraints you want.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamJoiner.scala
```diff
@@ -0,0 +1,283 @@
...
+      extends JoinFunction[K, V1, V2, R] {
+    def apply(k: K, left: Iterable[V1], right: Iterable[V2]): Iterable[R] =
+      left.flatMap(hj(k, _, right))
+  }
+
+  final case class ArbitraryJoin[K, V1, V2, R](fn: (K, Iterator[V1], Iterable[V2]) => Iterator[R])
+      extends JoinFunction[K, V1, V2, R] {
+    def apply(k: K, left: Iterable[V1], right: Iterable[V2]): Iterable[R] =
+      new Iterable[R] {
+        def iterator: Iterator[R] = fn(k, left.iterator, right)
```
johnynek, line 279: same comment about how the `Iterable` is used. If someone just calls `.iterator` repeatedly, this should be fine.

But if that's the case, returning `fn(k, left.iterator, right).toStream` would also be okay. Stream has the advantage of being lazy on filter, map, flatMap.

johnynek (commented):
It's unfortunate to have this duplication but it looks like the right direction to me.

I added some comments.

Does this solve the performance issue you had with the 7-way join?

tlazaro:
> It's unfortunate to have this duplication but it looks like the right direction to me.
> 
> I added some comments.
> 
> Does this solve the performance issue you had with the 7-way join?

Thanks for the review! Haven't tested yet, maybe next week we can do so. Will try out the iterator/stream.

tlazaro:
@johnynek the massive join succeeded with this change! 

Then the job went on to fail on a `shard(16)` that is totally unnecessary in this case on Dataflow. Not sure if we can and should have a rule on how to map that but I'm leaning towards not sharding at all, maybe just a Reshuffle on Beam but no custom grouping. That's a problem for another PR. In this case I think it's fine to just remove the sharding from the user job.

johnynek:
That's great news! Let's make sure to comment well so people don't wonder why this is done.

btw: note this: 

https://github.com/twitter/scalding/issues/1639

I think that is a real issue that should be addressed... We keep kind of punting it. It could possibly be related to your problem here.

https://github.com/twitter/scalding/blob/de2c7c20d7a36a209b81fefa0408a74ef947833c/scalding-core/src/main/scala/com/twitter/scalding/typed/functions/Functions.scala#L137

I don't see any reason to do `nextInt(modulus)`. I think it is only reducing the possible entropy. I think it is better to do `rng.nextInt()` and get the full entropy, and then use the modulus only at the end.

johnynek:
btw: it looks like if all we do with these iterables is convert them to java iterables then your current code is likely safe:

https://github.com/scala/scala/blob/4152c02cd8b222ee1782929ecbaf73e85bc8a214/src/library/scala/collection/convert/Wrappers.scala#L23

and will not cause any materializations.

nownikhil:
@johnynek This code worked with 7 way join and things look stable. Do you think we can merge this?

johnynek:
Yes, I think this is certainly shape to merge. Merge when you like!

PR #1960, opened 2022-01-26
Use BeamJoiner function in `planMapGroup`

nownikhil:
User job was OOMing at groupBy step and the job finished after making this change.


johnynek (approved):
you might to a grep for `toIterable`. That is going to fully materialize and you would not ever want to do that.

you would always want to use the approach you have here.

PR #1962, opened 2022-01-28
ToIterable execution in Beam runner

nownikhil:
We use the coder from the PCollection and store data as bytes in temp location using FileIO.
We later read this file using the same coder and get the data on the submitter node.
TESTS: Added unit tests + Tested with Dataflow Runner

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
@@ -3,14 +3,13 @@ package com.twitter.scalding.beam_backend
 import com.twitter.scalding.Execution.Writer
 import com.twitter.scalding.typed.{Resolver, TypedSink, TypedSource}
 import com.twitter.scalding.{Config, Mode, TextLine}
-import java.io.InputStream
+import java.io.{EOFException, InputStream}
 import java.nio.channels.{Channels, WritableByteChannel}
 import org.apache.beam.sdk.Pipeline
 import org.apache.beam.sdk.coders.Coder
 import org.apache.beam.sdk.io.{FileIO, TextIO}
 import org.apache.beam.sdk.options.PipelineOptions
 import org.apache.beam.sdk.values.PCollection
-import scala.util.{Failure, Success, Try}
 
 case class BeamMode(
     pipelineOptions: PipelineOptions,
@@ -54,7 +53,7 @@ object BeamSource extends Serializable {
 }
 
 trait BeamSink[-A] extends Serializable {
-  def write(pipeline: Pipeline, config: Config, pc: PCollection[_ <: A]): Unit
+  def write(pc: PCollection[_ <: A], config: Config): Unit
 }
 
 object BeamSink extends Serializable {
@@ -74,16 +73,15 @@ object BeamSink extends Serializable {
 
   def textLine(path: String): BeamSink[String] =
     new BeamSink[String] {
-      override def write(pipeline: Pipeline, config: Config, pc: PCollection[_ <: String]): Unit =
+      override def write(pc: PCollection[_ <: String], config: Config): Unit =
         pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
     }
 }
 
 class BeamFileIO[T](output: String) extends BeamSink[T] {
   override def write(
-      pipeline: Pipeline,
-      config: Config,
-      pc: PCollection[_ <: T]
+      pc: PCollection[_ <: T],
+      config: Config
   ): Unit = {
     val pColT: PCollection[T] = BeamFunctions.widenPCollection(pc)
 
@@ -92,7 +90,6 @@ class BeamFileIO[T](output: String) extends BeamSink[T] {
         .write()
         .via(new CoderFileSink(pColT.getCoder))
         .to(output)
-        .withNumShards(1)
     )
   }
 }
@@ -121,11 +118,11 @@ class InputStreamIterator[T](stream: InputStream, coder: Coder[T]) extends Itera
   }
 
   private def fetchNext(): Unit =
-    Try(coder.decode(stream)) match {
-      case Success(value) =>
-        nextRecord = value
-        hasNextRecord = true
-      case Failure(_) =>
+    try {
+      nextRecord = coder.decode(stream)
+      hasNextRecord = true
+    } catch {
+      case _: EOFException =>
         hasNextRecord = false
     }
 }
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
@@ -3,10 +3,14 @@ package com.twitter.scalding.beam_backend
 import com.twitter.scalding.Execution.Writer
 import com.twitter.scalding.typed.{Resolver, TypedSink, TypedSource}
 import com.twitter.scalding.{Config, Mode, TextLine}
+import java.io.InputStream
+import java.nio.channels.{Channels, WritableByteChannel}
 import org.apache.beam.sdk.Pipeline
-import org.apache.beam.sdk.io.TextIO
+import org.apache.beam.sdk.coders.Coder
+import org.apache.beam.sdk.io.{FileIO, TextIO}
 import org.apache.beam.sdk.options.PipelineOptions
 import org.apache.beam.sdk.values.PCollection
+import scala.util.{Failure, Success, Try}
 
 case class BeamMode(
     pipelineOptions: PipelineOptions,
@@ -72,5 +76,56 @@ object BeamSink extends Serializable {
     new BeamSink[String] {
       override def write(pipeline: Pipeline, config: Config, pc: PCollection[_ <: String]): Unit =
         pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
+    }
+}
+
+class BeamFileIO[T](output: String) extends BeamSink[T] {
+  override def write(
+      pipeline: Pipeline,
+      config: Config,
+      pc: PCollection[_ <: T]
+  ): Unit = {
+    val pColT: PCollection[T] = BeamFunctions.widenPCollection(pc)
+
+    pColT.apply(
+      FileIO
+        .write()
+        .via(new CoderFileSink(pColT.getCoder))
+        .to(output)
+        .withNumShards(1)
+    )
+  }
+}
+
+class CoderFileSink[T](coder: Coder[T]) extends FileIO.Sink[T] {
+  private var outputStream: java.io.OutputStream = _
+
+  override def open(channel: WritableByteChannel): Unit =
+    outputStream = Channels.newOutputStream(channel)
+
+  override def write(element: T): Unit = coder.encode(element, outputStream)
+  override def flush(): Unit = outputStream.flush()
+}
+
+class InputStreamIterator[T](stream: InputStream, coder: Coder[T]) extends Iterator[T] {
+  var hasNextRecord: Boolean = _
+  var nextRecord: T = _
+
+  fetchNext()
+  override def hasNext: Boolean = hasNextRecord
+
+  override def next(): T = {
+    val recordToReturn = nextRecord
+    fetchNext()
+    recordToReturn
+  }
+
+  private def fetchNext(): Unit =
+    Try(coder.decode(stream)) match {
+      case Success(value) =>
+        nextRecord = value
+        hasNextRecord = true
+      case Failure(_) =>
+        hasNextRecord = false
     }
 }
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
@@ -73,8 +73,10 @@ object BeamSink extends Serializable {
 
   def textLine(path: String): BeamSink[String] =
     new BeamSink[String] {
-      override def write(pc: PCollection[_ <: String], config: Config): Unit =
-        pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
+      override def write(pc: PCollection[_ <: String], config: Config): Unit = {
+        val stringPCollection: PCollection[String] = BeamFunctions.widenPCollection(pc)
+        stringPCollection.apply(TextIO.write().to(path))
+      }
     }
 }
 
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
@@ -4,6 +4,7 @@ import com.stripe.dagon.Rule
 import com.twitter.scalding.Execution.{ToWrite, Writer}
 import com.twitter.scalding.typed._
 import com.twitter.scalding.{CFuture, CancellationHandler, Config, Execution, ExecutionCounters}
+import java.io.InputStream
 import java.nio.channels.Channels
 import java.util.concurrent.atomic.AtomicLong
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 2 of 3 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -74,3 +78,54 @@ object BeamSink extends Serializable {
...
+  private var outputStream: java.io.OutputStream = _
+
+  override def open(channel: WritableByteChannel): Unit =
+    outputStream = Channels.newOutputStream(channel)
+
+  override def write(element: T): Unit = coder.encode(element, outputStream)
+  override def flush(): Unit = outputStream.flush()
+}
+
+class InputStreamIterator[T](stream: InputStream, coder: Coder[T]) extends Iterator[T] {
```
nownikhil, line 109: Maybe a cleaner way to do this?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -74,3 +78,54 @@ object BeamSink extends Serializable {
...
+      pc: PCollection[_ <: T]
+  ): Unit = {
+    val pColT: PCollection[T] = BeamFunctions.widenPCollection(pc)
+
+    pColT.apply(
+      FileIO
+        .write()
+        .via(new CoderFileSink(pColT.getCoder))
+        .to(output)
+        .withNumShards(1)
```
johnynek: seems like `1` should be a constructor param.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -74,3 +78,54 @@ object BeamSink extends Serializable {
...
+}
+
+class CoderFileSink[T](coder: Coder[T]) extends FileIO.Sink[T] {
+  private var outputStream: java.io.OutputStream = _
+
+  override def open(channel: WritableByteChannel): Unit =
+    outputStream = Channels.newOutputStream(channel)
+
+  override def write(element: T): Unit = coder.encode(element, outputStream)
+  override def flush(): Unit = outputStream.flush()
```
johnynek, line 106: is there no `close` in `FileIO.Sink`? Seems like we should have a way to close outputStream.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -74,3 +78,54 @@ object BeamSink extends Serializable {
...
+  override def hasNext: Boolean = hasNextRecord
+
+  override def next(): T = {
+    val recordToReturn = nextRecord
+    fetchNext()
+    recordToReturn
+  }
+
+  private def fetchNext(): Unit =
+    Try(coder.decode(stream)) match {
```
johnynek: this worries me. What if you see an error that is not EOF?

Also, boxing with Try is an extra cost you don't need.
```scala
try {
  nextRecord = coder.decode(stream)
  hasNextRecord = true
}
catch {
  case _: EOFException =>
    hasNextRecord = false
  // don't catch other errors.
}
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -4,14 +4,21 @@ import com.stripe.dagon.Rule
...
+import org.apache.beam.sdk.coders.Coder
+import org.apache.beam.sdk.io.FileSystems
 import scala.annotation.tailrec
 import scala.concurrent.{ExecutionContext, Future}
 
 class BeamWriter(val beamMode: BeamMode) extends Writer {
   private val state = new AtomicLong()
 
+  private val sourceCounter: AtomicLong = new AtomicLong(0L)
+  private val iterablePipes: scala.collection.mutable.Map[TypedPipe[_], (Coder[_], String)] =
```
johnynek: I think this needs to be thread-safe, so it may be best to use `ConcurrentHashMap` from java or use a mutex somehow.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -22,7 +29,27 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
...
+          override def iterator: Iterator[T] = {
+            // Single dir by default just matches the dir, we need to match files inside
+            val matchedResources = FileSystems.`match`(s"$path*").metadata()
+            matchedResources.size() match {
+              case 0 => Iterator.empty
+              case 1 =>
+                val resource = matchedResources.get(0).resourceId()
+                val inputStream = Channels.newInputStream(FileSystems.open(resource))
+                new InputStreamIterator(inputStream, c)
+              // We enforce num shards = 1, so exactly one file should exist
```
johnynek: why do we enforce this? we could easily concatenate the Iterators right?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -22,7 +29,27 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
...
       cec: ExecutionContext
-  ): Future[Iterable[T]] = ???
+  ): Future[Iterable[T]] =
+    iterablePipes.get(initial) match {
+      case Some((coder, path)) =>
+        val c: Coder[T] = coder.asInstanceOf[Coder[T]]
+        Future(new Iterable[T] {
+          override def iterator: Iterator[T] = {
+            // Single dir by default just matches the dir, we need to match files inside
+            val matchedResources = FileSystems.`match`(s"$path*").metadata()
```
johnynek: seems like we should move this up to me:
```scala
Future {
  // Single dir by default just matches the dir, we need to match files inside
  val matchedResources = FileSystems.`match`(s"$path*").metadata()
  
  new Iterable[T] {
    def iterator = matchedResources.asScala.iterator.flatMap { m =>
      new InputStreamIterator(
        Channels.newInputStream(FileSystems.open(m.resourceId())),
        c)
    }
}
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -48,7 +75,16 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
               }
               rec(xs)
             }
-            //TODO: handle Force and ToIterable
+            case OptimizedWrite(pipe, ToWrite.ToIterable(opt)) =>
+              val pcoll = planner(opt).run(pipeline)
+              val tempLocation = pcoll.getPipeline.getOptions.getTempLocation
+              assert(tempLocation != null, "Temp location cannot be null when using toIterableExecution")
```
johnynek: this should probably be a `require`. An assertion is something that *must* be true by construction, not a requirement.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -48,7 +75,16 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
               }
               rec(xs)
             }
-            //TODO: handle Force and ToIterable
+            case OptimizedWrite(pipe, ToWrite.ToIterable(opt)) =>
+              val pcoll = planner(opt).run(pipeline)
```
johnynek, line 88: just to be clear: this isn't blocking right? this adds it to a pipeline which is run later below right?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -74,3 +78,54 @@ object BeamSink extends Serializable {
         pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
     }
 }
+
+class BeamFileIO[T](output: String) extends BeamSink[T] {
+  override def write(
+      pipeline: Pipeline,
```
johnynek: you never use the `Pipeline` right? That is just because you are reusing the BeamSink interface? Does it need the pipeline?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -48,7 +75,16 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
               }
```
johnynek, line 84: just to be clear: the above isn't blocking when it calls `run` or `write` is it? It kind of looks like it. I think we need a comment.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -48,7 +75,16 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
...
               rec(xs)
             }
-            //TODO: handle Force and ToIterable
+            case OptimizedWrite(pipe, ToWrite.ToIterable(opt)) =>
+              val pcoll = planner(opt).run(pipeline)
+              val tempLocation = pcoll.getPipeline.getOptions.getTempLocation
+              assert(tempLocation != null, "Temp location cannot be null when using toIterableExecution")
+
+              val outputPath = BeamWriter.addPaths(tempLocation, sourceCounter.getAndIncrement().toString)
+              new BeamFileIO(outputPath).write(pipeline, conf, pcoll)
```
johnynek: this looks like it is blocking here. We don't want that. We want to run this concurrently with all the other operations I think. Can you add a comment?

I think the loop should build a set of actions that need to run before the future that returns the run ID an execution counters.

johnynek (commented):
Also, we have no idea if the tests pass since the CI is disabled. I think we really need to get that working. It is very dangerous to merge PRs without a working CI.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
nownikhil: It is handled by the underlying FileIO which uses FileIO.Sink

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
nownikhil: Done

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
nownikhil: It only builds up on the pipeline object which runs in the end.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
nownikhil: Refactored the interface. Pipeline can be accessed from PCollection via `pcol.getPipeline`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
nownikhil: Added comment

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -74,16 +73,15 @@ object BeamSink extends Serializable {
 
   def textLine(path: String): BeamSink[String] =
     new BeamSink[String] {
-      override def write(pipeline: Pipeline, config: Config, pc: PCollection[_ <: String]): Unit =
+      override def write(pc: PCollection[_ <: String], config: Config): Unit =
         pc.asInstanceOf[PCollection[String]].apply(TextIO.write().to(path))
```
johnynek: nit: can we use our safer `widen` function here (general cast is more dangerous).

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -34,21 +36,17 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
...
-              case 0 => Iterator.empty
-              case 1 =>
-                val resource = matchedResources.get(0).resourceId()
-                val inputStream = Channels.newInputStream(FileSystems.open(resource))
+          // Single dir by default just matches the dir, we need to match files inside
+          val matchedResources = FileSystems.`match`(s"$path*").metadata().asScala
+
+          override def iterator: Iterator[T] =
+            matchedResources
+              .map { resource =>
```
johnynek: nit: I think `matchedResources.iterator.flatMap { ` clearer than manually reimplementing via fold.

Now that I think about this though, we are never closing the input stream.

I think we may need something like this:

```scala
def iterator = {
  // an empty Iterator that closes an InputStream when it is iterated
  def closeIt(is: InputStream): Iterator[T] =
    new Iterator[T] {
      def hasNext = {
        is.close()
        false
      }
      def next = Iterator.empty.next
    }

  matchedResources
    .iterator
    .flatMap { resource =>
      val is = Channels.newInputStream(FileSystems.open(resource.resourceId()))
      (new InputStreamIterator(is, c)) ++ closeIt(is)
    }
}
```
At least that would close the resource when you iterate through all the items.

johnynek (commented):
one concern about closing the InputStreams, but other than that I think it looks good.

That said, I will reiterate my concern about not having CI and having no idea if the tests pass or if this even compiles.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -43,12 +68,23 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
...
+              val tempLocation = pcoll.getPipeline.getOptions.getTempLocation
+              require(tempLocation != null, "Temp location cannot be null when using toIterableExecution")
+
+              val outputPath = BeamWriter.addPaths(tempLocation, sourceCounter.getAndIncrement().toString)
+              // Here we add a sink transformation on the PCollection.
+              // This does not run till the final `pipeline.run` step
+              new BeamFileIO(outputPath).write(pcoll, conf)
+              iterablePipes += ((pipe, (pcoll.getCoder, outputPath)))
+
+            //TODO: handle Force
```
johnynek, line 98: force is almost the same as toIterable, you just materialize it and then make a PCollection to read it in the next pipeline, but you are smart to leave that to a follow up PR.

nownikhil:
Someone from Scalding team wants to work on setting up CI for scalding repo. I will hold on to this patch till the CI is set up and apply it internally in source to test on real dataflow jobs. Once the jobs on dataflow succeed and CI is setup, I will merge this patch.

Tests are succeeding on my local. But this is not urgent to merge right now so better to wait here.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
nownikhil: Done

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
nownikhil: Yup this will be done in a follow up patch

johnynek (approved):
sounds good!

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -388,6 +389,29 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+    assert(iter.toSet == input.toSet)
+  }
+
+  test("toIterableExecutionWithJoin") {
+    val bmode = BeamMode.default(pipelineOptions)
+    val tp1 = TypedPipe.from(1 to 10).map(x => (x, 1))
+    val tp2 = TypedPipe.from(1 to 10).map(x => (x, 2))
+    val output = tp1
+      .join(tp2)
+      .map(x => (x._1, x._2._1 + x._2._2))
```
johnynek: nit: `.mapValues { case (left, right) => left + right }` would be a bit more idiomatic (always prefer operating only on values if you can since you open up more opportunities for dag optimization.

johnynek:
please merge master after #1965 is merged because scalding-beam was currently not being tested.

johnynek:
(although master does pass for me, I assume this PR passes also).

PR #1967, opened 2022-03-07
Implement ForceToDisk for Beam runner

nownikhil:
We are defining a `TempSource` which captures output path and coder. We then add a resolver from `TempSource` to corresponding `BeamSource`.
Added unit test for `forceToDiskExecution`.

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
@@ -19,7 +19,6 @@ case class BeamMode(
     sources: Resolver[TypedSource, BeamSource],
     sink: Resolver[TypedSink, BeamSink]
 ) extends Mode {
-
   def newWriter(): Writer = new BeamWriter(this)
 }
 
@@ -116,6 +115,7 @@ class BeamTempFileSource[T](coder: Coder[T], output: String) extends BeamSource[
 case class TempSourceDoFn[T](coder: Coder[T]) extends DoFn[MatchResult.Metadata, T] {
   @ProcessElement
   def processElement(c: DoFn[MatchResult.Metadata, T]#ProcessContext): Unit = {
+    // We do not split the files produced in the previous stage and use a single thread per file
     val stream = Channels.newInputStream(FileSystems.open(c.element().resourceId()))
     val it = InputStreamIterator.closingIterator(stream, coder)
     while (it.hasNext) c.output(it.next())
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
@@ -7,8 +7,11 @@ import java.io.{EOFException, InputStream}
 import java.nio.channels.{Channels, WritableByteChannel}
 import org.apache.beam.sdk.Pipeline
 import org.apache.beam.sdk.coders.Coder
-import org.apache.beam.sdk.io.{FileIO, TextIO}
+import org.apache.beam.sdk.io.fs.MatchResult
+import org.apache.beam.sdk.io.{FileIO, FileSystems, TextIO}
 import org.apache.beam.sdk.options.PipelineOptions
+import org.apache.beam.sdk.transforms.DoFn.ProcessElement
+import org.apache.beam.sdk.transforms.{Create, DoFn, ParDo}
 import org.apache.beam.sdk.values.PCollection
 
 case class BeamMode(
@@ -16,6 +19,7 @@ case class BeamMode(
     sources: Resolver[TypedSource, BeamSource],
     sink: Resolver[TypedSink, BeamSink]
 ) extends Mode {
+
   def newWriter(): Writer = new BeamWriter(this)
 }
 
@@ -40,7 +44,8 @@ object BeamSource extends Serializable {
               case path :: Nil => Some(textLine(path))
               case _           => throw new Exception("Can not accept multiple paths to BeamSource")
             }
-          case _ => None
+          case TempSource(path, coder) => Some(new BeamTempFileSource(coder, path))
+          case _                       => None
         }
     }
   }
@@ -80,7 +85,7 @@ object BeamSink extends Serializable {
     }
 }
 
-class BeamFileIO[T](output: String) extends BeamSink[T] {
+class BeamTempFileSink[T](output: String) extends BeamSink[T] {
   override def write(
       pc: PCollection[_ <: T],
       config: Config
@@ -93,6 +98,27 @@ class BeamFileIO[T](output: String) extends BeamSink[T] {
         .via(new CoderFileSink(pColT.getCoder))
         .to(output)
     )
+  }
+}
+
+class BeamTempFileSource[T](coder: Coder[T], output: String) extends BeamSource[T] {
+  override def read(
+      pipeline: Pipeline,
+      config: Config
+  ): PCollection[_ <: T] =
+    pipeline
+      .apply(Create.of(s"$output/*"))
+      .apply(FileIO.matchAll())
+      .apply(ParDo.of(new TempSourceDoFn[T](coder)))
+      .setCoder(coder)
+}
+
+case class TempSourceDoFn[T](coder: Coder[T]) extends DoFn[MatchResult.Metadata, T] {
+  @ProcessElement
+  def processElement(c: DoFn[MatchResult.Metadata, T]#ProcessContext): Unit = {
+    val stream = Channels.newInputStream(FileSystems.open(c.element().resourceId()))
+    val it = InputStreamIterator.closingIterator(stream, coder)
+    while (it.hasNext) c.output(it.next())
   }
 }
 
@@ -127,4 +153,19 @@ class InputStreamIterator[T](stream: InputStream, coder: Coder[T]) extends Itera
       case _: EOFException =>
         hasNextRecord = false
     }
+}
+
+object InputStreamIterator {
+  // an empty Iterator that closes an InputStream when it is iterated
+  def closingIterator[T](stream: InputStream, coder: Coder[T]): Iterator[T] = {
+    def closeIt(is: InputStream): Iterator[T] =
+      new Iterator[T] {
+        def hasNext: Boolean = {
+          is.close()
+          false
+        }
+        def next = Iterator.empty.next
+      }
+    new InputStreamIterator[T](stream, coder) ++ closeIt(stream)
+  }
 }
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
@@ -14,6 +14,7 @@ import org.apache.beam.sdk.io.FileSystems
 import scala.annotation.tailrec
 import scala.collection.convert.decorateAsScala._
 import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.blocking
 import scala.collection.JavaConversions._
 
 case class TempSource[A](path: String, coder: Coder[A]) extends TypedSource[A] {
@@ -56,17 +57,20 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
     tempSources.get(initial) match {
       case Some(TempSource(path, coder)) =>
         val c: Coder[T] = coder.asInstanceOf[Coder[T]]
+        Future {
+          blocking {
+            // Single dir by default just matches the dir, we need to match files inside
+            val matchedResources = FileSystems.`match`(s"$path*").metadata().asScala
+            val records = matchedResources.iterator.flatMap { resource =>
+              val is = Channels.newInputStream(FileSystems.open(resource.resourceId()))
+              InputStreamIterator.closingIterator(is, c)
+            }.toList
 
-        // Single dir by default just matches the dir, we need to match files inside
-        val matchedResources = FileSystems.`match`(s"$path*").metadata().asScala
-        val records = matchedResources.iterator.flatMap { resource =>
-          val is = Channels.newInputStream(FileSystems.open(resource.resourceId()))
-          InputStreamIterator.closingIterator(is, c)
-        }.toList
-
-        Future(new Iterable[T] {
-          override def iterator: Iterator[T] = records.toIterator
-        })
+            new Iterable[T] {
+              override def iterator: Iterator[T] = records.toIterator
+            }
+          }
+        }
       case _ => Future.failed(new IllegalStateException(s"TypedPipe = $initial has no existing Iterable"))
     }
 
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
@@ -47,6 +47,7 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
   ): Future[TypedPipe[T]] =
     tempSources.get(initial) match {
       case Some(source) => Future.successful(TypedPipe.from(source).asInstanceOf[TypedPipe[T]])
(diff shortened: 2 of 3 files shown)


scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -389,6 +389,31 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
 
+  test("Force to Disk execution") {
+    val bmode = BeamMode.default(pipelineOptions)
+    val tmpPath1 = tmpPath("tp1")
+    val tmpPath2 = tmpPath("tp2")
+
+    val forcedExecution =
+      TypedPipe
+        .from(Seq(5, 3, 2, 6, 1, 4))
+        .forceToDiskExecution
```
nownikhil, line 434: Tested by adding println() statements and they got printed only once. Whats a good way to add a test which checks that logic is not computed twice?
I tried by defining AtomicLong() and incrementing it in a map function but that didn't work

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -115,6 +114,11 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
 }
 
 object BeamWriter {
+  // We create BeamWriter per Execution. To share forcedPipes across executions we are putting
```
nownikhil: Can it have some kind of unexpected behavior?

nownikhil:
@johnynek Any suggestions on how to implement forceToDisk?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamMode.scala
```diff
@@ -96,6 +101,27 @@ class BeamFileIO[T](output: String) extends BeamSink[T] {
...
+    pipeline
+      .apply(Create.of(s"$output/*"))
+      .apply(FileIO.matchAll())
+      .apply(ParDo.of(new TempSourceDoFn[T](coder)))
+      .setCoder(coder)
+}
+
+case class TempSourceDoFn[T](coder: Coder[T]) extends DoFn[MatchResult.Metadata, T] {
+  @ProcessElement
+  def processElement(c: DoFn[MatchResult.Metadata, T]#ProcessContext): Unit = {
```
johnynek, line 117: can you add a comment to the code here? It looks to me like everything in the `Iterator` is going to be read in a single thread.

I can't really see how that's not the case? That seems like it is going to really hard performance since you don't want to kill parallelism when using forceToDiskExecution.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
johnynek: yes. Why did you move it out of `BeamWriter`? The idea is all mutable state lives in the `Writer` for the scope of an execution.

We do not want to have these survive past a Writer lifetime.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -1,62 +1,59 @@
...
 class BeamWriter(val beamMode: BeamMode) extends Writer {
   private val state = new AtomicLong()
 
   private val sourceCounter: AtomicLong = new AtomicLong(0L)
-  private val iterablePipes: scala.collection.concurrent.Map[TypedPipe[_], (Coder[_], String)] =
-    new ConcurrentHashMap[TypedPipe[_], (Coder[_], String)]().asScala
 
   override def start(): Unit = ()
 
   override def finished(): Unit = ()
```
johnynek: I think you need to iterate through the forcedPipes here and delete them all. If we don't do this, you will probably have problems leaking space.

We had many issues like that in the past.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -84,19 +81,21 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
               }
               rec(xs)
             }
-            case OptimizedWrite(pipe, ToWrite.ToIterable(opt)) =>
+            case OptimizedWrite(pipe, toWrite) if !tempSources.contains(pipe) =>
```
johnynek: can we change this to:

```scala
case OptimizedWrite(pipe, toWrite @ (ToIterable(_) | Force(_)) if !tempSources.contains(pipe) =>
  // now we know matching below is total
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -84,19 +81,21 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
...
               // Here we add a sink transformation on the PCollection.
               // This does not run till the final `pipeline.run` step
-              new BeamFileIO(outputPath).write(pcoll, conf)
-              iterablePipes += ((pipe, (pcoll.getCoder, outputPath)))
-
-            //TODO: handle Force
-            case _ => ???
+              new BeamTempFileSink(outputPath).write(pcoll, conf)
+              tempSources += ((pipe, TempSource(outputPath, pcoll.getCoder)))
+            case _ => ()
```
johnynek: can we add a total match and comment:

```scala
case OptimizedWrite(_, ToIterable(_) | Force(_)) =>
  // we know that tempSources.contains(pipe) on this branch, which means it was already computed.
```

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
johnynek: this is a really good question.

I think atomic long isn't going to work because that assumes that there is no serialization of the job, which I think there likely is.

What might work, but will be ugly, would be to make a `TypedPipe.from(List(1)).map { x => /* create a specific file, or if it already exists fail */ }`

So, allocate a single random path at the start of the test, make sure the map function creates that file, then at the end of the test remove the path.

That would be my suggestion:

https://docs.oracle.com/javase/6/docs/api/java/io/File.html#createNewFile()

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -389,6 +389,31 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+
+    val tp1 = forcedExecution.flatMap { forced =>
+      forced.filter(_ % 2 == 0).map(_.toString).writeExecution(TextLine(tmpPath1))
+    }
+
+    val tp2 = forcedExecution.flatMap { forced =>
+      forced.filter(_ % 2 == 1).map(_.toString).writeExecution(TextLine(tmpPath2))
+    }
+
+    Execution.zip(tp1, tp2).waitFor(Config.empty, bmode)
```
johnynek: would be good to test somehow that the forceToDisk data has been cleaned by this point.

johnynek (commented):
also, it looks like this broke an existing test.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
nownikhil: When I tested it out I was not able to make it work.
In the run method of `Execution` we create a new writer. Now lets say I do
`val tp1 = TypedPipe(..).forceToDiskExecution`
`val tp2 = tp1.flatmap(..)` 

How do we share forcedPipe info across executions?

johnynek: note:

https://github.com/twitter/scalding/blob/1312b22b389b6ae226be6e7ab9a3af44bac2fde5/scalding-core/src/main/scala/com/twitter/scalding/Execution.scala#L151

we create a single `Writer` for the life of the final `Execution` evaluation. So, that `run` method is only called by `waitFor` in your code.

We do not want to share state across two `waitFor` calls. We actually should have tests that we *don't* do that.

If you want to share, you put it inside one `Execution` that's the whole point of `Execution`: creating an environment to model sequential and shared state. But after it is done, the only effects you should see are the things you finally wrote out and the result value of the `Execution`.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -406,6 +449,7 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
       .mapValues { case (left, right) => left + right }
       .filter(_._1 % 5 == 0)
       .toIterableExecution
+      .flatMap(iter => Execution.from(iter.toArray))
```
nownikhil: After adding logic of deletion, this started to fail with files not found. This is because iterator is reading the output files as it is. For what I know is that we don't use `toIterableExecution` as a final step, we always map the iterator to do another computation.

johnynek: oh... okay. I think the problem is that we are closing over a temporary file. So, a final `toIterableExecution` can't be lazy, it has to go ahead and materialize if we are going to clean the iterator outputs.

Do you think that makes sense?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -1,62 +1,70 @@
...
 
   def getIterable[T](conf: Config, initial: TypedPipe[T])(implicit
       cec: ExecutionContext
   ): Future[Iterable[T]] =
-    iterablePipes.get(initial) match {
-      case Some((coder, path)) =>
+    tempSources.get(initial) match {
+      case Some(TempSource(path, coder)) =>
         val c: Coder[T] = coder.asInstanceOf[Coder[T]]
         Future(new Iterable[T] {
```
johnynek: I think this isn't safe after the end of the `Execution`: if we `.waitFor` the Execution, then we have deleted all the files, so then this match won't find anything.

I think we have two possible approaches:

1. materialize the Iterator now into a Vector or List (Vector is more memory efficient, but not as good for recursive algorithms).
2. remove this path from the set of files to clean and just assume that some other tmp dir cleaning process will clean things up.

I don't see a good third way. We don't have a great way to see that an Iterable does not escape.

That said, I'm pretty sure cascading materializes the Iterable into memory in this case, so I don't think it is a regression.

What do you think?

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
```diff
@@ -389,12 +389,55 @@ class BeamBackendTests extends FunSuite with BeforeAndAfter {
...
+    val tmpPath2 = tmpPath("tp2")
+
+    val tmpDir = new File(tmpPath("forced"))
+    tmpDir.mkdirs()
+
+    val forcedExecution =
+      TypedPipe
+        .from(Seq(5, 3, 2, 6, 1, 4))
+        .map { e =>
+          // This tests that forcedExecution runs only once. If it runs twice createFile will return false
```
johnynek: did you verify that this test will fail if you remove the code that prevents rerunning?

A wise man once told me: you haven't seen a test pass until you have seen it fail.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
nownikhil: Yup that makes sense.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
nownikhil: We can also add logic that we materialize iterable sources on deletion but I think it will complicate things. Mostly we expect data to be small when bringing in on submitter so I think just materializing it is the safest option.

scalding-beam/src/test/scala/com/twitter/scalding/beam_backend/BeamBackendTests.scala
nownikhil: Haha good one.
Yeah I tested it without forceToDisk and it failed. I can add a test which fails.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -1,62 +1,70 @@
...
+      .flatMap(path => FileSystems.`match`(path).metadata().map(_.resourceId()))
+    FileSystems.delete(resources.toList)
+  }
 
   def getForced[T](conf: Config, initial: TypedPipe[T])(implicit
       cec: ExecutionContext
-  ): Future[TypedPipe[T]] = ???
+  ): Future[TypedPipe[T]] =
+    tempSources.get(initial) match {
+      case Some(source) => Future.successful(TypedPipe.from(source).asInstanceOf[TypedPipe[T]])
```
johnynek, line 50: actually, can we add the other case here:

`case None => Future.failed(new IllegalStateException(s"TypedPipe = $inital, has not yet been forced")`

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -1,62 +1,70 @@
...
-
+          override def iterator: Iterator[T] =
             matchedResources.iterator
               .flatMap { resource =>
                 val is = Channels.newInputStream(FileSystems.open(resource.resourceId()))
-                new InputStreamIterator(is, c) ++ closeIt(is)
+                InputStreamIterator.closingIterator(is, c)
               }
-          }
         })
```
johnynek: can we add a total match:
```scala
case _ => Future.failed(new IllegalStateException(s"TypedPipe = $initial has no existing Iterable"))
```

johnynek (changes requested):

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamWriter.scala
```diff
@@ -55,17 +56,18 @@ class BeamWriter(val beamMode: BeamMode) extends Writer {
     tempSources.get(initial) match {
       case Some(TempSource(path, coder)) =>
         val c: Coder[T] = coder.asInstanceOf[Coder[T]]
+
+        // Single dir by default just matches the dir, we need to match files inside
+        val matchedResources = FileSystems.`match`(s"$path*").metadata().asScala
+        val records = matchedResources.iterator.flatMap { resource =>
+          val is = Channels.newInputStream(FileSystems.open(resource.resourceId()))
+          InputStreamIterator.closingIterator(is, c)
+        }.toList
```
johnynek: this needs to be inside the Future or it will synchronously block other work.

```scala
import scala.concurrent.blocking

Future {
  blocking {
    // make the list here and the blocking construct will not block one of the CPU thread pools.

    ...
  }
}
```
see the discussion:

https://docs.scala-lang.org/overviews/core/futures.html

johnynek (changes requested):
one more minor change.

johnynek (approved):
Looks great!

Would be really nice to make a readme to show people how to use existing Executions or Jobs with this backend.

nownikhil:
Yeah good idea. We have some Twitter internal code for that. Though this is something I wanna revisit. For example in your patch you introduced `CascadingBackend` which adds implicit methods on mode and handles args parsing. Now we can do it for every backend, or a common entry point which uses the right mode based on args provided.

Currently we expect users to extend a custom trait `MyJob extends ExecutionApp with TwitterBeamExecution`, though ideally users should just be able to provide mode as an arg at runtime (which will of course require multiple packages to be available). Not super important at this point though.

Issue #1968, opened 2022-03-09
scalding-jbdc current use / discussion on support/deprecation


hazel-sudz:
I am wondering if there is still actual real use of the `scalding-jbdc` package in other orgs. My concern is that it pulls in a number of cascading ecosystem packages that are no longer supported (cascading-jdbc-core and cascading-jdbc-mysql). 

I see many years ago there was some activity on this: 

https://github.com/twitter/scalding/issues/1385
https://github.com/twitter/scalding/pull/1382

johnynek:
@nownikhil can you comment?

Does Twitter object? IIRC twitter hasn't used that code in some time (I think there is an internal version).

hazel-sudz:
I don't think we have usage at Foursquare @jonshea to confirm.

jonshea:
Confirming that Foursquare does not use `scalding-jbdc`.

hazel-sudz:
resolved by merge

PR #1969, opened 2022-03-10
remove scalding-jbdc

hazel-sudz:
based on https://github.com/twitter/scalding/issues/1968. We are not aware of any popular use of this module. Even if someone is using it it will likely need to be re-written in future scalding releases because: 

1. it is hard-wired to cascading 
2. it depends on cascading ecosystem packages that have not been maintained since cascading 2.X release

I would prefer to reduce support burden on things that do not provide value to most consumers. This is also something that can implemented on the consumer side if it is really required.


hazel-sudz:
I found `com.twitter.scalding.db` and looks like in the past there were plans to move jbdc in there but there was never any incentive to do so after a while.

hazel-sudz:
based on https://github.com/twitter/scalding/pull/1382 twitter used/uses an internal implementation of jbdc for scalding

tlazaro:
I don't remember, if jdbc was used at Twitter. `scalding-db` for sure was. If @germanschnyder or @nownikhil could chime in.

johnynek:
we can revert this if Twitter corrects us later. Let's keep scalding-db for now.

.github/workflows/CI.yml
```diff
@@ -1,13 +1,14 @@
 name: CI
 
-# run on all branches
 on:
   pull_request:
     branches:
       - '**'
+  # it's annoying to have CI double run on PR/push
+  # limit push to only final integration branches like develop
```
germanschnyder, line 8: Agree

nownikhil:
Sorry for the delay, was out for few days. Verified that its not being used at Twitter. Thanks @daniel-sudz

PR #1970, opened 2022-03-10
Use a null flow process when one cannot be found

navinvishy:
The Beam runner for Scalding does not work with Hadoop counters(Stat). When Scalding jobs that use the Stat API are run using the Beam runner, they result in the following error:

`Error in job deployment, the FlowProcess for unique id %s isn't available".format(uniqueId)`

It looks like currently it is not possible for a runner to be able to provide its own implementation of a stat, because the implementation has a dependency on a Cascading FlowProcess. Here we return a `NullFlowProcess` when a flow process cannot be found in the flow mapping store, instead of erroring out. This has the effect of turning the stat call into a noop, since the `NullFlowProcess` does nothing on a call to increment counters.

Ideally, we would be able to plug in a Beam counter for Stat. The change I have here may not be ideal, but the goal is to discuss what could be done here, and to understand if returning a NullFlowProcess could have other unintended consequences.

### DIFF
--- a/scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
+++ b/scalding-core/src/main/scala/com/twitter/scalding/Stats.scala
@@ -169,7 +169,10 @@ object RuntimeStats extends java.io.Serializable {
     } yield {
       flowProcess
     }).getOrElse {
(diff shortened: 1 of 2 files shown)


johnynek:
looks like we had test coverage for this, so you need to fix a test.

johnynek (approved):

PR #1972, opened 2022-03-11
add scalding-base, minimal dep base module

johnynek:
The idea here is to make an *almost completely* source compatible change that moves as many of the main API types into a scalding-base module that does not depend on any compute substrate (e.g. Hadoop).

Most jobs should require no source change, but some library functions or tests may need to do:

```
import com.twitter.scalding.typed.cascading_backend.CascadingExtensions._
```
To get some functions that were moved out of `scalding-base` and off the `Config, TypedPipe, Execution` types.

- [x] move as many of the tests into base as possible
- [x] get the tests passing

cc @daniel-sudz @tlazaro @nownikhil @jonshea


scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -40,6 +38,15 @@ import scala.concurrent.{
 import scala.util.{Failure, Success, Try}
 import scala.util.hashing.MurmurHash3
 
+trait Config {
```
johnynek: this needs to be built up a bit more and there is a duplicated class with the same name in core.

build.sbt
```diff
@@ -323,7 +335,7 @@ lazy val scaldingSpark = module("spark")
       "org.apache.spark" %% "spark-sql" % sparkVersion
     )
   )
-  .dependsOn(scaldingCore)
+  .dependsOn(scaldingBase)
```
johnynek: notice the scalding-spark no longer depends on cascading

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -322,20 +327,6 @@ object Execution {
...
 
-  /**
-   * Distributes the file onto each map/reduce node, so you can use it for Scalding source creation and
-   * TypedPipe, KeyedList, etc. transformations. Using the [[com.twitter.scalding.filecache.CachedFile]]
-   * outside of Execution will probably not work.
-   *
-   * For multiple files you must nested your execution, see docs of
-   * [[com.twitter.scalding.filecache.DistributedCacheFile]]
-   */
-  def withCachedFile[T](path: String)(fn: CachedFile => Execution[T]): Execution[T] =
```
johnynek, line 333: this will almost certainly need to be added back, and I think it can be, at least for cascading.

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -951,11 +942,6 @@ object Execution {
   /** Returns a constant Execution[Unit] */
   val unit: Execution[Unit] = from(())
 
-  /**
-   * This converts a function into an Execution monad. The flowDef returned is never mutated.
-   */
-  def fromFn(fn: (Config, Mode) => FlowDef): Execution[Unit] =
```
johnynek, line 957: this is not easy to add back, and I'd rather not. It's exposing too much of cascading even in cascading mode.

@nownikhil can you check to see if `Execution.fromFn` is used at all at Twitter? There are only a couple calls to it in this repo for some tests, but I think it's a bad idea.

scalding-base/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -217,14 +181,9 @@ object TypedPipe extends Serializable {
   final case class MergedTypedPipe[T](left: TypedPipe[T], right: TypedPipe[T]) extends TypedPipe[T]
   final case class ReduceStepPipe[K, V1, V2](@transient reduce: ReduceStep[K, V1, V2])
       extends TypedPipe[(K, V2)]
-  final case class SourcePipe[T](@transient source: TypedSource[T]) extends TypedPipe[T]
+  final case class SourcePipe[T](@transient source: Input[T]) extends TypedPipe[T]
   final case class SumByLocalKeys[K, V](input: TypedPipe[(K, V)], @transient semigroup: Semigroup[V])
       extends TypedPipe[(K, V)]
-  final case class TrappedPipe[T](
```
johnynek, line 223: @nownikhil can you check to see if anyone is using `.addTrap` on a TypedPipe?

If they are, you should migrate them off. It is super confusing. If you want to handle errors, use:
```scala
foo.map { input =>
  try Right {
 
  }
  catch {
   case NonFatal(e) =>
     Left { ... }
  }
}
```
which is to say, use Either (which kryo can serialize well) to signal if each row errored or not, and then handle accordingly. I doubt I can clearly and correctly explain how trap really works, and I'm pretty sure that's true of pretty much everyone except Chris Wensel who wrote it in cascading.

scalding-base/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
```diff
@@ -257,15 +216,6 @@ object TypedPipe extends Serializable {
     def distinct(implicit ord: Ordering[T]): TypedPipe[T] =
       pipe.asKeys.sum.keys
 
-    /**
-     * If any errors happen below this line, but before a groupBy, write to a TypedSink
-     */
-    @deprecated(
```
johnynek, line 263: it's been deprecated for some time!

johnynek:
I'll get the tests to pass tomorrow.

hazel-sudz (commented):
I'm still trying to understand this but I have a high level question: 

what happens to cascading plugins? For instance: custom cascading schemes ie. cascading-avro etc. I assume these will only be compatible on the cascading-backend now would that be a correct assumption? 

anyways just wanted to say I'm super excited by this :)

johnynek:
@daniel-sudz that's right. Those plugins already depend on cascading, so there isn't much we can do.

I think there are cases here:

1. someone who wants no hadoop or cascading on the classpath at all (it interferes with build, so some other concern). This approach addresses this fully.
2. someone who doesn't care if those things are on the classpath, especially if they aren't used at runtime, the old approach probably did well enough with that, but this approach can work even better.

So, for example, the Beam backend currently maps (using the Resolver type) hadoop writable sources to something Beam can use (maybe Hadoop Writable?). They don't mind having that on the classpath since they aren't running the code and it just becomes a kind of marker class to allow them to do the translation.

So, someone wanting to run on another platform can *choose* to reuse cascading types, but they are not forced to.

This does no magic of somehow making all cascading plugins just run without cascading. Some work on that could potentially be done for common types, but without loading the cascading classes could be tricky.

PS: I think it would actually be pretty fun to implement a scalding-hadoop3 backend that just directly runs on hadoop (and probably a bit faster than going via cascading because we wouldn't have to do the extra wrapping into cascading Tuple.

hazel-sudz (commented):
edit: ++ to scalding-hadoop3 didn't see your message before I wrote this.

I would like some discussion around scala version as well. Modern spark will require `2.12` and in the future maybe even `2.13`. 

Do we keep enforcing a global scala version like we do now? That would entail dropping scala `2.11` support for use case that @jonshea and I need (either spark 3.X or hadoop 3.X).

One alternative: we keep set scala version per module though I think that would be messy real quick with third party dependencies and I don't like it as much. 

Second alternative: we can also write a pure hadoop backend. I think that would be a good compromise for @jonshea and I. From what I see hadoop itself does not have any relation to scala. And we could keep updating our little backend for as long as we have services running on scalding.

We don't care which one of hadoop MR/spark we use. We have an AWS EMR environment which for the most part supports hadoop MR and spark equally well. Our only limitation is that EMR pre-loads a bunch of hadoop libs on the classpath. So we are forced to match their libs exactly by version or else weird internal things start to break with hadoop.

johnynek:
We cross build for 2.11 and 2.12 now. I think all downstreams are on 2.13 too, or should be. I think we can just add 2.13 to the CI.

I don't want to limit versions unless we really need to. I think with a minimum of sbt config we can support 2.11 to 3 (I have done that for a number of projects).

hazel-sudz:
Moving my discussion about hadoop3 to https://github.com/twitter/scalding/issues/1974 to not clobber up this thread.

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
tlazaro: This I know is used at Twitter, although maybe not much. For some things like credentials I made it part of the external tooling to run the Hadoop command with distributed files, so it might be possible to remove it from Scalding if all uses followed that pattern.

scalding-base/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
tlazaro: Pretty sure trapping is discouraged but there might be a few uses. cc @navinvishy @germanschnyder

tlazaro:
This looks great, seems at worst importing `CascadingExtensions` in a few places, maybe adding few `fromPipe` and getting rid of trapping.

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
navinvishy: I see a few uses of it at Twitter, which I suspect are legacy jobs. But more importantly, it's used in Summingbird [here](https://github.com/twitter/summingbird/blob/develop/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala#L610).

scalding-base/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
navinvishy: Yeah there are a few uses of this, and the pattern is always the same:
`source.read.addTrap()`

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: okay, I added a more general mechanism which isn't tied to cascading and implemented fromFn with that.

Also added back the distributed cache support on the CascadingExtensions.

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
```diff
@@ -548,7 +533,7 @@ object Execution {
             .map(v => (v, CancellationHandler.empty)) // map this to the right shape
             .recoverWith {
               val flowStop: PartialFunction[Throwable, Future[Nothing]] = {
-                case t: FlowStopException => // do not recover when the flow was stopped
+                case t: FatalExecutionError => // do not recover when the flow was stopped
```
johnynek, line 534: not tied to cascading, but that exception now extends this so it works the same. A backend can throw the FatalExecutionError and it cannot be recovered.

scalding-base/src/main/scala/com/twitter/scalding/Execution.scala
johnynek: Okay, added it back.

scalding-base/src/main/scala/com/twitter/scalding/serialization/Externalizer.scala
```diff
@@ -15,26 +15,37 @@ limitations under the License.
...
-
-import com.twitter.chill.config.ScalaAnyRefMapConfig
+import com.twitter.chill.{Externalizer => ChillExtern}
+import com.twitter.chill.{KryoInstantiator, ScalaKryoInstantiator}
 
 /**
  * We need to control the Kryo created
  */
-object Externalizer {
+object Externalizer extends Serializable {
```
johnynek: it would be better if we could get base to not depend on chill. It seems very lightly needed.

scalding-base/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala
johnynek: okay, I've re-added it.

johnynek (commented):
Okay, I think this is ready to review.

tlazaro (commented):
FWIW looks good to me although I no longer have a stake.

Having the backends not depending on Cascading is a great move. Would be nice to have 'migration steps', probably just importing cascading backend extensions?

nownikhil:
Trying to understand the final state. Are we planning to have both scalding-base and scalding-core or just scalding-base and backend implementations?

johnynek:
Both. Core depends on base. The idea is the minimal change we can to:

1. Have zero source incompatiblities for almost all users.
2. No change required to build dependencies
3. Allow backends to not be tied to cascading or a particular version of hadoop.

3 is the motivation, 1 and 2 are what makes it safe.

johnynek:
okay,

I think this is ready for review.

I am fairly confident given our rather good test coverage that this is not a breaking change.

It does require some users to add an import, but if they don't they won't get buggy jobs, they will get a failure to compile until they add:

```scala
import com.twitter.scalding.typed.cascading_backend.CascadingExtensions._
```

Note, all those extensions are mixed into Job already, so most users should not see an issues. Our test code saw more issues because many of the functions are called outside of a `Job` subclass.

I think merging this is nice since it enables backends with a much lower dependency graph.

For instance, I think we could rework the beam backend to not depend on cascading at all. Internal to Twitter, it likely would depend on cascading still (since you all have a ton of TypedSource/TypedSink you want to reuse that currently work for both), but you could migrate those at your own pace.

nownikhil:
LGTM! Happy to merge this.

johnynek:
let me merge in @daniel-sudz 's work.

scalding-core/src/main/scala/com/twitter/scalding/typed/cascading_backend/CascadingBackend.scala
```diff
@@ -46,6 +46,37 @@ import org.slf4j.LoggerFactory
...
+        TupleGetter.Casting(),
+        TupleGetter.Casting(),
+        TupleGetter.Casting(),
+        TupleGetter.Casting(),
+        TupleGetter.Casting()))
+      case _ =>
+        // TODO: we should generate all 22 of these functions and the inverses
+        None
+    }).asInstanceOf[Option[TupleConverter[A]]]
+
```
hazel-sudz: is this todo still relevant?

johnynek:
Okay, I think this is ready to merge.

I generated the functions I mentioned in the TODO and merged with develop.

johnynek:
Okay I'm pretty sure the code coverage isn't merging across projects correctly since this PR tests a lot of base in core and note the massive drop in coverage as a result.

johnynek:
I tried to make scalding-beam only depend on scalding-base, but there are two issues:

1. it uses TextLine in the resolver you use for testing. I think this can be resolved by just making a new TextInput class and using that, or making a test depedency on core since I guess internally you have a different Resolver.
2. Beam is sharing a lot of the serialization configuration code with scalding-core, even though that is not cascading related (basically it is chill and kryo code). To solve this, I suggest we make a scalding-chill package or something that depends on chill and kryo, and we would probably reuse that from maybe spark, beam and future tez or hadoop backends.

johnynek:
well, there were no objections here, so merging.

Issue #1973, opened 2022-03-11
Set up automatic publishing


johnynek:
it would be really nice to set up automatic publishing as was done:

https://github.com/twitter/algebird/pull/867

this would enable a much faster cadence of release.

hazel-sudz:
I can port a PR. Feel free to assign this to me. Would need someone to load the secrets into the github repo settings though but that would be a simple step independent of the actual workflow.

johnynek:
this plugin is probably worth adopting:

https://github.com/typelevel/sbt-typelevel

I think it streamlines a lot of this and the maintainer is very helpful.

hazel-sudz:
sbt-typelevel looks promising, looking into it. I'm usually weary of unnecessary dependencies but there's very high community support so seems fine to me.

hazel-sudz:
I looked at sbt-typelevel and IMO is tries to do way too much. The generative aspect of the CI workflows makes it impossible to do any custom modifications and I don't think it's a good fit here. 

And we already have a workflow for running tests that works fine. I'm thinking of something like https://github.com/sbt/sbt-ci-release to handle just the release part separately. Don't see the value in re-organizing everything just to be able to use sbt-typelevel as a managed platform.

hazel-sudz:
https://github.com/twitter/scalding/pull/1975 looking for someone to help with adding repo secrets for sonatype and GPG

nownikhil:
I think I accidentally pushed the wrong commit in develop. Publish is failing. I think daniel pushed a commit as I did a squash and merge.

hazel-sudz:
@nownikhil the commit seems fine to me. Publish is failing on GPG signing https://github.com/twitter/scalding/runs/5573979628?check_suite_focus=true

do we have GPG keys in the secrets section? That's probably the cause

hazel-sudz:
you can see that the command passed to GPG does not include a passphrase: 

```
'gpg --batch --pinentry-mode loopback --passphrase  --detach-sign
```

typically it would read in the logs as 
```
'gpg --batch --pinentry-mode loopback --passphrase <****** redacted passphrase here in CI>  --detach-sign
```

johnynek:
there are no secrets at all set up.

I am still an admin of this repo, so I can add them it seems.

hazel-sudz:
[...]

PR #1975, opened 2022-03-11
[Ready for Review] Add Snapshot Workflow for Develop

hazel-sudz:
I looked at some tools and ended up deciding that just re-using the existing publish workflow makes the most sense without having to change the existing CI workflow for tests or causing a lot of diffs in general.


build.sbt
```diff
@@ -118,7 +118,6 @@ val sharedSettings = Seq(
     setNextVersion,
     commitNextVersion,
     ReleaseStep(action = Command.process("sonatypeReleaseAll", _)),
-    pushChanges
   ),
```
hazel-sudz: todo: I will make this depend conditionally on if we are pushing a snapshot or not

build.sbt
```diff
@@ -106,19 +106,18 @@ val sharedSettings = Seq(
...
+      Seq[ReleaseStep](
+        inquireVersions,
+        runClean,
+        publishArtifacts,
+        ReleaseStep(action = Command.process("sonatypeReleaseAll", _)),
+      )
+    // todo : implement workflow for final tagged releases that are not snapshots
+    else Seq[ReleaseStep](
+
+    )
```
hazel-sudz: this will be in follow-up PR once we have snapshots working well off develop

hazel-sudz: I don't think we need to run tests as we already run tests on each PR so that should be fine

version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "0.18.0-SNAPSHOT"
+ThisBuild / version := "0.18.0"
```
hazel-sudz: on snapshot publish, this will get prepended with SNAPSHOT by the CI. This is also how we can differentiate snapshot vs non-snapshot release on build.sbt file.

hazel-sudz:
I need the following loaded into the repo secrets to test this any further: 

| Github Repo Secret  | Used by Plugin |
| ------------- | ------------- |
| SONATYPE_PASSWORD  | https://github.com/xerial/sbt-sonatype |
| SONATYPE_USERNAME  | https://github.com/xerial/sbt-sonatype |
| PGP_PASSPHRASE  |https://github.com/sbt/sbt-pgp |
| PGP_SECRET  | https://github.com/sbt/sbt-pgp |

version.sbt
johnynek: I wonder if we can use:

https://github.com/sbt/sbt-dynver

hazel-sudz: sounds fine to me will implement I like the "fetch from recent tag" idea.

build.sbt
```diff
@@ -102,25 +102,26 @@ val sharedSettings = Seq(
...
-    runTest,
-    setReleaseVersion,
-    commitReleaseVersion,
-    tagRelease,
-    publishArtifacts,
-    setNextVersion,
-    commitNextVersion,
-    ReleaseStep(action = Command.process("sonatypeReleaseAll", _)),
-    pushChanges
-  ),
```
hazel-sudz, line 122: any kind of versioning through `version.sbt` file is now no longer needed thanks to https://github.com/sbt/sbt-dynver

final versions can now be trigged trough the Github Releases UI in which you can select a branch and a tag.

version.sbt
```diff
@@ -1 +0,0 @@
-ThisBuild / version := "0.18.0-SNAPSHOT"
```
hazel-sudz, line 1: `develop` currently resolves to `0.16.1-RC2` based on tags in the git history. Would need to push a new tag for the snapshot versions to start making sense

johnynek:
https://github.com/sbt/sbt-buildinfo

if we add that, we can remove any code that depends on version in sbt and instead get it from a generated object.

johnynek:
actually using this we can also add the gitsha:

https://brainchunk.blogspot.com/2016/11/embed-git-commit-hash-scala-project.html;

version.sbt
johnynek: yeah, I can take that. https://search.maven.org/artifact/com.twitter/scalding-core_2.12 looks like 0.18.0-RC1 was the last published.

hazel-sudz:
> actually using this we can also add the gitsha:
> 
> https://brainchunk.blogspot.com/2016/11/embed-git-commit-hash-scala-project.html;

gitsha can be handled by `sbt-dynver`. It is written in the version by default if the HEAD commit does not have a tag (ie. for SNAPSHOTS)

hazel-sudz:
> https://github.com/sbt/sbt-buildinfo
> 
> if we add that, we can remove any code that depends on version in sbt and instead get it from a generated object.

++ I think there's one place doing something with scalding version (but hasn't been updated in a while so probably not critical?). I'm currently having some problems with scald.rb not finding the proper versioned artifacts working on resolving that.

hazel-sudz:
was fixing some test issues caused by the dynamic version. Will add changes in a bit. Here is what the versioning behaviour will look like afterwards:


```scala
// head commit has been tagged with v2.0
dsudzilouski@dsudzilouski scalding % ./sbt -Dsbt.log.noformat=true -Dsbt.supershell=false "print scalding-core / version" -error
2.0

// if head commit is tagged should always assume a final release even if git history is dirty
dsudzilouski@dsudzilouski scalding % touch TEST_FILE_DIRTY
dsudzilouski@dsudzilouski scalding % git add .
dsudzilouski@dsudzilouski scalding % git status
On branch debug_CI
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	new file:   TEST_FILE_DIRTY

dsudzilouski@dsudzilouski scalding % ./sbt -Dsbt.log.noformat=true -Dsbt.supershell=false "print scalding-core / version" -error
2.0

// head commit is untagged, should assume SNAPSHOT release
dsudzilouski@dsudzilouski scalding % git commit -m "no tag so should drop to snapshot against 2.0 with commit sha"
[debug_CI 7ed1fdbb] no tag so should drop to snapshot against 2.0 with commit sha
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 TEST_FILE_DIRTY

dsudzilouski@dsudzilouski scalding % ./sbt -Dsbt.log.noformat=true -Dsbt.supershell=false "print scalding-core / version" -error
2.0-1-7ed1fdbb-SNAPSHOT
```

scalding-core/src/main/scala/com/twitter/package.scala
```diff
@@ -34,7 +34,7 @@ package object scalding {
   /**
    * Make sure this is in sync with version.sbt
    */
-  val scaldingVersion: String = "0.17.2"
+  val scaldingVersion: String = "0.18.0"
```
hazel-sudz: I believe this is the only place that we refer to scalding version within src. ~~As mentioned above, this can be addressed with something like https://github.com/sbt/sbt-buildinfo. But considering that it is already out of date and clearly not critical, I think it would be fair for a followup~~ (done) CC @johnynek

hazel-sudz:
@nownikhil any chance we could have help with creds being imported to the repo secrets section?

scalding-core/src/main/scala/com/twitter/package.scala
hazel-sudz: I have moved this change to https://github.com/twitter/scalding/pull/1976 as it is independent of what is being proposed here

nownikhil:
Apparently both me and Navin don't have the right permissions. There is twitter internal process to get access to publish to external repo. @navinvishy is working on that.

johnynek:
can you merge in develop?

johnynek (approved):

daniel-sudz:
we can merge it's just this will fail until we have the publish secrets available. It's probably not a big deal (and also might be helpful to sort other bugs if there any).

hazel-sudz:
@nownikhil are there any updates on publish creds? We would really love to have some RC artifacts to do integration with real production jobs.

nownikhil:
I got the impression that Oscar (@johnynek) is gonna setup up the secrets since he is the admin. He commented on https://github.com/twitter/scalding/issues/1973

I will follow up on the internal process. The internal ticket is still unassigned, let me ping some folks here and see if they know any better.

johnynek:
I can take this. Sorry I just haven't gotten to it.

Issue #1977, opened 2022-03-18
Re-enable code coverage


johnynek:
https://github.com/twitter/scalding/pull/1922

that PR was started, but went stale.

I don't think it is much work.

@daniel-sudz @nownikhil

hazel-sudz:
@johnynek I can look into this next week. We are also still waiting on the situation with publishing keys for sonatype and GPG

hazel-sudz:
they have a binding for github-actions that is credential-less: https://github.com/marketplace/actions/codecov
I will try to get something out today

hazel-sudz:
I think we need more aggressive timeouts on the codecov uploader. Based on what I've seen in https://github.com/codecov/codecov-action/issues the symptom seems to be hitting github rate limits on their API. 

Another resolution for some people appears to be adding an auth token. This is optional for public repos, but seems to help some people to not get upload failures.

PR #1981, opened 2022-03-22
Enable Codecoverage Upload (Jacoco/Codecov)

hazel-sudz:
- [x] verify that it is actually running
- [x] codecov.yml file probably good idea
- [x] clean up for review
- [x] verify that codecov seems reasonable 

basically moving changes over from https://github.com/twitter/scalding/pull/1922 based on https://github.com/twitter/scalding/issues/1977


.github/workflows/CI.yml
```diff
@@ -130,3 +130,12 @@ jobs:
...
           ${{ matrix.script }}
+
+
+      - uses: actions/checkout@master
+      - uses: codecov/codecov-action@v2
+        with:
+          # files: ./coverage1.xml,./coverage2.xml # optional
+          name: codecov
+          fail_ci_if_error: true
+          verbose: true
```
hazel-sudz: the documentation assures that no credentials are needed using the provided action. I'm not sure how the uploaded works and if it can auto-detect the generated coverage files

scripts/run_test.sh
```diff
@@ -5,40 +5,30 @@ BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/.. && pwd )"
...
+echo "Running test suite"
 export JVM_OPTS="-XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -XX:ReservedCodeCacheSize=128m -XX:+TieredCompilation -XX:MaxPermSize=256m -Xms256m -Xmx768m -Xss2m"
-echo "calling ... "
-echo "time ./sbt ++$TRAVIS_SCALA_VERSION $(withCmd test)"
-time ./sbt -Dhttp.keepAlive=false -Dsbt.repository.secure=true  ++$TRAVIS_SCALA_VERSION "$(withCmd test)"
+time ./sbt -Dhttp.keepAlive=false -Dsbt.repository.secure=true ++$TRAVIS_SCALA_VERSION coverage "$(withCmd test) coverageReport"
 TST_EXIT_CODE=$?
 
 echo "Running mima checks ... "
-echo "time ./sbt ++$TRAVIS_SCALA_VERSION $(withCmd mimaReportBinaryIssues)"
```
hazel-sudz, line 33: we already have the bash flags that print out commands before they are run so this isn't really needed

scripts/run_test.sh
```diff
@@ -5,40 +5,30 @@ BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/.. && pwd )"
...
 
 echo "Running compile:doc ... "
 echo "time ./sbt ++$TRAVIS_SCALA_VERSION $(withCmd compile:doc)"
 time ./sbt -Dhttp.keepAlive=false -Dsbt.repository.secure=true  ++$TRAVIS_SCALA_VERSION "$(withCmd compile:doc)"
-COMPILE_DOC_EXIT_CODE=$?
 
-echo "all done"
 
-exit $(( $TST_EXIT_CODE || $MIMA_EXIT_CODE || $COMPILE_DOC_EXIT_CODE ))
+echo "all test checks done"
```
hazel-sudz, line 30: we already have bash flags to fail-fast so we don't explicitly need to collect exit codes

johnynek:
Coverage seems quite low. Are we sure it is aggregating all the projects correctly? Many subprojects wind up testing the others.

hazel-sudz:
> Coverage seems quite low. Are we sure it is aggregating all the projects correctly? Many subprojects wind up testing the others.

there's currently one part in the CI that is failing. It should be a bit higher because of it. I'm not 100% sure why it's failing just get a "step failed" in cascading. I suspect it's due to very high memory usage due to coverage. I can currently reproduce the failure locally when switching coverage on/off before running the test. I'm playing with memory settings but currently not much success. See https://github.com/scoverage/sbt-scoverage/issues/50

hazel-sudz:
> > Coverage seems quite low. Are we sure it is aggregating all the projects correctly? Many subprojects wind up testing the others.
> 
> there's currently one part in the CI that is failing. It should be a bit higher because of it. I'm not 100% sure why it's failing just get a "step failed" in cascading. I suspect it's due to very high memory usage due to coverage. I can currently reproduce the failure locally when switching coverage on/off before running the test. I'm playing with memory settings but currently not much success. See [scoverage/sbt-scoverage#50](https://github.com/scoverage/sbt-scoverage/issues/50)

~~after tuning the JVM opts, jacking the stack frame size to 64M (crazy!) seemed to let the failing test pass. It looks like others have had similar issues: https://github.com/scoverage/sbt-scoverage/issues/276 https://github.com/scoverage/sbt-scoverage/issues/422~~

edit: I though I had coverage enabled for above but I did not so I still have no resolution for sbt-scoverage

johnynek:
huh:

https://github.com/twitter/scalding/runs/5664208563?check_suite_focus=true#step:7:1367

seems one of the jobs fails (in both 2.11 and 2.12) with coverage... I wonder if somehow the instrumentation is making something non-serializable...

hazel-sudz:
> huh:
> 
> https://github.com/twitter/scalding/runs/5664208563?check_suite_focus=true#step:7:1367
> 
> seems one of the jobs fails (in both 2.11 and 2.12) with coverage... I wonder if somehow the instrumentation is making something non-serializable...

that's completely possible. I've tested https://www.scala-sbt.org/sbt-jacoco/index.html as an alternative on the failing test and it seems to work. I'm going to try it in CI. 

The downside of jacoco is that it knows nothing about scala (targets bytecode) so it might be slightly less accurate.

scripts/run_test.sh
```diff
@@ -4,41 +4,27 @@
 BASE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/.. && pwd )"
 cd $BASE_DIR
 
-export JVM_OPTS="-XX:+CMSClassUnloadingEnabled -XX:+UseConcMarkSweepGC -XX:ReservedCodeCacheSize=96m -XX:+TieredCompilation -XX:MaxPermSize=256m -Xms256m -Xmx512m -Xss2m"
-
-
-INNER_JAVA_OPTS="set javaOptions += \"-Dlog4j.configuration=file://$TRAVIS_BUILD_DIR/project/travis-log4j.properties\""
```
hazel-sudz, line 10: I see no references to this variable while searching through the codebase. I don't think it's used anymore

hazel-sudz:
@johnynek does 40% sound reasonable? It's recording 0% for maple (there is no test folder) and 0% for scalding-avro (no test folder) and 0% for scalding-thrift-macros (there are some tests but maybe it can't trace macros well). 

I think this is similar to what scoverage was reporting. 

Report is here: https://codecov.io/gh/twitter/scalding/tree/5db2df2a50b4a92c0fbf934a5dc13b06229624e9
Screen Shot 2022-03-24 at 12 28 30 PM


the reference from 2 years ago was 68%: https://codecov.io/gh/twitter/scalding/tree/f1e9670175047724bf23727ab17f807303281150

Screen Shot 2022-03-24 at 12 34 38 PM

johnynek:
I don't see how it could have fallen so much, but maybe.

I was looking through, it marks a lot of lines as partially covered, e.g.:

https://app.codecov.io/gh/twitter/scalding/blob/5db2df2a50b4a92c0fbf934a5dc13b06229624e9/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala

are those being counted?

Anyway, this is an improvement, and we can iterate. Thanks for spending the time on this. We can later see if we can find examples of tests that definitely exercise code, but it doesn't see and add an issue at that time.

README.md
```diff
@@ -1,7 +1,7 @@
 # Scalding
 
 [![Build status](https://github.com/twitter/scalding/actions/workflows/CI.yml/badge.svg?branch=develop)](https://github.com/twitter/scalding/actions)
-[![Coverage Status](https://coveralls.io/repos/twitter/scalding/badge.png?branch=develop)](https://coveralls.io/r/twitter/scalding?branch=develop)
+[![Coverage Status](https://img.shields.io/codecov/c/github/twitter/scalding/develop.svg?maxAge=3600)](https://codecov.io/github/twitter/scalding)
```
johnynek, line 4: max age is 1h? seems like maybe that should be longer no?

hazel-sudz: I'm fine with whatever. Although github also does their own caching so it's probably a bit irrelevant

hazel-sudz:
> I don't see how it could have fallen so much, but maybe.
> 
> I was looking through, it marks a lot of lines as partially covered, e.g.:
> 
> https://app.codecov.io/gh/twitter/scalding/blob/5db2df2a50b4a92c0fbf934a5dc13b06229624e9/scalding-core/src/main/scala/com/twitter/scalding/typed/memory_backend/MemoryBackend.scala
> 
> are those being counted?
> 
> Anyway, this is an improvement, and we can iterate. Thanks for spending the time on this. We can later see if we can find examples of tests that definitely exercise code, but it doesn't see and add an issue at that time.

my guess would be that the partial lines are the ones that can throw exceptions. Ie. they are being executed but not all paths covered

johnynek:
Thanks!

PR #1983, opened 2022-04-08
Beam backend: use TypedPipe descriptions as names for PTransforms

navinvishy:
Problem: When debugging Beam jobs running on Dataflow, it is hard to map a failure in the Dataflow UI back to the corresponding line in the Scalding code.

Solution: Assign the description of the TypedPipe as the name of the PTransform(this name appears in the Dataflow UI). This allows mapping the error back to the line in the Scalding code since by default, the descriptions contain the line numbers.

In order to make this work, I changed the Beam backend optimization rules to not use the `DescribeLater` rule, which moves a `WithDescriptionTypedPipe` higher in the AST. This was necessary in order to preserve the location of the description in the AST and apply it to the right PTransform in Beam. This possibly introduces a few unintended consequences for the optimizer: 
* Some composition rules such as `ComposeMap`, `ComposeFlatMap` etc. may not be applicable since those rely on `DescribeLater` being run in an earlier phase. While such optimizations dont seem to exist in Beam, they do have something along these lines in [Dataflow](https://cloud.google.com/dataflow/docs/guides/deploying-a-pipeline#fusion-optimization). So this should ideally not have any performance implications, at least in Dataflow.
* `DescribeLater` also allows `ComposeDescriptions` to work better, which reduces the number of nodes in the AST. Its possible that removing `DescribeLater` can make the AST very large.

### DIFF
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
@@ -76,7 +76,12 @@ object BeamPlanner {
           BeamOp.FromIterable(iterable, kryoCoder)
         case (wd: WithDescriptionTypedPipe[_], rec) => {
           val op = rec(wd.input)
-          op.withName(wd.descriptions.map(_._1).head)
+          wd.descriptions match {
+            case head :: _ =>
+              op.withName(head._1)
+            case Nil =>
+              op
+          }
         }
         case (SumByLocalKeys(pipe, sg), rec) =>
           val op = rec(pipe)
--- a/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
+++ b/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
@@ -1,20 +1,33 @@
 package com.twitter.scalding.beam_backend
 
-import com.twitter.scalding.dagon.{FunctionK, Memoize, Rule}
 import com.twitter.chill.KryoInstantiator
 import com.twitter.chill.config.ScalaMapConfig
 import com.twitter.scalding.Config
-import com.twitter.scalding.beam_backend.BeamOp.{CoGroupedOp, MergedBeamOp}
+import com.twitter.scalding.beam_backend.BeamOp.CoGroupedOp
+import com.twitter.scalding.beam_backend.BeamOp.MergedBeamOp
+import com.twitter.scalding.dagon.FunctionK
+import com.twitter.scalding.dagon.Memoize
+import com.twitter.scalding.dagon.Rule
 import com.twitter.scalding.serialization.KryoHadoop
+import com.twitter.scalding.typed.CoGrouped.WithDescription
+import com.twitter.scalding.typed.OptimizationRules.AddExplicitForks
+import com.twitter.scalding.typed.OptimizationRules.ComposeDescriptions
+import com.twitter.scalding.typed.OptimizationRules.DeferMerge
+import com.twitter.scalding.typed.OptimizationRules.DiamondToFlatMap
+import com.twitter.scalding.typed.OptimizationRules.FilterKeysEarly
+import com.twitter.scalding.typed.OptimizationRules.IgnoreNoOpGroup
+import com.twitter.scalding.typed.OptimizationRules.MapValuesInReducers
+import com.twitter.scalding.typed.OptimizationRules.RemoveDuplicateForceFork
+import com.twitter.scalding.typed.OptimizationRules.RemoveUselessFork
+import com.twitter.scalding.typed.OptimizationRules.composeIntoFlatMap
+import com.twitter.scalding.typed.OptimizationRules.composeSame
+import com.twitter.scalding.typed.OptimizationRules.simplifyEmpty
 import com.twitter.scalding.typed._
-import com.twitter.scalding.typed.functions.{
-  FilterKeysToFilter,
-  FlatMapValuesToFlatMap,
-  MapValuesToMap,
-  ScaldingPriorityQueueMonoid
-}
-
 import com.twitter.scalding.typed.cascading_backend.CascadingExtensions.ConfigCascadingExtensions
+import com.twitter.scalding.typed.functions.FilterKeysToFilter
+import com.twitter.scalding.typed.functions.FlatMapValuesToFlatMap
+import com.twitter.scalding.typed.functions.MapValuesToMap
+import com.twitter.scalding.typed.functions.ScaldingPriorityQueueMonoid
 
 object BeamPlanner {
   def plan(
@@ -61,8 +74,10 @@ object BeamPlanner {
           BeamOp.Source(config, src, srcs(src))
         case (IterablePipe(iterable), _) =>
           BeamOp.FromIterable(iterable, kryoCoder)
-        case (wd: WithDescriptionTypedPipe[a], rec) =>
-          rec[a](wd.input)
+        case (wd: WithDescriptionTypedPipe[_], rec) => {
+          val op = rec(wd.input)
+          op.withName(wd.descriptions.map(_._1).head)
+        }
         case (SumByLocalKeys(pipe, sg), rec) =>
           val op = rec(pipe)
           config.getMapSideAggregationThreshold match {
@@ -97,7 +112,10 @@ object BeamPlanner {
             uir.evidence.subst[BeamOpT](sortedOp)
           }
           go(ivsr)
-        case (ReduceStepPipe(ValueSortedReduce(keyOrdering, pipe, valueSort, reduceFn, _, _)), rec) =>
+        case (
+              ReduceStepPipe(ValueSortedReduce(keyOrdering, pipe, valueSort, reduceFn, _, _)),
+              rec
+            ) =>
           val op = rec(pipe)
           op.sortedMapGroup(reduceFn)(keyOrdering, valueSort, kryoCoder)
         case (ReduceStepPipe(IteratorMappedReduce(keyOrdering, pipe, reduceFn, _, _)), rec) =>
@@ -116,7 +134,7 @@ object BeamPlanner {
             val ops: Seq[BeamOp[(K, Any)]] = cg.inputs.map(tp => rec(tp))
             CoGroupedOp(cg, ops)
           }
-          go(cg)
+          if (cg.descriptions.isEmpty) go(cg) else go(cg).withName(cg.descriptions.last)
         case (Fork(input), rec) =>
           rec(input)
         case (m @ MergedTypedPipe(_, _), rec) =>
@@ -137,7 +155,21 @@ object BeamPlanner {
 
   def defaultOptimizationRules(config: Config): Seq[Rule[TypedPipe]] = {
     def std(forceHash: Rule[TypedPipe]) =
-      OptimizationRules.standardMapReduceRules :::
+      List(
+        // phase 0, add explicit forks to not duplicate pipes on fanout below
+        AddExplicitForks,
+        RemoveUselessFork,
+        // phase 1, compose flatMap/map, move descriptions down, defer merge, filter pushup etc...
+        IgnoreNoOpGroup.orElse(composeSame).orElse(FilterKeysEarly).orElse(DeferMerge),
+        // phase 2, combine different kinds of mapping operations into flatMaps, including redundant merges
+        composeIntoFlatMap
+          .orElse(simplifyEmpty)
+          .orElse(DiamondToFlatMap)
+          .orElse(ComposeDescriptions)
+          .orElse(MapValuesInReducers),
+        // phase 3, remove duplicates forces/forks (e.g. .fork.fork or .forceToDisk.fork, ....)
+        RemoveDuplicateForceFork
+      ) :::
         List(
           OptimizationRules.FilterLocally, // after filtering, we may have filtered to nothing, lets see
           OptimizationRules.simplifyEmpty,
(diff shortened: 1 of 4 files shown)


scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -1,20 +1,33 @@
...
 import com.twitter.scalding.Config
-import com.twitter.scalding.beam_backend.BeamOp.{CoGroupedOp, MergedBeamOp}
+import com.twitter.scalding.beam_backend.BeamOp.CoGroupedOp
+import com.twitter.scalding.beam_backend.BeamOp.MergedBeamOp
+import com.twitter.scalding.dagon.FunctionK
+import com.twitter.scalding.dagon.Memoize
+import com.twitter.scalding.dagon.Rule
 import com.twitter.scalding.serialization.KryoHadoop
+import com.twitter.scalding.typed.CoGrouped.WithDescription
+import com.twitter.scalding.typed.OptimizationRules.AddExplicitForks
```
johnynek: nit: can we use a grouped import here rather than one-at-a-time java style:
```scala
import com.twitter.scalding.typed.OptimizationRules.{
  ComposeDescriptions, DeferMerge, Dia...
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -61,8 +74,10 @@ object BeamPlanner {
           BeamOp.Source(config, src, srcs(src))
         case (IterablePipe(iterable), _) =>
           BeamOp.FromIterable(iterable, kryoCoder)
-        case (wd: WithDescriptionTypedPipe[a], rec) =>
-          rec[a](wd.input)
+        case (wd: WithDescriptionTypedPipe[_], rec) => {
+          val op = rec(wd.input)
+          op.withName(wd.descriptions.map(_._1).head)
```
johnynek: `.head` can throw. Can we instead do:
```scala
wd.descriptions match {
  case head :: _ =>
    op.withName(head._1)
  case Nil =>
    op
}
```

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -116,7 +134,7 @@ object BeamPlanner {
             val ops: Seq[BeamOp[(K, Any)]] = cg.inputs.map(tp => rec(tp))
             CoGroupedOp(cg, ops)
           }
-          go(cg)
+          if (cg.descriptions.isEmpty) go(cg) else go(cg).withName(cg.descriptions.last)
```
johnynek, line 129: why using `last` here but `head` above?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -137,7 +155,21 @@ object BeamPlanner {
 
   def defaultOptimizationRules(config: Config): Seq[Rule[TypedPipe]] = {
     def std(forceHash: Rule[TypedPipe]) =
-      OptimizationRules.standardMapReduceRules :::
+      List(
```
johnynek, line 150: why the change here? Is this copying the same cascading optimizations?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
```diff
@@ -6,9 +6,10 @@ import com.twitter.scalding.Config
 import com.twitter.scalding.beam_backend.BeamFunctions._
 import com.twitter.scalding.beam_backend.BeamJoiner.MultiJoinFunction
 import com.twitter.scalding.serialization.Externalizer
+import com.twitter.scalding.typed.Input
 import com.twitter.scalding.typed.functions.ComposedFunctions.ComposedMapGroup
 import com.twitter.scalding.typed.functions.{EmptyGuard, MapValueStream, ScaldingPriorityQueueMonoid, SumAll}
-import com.twitter.scalding.typed.{CoGrouped, Input}
+import com.twitter.scalding.typed.{CoGrouped, TypedSource}
```
johnynek: Why the change back from Input?

(I think I know the answer: you all are actually using a fork... I really hope you will get onto the mainline branch. I am contributing my time to help you, I hope you will contribute your energy to help others by actually using the open source version).

johnynek (commented):
Just had a couple of minor comments.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: Done. Thanks!

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: Done. I thought that for a `WithDescriptionTypedPipe` this could not be empty.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: I meant to ask you about this myself. If you look at the test case in this PR, The cogrouped expression has two descriptions "Count words" and "Join with t1", both of which appear in the descriptions of `CoGrouped`. But since this appears to be appending descriptions for eg. in `CoGrouped.Pair` and the optimization rule `ComposeDescriptions`, I chose to keep the last one. Hence no assertion for "Count words" even in that test case.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: It really is the `standardMapReduceRules` with `DescribeLater` excluded. I've mentioned in the summary of this PR that I chose to exclude that rule in order to capture the line numbers at the right points, without the optimizer changing the AST. I'm curious if there's a better way to do this though.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
navinvishy: I didn't really change it back...this was just a missed import, but you're right that we're using a fork. Thank you for investing your time in this, it really is much appreciated! I have it on my todo list to get us on the mainline branch. I do need to go through your earlier refactor more closely though.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
johnynek: sorry I missed that. Note: the line numbers are already captured by the time that rule runs. The line numbers are collected on the `TypedPipe`. I don't see how removing that rule helps descriptions.

If I were you, I would look at all the descriptions and add the full list.

The problem with removing that rule is that it will block merging nodes together. It may be fine, maybe Beam will follow up with optimizations, but I would be careful: scalding may do some optimizations that beam doesn't.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -1,21 +1,21 @@
...
 import com.twitter.scalding.typed._
+import com.twitter.scalding.typed.cascading_backend.CascadingExtensions.ConfigCascadingExtensions
 import com.twitter.scalding.typed.functions.{
   FilterKeysToFilter,
   FlatMapValuesToFlatMap,
   MapValuesToMap,
   ScaldingPriorityQueueMonoid
 }
 
-import com.twitter.scalding.typed.cascading_backend.CascadingExtensions.ConfigCascadingExtensions
```
johnynek, line 17: I'm glad if we don't need this, but I'm also surprised. I meant to remove any of these that were unneeded.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
johnynek: Why not combine all of them? Why not just `cg.descriptions.mkString("\n")` or something?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: The line numbers are in the descriptions though, if users don't explicitly add them with a call to `withDescription`. 
Here's what I'm trying to do: if we have a `TypedPipe` that looks something like this:
`WithDescription(Mapped(WithDescription(FlatMapped(), "a")), "b") `
I'm trying to associate the description "a" with the PTransform corresponding to the `FlatMapped` and the description "b" with the PTransform corresponding to the `Mapped`.  By default, the "a" and "b" here are line numbers unless the user explicitly added their own descriptions.
If I let `DescribeLater` run, I think we'll end up with something like this:
`WithDescription(FlatMapped(), ["a","b"])`
(since this would also compose the `Mapped` and `FlatMapped`, and then also run `ComposeDescriptions`)
This would be hard to debug, when there is an error in this stage in the job since it's hard to map back to the code.
But as I write this, I'm thinking if its probably better to just concatenate the line numbers instead...
Dataflow (which is the Beam runner we're using) does apply similar optimizations (I've mentioned this in the summary), but I'm debating now if its a good idea to change what the optimizer does, since it definitely makes the optimizer less useful. What do you think?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: This hasn't been removed...just moved up with the ordering of imports. I think we need this.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
johnynek: it maybe can't but we didn't make that clear in the types... :/

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
johnynek: I don't really know the answer. So, line numbers are added to descriptions when the user constructs the TypedPipe, see:
https://github.com/twitter/scalding/blob/64343489135111ca205e5f5e894367cdd545f765/scalding-base/src/main/scala/com/twitter/scalding/typed/TypedPipe.scala#L533

The idea of `DescribeLater` is to try to merge as many operations into one step and then put all the line numbers/descriptions on that single step. By taking just a single item you are hiding which actions might be combined.

As to what to do, I don't know. If beam's optimizer is very good, it maybe doesn't matter. Maybe you should try to separate runs with a somewhat complex job and compare?

Also, I can imagine a `Config` setting like `scalding.preserve_description_order` that removes applying that optimization. Then if you were debugging, you could set that to true, but for a production job you could set it to false.

I would probably bias to just combining all the descriptions into a single beam description unless you actually see problems. You can always come back and add that setting. I would personally bias to maximizing the utility of the optimizer.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
johnynek: I don't follow. I changed this code to using `Input` which is a subclass and replacement for `TypedSource`. It seems the diff changes it back from Input to TypedSource. Can you restate what you meant that this is a "missed import"?

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
```diff
@@ -61,8 +61,15 @@ object BeamPlanner {
           BeamOp.Source(config, src, srcs(src))
         case (IterablePipe(iterable), _) =>
           BeamOp.FromIterable(iterable, kryoCoder)
-        case (wd: WithDescriptionTypedPipe[a], rec) =>
-          rec[a](wd.input)
+        case (wd: WithDescriptionTypedPipe[_], rec) => {
+          val op = rec(wd.input)
+          wd.descriptions match {
```
johnynek, line 66: actually, as commented, I think this is probably running the risk of dropping some descriptions.

I would do:
```scala
op.withName(wd.descriptions.map(_._1).mkString(", "))
```
otherwise I think you will wind up with cases where you lose line numbers.

johnynek (commented):
this generally looks good, I'm just worried about two things:

1. disabling moving descriptions deactivating many (most?) optimizations
2. not including all the descriptions (I think you will lose line numbers).

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: Thanks. Let me try out some jobs with the rule enabled and showing all the descriptions.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamOp.scala
navinvishy: Sorry I wasn't clear. This change is now outdated - I updated it to remove the reference to `TypedSource`. The reason it was there in the first place was because I brought it over from our fork, but I hadn't removed unused imports, which is why it showed up. You should see no reference to `TypedSource` now in the diff.

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: Yes, let me try this out on a job first. Thanks!

scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala
navinvishy: Thanks, I'll try this out.

PR #1984, opened 2022-04-09
no op change to test publish and see if something breaks

hazel-sudz:
added new line to readme @johnynek

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,3 +156,4 @@ A full list of [contributors](https://github.com/twitter/scalding/graphs/contrib
 Copyright 2016 Twitter, Inc.
 
 Licensed under the [Apache License, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0)
+


johnynek (approved):

johnynek:
let's see if this publishes!

@daniel-sudz let's add a comment to the artifact we publish here (I hope a snapshot with a sha derived version).

johnynek:
https://github.com/twitter/scalding/actions/runs/2142207323

queued...

hazel-sudz:
hmmm looks like we still have a GPG issue. I wonder if this is related: https://github.com/sbt/sbt-pgp/issues/173

edit: we seem to have the latest version of sbt-pgp so unlikely

johnynek:
```
[error] java.lang.RuntimeException: Failure running 'gpg --batch --pinentry-mode loopback --passphrase *** --detach-sign --armor --use-agent --output /home/runner/work/scalding/scalding/scalding-avro/target/scala-2.12/scalding-avro_2.12-0.16.1-RC2-360-ebbfeb71-SNAPSHOT.pom.asc /home/runner/work/scalding/scalding/scalding-avro/target/scala-2.12/scalding-avro_2.12-0.16.1-RC2-360-ebbfeb71-SNAPSHOT.pom'.  Exit code: 2
```
two issues:

1. it is trying to publish 0.16.1-RC2 which I think is very old. I think we want something like 0.18.0-RC2: https://search.maven.org/artifact/com.twitter/scalding-core_2.12
2. I can't see any particular error, but the flag they mention seems to be added.

I wonder if I somehow goofed up the secrets? I followed these instructions: https://typelevel.org/sbt-typelevel/secrets.html and did the last option:
```
gpg --armor --export-secret-keys $LONG_ID | base64
```
and set that as the the secret.

hazel-sudz:
yes, let's push a new tag
I can try to repro PGP issue locally with your command nothing obvious I can think of for now

johnynek:
yeah, those pgp instructions are wrong, you don't want the trailing `| base64` I think. I think you just want to copy the base64 output.

johnynek:
I just updated the secret. We can try again. But we should fix the tag.

hazel-sudz:
I also wonder if we need the PGP_PASSPHRASE the doc you linked seems to suggest that it is not required (only PGP_SECRET). If that isn't used we should remove it could be tripping up pgp

johnynek:
cc @armanbilge (maintainer of sbt-typelevel: I think those instructions are double base64 encoding the keys, but I'm not sure since I haven't made it work yet).

I don't know? I added a passphrase to the key, but when I exported it maybe it removes it? idk.

johnynek:
also, where do we need the tags? we do have a later tag:

https://github.com/twitter/scalding/tree/v0.18.0-RC1

armanbilge:
> If that isn't used we should remove it could be tripping up pgp

If your key is passphrased, you need it. If your key is not passphrased, then you should remove it.

One easy fix may just be to generate a new keypair using the helper:
https://typelevel.org/sbt-typelevel/secrets.html#option-1-generate-a-key-pair-in-your-browser

hazel-sudz:
@johnynek I will make a PR to remove passphrase it seems redundant for our purpose
let's use the generator as suggested above

johnynek:
Wait well, I made a passphrased key and used the base64 encoding the first time and it didn't work?

armanbilge:
Ummm ... is this project using sbt-typelevel? None of those instructions apply if it's not ...

hazel-sudz:
yeah we are not using typelevel I was meaning to mention that. We use the base plugin: 
```addSbtPlugin("com.github.sbt" % "sbt-release" % "1.1.0")```
```addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")```

the environment variable should be the same though so I assume the generator on that site should still work for us

**Comment by armanbilge** [comment]
Yeah, don't know why I was pinged, I have no stake in anything here :(

**Comment by daniel-sudz** [comment]
@johnynek I see the following instructions for the plugin we use: 
https://github.com/sbt/sbt-ci-release#secrets
```
Add the following secrets:

PGP_PASSPHRASE: The randomly generated password you used to create a fresh gpg key. For Travis Only: If the password contains bash special characters, make sure to escape it by wrapping it in single quotes 'my?pa$$word', see [Travis Environment Variables](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings).
PGP_SECRET: The base64 encoded secret of your private key that you can export from the command line like here below
# macOS
gpg --armor --export-secret-keys $LONG_ID | base64 | pbcopy
# Ubuntu (assuming GNU base64)
gpg --armor --export-secret-keys $LONG_ID | base64 -w0 | xclip
# Arch
gpg --armor --export-secret-keys $LONG_ID | base64 | sed -z 's;\n;;g' | xclip -selection clipboard -i
# FreeBSD (assuming BSD base64)
gpg --armor --export-secret-keys $LONG_ID | base64 | xclip
# Windows
gpg --armor --export-secret-keys %LONG_ID% | openssl base64
```

hazel-sudz:
@johnynek I think the situation is this: 

with https://github.com/sbt/sbt-release, it just forwards to sbt-pgp which dosn't handle the env variables nicely 
we can use https://github.com/sbt/sbt-ci-release that should handle everything nicely 

the situation with typelevel was that it would require a lot of changes for all of our config setups so it didn't seem worth it

johnynek:
Sorry @armanbilge i pinged you because I thought there may be a typo in your instructions I copied. I now realize that like most things in scala build it is massively bespoke and it appears there is a ton of diversity probably for no sane reason except publishing was never made excellent in sbt by default.

Sorry for my mistake on guessing your instructions may have a typo.

hazel-sudz:
https://github.com/twitter/scalding/pull/1985

PR #1985, opened 2022-04-10
Try sbt-publish-ci

hazel-sudz:
it handles the GPG part nicely (should work with the exiting ENV variable). It might take some iterations to get this to work properly as it's hard to test. 

### testing
I created some dummy GPG creds that should simulate how to CI environment will look like: 
```
# simulate keys that will exist in CI with a dummy keychain that I made
export LONG_ID=5E7939893DB7189A35E0FB7B8D866B98E9211F8F
export PGP_SECRET=`gpg --pinentry-mode loopback --passphrase test --armor --export-secret-keys $LONG_ID`
export PGP_PASSPHRASE="test"
export SONATYPE_PASSWORD="upload should fail"
export SONATYPE_USERNAME="upload should fail"
```

### running `./sbt "ci-release"`
It does not fail on GPG signing anymore. I do have some errors from sonatype network requests that look suspicious. I would expect something like "forbidden" on the network requests but I get: 

```
Caused by: java.net.ProtocolException: Server redirected too many  times (20)
```

edit: I think this is because of malformed creds but I cannot confirm as I don't have real creds

### sbt-denver
sbt-publish-ci still uses sbt-denver so the tag/snapshot workflow should still work. I kept my overrides for sbt-denver to disable adding timestamps to the snapshots because that will break tests.


build.sbt
```diff
@@ -63,6 +62,7 @@ val sharedSettings = Seq(
   crossScalaVersions := Seq(scalaVersion.value, "2.12.14"),
   javacOptions ++= Seq("-source", "1.8", "-target", "1.8"),
   doc / javacOptions := Seq("-source", "1.8"),
+  versionScheme := Some("early-semver"),
```
hazel-sudz, line 65: sbt throws some annoying warnings without this I don't think it does anything functional though

hazel-sudz:
@johnynek sorry if this takes a few iterations it is hard to test. But I think this should solve our GPG problem at the least

hazel-sudz:
also if it would be possible to run `./sbt "ci-release"` locally with real creds loaded for sonatype if might be helpful to debug any issues with a faster feedback loop here instead of constantly making new PRs.

johnynek (approved):
Do you know what tags we need to add to get it to pick up the right version?

johnynek:
Unfortunately I'm away from my machine now.

hazel-sudz:
> Unfortunately I'm away from my machine now.

let's just merge and see what changes we need after that then. Can't break something that's not working anyways is my logic. 

re tags: not sure I will look at the git history locally when I get the chance. Wild guess but the tag might be attributed to a different branch?

johnynek:
pushed a tag: https://github.com/twitter/scalding/tree/v0.18.0-RC2

let's see if this works.

johnynek:
publish action: https://github.com/twitter/scalding/actions/runs/2142743724

hazel-sudz:
it's a miracle!!! https://oss.sonatype.org/content/repositories/snapshots/com/twitter/maple/0.18.0-RC2-1-64343489-SNAPSHOT/maple-0.18.0-RC2-1-64343489-SNAPSHOT-javadoc.jar

going to try these out next week I'm sure something might come up @johnynek

daniel-sudz:
https://search.maven.org/artifact/com.twitter/scalding-commons_2.12/0.18.0-RC1/jar

johnynek:
It's not a miracle, it's your hard work! Thank you!

PR #1986, opened 2022-04-10
cross build spark 2.X (2.11) and spark 3.X (2.12) on spark-backend

### LINKED ISSUE #1974: scalding-hadoop3 backend OR spark backend upgrade to 3.X
**Is your feature request related to a problem? Please describe.**
For the use cases that @jonshea and I need, we require hadoop3 or spark3. It doesn't matter which one but he have a managed runtime (AWS EMR) that forces us to use newer versions of either. 

**Describe the solution you'd like**
We would like to either: 

1. upgrade the existing spark-backed to spark3.X and scala 2.12
2. if 1 is not possible then we would like to write a hadoop backend that will use hadoop3.

**Questions**
Who uses the spark backend? What was it written for and is in production anywhere? Can we upgrade it or will it be problematic?

hazel-sudz:
build spark 2.X on scala 2.11 and 3.X on scala 2.12. Tests pass cleanly without further changes: 

```bash
./sbt +scalding-spark/test
```

If there's no objection to this from any current users of the spark backend it would be helpful for the use case that I need. Big thanks to the work done in https://github.com/twitter/scalding/pull/1972 by @johnynek that makes this work without pulling in transitive hadoop dependencies from cascading. 

closes #1974

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,6 @@ val scalameterVersion = "0.8.2"
 val scalaCheckVersion = "1.13.4"
 val scalaTestVersion = "3.0.1"
 val scroogeVersion = "19.8.0"
-val sparkVersion = "2.4.0"
 val beamVersion = "2.29.0"
 val slf4jVersion = "1.7.30"
 val thriftVersion = "0.9.3"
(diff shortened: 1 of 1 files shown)


johnynek (commented):
Is that the latest version supported by 2.11? I think we should presume the latest version will work and folks who can't live with that can file issues and/or recompile.

I would do that before complicating the build.

hazel-sudz:
the spark 2.X line support scala 2.11 and scala 2.12 with the latest version being spark-2.4.8
the spark 3.X line support scala 2.12+ and drops scala 2.11 support

edit: so yes 2.4.8 is the latest (and last) version supported by 2.11

johnynek (approved):

PR #1987, opened 2022-04-11
Move ReferencedClassFinder to base

johnynek:
Ideally anything that does not increase the dependencies should be moved to base. By removing a function it looks like no one was using (and is just a trivial redirection to Config) we can do that here (although the tests do reference things in code).

cc @navinvishy @daniel-sudz I think we could use this code in the beam and spark backends in each stage of writes to try to create tokens for all the classes we can see. In kryo, this is registering classes so they have names. Note, I would sort the classes first so we assign the classes in a reproducible way that never depends on any iteration order of a hash set.

### DIFF
--- a/scalding-base/src/main/scala/com/twitter/scalding/ReferencedClassFinder.scala
+++ b/scalding-base/src/main/scala/com/twitter/scalding/ReferencedClassFinder.scala
@@ -1,6 +1,6 @@
 package com.twitter.scalding
 
-import com.twitter.scalding.typed.CoGroupable
+import com.twitter.scalding.typed.{TypedPipe, Input, Output, CoGroupable, KeyedList}
 import org.slf4j.LoggerFactory
 import scala.reflect.runtime.universe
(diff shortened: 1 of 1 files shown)


navinvishy:
In the beam backend, I see that we're getting a kryo instantiator [here](https://github.com/twitter/scalding/blob/e44f2f21fa90f669d6bf81af607c5d537f19e93c/scalding-beam/src/main/scala/com/twitter/scalding/beam_backend/BeamBackend.scala#L133). We're using `KryoHadoop` by default which gets Cascading serialization tokens and registers those classes with Kryo. Are you suggesting removing the dependency on `KryoHadoop`?

hazel-sudz (commented):
this change seems fine to me

I don't quite understand this part though: 

> I think we could use this code in the beam and spark backends in each stage of writes to try to create tokens for all the classes we can see. In kryo, this is registering classes so they have names. Note, I would sort the classes first so we assign the classes in a reproducible way that never depends on any iteration order of a hash set.

what is the use case for this? Having writers be aware of all exiting classes?

PR #1991, opened 2022-04-11
add format check

### LINKED ISSUE #1990: [bug/nit] code format enforcement lost
**Describe the bug**
When we migrated the CI, we lost the auto-formatting CI check: https://github.com/twitter/scalding/pull/1954

I prefer to not worry about formatting when working because it's a waste of time. Todo: add back format enforcement on CI, re-run format on PR. I can take care of this.

hazel-sudz:
closes #1990 

should first fail on develop. Once I push a format commit to the PR it should pass.


scalding-core/src/main/scala/com/twitter/package.scala
```diff
@@ -32,8 +32,8 @@ package object scalding {
   type Grouped[K, +V] = com.twitter.scalding.typed.Grouped[K, V]
 
   /**
-   * scaldingVersion is logged sometimes to inform a user of what scalding version they are using
-   * The value is obtained through code gen with https://github.com/sbt/sbt-buildinfo
+   * scaldingVersion is logged sometimes to inform a user of what scalding version they are using The value is
```
hazel-sudz, line 35: seems like autoformat can break some multiline commends maybe not a big deal?

hazel-sudz:
on the topic of CI improvements: we should try https://github.com/twitter/scalding/issues/1977#issuecomment-1095301691. It's really annoying that codecov upload is so flaky and if we can't make it work 99% of the time maybe we should switch it to something else 

you can see through the CI build history that it seems to fail pretty often for it to be a burden. This is also why we sometimes get such large diffs. It's simply because coverage files were not uploaded.

johnynek:
you are racing to the top of the git commit stats!

hazel-sudz:
really weird thing .. scalafmt takes two passes to stabilize on the formatted code. I think this is a bug as formatting should be a pure function in general

I replicated on both latest version of scalafmt and and the version we have now 🤷

.scalafmt.conf
```diff
@@ -1,8 +1,17 @@
...
+rewrite.redundantBraces.generalExpressions = false
+
+# scalafmt can only choose one scala version target per file to format
+# we have to use 212 for build.sbt or else we get failures
+runner.dialect = scala211
+fileOverride {
+  "glob:**build.sbt" {
+    runner.dialect = scala212
+  }
+}
```
hazel-sudz: figured I would use the latest version while I'm at it

johnynek (approved):

hazel-sudz:
@johnynek can we merge this format PR or are we waiting on something?

johnynek:
There was a failing CI. I was waiting for it to go green. Otherwise yea.

hazel-sudz:
ok the failing CI is the codecov failing upload sometimes. I need to jack up the retry timeout (when it fails, it seems to because of rate limits because a whole cluster of uploads).

hazel-sudz:
https://github.com/twitter/scalding/issues/1993

johnynek:
Congrats on your new position on the contributor charts! ;)

PR #1992, opened 2022-04-13
spark-backend CounterPipe

hazel-sudz:
so counterPipe looks to be good... the discussion point I think is the best way to make `addForce` a blocking call.

### DIFF
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
@@ -2,11 +2,13 @@ package com.twitter.scalding.spark_backend
 
 import com.twitter.scalding.dagon.{FunctionK, Memoize}
 import com.twitter.algebird.Semigroup
-import com.twitter.scalding.Config
+import com.twitter.scalding.{Config, StatKey}
 import com.twitter.scalding.typed._
 import com.twitter.scalding.typed.functions.{DebugFn, FilterKeysToFilter}
+
 import java.util.{LinkedHashMap => JLinkedHashMap, Map => JMap}
 import org.apache.spark.storage.StorageLevel
+
 import scala.collection.mutable.{ArrayBuffer, Map => MMap}
 
 object SparkPlanner {
@@ -62,7 +64,11 @@ object SparkPlanner {
   /**
    * Convert a TypedPipe to an RDD
    */
-  def plan(config: Config, srcs: Resolver[Input, SparkSource]): FunctionK[TypedPipe, Op] =
+  def plan(
+      config: Config,
+      srcs: Resolver[Input, SparkSource],
+      counter: SparkCountersInternal
+  ): FunctionK[TypedPipe, Op] =
     Memoize.functionK(new Memoize.RecursiveK[TypedPipe, Op] {
       import TypedPipe._
 
@@ -72,9 +78,7 @@ object SparkPlanner {
             val resolve = rec(p.pipe)
             resolve.map { x =>
               // increment counter for values in pipe
-              x._2.foreach(counterAction =>
-                SparkCounters.addToCounter(counterAction._1._1, counterAction._1._2, counterAction._2)
-              )
+              counter.add(x._2.iterator)
               // pass through values in pipe
               x._1
             }
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
@@ -2,7 +2,7 @@ package com.twitter.scalding.spark_backend
 
 import com.twitter.scalding.dagon.{FunctionK, Memoize}
 import com.twitter.algebird.Semigroup
-import com.twitter.scalding.{Config, StatKey}
+import com.twitter.scalding.Config
 import com.twitter.scalding.typed._
 import com.twitter.scalding.typed.functions.{DebugFn, FilterKeysToFilter}
 
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
@@ -68,9 +68,17 @@ object SparkPlanner {
 
       def toFunction[A] = {
         case (cp @ CounterPipe(_), rec) =>
-          // TODO: counters not yet supported
-          def go[A](p: CounterPipe[A]): Op[A] =
-            rec(p.pipe).map(_._1)
+          def go[A](p: CounterPipe[A]): Op[A] = {
+            val resolve = rec(p.pipe)
+            resolve.map { x =>
+              // increment counter for values in pipe
+              x._2.foreach(counterAction =>
+                SparkCounters.addToCounter(counterAction._1._1, counterAction._1._2, counterAction._2)
+              )
+              // pass through values in pipe
+              x._1
+            }
+          }
           go(cp)
         case (cp @ CrossPipe(_, _), rec) =>
           def go[A, B](cp: CrossPipe[A, B]): Op[(A, B)] =
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
@@ -75,13 +75,18 @@ object SparkPlanner {
       def toFunction[A] = {
         case (cp @ CounterPipe(_), rec) =>
           def go[A](p: CounterPipe[A]): Op[A] = {
+            // spark only guarantees accurate accumulators inside of RDD actions
+            // so instead of a straight map, we double loop with forEach
             val resolve = rec(p.pipe)
-            resolve.map { x =>
-              // increment counter for values in pipe
-              counter.add(x._2.iterator)
-              // pass through values in pipe
-              x._1
-            }
+            resolve
+              .forEachIdentity { x =>
+                // increment counter for values in pipe
+                counter.add(x._2.iterator)
+              }
+              .map { x =>
+                // pass through values in pipe
+                x._1
+              }
           }
           go(cp)
         case (cp @ CrossPipe(_, _), rec) =>
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -37,14 +37,17 @@ class SparkCounters(sparkSession: SparkSession, counterPrefix: String) {
   sparkSession.sparkContext.register(accumulator, counterPrefix)
 
   /**
-   * We return an ExecutionCounters interface. The results will not be stable unless spark execution has
-   * finished.
+   * We return an ExecutionCounters interface and reset the counter state. This should only be called at
+   * SparkWriter once execution has completely finished.
    */
   def asExecutionCounters(): ExecutionCounters = {
     val copyState = MutableMap[StatKey, Long]()
-    for (kvPair <- accumulator.value) {
-      val keyEntry = StatKey(kvPair._1._2, kvPair._1._1)
-      copyState(keyEntry) = kvPair._2
+    accumulator.synchronized {
+      for (kvPair <- accumulator.value) {
+        val keyEntry = StatKey(kvPair._1._2, kvPair._1._1)
+        copyState(keyEntry) = kvPair._2
+      }
+      accumulator.reset()
     }
     val immutableCopyState = copyState.toMap
     new ExecutionCounters {
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -17,17 +17,16 @@ class SparkCounters(sparkSession: SparkSession, counterPrefix: String = java.uti
   val accumulator: SparkCountersInternal = new SparkCountersInternal
 
   /**
-   * register for spark to be aware of accumulator. Note that accumulators are now automatically garbage
-   * collected: https://github.com/apache/spark/pull/4021 so we don't need to explicitly release them
+   * Register for spark to be aware of accumulator. Note that accumulators are now automatically garbage
+   * collected: https://github.com/apache/spark/pull/4021 so we don't need to explicitly release them.
    */
   {
     sparkSession.sparkContext.register(accumulator, counterPrefix)
   }
 
   /**
-   * we return an ExecutionCounters interface.
-   *
-   * Values returned from keys/get will not be stable until spark execution is complete
+   * We return an ExecutionCounters interface. The results will not be stable unless spark execution has
+   * finished.
    */
   def asExecutionCounters(): ExecutionCounters = {
     val copyState = MutableMap[StatKey, Long]()
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -119,9 +119,10 @@ sealed class SparkCountersInternal(
   override def copyAndReset(): SparkCountersInternal = new SparkCountersInternal
 
   // internal for spark do not call
-  override def merge(other: AccumulatorV2[((String, String), Long), ImmutableMap[(String, String), Long]]) =
+  override def merge(other: AccumulatorV2[((String, String), Long), ImmutableMap[(String, String), Long]]) = {
+    val otherMapState = other.value
     mapState.synchronized {
-      for (kvPair <- other.value) {
+      for (kvPair <- otherMapState) {
         val prev = mapState.get(kvPair._1) match {
           case Some(v) => v
           case None    => 0L
@@ -129,6 +130,7 @@ sealed class SparkCountersInternal(
         mapState(kvPair._1) = prev + kvPair._2
       }
     }
+  }
 
   // internal for spark do not call
   override def isZero(): Boolean = mapState.synchronized {
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -7,10 +7,7 @@ import org.apache.spark.util.AccumulatorV2
 import scala.collection.mutable.{Map => MutableMap}
 
 /**
- * Spark Counters is a singleton responsible for managing counters in the spark backend
- *
- * Inspiration has been taken from
- * https://stackoverflow.com/questions/66438570/create-accumulator-on-executor-dynamically
+ * We use a new SparkCounters instance for each SparkWriter to keep track of scalding counters
  */
 class SparkCounters(sparkSession: SparkSession, counterPrefix: String = java.util.UUID.randomUUID.toString) {
 
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -23,7 +23,12 @@ class SparkCounters(sparkSession: SparkSession, counterPrefix: String) {
    * implementation here:
    * https://github.com/apache/spark/blob/87c744b60507f82e1722f1488f1741cb2bb8e8e5/core/src/main/scala/org/apache/spark/util/AccumulatorV2.scala#L165
    */
-  @transient val accumulator: SparkCountersInternal = new SparkCountersInternal
+  @transient private val accumulator: SparkCountersInternal = new SparkCountersInternal
+
+  /**
+   * This should only be called inside of SparkWriter execution.
+   */
+  def getAccumulator: SparkCountersInternal = accumulator
 
   /**
    * Register for spark to be aware of accumulator. Note that accumulators are now automatically garbage
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -36,17 +36,20 @@ object SparkCounters {
   def addToCounter(counterGroup: String, counterName: String, diffVal: Long): Unit =
     getAccumulator.add(counterGroup, counterName, diffVal)
 
-  def exportExecutionCounters(): ExecutionCounters = {
-    val statKeyForm =
-      getAccumulator.value.foldLeft(Set[StatKey]())((ssk, kvPair) =>
-        ssk + StatKey(kvPair._1._1, kvPair._1._2)
-      )
+  /**
+   * we return an ExecutionCounters interface.
+   *
+   * Values returned from keys/get will not be stable until spark execution is complete
+   */
+  def lazyEvaluateAsExecutionCounters(): ExecutionCounters =
     new ExecutionCounters {
-      def keys = statKeyForm
+      def keys =
+        getAccumulator.value.foldLeft(Set[StatKey]())((ssk, kvPair) =>
+          ssk + StatKey(kvPair._1._1, kvPair._1._2)
+        )
       def get(key: StatKey) =
         getAccumulator.value.get((key.group, key.counter))
     }
-  }
 
   private def getAccumulator(): SparkCountersInternal = {
     if (accumulator == null) {
--- a/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
+++ b/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
@@ -12,66 +12,57 @@ import scala.collection.mutable.{Map => MutableMap}
  * Inspiration has been taken from
  * https://stackoverflow.com/questions/66438570/create-accumulator-on-executor-dynamically
  */
-object SparkCounters {
+class SparkCounters(sparkSession: SparkSession, counterPrefix: String = java.util.UUID.randomUUID.toString) {
 
-  var accumulator: SparkCountersInternal = null
+  val accumulator: SparkCountersInternal = new SparkCountersInternal
 
   /**
-   * This should be called once prior to starting execution on the spark backend. The spark context must
-   * already be active in the JVM.
+   * register for spark to be aware of accumulator. Note that accumulators are now automatically garbage
+   * collected: https://github.com/apache/spark/pull/4021 so we don't need to explicitly release them
    */
-  def register(): Unit = {
-    val sparkContext = SparkSession.getActiveSession
-      .getOrElse(throw new IllegalStateException("SparkSession is not active!"))
-      .sparkContext
-    val accum = new SparkCountersInternal
-    sparkContext.register(accum, "[scalding/spark-backend] accumulator")
-    accumulator = accum
+  {
+    sparkSession.sparkContext.register(accumulator, counterPrefix)
   }
-
-  /**
-   * Adds a long value to a (counterGroup, counterName). Should be used inside of executors for side-effect
-   * free counter implementation
-   */
-  def addToCounter(counterGroup: String, counterName: String, diffVal: Long): Unit =
-    getAccumulator.add(counterGroup, counterName, diffVal)
 
   /**
    * we return an ExecutionCounters interface.
    *
    * Values returned from keys/get will not be stable until spark execution is complete
    */
-  def lazyEvaluateAsExecutionCounters(): ExecutionCounters =
+  def asExecutionCounters(): ExecutionCounters = {
+    val copyState = MutableMap[StatKey, Long]()
+    for (kvPair <- accumulator.value) {
+      val keyEntry = StatKey(kvPair._1._2, kvPair._1._1)
+      copyState.getOrElseUpdate(keyEntry, 0L)
+      copyState(keyEntry) += kvPair._2
+    }
     new ExecutionCounters {
-      def keys =
-        getAccumulator.value.foldLeft(Set[StatKey]())((ssk, kvPair) =>
-          ssk + StatKey(kvPair._1._1, kvPair._1._2)
-        )
-      def get(key: StatKey) =
-        getAccumulator.value.get((key.group, key.counter))
+      def keys = copyState.keys.toSeq.toSet
+      def get(key: StatKey) = copyState.get(key)
     }
-
-  private def getAccumulator(): SparkCountersInternal = {
-    if (accumulator == null) {
-      throw new IllegalStateException("SparkCounters.register has not been called!")
-    }
-    accumulator
   }
 
 }
 
 sealed class SparkCountersInternal(
-    val mapState: MutableMap[(String, String), Long] = MutableMap[(String, String), Long]()
-) extends AccumulatorV2[(String, String, Long), MutableMap[(String, String), Long]]
+    mapState: MutableMap[(String, String), Long] = MutableMap[(String, String), Long]()
+) extends AccumulatorV2[((String, String), Long), MutableMap[(String, String), Long]]
     with Serializable {
 
+  /**
+   * Should be used inside of spark executor actions for a side-effect free counter implementation.
+   */
+  def add(in: Iterator[((String, String), Long)]): Unit =
+    while (in.hasNext) {
+      val count = in.next
+      mapState.getOrElseUpdate(count._1, 0L)
+      mapState(count._1) += count._2
+    }
+
   // internal for spark do not call
-  override def add(in: (String, String, Long)): Unit = {
-    val counterGroup = in._1
-    val counterName = in._2
-    val counterDiff = in._3
-    mapState.getOrElseUpdate((counterGroup, counterName), 0)
-    mapState((counterGroup, counterName)) += counterDiff
+  override def add(in: ((String, String), Long)): Unit = {
+    mapState.getOrElseUpdate(in._1, 0L)
+    mapState(in._1) += in._2
   }
 
   // internal for spark do not call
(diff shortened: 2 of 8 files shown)


scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -219,6 +220,8 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
       (nextState.copy(id = nextState.id + 1), (nextState.id, acts))
     }
     // now we run the actions:
-    CFuture.uncancellable(Future.traverse(acts)(fn => fn()).map(_ => (id, ExecutionCounters.empty)))
+    CFuture.uncancellable(
+      Future.traverse(acts)(fn => fn()).map(_ => (id, SparkCounters.lazyEvaluateAsExecutionCounters()))
```
hazel-sudz: I am currently confused why getting the SparkCounters state is empty inside of this map scope but if I compute it lazily it has the correct values later when asserting in tests.

hazel-sudz: when placing breakpoints, the map block above hits before the counterPipe is processed in spark

hazel-sudz: @johnynek any tips?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
```diff
@@ -68,9 +68,17 @@ object SparkPlanner {
 
       def toFunction[A] = {
         case (cp @ CounterPipe(_), rec) =>
-          // TODO: counters not yet supported
-          def go[A](p: CounterPipe[A]): Op[A] =
-            rec(p.pipe).map(_._1)
+          def go[A](p: CounterPipe[A]): Op[A] = {
+            val resolve = rec(p.pipe)
+            resolve.map { x =>
+              // increment counter for values in pipe
```
johnynek: perf nits:

1. access `counterAction._1` once
2. iterator is better since it is always inlined

```scala
resolve.map { x =>
  val counts = x._2.iterator
  while (counts.hasNext) {
    val ((g, c), amt) = counts.next()
    SparkCounters.addToCounter(g, c, amt)
  }
  x._1
}
```

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Spark Counters is a singleton responsible for managing counters in the spark backend
+ *
+ * Inspiration has been taken from
+ * https://stackoverflow.com/questions/66438570/create-accumulator-on-executor-dynamically
+ */
+object SparkCounters {
+
+  var accumulator: SparkCountersInternal = null
```
johnynek: 1. can we make this private?
2. I would probably instead do `@transient private lazy val accumulator = new AtomicReference[SparkCountersInternal](null)` and use atomic operations since I think you can have multithreading in spark.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,104 @@
...
+    val accum = new SparkCountersInternal
+    sparkContext.register(accum, "[scalding/spark-backend] accumulator")
+    accumulator = accum
+  }
+
+  /**
+   * Adds a long value to a (counterGroup, counterName). Should be used inside of executors for side-effect
+   * free counter implementation
+   */
+  def addToCounter(counterGroup: String, counterName: String, diffVal: Long): Unit =
```
johnynek: if you do this in a loop you are calling `getAccumulator` many times. I would also add a method that accepts `Iterable[((String, String), Long)]` or something so you can get once, then add everything.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,104 @@
...
+    if (accumulator == null) {
+      throw new IllegalStateException("SparkCounters.register has not been called!")
+    }
+    accumulator
+  }
+
+}
+
+sealed class SparkCountersInternal(
+    val mapState: MutableMap[(String, String), Long] = MutableMap[(String, String), Long]()
```
johnynek: why make this a val?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,104 @@
...
+) extends AccumulatorV2[(String, String, Long), MutableMap[(String, String), Long]]
```
johnynek: if you make this `AccumulatorV2[((String, String), Long), ...]` can you avoid reboxing the counters?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,104 @@
...
+        newMap(kvPair._1) += kvPair._2
+        newMap
+    }
+    new SparkCountersInternal(copyState)
+  }
+
+  // internal for spark do not call
+  override def merge(other: AccumulatorV2[(String, String, Long), MutableMap[(String, String), Long]]) =
+    for (kvPair <- other.value) {
+      mapState.getOrElseUpdate(kvPair._1, 0)
```
johnynek: `0L` would be better to be more specific.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -146,6 +146,7 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
   def execute(conf: Config, writes: List[ToWrite[_]])(implicit
       cec: ExecutionContext
   ): CFuture[(Long, ExecutionCounters)] = {
+    SparkCounters.register // initialize counter tracker for spark backend
```
johnynek: I think this is probably better done in `newWriter` in `SparkMode` or in the constructor here.

You can have many stages to an `Execution`. Do you want to do this on every stage? I don't think so.

Also, I think ideally we could have a cleanup where maybe we unregister when the Writer is closed?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
johnynek: These futures complete when spark has returned an RDD in many cases. That may not be when the job is finished.

That's an issue. That hasn't been a problem that I know of so far. So we really need to make sure this Future is only complete when *all* of these writes have fully occurred.

I think the problem is probably with `def force`. So, this part of the future should occur earlier:

https://github.com/twitter/scalding/blob/e44f2f21fa90f669d6bf81af607c5d537f19e93c/scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala#L90

and in the case of `forceToDiskExecution` we need some way for spark to tell us when the caching/materialization is done. I don't know how to do that off the top of my head.

Alternatively, rather than using cache, we could make a temp file ourselves, write to that, which is a spark action and we can block on that, then we can re-read that path ourselves.

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBackendTests.scala
```diff
@@ -67,6 +67,12 @@ class SparkBackendTests extends FunSuite with BeforeAndAfter {
       t.toIterableExecution.waitFor(Config.empty, MemoryMode.empty).get
     )
 
+  def sparkRetrieveCounters[A: Ordering](t: TypedPipe[A], conf: Config = Config.empty) = {
+    val smode = SparkMode.default(session)
+    val (eiter, ecounters) = t.toIterableExecution.getCounters.waitFor(conf, smode).get
```
johnynek: we are using `toIterableExecution` here, so the fact that the `toLocalIterator` is being done in the wrong place is likely your issue.

hazel-sudz:
thank you so much! I will look into making the changes

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
hazel-sudz: > These futures complete when spark has returned an RDD in many cases. That may not be when the job is finished.

same conclusion I came to I will look into trying to solve this issue I might also split it off into a separate PR if I figure it out

hazel-sudz: there is a pattern that was historically used inside of spark sql of using the `count` operation to force an RDD computation: 

https://github.com/apache/spark/pull/30403/files#diff-aa467582590608ec702c1caf6208e79af5e6f3ee2cfd58a53bdf2601cd5015d6R65

this is one way to achieve a blocking `cache` operation though it does carry some minimal overhead of re-reading the RDD. 

it's surprising that there's a blocking version of `unpersist` but no blocking `persist` on RDD

johnynek: okay, that makes sense for the Force case, but I think being more direct and fixing the bug with toIterable.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkBackend.scala
hazel-sudz: done

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -161,19 +168,32 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
...
-        val rddF = op.run(session)
+        val rddF = op
+          .run(session)
         promise.completeWith(rddF)
-        rddF.map(_ => ())
+        forcedRDD.map { x =>
+          x.count // any cheap RDD action to force DAG execution works here
+          ()
+        }
       }
```
hazel-sudz: okay I currently have this hack to make the addForce method be blocking. I'm sure there is a better way but I currently can't think of one. Is this alright for now?

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBackendTests.scala
hazel-sudz: I believe the call to toLocalIterator happens after the `execute` call for `SparkWriter`. So I'm not sure how we can lift that call for `execute` to be aware of it. 

But... that would definitely be the cleanest solution to use instead of the `count` hack I currently have

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBackendTests.scala
```diff
@@ -145,6 +151,44 @@ class SparkBackendTests extends FunSuite with BeforeAndAfter {
...
+              .tallyBy("inner")(x => (if (x % 3 == 0) "divisible by 3" else "not divisible by 3"))
+          )
+          .tallyBy("outer")(x => (if (x._2 % 3 == 0) "divisible by 3" else "not divisible by 3"))
+      val cresult2 = sparkRetrieveCounters(cpipe3)
+      assert(cresult2.toMap.size == 5)
+      assert(cresult2.get(StatKey("divisible by 4", "something interesting")).get == 25)
+      assert(cresult2.get(StatKey("divisible by 3", "inner")).get == 4)
+      assert(cresult2.get(StatKey("not divisible by 3", "inner")).get == 7)
+      assert(cresult2.get(StatKey("divisible by 3", "outer")).get == 25 * 4)
+      assert(cresult2.get(StatKey("not divisible by 3", "outer")).get == 25 * 7)
```
hazel-sudz: added a more tricky test case

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
hazel-sudz: done

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
hazel-sudz: moved into SparkWriter constructor and fixed for it to be compatible with multistage execution.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
hazel-sudz: done

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
hazel-sudz: done

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
hazel-sudz: fixed

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
hazel-sudz: I don't think 2 is necessary. Check out this explanation by someone familiar with spark internals: 
https://stackoverflow.com/a/43912747

> Accumulators are not thread-safe, in fact they do not need to be thread-safe. For executors, accumulators are write only variables, they can be added by executors and they can be read by the driver. Driver makes use of DAGScheduler.updateAccumulators method to update values of accumulators after task is completed, and this method is called only from a thread that runs scheduling loop. At a time, only one task completion event is handled. That's why there is no need for accumulators to be thread-safe.

hazel-sudz: for 1 I changed the implementation so not sure if it's still applicable

johnynek:
can you merge develop now that I've merged the format?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+ */
+class SparkCounters(sparkSession: SparkSession, counterPrefix: String = java.util.UUID.randomUUID.toString) {
+
+  val accumulator: SparkCountersInternal = new SparkCountersInternal
+
+  /**
+   * Register for spark to be aware of accumulator. Note that accumulators are now automatically garbage
+   * collected: https://github.com/apache/spark/pull/4021 so we don't need to explicitly release them.
+   */
+  {
```
johnynek: let's remove the `{ }` please.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.spark.util.AccumulatorV2
+
+import scala.collection.mutable.{Map => MutableMap}
+
+/**
+ * We use a new SparkCounters instance for each SparkWriter to keep track of scalding counters
+ */
+class SparkCounters(sparkSession: SparkSession, counterPrefix: String = java.util.UUID.randomUUID.toString) {
+
+  val accumulator: SparkCountersInternal = new SparkCountersInternal
```
johnynek: can this be private? Can you add a comment on why this mutable value is safe? Is it never used in a multi-threaded context?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+import com.twitter.scalding.{ExecutionCounters, StatKey}
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.util.AccumulatorV2
+
+import scala.collection.mutable.{Map => MutableMap}
+
+/**
+ * We use a new SparkCounters instance for each SparkWriter to keep track of scalding counters
+ */
+class SparkCounters(sparkSession: SparkSession, counterPrefix: String = java.util.UUID.randomUUID.toString) {
```
johnynek: Can we remove the side-effecting default value. Instead can you make a method on the companion:

```scala
object SparkCounters {
  def withRandomContextPrefix(sparkSession: SparkSession): SparkCounters =
    new SparkCounters(sparkSession, java.util.UUID.randomUUID.toString)
}
```

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+   */
+  def asExecutionCounters(): ExecutionCounters = {
+    val copyState = MutableMap[StatKey, Long]()
+    for (kvPair <- accumulator.value) {
+      val keyEntry = StatKey(kvPair._1._2, kvPair._1._1)
+      copyState.getOrElseUpdate(keyEntry, 0L)
+      copyState(keyEntry) += kvPair._2
+    }
+    new ExecutionCounters {
+      def keys = copyState.keys.toSeq.toSet
```
johnynek: why doesn't `def keys = copyState.keySet` work here? copystate never changes after this method right?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+) extends AccumulatorV2[((String, String), Long), MutableMap[(String, String), Long]]
+    with Serializable {
+
+  /**
+   * Should be used inside of spark executor actions for a side-effect free counter implementation.
+   */
+  def add(in: Iterator[((String, String), Long)]): Unit =
+    while (in.hasNext) {
+      val count = in.next
+      mapState.getOrElseUpdate(count._1, 0L)
```
johnynek: unfortunately, `getOrElseUpdate` allocates since the `else` clause is lazy. Can we move to:

```scala
val prev =
  mapState.get(count._1) match {
    case Some(v) => v
    case None => 0L
  }
mapState(count._1) = prev + count._2
```

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+    mapState: MutableMap[(String, String), Long] = MutableMap[(String, String), Long]()
+) extends AccumulatorV2[((String, String), Long), MutableMap[(String, String), Long]]
+    with Serializable {
+
+  /**
+   * Should be used inside of spark executor actions for a side-effect free counter implementation.
+   */
+  def add(in: Iterator[((String, String), Long)]): Unit =
+    while (in.hasNext) {
+      val count = in.next
```
johnynek: convention in scala is to use `()` for side-effecting things, like this: `val count = in.next()`

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+      mapState.getOrElseUpdate(count._1, 0L)
+      mapState(count._1) += count._2
+    }
+
+  // internal for spark do not call
+  override def add(in: ((String, String), Long)): Unit = {
+    mapState.getOrElseUpdate(in._1, 0L)
```
johnynek: same comment as above about using the match pattern.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  // internal for spark do not call
+  override def value() = mapState
+
+  // internal for spark do not call
+  override def copy(): SparkCountersInternal = {
+    val copyState = MutableMap[(String, String), Long]()
+    for (kvPair <- mapState) {
+      copyState.getOrElseUpdate(kvPair._1, kvPair._2)
```
johnynek: mapState is already a map, so I don't think you need this `getOrElseUpdate`. I think:

```scala
val bldr = MutableMap.newBuilder[(String, String), Long]()
bldr ++= mapState
bldr.result()
```
is the more efficient way to do this.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+    }
+    new SparkCountersInternal(copyState)
+  }
+
+  // internal for spark do not call
+  override def merge(other: AccumulatorV2[((String, String), Long), MutableMap[(String, String), Long]]) =
+    for (kvPair <- other.value) {
+      mapState.getOrElseUpdate(kvPair._1, 0L)
```
johnynek: I think it is better to do:
```scala
val prev = mapState.get(kvPair._1) {
  case Some(v) => v
  case None => 0L
}
mapState(kvPair._1) = prev + kvPair._2
```

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
```diff
@@ -63,9 +65,9 @@ class SparkWriter(val sparkMode: SparkMode) extends Writer {
           val newForced = forcedPipes + ((c, opt) -> workVal)
           val newInitToOpt = initToOpt + ((c, init) -> opt)
 
-          (copy(sources = newSources, forcedPipes = newForced, initToOpt = newInitToOpt), true)
+          (copy(sources = newSources, forcedPipes = newForced, initToOpt = newInitToOpt), true, forcedRdd)
         case Some(_) =>
```
johnynek: shouldn't we get the `forcedRdd` from the `WorkVal` you have here? I think so. If not, please add a comment why not.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
johnynek: I think we actually want:

```scala
forcedRdd.flatMap { x =>
  Future {
    scala.concurrent.blocking {
      x.count()
      ()
    }
  }
}
```

Because we don't want to block the mapping thread. (we should review that we are using `scala.concurrent.blocking` on blocking operations.

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBackendTests.scala
```diff
@@ -67,6 +67,12 @@ class SparkBackendTests extends FunSuite with BeforeAndAfter {
       t.toIterableExecution.waitFor(Config.empty, MemoryMode.empty).get
     )
 
+  def sparkRetrieveCounters[A: Ordering](t: TypedPipe[A], conf: Config = Config.empty) = {
```
johnynek: do you actually need the `Ordering` here? I don't see why.

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBackendTests.scala
```diff
@@ -145,6 +151,44 @@ class SparkBackendTests extends FunSuite with BeforeAndAfter {
...
+
+      val cpipe3 =
+        cpipe2
+          .cross(
+            TypedPipe
+              .from(0 to 10)
+              .tallyBy("inner")(x => (if (x % 3 == 0) "divisible by 3" else "not divisible by 3"))
+          )
+          .tallyBy("outer")(x => (if (x._2 % 3 == 0) "divisible by 3" else "not divisible by 3"))
+      val cresult2 = sparkRetrieveCounters(cpipe3)
```
johnynek: Can we also do a version where you write to a null-sink or text file or something and make sure the `writeExecution` version also works?

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,89 @@
...
+    mapState(in._1) += in._2
+  }
+
+  // internal for spark do not call
+  override def value() = mapState
+
+  // internal for spark do not call
+  override def copy(): SparkCountersInternal = {
+    val copyState = MutableMap[(String, String), Long]()
+    for (kvPair <- mapState) {
```
johnynek: this isn't thread-safe: mapState can change while iterating.

Can you comment why this is safe?

Alternatives:
1. use `mapState.synchronized { }` before reading/writing
2. use an `immutable.Map[(String, String), Long]` and an `AtomicReference` so taking a copy is very cheap (although updates will be maybe twice as slow, so maybe the mutex approach above is better).

hazel-sudz: all of these methods on accumulator operate on a "share nothing" model. The driver maintains the "true" state of the accumulator and merges updates from other accumulator states returned from executors. I am 99% sure that the merging process does not require thread safety to function. 

If you look at the implementation for what is passed to executors: 
```scala
  // Called by Java when serializing an object
  final protected def writeReplace(): Any = {
    if (atDriverSide) {
      if (!isRegistered) {
        throw new UnsupportedOperationException(
          "Accumulator must be registered before send to executor")
      }
      val copyAcc = copyAndReset()
      assert(copyAcc.isZero, "copyAndReset must return a zero value copy")
      val isInternalAcc = name.isDefined && name.get.startsWith(InternalAccumulator.METRICS_PREFIX)
      if (isInternalAcc) {
        // Do not serialize the name of internal accumulator and send it to executor.
        copyAcc.metadata = metadata.copy(name = None)
      } else {
        // For non-internal accumulators, we still need to send the name because users may need to
        // access the accumulator name at executor side, or they may keep the accumulators sent from
        // executors and access the name when the registered accumulator is already garbage
        // collected(e.g. SQLMetrics).
        copyAcc.metadata = metadata
      }
      copyAcc
    } else {
      this
    }
  }
``` 

executor threads receive a state-less version of the accumulator.

hazel-sudz: the only time we read the value of the driver accumulator is at the end after execution has already ended and all updates to the accumulator have been merged by spark.

johnynek: Okay, can you add that comment to the code around the class something like:

```scala
/**
 * regarding thread-safety: .... 
 * permalink to this code: ... 
```

hazel-sudz: hmmm maybe not I'm starting to doubt based on the official example: https://github.com/apache/spark/blob/0494dc90af48ce7da0625485a4dc6917a244d580/core/src/main/scala/org/apache/spark/util/AccumulatorV2.scala#L451

I want to try to write a unit test to prove if this is necessary because synchronizing threads is going to be a big perf hit.

johnynek: Sync won't be a big perf hit: you are copying here. The added cost of synchronization will not be large. Also this probably happens very rarely (at some kind of boundary, such as the task finishing). I would be surprised if you could measure a hit.

hazel-sudz: good point this only happens after an rdd actions starts/returns. So assuming a very complicated job has ~100-1000 rdd actions this is no problem

 ok let's just do it better safe than sorry

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
hazel-sudz: changed all methods to be thread safe based on reported crashes and resolutions in https://issues.apache.org/jira/browse/SPARK-17463 after digging for info

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
johnynek: I wonder if your failing test is due to this comment? You aren't returning a previously forced value.

hazel-sudz: I have not looked at that yet. But one other thing I suspect: the counter increment is done through map transform on RDD. Spark does not guarantee counter consistency in transformation (only actions).

is we want strict consistency we would need a blocking action like `forEach`

hazel-sudz: the reason I suspect this is because the result is a strict multiple of the expected value (200 returned vs 100 expected). This leads me to think that the map phase was recomputed twice by the spark engine.

johnynek: Yes, but it could also be because the planner asked Spark to compute it twice. Not clear to me. I would investigate and try to falsify both.

hazel-sudz: > shouldn't we get the `forcedRdd` from the `WorkVal` you have here? I think so. If not, please add a comment why not.

I am a bit confused as to what is being referred to here. `WorkVal` is just `val workVal: WorkVal[T] = (src, forcedRdd)` no?.

Anyways it does appear that we have a double computation somewhere in the planner. Specifically happens in the case that this line is included: `.flatMap(_ => TypedPipe.from(sinkExample).toIterableExecution)` after an execution.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,140 @@
...
+    for (kvPair <- accumulator.value) {
+      val keyEntry = StatKey(kvPair._1._2, kvPair._1._1)
+      copyState(keyEntry) = kvPair._2
+    }
+    val immutableCopyState = copyState.toMap
+    new ExecutionCounters {
+      def keys = immutableCopyState.keySet
+      def get(key: StatKey) = immutableCopyState.get(key)
+    }
+  }
```
hazel-sudz: I believe I located the problem with the failing test. We need to reset the counter state before we return from here. I think what's going on is that flapMapping the execution re-uses the same writer which duplicates the counter values.

I confirmed that the actual counter incrementation occurs only once by stepping through breakpoints and checking on the counter state.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkCounters.scala
```diff
@@ -0,0 +1,143 @@
...
+    new SparkCountersInternal(copyState)
+  }
+
+  // internal to spark do not call
+  override def copyAndReset(): SparkCountersInternal = new SparkCountersInternal
+
+  // internal for spark do not call
+  override def merge(other: AccumulatorV2[((String, String), Long), ImmutableMap[(String, String), Long]]) =
+    mapState.synchronized {
+      for (kvPair <- other.value) {
```
johnynek: nit:

```scala
// get this without holding the lock here, so we avoiding holding two locks simultaneously
val thatMapState = other.value
mapState.synchronized {
 for (kvPair <- thatMapState) { ... 
}
```

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
johnynek: yeah `WorkVal` is that pair, but if you have a cached value, you are ignoring the cached value and returning the input. I think you need to use that cached value. Don't you? If not, can you explain why we can safely ignore an existing `Future[RDD[_ <: A]]` given that we already have another equivalent computation?

I don't see why, but it could be true.

Basically: if we have two identical optimized typedpipes, they should share the exact same `Future[RDD...`.

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
johnynek: We still need to address this. Otherwise this call may block a compute thread which can really harm concurrency.

scalding-spark/src/test/scala/com/twitter/scalding/spark_backend/SparkBaseTest.scala
```diff
@@ -0,0 +1,80 @@
...
+      t.toIterableExecution.waitFor(Config.empty, MemoryMode.empty).get
+    )
+
+  def sparkRetrieveCounters[A](t: TypedPipe[A]) = {
+    val smode = SparkMode.default(session)
+    val (eiter, ecounters) = t.toIterableExecution.getCounters.waitFor(Config.empty, smode).get
+    ecounters
+  }
+
+  def sparkRetrieveCounters[A](t: Execution[_], conf: Config = Config.empty) = {
```
johnynek: can we remove `[A]` and use `Execution[Any]`

scalding-spark/src/main/scala/com/twitter/scalding/spark_backend/SparkWriter.scala
hazel-sudz: ok yeah if I'm understanding correctly we should do this. Sounds good to me if this is the case: 

```scala
-        case Some(_) =>
-          (copy(initToOpt = initToOpt + ((c, init) -> opt)), false, rdd)
+        case Some(cacheHit) =>
+          (copy(initToOpt = initToOpt + ((c, init) -> opt)), false, cacheHit._2)

```

johnynek: yeah, I think so, or I would write `case Some((_, frdd)) => ...`

johnynek (approved):

johnynek:
This is great! Thanks.

PR #1996, opened 2022-04-28
publish for scalding base missing

hazel-sudz:
I noticed when pulling down the `scalding-core` snapshot the `scalding-base` module was not published. This can be confirmed by looking at https://oss.sonatype.org/content/repositories/snapshots/com/twitter/

I'm not 100% on how sbt works but I think it needs to be listed under the aggregation for 
```
lazy val scalding = Project(id = "scalding", base = file("."))
```

I tested with `./sbt publishLocal ` and it seemed to do the trick locally.


hazel-sudz:
slightly awkward situation: the java8 runtime has dropped some classes that hadoop relies on see: https://issues.apache.org/jira/browse/HADOOP-16590. 

this means presumably that any managed cloud services are quickly going to become completely incompatible with scalding if they use recent java8 versions. This is especially problematic because we are pinned to old cascading + old hadoop so we can't pull in the upstream patch. 

we can lock the java8 version in CI for now but this is not great. At some point we are going to have to drop support for cascading backend it just seems inevitable. 

CC @nownikhil @navinvishy curious on your thoughts. Is it reasonable in the future to expect only support of beam/spark backends?

.github/workflows/publish.yml
```diff
@@ -4,8 +4,8 @@ on:
   push:
     branches:
       - "develop"
-    release:
-      types: [created]
+    tags:
+      - "v*"
```
hazel-sudz, line 8: I think we want tags here instead of release. The last tag push didn't trigger publish because of this.

johnynek (approved):
Thanks!

